PR #764, opened 2016-05-25
Add solarized dark syntax highlighting

lucperkins:
This PR adds [Solarized Dark](http://ethanschoonover.com/solarized) syntax highlighting to the code samples, which in my opinion looks much better. It also adds the appropriate build steps to the `Makefile`. Here's an example:

screen shot 2016-05-24 at 8 14 07 pm

I also added the appropriate URLs to the footer (for "Support," "Governance," etc.).

### DIFF
--- a/website/scripts/setup.sh
+++ b/website/scripts/setup.sh
@@ -5,6 +5,7 @@ SOLARIZED_PYGMENT_DIR=assets/solarized-pygment
 brew list hugo || brew install hugo
 npm install
 
+pip uninstall requests -y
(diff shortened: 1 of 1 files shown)


lucperkins:
@billonahill It broke my environment, too. These pip permissions issues are so thorny and almost completely inscrutable to me. My most-likely-suboptimal solution was to uninstall and reinstall pip and `sudo`ize all the pip stuff in those scripts. I'll see if I can get some help from a Python-savvy engineer (which I am not).

PR #786, opened 2016-05-25
Correct the webpages of heron internal configuration

maosongfu:


website/content/docs/operators/configuration/config-intro.md
```diff
@@ -1,17 +1,18 @@
 ---
-title: Intro to Heron Configuration
+title: Intro to Heron Internal Configuration
 ---
 
-Heron can be configured at two levels:
+Heron internal configuration can be configured at two levels:
```
lucperkins: What do you mean by "internal" here? Make sure to clarify because there are a lot of moving pieces at work in Heron.

maosongfu: For "internal", I mean it is config applies to the whole cluster, and is transparent to topology developers (it is impossible for them to override it)

billonahill:
👍

website/content/docs/operators/configuration/config-intro.md
lucperkins: I'm not sure how others feel about this, but it might be better to call this doc "Intro to Heron Cluster Configuration." "Internal" is just a bit too vague for my tastes. Then, the first few sentences could be something like this:

"Heron clusters can be configured at two different levels:

1. **The system level** --- System-level configurations apply to the whole Heron cluster rather than to any specific topology (e.g. logging configs).
2. **The component level** --- Component-level configurations enable you to establish default configurations for different topology components.

Neither system- nor component-level configurations can be overridden by topology developers."

maosongfu: sounds good. updated.

lucperkins:
:+1:

Issue #797, opened 2016-05-25
links to java docs on documentation page broken when viewing on local server


objmagic:
current solution suggested by @lucperkins is that one can do ``cp -rf public/ static/`` to make Java docs API appear under a ``website/static`` directory. Lucas is trying to figure out a better solution.

objmagic:
cf #798

objmagic:
resolved via https://github.com/twitter/heron/pull/943

PR #798, opened 2016-05-25
Temporary solution to Javadocs issue (#797)

lucperkins:
I'm not particularly fond of this solution and I hope that we can find a better one, but this is *a* solution that eliminates some unpleasant surprises for new users and seems to have very few potential downsides. Some background:

* As far as I can tell, in order to view the Javadocs locally when running `make serve` (which starts up the Hugo server to view the docs locally), they need to be in the `website/static` directory. This is important for developing the docs, fixing links, etc.
* But if the Javadocs are in `website/static` and not in `website/public`  then they can't be published to the `gh-pages` branch (and thus cannot go live on the site).
* Having the Javadocs in both locations provides a solution but it's not [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and feels kind of janky

Suggestions welcome!

### DIFF
--- a/website/scripts/javadocs.sh
+++ b/website/scripts/javadocs.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 #set -e TODO: figure out why this breaks CI and re-enable (https://github.com/twitter/heron/issues/766)
 
-JAVADOC=javadoc
 FLAGS="-quiet"
 
 HERON_ROOT_DIR=$(git rev-parse --show-toplevel)
@@ -30,8 +29,9 @@ CLOSURE_CLASSES="$HERON_ROOT_DIR/bazel-bin/heron/storm/src/java/_javac/storm-com
 
 export CLASSPATH=$BIN_JARS:$GEN_JARS:$SCRIBE_JARS:$PROTO_JARS:$CLOSURE_CLASSES
 
-$JAVADOC $FLAGS -d $JAVADOC_OUTPUT_DIR $GEN_FILES $HERON_SRC_FILES $BACKTYPE_SRC_FILES $APACHE_SRC_FILES
-cp -rf $JAVADOC_OUTPUT_DIR $JAVADOC_STATIC_OUTPUT_DIR/api
+javadoc $FLAGS -d $JAVADOC_OUTPUT_DIR $GEN_FILES $HERON_SRC_FILES $BACKTYPE_SRC_FILES $APACHE_SRC_FILES
+
+ln -s $JAVADOC_OUTPUT_DIR $JAVADOC_STATIC_OUTPUT_DIR
 
 echo "Javdocs generated at $JAVADOC_OUTPUT_DIR"
 exit 0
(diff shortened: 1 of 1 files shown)


nlu90:
👍

objmagic:
@billonahill this is the temporary solution I said. what do you think?

billonahill:
If hugo is flaky couldn't one just view them as files in the browser? This is how I've typically viewed local javadocs.

```
open website/public/api/index.html
```

objmagic:
when you are on Heron introduction page and follow link leading you to API's javadoc, you'll get 404. So yes, the only way is to ``open website/public/api/index.html`` and manually find the doc of class you want to read.

lucperkins:
As someone who worked on docs about things like implementing custom
schedulers and metrics, it would've been immensely beneficial to be able to
move seamlessly between the locally running site and the Javadocs.

As for Hugo, it's not being flaky, it's just not built to do what we'd like
it to without a workaround. I'm going to continue to look for a better
solution but I think this is an approach that's both beneficial to doc
writers and has no real drawbacks.

billonahill:
The issue as I see it re hugo is that while many of us can browse from docs to javadocs without issue (for example you can go from http://localhost:1313/heron/docs/developers/data-model/ to http://localhost:1313/heron/api/com/twitter/heron/api/tuple/Tuple.html and integration is seamless), some people get 404s with the javadocs links and we don't know why. And then the issue mysteriously goes away. Let me know if I'm mis-assessing the issue here, but if that is our issue we should understand why we see that behavior.

It would be great to understand why hugo can load javadocs from website/static, but not website/public/api, but if that's not possible, could we instead symlink from the latter to the former instead of copy?

objmagic:
Also, I can go from Heron page to Java doc API page if ``static/api`` exists. However, when I ``rm -rf static/api``, and ``make serve`` again, I can still go to Java doc API page from Heron page. I can only imagine there is some caching mechanism going on inside hugo

lucperkins:
@objmagic You could be right that there's some caching mechanism in place. I'll dig into some Hugo internals to see what's going on there. What I know for sure is that removing the `static` dir and running `make site && make serve` will *not* bring about the desired result, and that's the result I'm working toward.

@billonahill I have no problem with symlinking or any other approach as long as I can clone, `cd website && make site && make serve` and be able to navigate between universes without issue. I'll explore this approach today and see if I can get it to go.

lucperkins:
@billonahill So I tried out a symlink approach and bizarrely enough it works just fine and yet the Hugo server throws `Symbolic links for directories not supported`, apparently because [Go's filesystem API is weird about symlinks](https://discuss.gohugo.io/t/symbolic-links-not-supported/590). See if you can reproduce using these steps:

```bash
$ cd website
$ rm -rf public static
$ make site
$ make serve
$ open http://localhost:1313/heron/api
```

objmagic:
@lucperkins I pulled and tested your branch. I can still open the symlinked javadoc API page using hugo (I can open ``http://localhost:1313/heron/api``). Hugo showed same errors though.

I am not sure if this is the case for other people. Either symlinking or ``cp -rf`` is fine to me, as long as one can go to API page from Heron page when developing on local machine.

objmagic:
close via #943

PR #803, opened 2016-05-26
Apache Mesos Native Scheduler

joestein:


DarinJ:
Joe,
Started looking at this, tried running it and noticed the executor kept crashing due to some double quotes. 

Run command is pasted below for reference.
```
Registered executor on ip-172-31-2-167.eu-central-1.compute.internal
Starting task task:container_1_2a30610b-168f-402c-add8-73a39be9c6fe:1464924583725:1336
Forked command at 6469
sh -c 'tar -xvf test17-ubuntu-tag-0-4204049127527475993 && rm test17-ubuntu-tag-0-4204049127527475993 heron-core-release.tar.gz && ./heron-core/bin/heron-executor 1 "test17" "test17af4487e3-bf2a-44a1-aec4-93b42d5c55d1" "test17.defn" "1:word:2:0:exclaim1:1:0" "172.31.11.31:2181" "/heron" "./heron-core/bin/heron-tmaster" "./heron-core/bin/heron-stmgr" "./heron-core/lib/metricsmgr/*" ""LVhYOitIZWFwRHVtcE9uT3V0T2ZNZW1vcnlFcnJvcg&equals;&equals;"" "heron-examples.jar" "31004" "31000" "31005" "./heron-conf/heron_internals.yaml" "exclaim1:536870912,word:536870912" """" "jar" "heron-examples.jar" "/usr/lib/jvm/java-8-oracle" "31003" "./heron-core/bin/heron-shell" "31002" "mesos" "ubuntu" "default" "./heron-core/lib/instance/*" "./heron-conf/metrics_sinks.yaml" "./heron-core/lib/scheduler/*:./heron-core/lib/packing/*:./heron-core/lib/statemgr/*" "31001"'
Command exited with status 2 (pid: 6469)
Shutting down
```

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,404 @@
...
+        "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\" "
+            + "\"%s\" \"%s\" \"%s\"",
```
DarinJ: Here was the issue.  s/\\"\"/gc and the heron-executor ran.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,404 @@
...
+  ) {
+    return getHeronExecutorArguments("" + port1, "" + port2,
+        "" + port3, "" + shellPort, "" + metricsmgrPort, "" + schedulerPort);
+  }
+
+  protected String formatJavaOpts(String javaOpts) {
+    String javaOptsBase64 = DatatypeConverter.printBase64Binary(
+        javaOpts.getBytes(Charset.forName("UTF-8")));
+
+    return String.format("\"%s\"", javaOptsBase64.replace("=", "&equals;"));
```
DarinJ: This is why it was double escaped.

DarinJ:
Figured that bit out, see inline comments.  I'm able to now run the exclamation topology.

kramasamy:
@DarinJ - thanks for giving it a spin. I was wondering if you had a chance to try heron tracker and heron ui. If not, please do and see if the topologies that you launched show up in UI.

DarinJ:
@kramasamy yes they appear, and I'm seeing tuples pass through.  I can also confirm the I can kill the topology and it is torn down properly in mesos.  The main thing I'm looking at now is the HdfsUploader, but that'll be another day.

kramasamy:
@DarinJ - Awesome. Let us know if you run into issues with HDFS uploader.

maosongfu:
@joestein 
Correct me if i'm wrong, It seems, the implementation launches the SchedulerMain(including MesosScheduler and Tmaster) locally, and then talk with Mesos Master to start other containers on Mesos Slave's machines?

aShevc:
@maosongfu
Think it would be fair enough for me to reply on this. Yes, MesosLauncher is pretty much based off Heron’s basic LocalLauncher, therefore it launches SchedulerMain process within itself. It has a parameter `heron.scheduler.background` to regulate whether to launch it in a default LocalLauncher fashion in a background, or to launch in a foreground. The big picture idea is that the users who use Mesos on their clusters will most probably want to use some high level Mesos scheduler to have Heron-Mesos scheduler itself under some sort of Mesos supervision. If we go back from where we started, Heron Mesos scheduler, that was initially provided from the Twitter side, had a high level Mesos scheduler for launching an arbitrary tasks, and it was used by MesosLauncher to submit topologies to. This didn’t look easy to setup and launch comparing to other Heron schedulers, so we gave up this approach in favor of preserving the simplicity of launching topologies via `heron submit` command. And yes, after launching HTTP server and Tmaster, it launches other containers on slaves.

maosongfu:
@aShevc 
Thanks for the reply. But I am wondering how to guarantee fault tolerance of Scheduler & TMaster? Let's say, what happen when Scheduler & TMaster fail?

aShevc:
@maosongfu 
Sure, on the fault tolerance side, there are two obvious updates that yet need to be done. First, yes, Tmaster should be definitely started as a job on a Mesos slave, like the other Heron containers. That way it would be simply supervised with the Heron Mesos scheduler. Second is a bit trickier and involves slight changes to Heron’s SubmitterMain, so would be nice to hear your thoughts on this. Currently Heron is not allowing to “pick up” the failed topology with the same `heron submit` command obviously. Like, if we would put the `heron submit` under the Mesos supervision as is, SubmitterMain will yell at us at this point: https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java#L440. What would be nice to do here is the following. Let's say, Heron would acknowledge through state manager that there is a Scheduler registered for the topology, and the topology is found in a running state. So if when it pings the scheduler’s address and it responds as being unreachable, it still launches the topology through the specified Launcher, but it also passes some config parameter to let know the Launcher that it should do a recover, and not start everything from scratch. Generally speaking, this solution will allow, should one want to start the scheduler under supervision, to set `heron submit` under any high level Mesos scheduler and get a failover tolerant design.
Also, sure, the submitting machine should have Mesos master and Zookeeper reachable. Mesos slaves are not required to be able to reach the Heron Mesos scheduler back, however commands like  `heron kill` should be performed from the machines that can reach it. The mentioned updates hopefully to ship soon.

tysonnorris:
Hi @aShevc - in this case (where submit can become a recover), who would invoke the heron submit recover? Couldn't that submitter could end up clobbering a running topology that did not need recovering? 

I'm just starting to look at MesosScheduler, and was expecting more of the previous design ("had a high level Mesos scheduler for launching an arbitrary tasks". I think the reason the other launchers do not have this design is that they depend on other frameworks that expose APIs for allowing some additional semantics around creating, monitoring, and recovering mesos(or slurm, etc) tasks. 

Do you have more info on how you plan to support starting the scheduler under supervision or running heron submit on a different mesos scheduler? 

Thanks!

mhausenblas:
This LGTM from my end @kramasamy and we'd be willing to invest further cycles to make this happen. Now hoping @joestein finds some cycles to resolve the conflicts and I'd be working with him (we had a quick sync today) to turn this also into a DC/OS package.

aShevc:
@tysonnorris, sure, your concern on not introducing high-level Mesos scheduler from our side is well understood. However, the willingness to somehow comply with the existing submitting semantics is not the only reason on moving away from introducing high-level Mesos scheduler from our side, it’s rather kind of a side effect. So, consider the case that would take place should we introduce a high level scheduler here, as the initial implementation suggests. The question is what would happen when this high level scheduler itself would fail? Should there be another Mesos scheduler layer to recover this scheduler? I think you should get the point by now. So obviously, at some point you should have a Mesos scheduler which features highly availability mode, therefore can run multiple scheduler copies on the cluster and having only one leader which actually is registered with Mesos. It goes without saying that leader election algorithm should be implemented here. There are also some suggestions on implementing such a feature in the Mesos docs just to get a full picture: http://mesos.apache.org/documentation/latest/high-availability-framework-guide/. So, in short, if we would use the initial implementation we were given, we wouldn’t succeed in creating failover ready solution because high availability mode wasn’t implemented there. Generally, of course, we could implement it in the Heron-Mesos, but, first, this is partly a reinventing the wheel, as there are proven solutions for this kind of problems, we could rely on, for example, Marathon and Singularity. Another thing is that it’s not only obviously a significant effort in terms of resources, but it’s also a part which in case of some implementation issues can make a real mess. I remember Marathon have had a bug some time ago, where the ZK data was in some rare cases messed up badly during the leadership election, and this caused harm not only to Marathon, but Mesos cluster itself as well. So, for us the way to go here was actually using Marathon, feeding it the “heron submit” command. As you may notice, we also use Zookeeper as a data storage for the scheduler, so when it is recovered, it just can pick up where things were left off. The only thing, probably worth fixing in the approach is that we should maybe be more explicit in pointing out to long-running processes schedulers like Marathon and Singularity in the Mesos scheduler docs.
Another concern you pointed out about Heron possibly trying to recover a running job was actually addressed in one of my previous comments. In short, yes, the logic on recovering is not yet implemented, the thing to note here is that Mesos launcher, if asked to recover, will try to do a healthcheck on the scheduler HTTP server, and will take actions on recovering only if the healthcheck fails.

kramasamy:
@joestein - would you have some bandwidth to resolve the conflicts?

kramasamy:
@DarinJ - just wondering if you had a chance to spin this up on a mesos cluster and see if everything works well?

DarinJ:
@kramasamy was able to get working with a http or shared disk, using a patch described above (some arguments were being double escaped with quotes and broke the executor command line on at least Ubuntu + mesos 28.1). 

I also had issues with the hdfsuploader, but that may be a user error and I didn't have time to investigate.  It seemed curl was being called to download the artifacts stored in hdfs.  I've got some cycles next week and could check it out again.  If someone's succeeded with the hdfsuploader I would love to hear about it.

joestein:
@kramasamy unfortunately I won't though I thought I would sorry @mhausenblas if you or @DarinJ or someone else can help getting this over the goal line maybe @aShevc could maybe too dunno but I won't be able to myself.

tysonnorris:
@aShevc Thanks for your response - can you elaborate on "the way to go here was actually using Marathon, feeding it the “heron submit” command"? On the surface it looks like the submit command returns immediately, so I'm not following what you expect marathon to do with it? Maybe this is what you are getting at missing in the Mesos scheduler docs.

aShevc:
@tysonnorris we actually made `heron submit` launch a Mesos scheduler's HTTP server in it's foreground, so in Mesos case, `heron submit` would be a long-running command one may use for injecting this to Marathon, Singularity, etc. And, I agree this is also a part to point out in the docs

maosongfu:
Close it due to replicated with #1067

PR #815, opened 2016-05-26
Add Facebook Open Graph and Twitter Cards metadata

lucperkins:
This PR adds some basic Open Graph and Twitter Cards metadata to each page. I make a distinction here between the index page and documents (i.e. material in `/docs`). Metadata that applies to all pages is templated in `layouts/partials/meta.ace` and supplied via the `params` in `config.yaml`; metadata that only applies to the index page is templated in `layouts/partials/index/index-meta.ace`; and metadata for documents is templated in `layouts/partials/docs/page-meta.ace`.

Issue #363, opened 2016-04-15
Document release process


kramasamy:
Document how we make an external release. This process includes a number of steps. One of them is to update the heron version in `website/config.yaml` before making the release git tag.

PR #827, opened 2016-05-27
Dynamically fetch the Heron version for website creation

lucperkins:


lucperkins:
@kramasamy Actually, hold off on shipping this. I'm realizing now that the script currently alphabetizes all of the `config.yaml` params, which we really don't want. I'll see if I can undo that behavior.

objmagic:
This PR looks very promising. It solves [the problem](https://github.com/twitter/heron/pull/815/files/f9ba523479345d807c5a02af84c19bee6743f748#r64969974) by @kramasamy. 
Thanks!

objmagic:
if I understand this PR correctly, one can put arbitrary version number in ``config.yaml`` and CI will update it, but CI is not able to commit the corrected ``config.yaml``. This means version number field is necessary and could be omitted on ``upstream``, but we cannot omit it locally since we need to build and view the website. This situation is awkward to me.

website/scripts/heron-version.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+
```
billonahill: Please provide some human-speak comment that gives an overview of what this script does.

lucperkins: @billonahill Done

billonahill:
Thanks for clarifying. Really we want to have the release in the docs align with the git tag. So if you checkout a git tag for a release, the docs are properly labeled. With this approach they won't be. Could we instead update the release process to include changing the version in `config.yaml`? Either automatically via a release script (ideal) or manually via documented process.

kramasamy:
@lucperkins  - if you can resolve the conflicts and we are ready to merge.

billonahill:
@kramasamy I don't think we should merge this as is. See my comments above.

lucperkins:
@billonahill How is this approach different from what you're proposing? This approach determines the current Heron version via Git tags and then updates the version value in the website config.

Whether this script is called during the broader Heron release process is up to others, but as for the `make site` process this seems to do precisely what you're calling for.

billonahill:
The release tag in git should already include an updated config.yaml file to have a matching version.

lucperkins:
@billonahill But the issue with that approach is that there's nothing to ensure that that happens. @kramasamy explicitly asked me to create a script that automates keeping the two in sync. To be honest, I don't really mind either way; I'm just taking cues from the project lead.

billonahill:
Confirmed with @kramasamy that we will update the site version as part of the release process before we create the release tag. To be documented as part of #363. Closing this out.

PR #828, opened 2016-05-28
Fix jQuery URL

lucperkins:
The current jQuery URL (as set in `website/config.yaml`, under `params:assets:js`) is HTTP when it should be HTTPS, which may cause issues in HTTPS only environments.

### DIFF
--- a/website/config.yaml
+++ b/website/config.yaml
@@ -34,7 +34,7 @@ params:
       - /css/font-awesome.min.css
       - /css/pygments.css
     js:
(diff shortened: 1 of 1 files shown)


lucperkins:
@objmagic Try a running version of the site with this change here: http://lucperkins.github.io/heron/docs/getting-started/

objmagic:
@lucperkins  Nice. It works fine for me now.

objmagic:
Thanks. Built and published!

PR #835, opened 2016-05-30
Fix sidebar accordion issue

lucperkins:
Previously, the sidebar accordion nav in the docs worked only inconsistently. I believe that I've found a much better solution here that also uses local storage to maintain state across refreshes.

A working example can be found here:

http://lucperkins.github.io/heron/docs/getting-started


saileshmittal:
@lucperkins This is better. But the sidebar still scrolls to the top after a click. Can it remain at its current position?

lucperkins:
@saileshmittal Could you explain your issue a little bit further? What exactly is happening?

saileshmittal:
If you scroll down to the last dropdown menu, and select a topic from there, the sidebar will jump back to the top. Although the dropdown would still be open, you would need to scroll down to go to the selected topic.

lucperkins:
@saileshmittal I added a `scrollTo` line in `sidebar.js` to address this issue. Give it a try at the same link above. It's not perfect but I think that it does help.

saileshmittal:
👍 
This is much better. There is still a small delay when it opens the dropdown though.

lucperkins:
@saileshmittal Unfortunately, there's not much that can be done about that.

saileshmittal:
@lucperkins okay, lgtm. shipit.

Issue #724, opened 2016-05-20
clean out the duplicated instance code from localmode


nlu90:
now we packaged everything in api and storm jars, we can remove the duplicated instance code in localmode

PR #843, opened 2016-05-31
remove duplicated code in simulator

nlu90:
#724


heron/simulator/src/java/BUILD
```diff
@@ -8,6 +8,7 @@ simulator_deps_files = \
     heron_java_proto_files() + [
         "//heron/api/src/java:api-java",
         "//heron/common/src/java:common-java",
+        "//heron/instance/src/java:instance-java",
```
maosongfu: This leads to the issue that storm-compatibility, (or anything with simulator), contains the whole heron-instance dependencies, which is not expected since it complicates the storm-compatibility's dependencies a lot.

nlu90: As a discussion result, we deferred this issue later. Will close the pr.

PR #857, opened 2016-06-03
Double jobs parallelism from 25 to 50

billonahill:


billonahill:
This made no difference. Closing.

Issue #853, opened 2016-06-02
Diagnose missing logs for heron-executor.stderr


maosongfu:
Sometimes heron-executor fails to redirect the stderr output to heron-executor.stderr.
#848

maosongfu:
#862

PR #862, opened 2016-06-04
Log down the stdout & stderr for processes spawned by heron-executor

maosongfu:
#853 

Currently, the stderr & stdout of processes spawned are missed. It makes hard to diagnose on processes failed to start. For instance, hard to figure out the lacking dependencies to start stream mgr or tmaster.


nlu90:
👍

PR #872, opened 2016-06-07
Fix broken Dockerfile code sample syntax

lucperkins:
At the moment, the `Dockerfile` syntax highlighting on [this page](http://twitter.github.io/heron/docs/developers/compiling/docker/) is broken because it was generated using an outdated version of [Pygments](http://pygments.org/), which is used by [Hugo](http://gohugo.io/) to generate those snippets. This PR changes the website setup script to include installation of Pygments 2.1.3, which supports the `Dockerfile` syntax.

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -2,7 +2,7 @@ setup:
 	@brew list hugo || brew install hugo
 	@npm install
 	@pip uninstall requests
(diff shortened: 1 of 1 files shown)


kramasamy:
ship it 👍

PR #878, opened 2016-06-07
Remove dependency on log4j-over-slf4j

### LINKED ISSUE #620: Dependency on log4j-over-slf4j can cause conflict at runtime
I am seeing a log4j conflict in Yarn scheduler. Heron packs log4j-over-slf4j. Yarn applications have log4j-over-slf4j bridge in the classpath. slf4j detects this loops and fails with the following error:
`SLF4J: Detected both log4j-over-slf4j.jar AND slf4j-log4j12.jar on the class path, preempting StackOverflowError.
SLF4J: Found binding in [jar:file:...filecache/10/reef-job-4089997145675586599.jar/global/heron-zookeeper-statemgr.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:...hadoop/lib/slf4j-log4j12-1.7.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Actual binding is of type [org.slf4j.impl.JDK14LoggerFactory]
`
I have a workaround for this problem. However I was wondering if heron needs to depend on log4j-over-slf4j?

ashvina:
This fixes #620. Heron does not use log4j api and hence does not need
log4j-over-slf4j. Packing this jar can result in conflicts in cases where
slf4j-log4j binding is present.

### DIFF
--- a/3rdparty/java/BUILD
+++ b/3rdparty/java/BUILD
@@ -69,7 +69,6 @@ java_library(
     deps = [
         "@slf4j-api//jar",
         "@slf4j-jdk//jar",
-        "@log4j-over-slf4j//jar",
     ],
 )
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -137,11 +137,6 @@ maven_jar(
 )
 
 maven_jar(
-  name = "log4j-over-slf4j",
-  artifact = "org.slf4j:log4j-over-slf4j:" + slf4j_version
-)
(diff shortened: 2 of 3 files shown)


maosongfu:
@nlu90 Can u review this and make sure it also works in Twitter Env?

nlu90:
👍

PR #884, opened 2016-06-08
Always log spawned process's stderr

maosongfu:
Currently only topology submission with flag "--verbose" will log the stdout && stderr
of a process spawned by heron. This pull request always logs the stderr even without "--verbose".

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/common/ShellUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/common/ShellUtils.java
@@ -73,14 +73,11 @@ public final class ShellUtils {
   public static int runSyncProcess(
       boolean verbose, boolean isInheritIO, String[] cmdline, StringBuilder stdout,
       StringBuilder stderr, File workingDirectory) {
-    // TODO(nbhagat): Update stdout and stderr
-
     StringBuilder pStdOut = stdout;
     StringBuilder pStdErr = stderr;
     try {
-      if (verbose) {
-        LOG.info("$> " + Arrays.toString(cmdline));
-      }
+      // Log the command for debugging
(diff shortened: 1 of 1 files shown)


billonahill:
@maosongfu is this only for processes on the scheduler, or those on the client? If the latter can you show what the console logs look like on the heron client with this change? Seems like this could potentially produce a lot of output.

heron/spi/src/java/com/twitter/heron/spi/common/ShellUtils.java
```diff
@@ -92,10 +89,12 @@ public static int runSyncProcess(
       }
       pStdOut.append(inputstreamToString(process.getInputStream()));
       pStdErr.append(inputstreamToString(process.getErrorStream()));
-      if (verbose) {
-        LOG.info(pStdOut.toString());
-        LOG.info(pStdErr.toString());
-      }
+
+      LOG.log(Level.FINE, "\tSTDOUT: {0}", pStdOut.toString());
```
billonahill: Seems like Level.FINE logging should only be done when a verbose flag is passed. If there's general info logs that we want to show, could we instead change those to Level.INFO, instead of logging all of Level.FINE?

maosongfu: We determine whether the logging level is Level.FINE or Level.INFO, basing whether a verbose flag is passed. 

By design, we don't want to show the stdout of a spawned process unless a verbose flag is passed, since:
1. We should provide more meaningful INFO in itself, rather than rely on the stdout of a spawned process.
2. Also it is more controllable, since we have not ideas what the stdout can be. (potentially very large as you mentioned)

But we want to show stederr no matter the verbose flag being passed or not, since it is more considered as a kind of error message. (Level.SERVER)

PR #891, opened 2016-06-09
Increase the minimal queue size for Communicator

maosongfu:
Currently we use a strategy tuning the soft capacity of Communicator,
which can trigger back-pressure too easily.
In this pull request, we increase the minimal queue size from 0 to 1,
enhancing the concurrent processing between Gateway and Slave thread.

End-to-end tested with LocalScheduler.


nlu90:
👍

PR #913, opened 2016-06-14
Added docs and script for integration tests

taishi8117:
- Added docs for running functional integration tests locally
- Added a simple shell script to run the functional integration tests locally

### DIFF
--- a/scripts/run_integration_test.sh
+++ b/scripts/run_integration_test.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+#
+# Script to locally run the functional integration test.
+# This script assumes that integration tests are already built under ./bazel-bin directory
+# ($ bazel build --config=PLATFORM integration-test/src/...)
+#
+
+./bazel-bin/integration-test/src/python/http_server/http-server 8080 &
+http_server_id=$!
+trap "kill -9 $http_server_id" SIGINT SIGTERM EXIT
+
+./bazel-bin/integration-test/src/python/test_runner/test-runner.pex \
+  -hc heron -tj bazel-genfiles/integration-test/src/java/integration-tests.jar \
+  -rh localhost -rp 8080\
+  -tp integration-test/src/java/com/twitter/heron/integration_test/topology/ \
+  -cl local -rl heron-staging -ev devel -pi 'file://${HERON_DIST}/heron-core.tar.gz'
--- a/scripts/run_integration_test.sh
+++ b/scripts/run_integration_test.sh
@@ -5,12 +5,14 @@
 # ($ bazel build --config=PLATFORM integration-test/src/...)
 #
 
-declare -a required_files=("./bazel-bin/integration-test/src/python/http_server/http-server" \
-                           "./bazel-bin/integration-test/src/python/test_runner/test-runner.pex");
+HTTP_SERVER="./bazel-bin/integration-test/src/python/http_server/http-server"
+TEST_RUNNER="./bazel-bin/integration-test/src/python/test_runner/test-runner.pex"
+
+declare -a required_files=(${HTTP_SERVER} ${TEST_RUNNER});
 
 for file in ${required_files[@]}; do
   if [ ! -f ${file} ]; then
-    echo "Required files do not exist." >&2
+    echo "Required file ${file} does not exist." >&2
     echo "Make sure that the integration tests are built prior to" >&2
     echo "running this script using the following command." >&2
     echo "  bazel build --config={PLATFORM} integration-test/src/..." >&2
@@ -18,11 +20,11 @@ for file in ${required_files[@]}; do
   fi
 done
 
-./bazel-bin/integration-test/src/python/http_server/http-server 8080 &
+${HTTP_SERVER} 8080 &
 http_server_id=$!
 trap "kill -9 $http_server_id" SIGINT SIGTERM EXIT
 
-./bazel-bin/integration-test/src/python/test_runner/test-runner.pex \
+${TEST_RUNNER} \
   -hc heron -tj bazel-genfiles/integration-test/src/java/integration-tests.jar \
   -rh localhost -rp 8080\
   -tp integration-test/src/java/com/twitter/heron/integration_test/topology/ \
--- a/scripts/run_integration_test.sh
+++ b/scripts/run_integration_test.sh
@@ -26,6 +26,7 @@ else
 fi
 
 # building tar packages
+bazel run --config=${PLATFORM} -- scripts/packages:heron-client-install.sh --user
 bazel build --config=${PLATFORM} {heron/...,scripts/packages:tarpkgs,integration-test/src/...}
 
 # run the simple http server
--- a/scripts/run_integration_test.sh
+++ b/scripts/run_integration_test.sh
@@ -5,6 +5,19 @@
 # ($ bazel build --config=PLATFORM integration-test/src/...)
 #
 
+declare -a required_files=("./bazel-bin/integration-test/src/python/http_server/http-server" \
+                           "./bazel-bin/integration-test/src/python/test_runner/test-runner.pex");
+
+for file in ${required_files[@]}; do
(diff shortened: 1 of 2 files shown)


website/content/docs/contributors/testing.md
```diff
@@ -49,17 +49,35 @@ $ bazel query 'kind("pex_test rule", ...)'
...
 
-To run the failure integration tests on your Mac OS X, do the following:
+    These integration tests are designed for testing the functionality of 
+    Heron, such as topologies and groupings.
+    To run the functional integration tests on your Mac OS X, do the following:
 
-```
-bazel run --config=darwin -- scripts/packages:heron-client-install.sh --user
+    ```bash
+    bazel build --config=dawrin integration-test/src/...
```
billonahill: darwin

website/content/docs/contributors/testing.md
```diff
@@ -49,17 +49,35 @@ $ bazel query 'kind("pex_test rule", ...)'
...
-python integration-test/src/python/local_test_runner/main.py
-```
+* Failure integration tests
+
+    These integration tests are designed for testing recovery from failure/restart
+    in certain processes, such as Topology Master and Metrics Manager.
+    To run the failure integration tests on your Mac OS X, do the following:
+
+    ```bash
+    bazel build --config=dawrin integration-test/src/...
```
billonahill: darwin

website/content/docs/contributors/testing.md
```diff
@@ -49,17 +49,35 @@ $ bazel query 'kind("pex_test rule", ...)'
 
 ### Running Integration Tests
 
-Integration tests are divided into two categories
+In order to run the integration tests on your Mac OS X, you need to first
```
billonahill: on a Mac OS X

Same for two other locations below.

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+# Script to locally run the functional integration test.
+# This script assumes that integration tests are already built under ./bazel-bin directory
```
billonahill: We should test that the required build artifacts exist. If they don't, let's log output explaining what bazel targets need to be run and exit 1.

taishi8117:
@billonahill Fixed.

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,29 @@
...
+# ($ bazel build --config=PLATFORM integration-test/src/...)
+#
+
+declare -a required_files=("./bazel-bin/integration-test/src/python/http_server/http-server" \
```
billonahill: We should define these as constants and reuse here and below, which guarantees that what we check for and what we use stay in sync. e.g.,
```
HTTP_SERVER="./bazel-bin/integration-test/src/python/http_server/http-server"
```

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,29 @@
...
+                           "./bazel-bin/integration-test/src/python/test_runner/test-runner.pex");
+
+for file in ${required_files[@]}; do
+  if [ ! -f ${file} ]; then
+    echo "Required files do not exist." >&2
```
billonahill: Required file $file does not exist.

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,31 @@
...
+
+${HTTP_SERVER} 8080 &
+http_server_id=$!
+trap "kill -9 $http_server_id" SIGINT SIGTERM EXIT
+
+${TEST_RUNNER} \
+  -hc heron -tj bazel-genfiles/integration-test/src/java/integration-tests.jar \
+  -rh localhost -rp 8080\
+  -tp integration-test/src/java/com/twitter/heron/integration_test/topology/ \
+  -cl local -rl heron-staging -ev devel -pi 'file://${HERON_DIST}/heron-core.tar.gz'
```
billonahill: where is HERON_DIST set and instead can the tar be located relatively, like we do with server and runner?

taishi8117: @billonahill It seems that ${HERON_DIST} is resolved internally at runtime, so no need to set this environment variable --- 'file://${HERON_DIST}/heron-core.tar.gz' is provided to `heron submit` command as a configuration argument "heron.package.core.uri", and com.twitter.heron.spi.common.Misc resolves it at runtime.

If the tar file is located at another location, it can be specified by modifying ${HERON_DIST} part. But I believe with default installation of heron-core, this script would work fine.

billonahill: Ideally, the runs a bazel build command, which produces `heron-core.tar.gz` somewhere and then runs this script, which knows how to find it at the that location, without requiring the user to know that location and set it as HERON_DIST.

Could we set it to a reasonable default and allow the ability to override it, like this:
```
HERON_DIST=${HERON_DIST:−some/default/path}
```

where `some/default/path` is the location under bazel-*/. that contains the dist?

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 1;
+  fi
+done
+
+${HTTP_SERVER} 8080 &
+http_server_id=$!
+trap "kill -9 $http_server_id" SIGINT SIGTERM EXIT
+
+${TEST_RUNNER} \
+  -hc heron -tj bazel-genfiles/integration-test/src/java/integration-tests.jar \
```
billonahill: This file is also generated by the build and should be included in `required_files`.

scripts/run_integration_test.sh
taishi8117: @billonahill I don't know if that's really necessary --- despite this format, HERON_DIST doesn't seem to be a shell environment variable (I didn't find any code that sets this env var, and also `${HERON_DIST}` needs to be surrounded by a single quote in the script), meaning that if some users want to override, they need to fully replace the `-pi` option with their customized location. Also, `heron-core.tar.gz` is by default produced under the `~/.heron/dist` directory when heron is installed, rather than when the `bazel build --config=PLATFORM integration-test/src/...` command was run. Users neither need to know the location of `heron-core.tar.gz

billonahill: I've got two concerns:
1. It's not clear who/what sets HERON_DIST and as you point out, you haven't found any code that does, so it's not clear to me why we use it in this script.
2. This test is misleading, because the user is instructed to run bazel to build the project and then run this script to test, but the test doesn't use what the user just built with bazel. Instead it tests against a pre-installed version of heron in ~/.heron. This behavior is not clearly conveyed to the user and seems misleading, since they're not testing the code they just built.

Let's discuss more in person if that's easier.

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,41 @@
...
+if [ $# -eq 1 ] && \
+   [[ "$1" == "darwin" || "$1" == "ubuntu" || "$1" == "centos5" ]];
+then
+  PLATFORM="$1"
+else
+  usage
+fi
+
+# building tar packages
+bazel build --config=${PLATFORM} heron/...
```
billonahill: does this work as a single command?
```
+bazel build --config=${PLATFORM} {heron/...,scripts/packages:tarpkgs,integration-test/src/...}
```

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+#
+# Script to locally run the functional integration test.
+#
+
```
billonahill: we probably want to also do `set -e` to fail fast upon internal bash errors

website/content/docs/contributors/testing.md
```diff
@@ -49,17 +49,35 @@ $ bazel query 'kind("pex_test rule", ...)'
 
 ### Running Integration Tests
 
-Integration tests are divided into two categories
+In order to run the integration tests on a Mac OS X, you need to first
```
billonahill: This is not longer true right?

billonahill:
👍

PR #916, opened 2016-06-14
Added Java-serialization integration test

taishi8117:
Added a functional integration test for custom class using Java-based serialization


integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/resources/SerializationTopology.json
```diff
@@ -0,0 +1 @@
+[10]
```
billonahill: I'm not that familiar with how this test framework works, but if we emit the following from the spouts:
```
new CustomObject("A", 10)
new CustomObject("B", 20)
```
and then do an increment bolt, why would we expect to get just [10]?

integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/CustomObject.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * An example Java-serializable custom object.
+ */
+public class CustomObject implements Serializable {
+
+  private static final long serialVersionUID = -7502839214902002412L;
+  private String name;
+  private Integer age;
+
+  public CustomObject(String name, Integer age) {
```
billonahill: if we're testing serialization are there other types besides String and Integer that we should verify?

integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/resources/SerializationTopology.json
taishi8117: This is because spouts used for integration tests (com.twitter.heron.integration_test.core.IntegrationTestSpout) executes 10 times as default.

integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/CustomObject.java
taishi8117: @billonahill Yes, I will add other types, also make sure serialization/deserialization works fine in the intermediate bolts too.

integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/resources/SerializationTopology.json
billonahill: So [10] is the number of records we expect? Seems we'd also want to assert the data itself if possible.

How is IntegrationTestSpout used here, I only see CustomSpout in use?

taishi8117: @billonahill I just pushed a commit that asserts the data itself in bolts.
Also, CustomSpout is wrapped by IntegrationTestSpout when TestTopologyBuilder creates a topology.

integration-test/src/java/com/twitter/heron/integration_test/topology/serialization/CustomCheckBolt.java
```diff
@@ -0,0 +1,48 @@
...
+  private CustomObject[] inputObjects;
+
+  public CustomCheckBolt(CustomObject[] inputObjects) {
+    this.nItems = 0;
+    this.inputObjects = inputObjects;
+  }
+
+  @Override
+  public void execute(Tuple input, BasicOutputCollector collector) {
+    System.out.println("Got: " + input.getValueByField("custom").toString());
```
billonahill: s/Got/Received input tuple/

billonahill:
Looks good. Once last comment left, fix it and ship it.

👍

Issue #920, opened 2016-06-15
version and timestamp in release.yaml is not loaded as String


nlu90:
A recent travis ci job failed with the following exception:
```bash
Exception in thread "main" java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
	at com.twitter.heron.spi.common.Config.getStringValue(Config.java:61)
	at com.twitter.heron.spi.common.Context.buildVersion(Context.java:43)
	at com.twitter.heron.scheduler.LaunchRunner.createExecutionState(LaunchRunner.java:56)
	at com.twitter.heron.scheduler.LaunchRunner.call(LaunchRunner.java:144)
	at com.twitter.heron.scheduler.SubmitterMain.callLauncherRunner(SubmitterMain.java:465)
	at com.twitter.heron.scheduler.SubmitterMain.submitTopology(SubmitterMain.java:418)
	at com.twitter.heron.scheduler.SubmitterMain.main(SubmitterMain.java:315)
```

It's due to the value of `heron.build.version` contains only numbers

```bash
See http://heronstreaming.io/docs/getting-started for how to use Heron.
heron.build.version : 5943249
heron.build.time : Wed Jun 15 00:56:13 UTC 2016
heron.build.timestamp : 1465952173000
heron.build.host : testing-gce-0bbbc5a5-20a6-4849-ae8c-ff20341cdc6c
heron.build.user : travis
heron.build.git.revision : 59432499872f8488df79ae976cb63e9f6ac8d8de
heron.build.git.status : Modified
```

Since `5943249` contains only numbers, yaml loads it as an Integer instead of a String. Then when it's used later as a string, the above exception happens. 

The `heron.build.timestamp` has the similar issue, yaml loads it as a Long instead of a String.

The solution is wrapping the values of `heron.build.version` and `heron.build.timestamp` with single or double quotes to force them to be String.

kramasamy:
@nlu90 - thanks for tracking this. can you fix and send a PR?

billonahill:
Should timestamp actually be a number?

PR #921, opened 2016-06-15
force version and timestamp to be string

nlu90:
#920 

Tested locally:
```
Heron is now installed!

Make sure you have "/Users/nlu/bin" in your path.

See http://heronstreaming.io/docs/getting-started for how to use Heron.

heron.build.version : 'neng/fix-version-type'
heron.build.time : Tue Jun 14 21:10:18 PDT 2016
heron.build.timestamp : '1465963818000'
heron.build.host : tw-mbp-nlu
heron.build.user : nlu
heron.build.git.revision : e53c0cf173799dd7802ac3f3bb4e16fad8cdfa04
heron.build.git.status : Modified
```

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/common/Context.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/common/Context.java
@@ -47,8 +47,8 @@ public class Context {
     return cfg.getStringValue(ConfigKeys.get("BUILD_TIME"));
   }
 
-  public static String buildTimeStamp(Config cfg) {
-    return cfg.getStringValue(ConfigKeys.get("BUILD_TIMESTAMP"));
+  public static Long buildTimeStamp(Config cfg) {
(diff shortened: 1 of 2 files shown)


billonahill:
Why would we not want timestamp to be numeric?

nlu90:
@billonahill In the `Context.java` code, the timestamp is fetched as follows:

```java
  public static String buildTimeStamp(Config cfg) {
    return cfg.getStringValue(ConfigKeys.get("BUILD_TIMESTAMP"));
  }
```

We either make it a string in yaml or fetch it as a long in the code. I'm fine with either solution, let me know your idea.

billonahill:
Seems like it should always be a long to me.

nlu90:
@billonahill Updated. Timestamp now is Long.

billonahill:
👍

PR #925, opened 2016-06-15
New website build

lucperkins:
cc @billonahill

See http://lucperkins.github.io/heron/ for my fork of the site


billonahill:
Thanks @lucperkins.

As a follow-up we should probably clean up the layout of this page (e.g., bullets and indents):
http://lucperkins.github.io/heron/docs/getting-started-troubleshooting/

PR #926, opened 2016-06-16
Rewrite RoundRobinPacking class to fix bugs

maosongfu:
Current RoundRobinPacking can not calculate the ram for each instance correctly,
when the container_ram_request is specified. Also, it lacks necessary unit tests to prove its
correctness.

This pull request rewrites the RoundRobinPacking and adds unit tests to make sure
it behaves as expected.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
@@ -18,7 +18,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import com.twitter.heron.api.generated.TopologyAPI;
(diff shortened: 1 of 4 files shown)


kramasamy:
👍

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -27,44 +26,263 @@
...
+ * - (ram used by instances within TOPOLOGY_COMPONENT_RAMMAP config))) /
+ * (the # of instances in container not specified in TOPOLOGY_COMPONENT_RAMMAP config)
+ * 7. The pack() return null if PackingPlan fails to pass the safe check, for instance,
+ * the size of ram for an instance is less than the minimal required value.
  */
 public class RoundRobinPacking implements IPacking {
   private static final Logger LOG = Logger.getLogger(RoundRobinPacking.class.getName());
-  private static final long DEFAULT_DISK_PADDING = 12 * Constants.GB;
+
+  // TODO(mfu): Read these values from Config
```
billonahill: please link to an issue in the comment

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -27,44 +26,263 @@
...
+  public static final long DEFAULT_DISK_PADDING_PER_CONTAINER = 12L * Constants.GB;
+  public static final long DEFAULT_RAM_PADDING_PER_CONTAINER = 2L * Constants.GB;
+  public static final double DEFAULT_CPU_PADDING_PER_CONTAINER = 1;
+  public static final long MIN_RAM_PER_INSTANCE = 192L * Constants.MB;
+
+  // Use as a stub as default number value when getting config value
+  public static final String NOT_SPECIFIED_NUMBER_VALUE = "-1";
```
billonahill: Could we use java java.util.Optional instead to indicate something hasn't been provided?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -27,44 +26,263 @@
...
 
   @Override
   public PackingPlan pack() {
-    Map<String, List<String>> containerInstancesMap = getBasePacking();
-    return fillInResource(containerInstancesMap);
+    // Get the instances' round-robin allocation
+    Map<String, List<String>> roundRobinAllocation = getRoundRobinAllocation();
+
+    // Get the ram map for every instance
+    Map<String, Map<String, Long>> instancesRamMap =
```
billonahill: Working with maps of Strings to Maps of Strings to Longs can get confusing and could lead to bugs. We should consider using more strongly typed objects, like a Container that has a Collection of Instances, which each has requiredRam. It would make the code much readable and have better compile time checks.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -96,212 +311,94 @@ public double getContainerCpuHint(Map<String, List<String>> packing) {
...
   /**
-   * Provide default ram for instances of component whose Ram requirement is not specified.
+   * Provide ram per container.
    *
-   * @return default ram in bytes.
+   * @param allocation packing
+   * @return Container ram requirement
    */
-  public long getDefaultInstanceRam(Map<String, List<String>> packing) {
+  public long getContainerRamHint(Map<String, List<String>> allocation) {
```
billonahill: is allocation used in this method?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: https://github.com/twitter/heron/issues/945

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: https://github.com/twitter/heron/issues/945

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: True. The PackingPlan (https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/common/PackingPlan.java) is exactly what you want: strongly typed objects, like a Container that has a Collection of Instances, which each has typed "Resource".

Currently PackingPlan is constructed by providing the actual value directly like:
```
public ContainerPlan(String id,
                         Map<String, InstancePlan> instances,
                         Resource resource) {
      this.id = id;
      this.instances = instances;
      this.resource = resource;
    }
```
so we need to come up with some intermediate variables for conversion, i.e. "maps of Strings to Maps of Strings to Longs" you mentioned. It can be a good idea to build PackingPlan via Builder patter, allowing people to edit it with typed safety.
Issue created here:
https://github.com/twitter/heron/issues/946

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: It was but not now. It is possible to be used in future. I can remove it if you want.

billonahill: If it's dead code without an immediate resurrection in view, I'd say we remove it.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
billonahill: Thanks. Let's add the issue into the comment in the code, here and elsewhere in this patch.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
billonahill: Yeah, looking at the code it seems we've got a lot of the building blocks. Builder pattern makes sense.

PR #927, opened 2016-06-16
Fix the incorrect ram map specification for heron-executor

maosongfu:
Currently when constructing the ram map for heron-executor in scheduler,
it just considers the component_ram_map config, i.e. ignores the container_ram_requested
config.

This pull request fixs this issue: it constructs the actual ram map from the actual packing plan,
which represents the processed value.


kramasamy:
👍

Issue #935, opened 2016-06-16
Provide heron-spi artifact in maven central


ajorgensen:
There is documentation referencing the use of heron-spi for building custom schedulers, packing algorithms, etc (http://twitter.github.io/heron/docs/contributors/custom-scheduler/) however it is not provided in [maven central](http://mvnrepository.com/artifact/com.twitter.heron). We should add this artifact so that it is available for people to easily include as a dependency.

lewiskan:
@kramasamy building standalone "heron-spi" with ```bazel build --config=darwin heron/spi/src/java:heron-spi```  generates "libheron-spi.jar" located in ```bazel-out/local_darwin-fastbuild/bin/heron/spi/src/java```  

Is "libheron-spi.jar" correct jar to publish, if so, do we rename to "heron-spi" to match documentation, or do we need a new BUILD genrule similar to "heron-api" 

```
genrule(
    name = "heron-api",
    srcs = [":api-unshaded_deploy.jar"],
    outs = ["heron-api.jar"],
    cmd  = "cp $< $@",
)
```

lewiskan:
@ajorgensen I see the PR https://github.com/twitter/heron/pull/937/files with genrule added, great, I'll wait for the merge then complete Maven Central publish, thanks.

ajorgensen:
I was thinking that we just make it look the same as heron-api and treat it the same way. Using the genrul in the PR i was able to generate the jar and upload it to a private artifactory instance and include it in a project.

lewiskan:
cool, my thoughts exactly!  thanks for generating the PR, i'll wait for merge before uploading to Maven.

ajorgensen:
Sounds great. Thanks!

PR #937, opened 2016-06-16
Export heron-spi jar

ajorgensen:
The heron-spi jar is needed for developing custom schedulers, packing
algorithms, etc. We will want to upload this jar to maven central along
with the api and storm compatibility layer.

Related to: https://github.com/twitter/heron/issues/935

### DIFF
--- a/heron/spi/src/java/BUILD
+++ b/heron/spi/src/java/BUILD
@@ -89,7 +89,7 @@ java_library(
         "//heron/common/src/java:basics-java",
         "//heron/common/src/java:config-java",
     ],
-    resources = common_resource_files, 
+    resources = common_resource_files,
 )
 
 java_library(
@@ -97,7 +97,7 @@ java_library(
     srcs = glob([
         "**/spi/utils/**/*.java",
     ]),
-    deps = utils_deps_files, 
+    deps = utils_deps_files,
 )
 
 java_library(
@@ -105,7 +105,7 @@ java_library(
     srcs = glob([
         "**/spi/packing/**/*.java",
     ]),
-    deps = packing_deps_files, 
+    deps = packing_deps_files,
 )
 
 java_library(
@@ -113,7 +113,7 @@ java_library(
     srcs = glob([
         "**/spi/scheduler/**/*.java",
     ]),
-    deps = scheduler_deps_files, 
+    deps = scheduler_deps_files,
 )
 
 java_library(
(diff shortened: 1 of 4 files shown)


heron/spi/src/java/BUILD
```diff
@@ -146,3 +146,19 @@ java_library(
...
+    ]),
+    deps = utils_deps_files,
+)
+
+genrule(
+    name = "heron-spi-jar",
+    srcs = [":spi-unshaded_deploy.jar"],
+    outs = ["heron-spi.jar"],
+    cmd  = "cp $< $@",
+)
```
kramasamy: This might not be enough - we might have to package this jar along with storm and api jars as well - so that we can publish to maven and goes into the install scripts.

ajorgensen: My thought is that this would live along side the api and storm jars and can be added as dependency the same way as those two. I think generally anything that coding against the spi interfaces will be expected to be run from the heron-cli or something like that right?

kramasamy: We need to get this jar into heron-api install scripts - so that when a user installs he/she gets everything.

ajorgensen: Ah sorry I misunderstood. I can look into adding it there as well.

On Friday, June 17, 2016, Karthik Ramasamy <notifications@github.com> wrote:

> In heron/spi/src/java/BUILD
> <https://github.com/twitter/heron/pull/937#discussion_r67466587>:
>
> > +
> > +java_binary(
> > +    name = "spi-unshaded",
> > +    srcs = glob([
> > +        "**/spi/utils/**/*.java",
> > +        "**/spi/common/**/*.java",
> > +    ]),
> > +    deps = utils_deps_files,
> > +)
> > +
> > +genrule(
> > +    name = "heron-spi-jar",
> > +    srcs = [":spi-unshaded_deploy.jar"],
> > +    outs = ["heron-spi.jar"],
> > +    cmd  = "cp $< $@",
> > +)
>
> We need to get this jar into heron-api install scripts - so that when a
> user installs he/she gets everything.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/twitter/heron/pull/937/files/b4087fad325028efed0c3a317bd786894a4f1bd2#r67466587>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAfHUyQ3gTVlLHq9HCXpDGZTr94GEPD_ks5qMkNXgaJpZM4I3zHj>
> .
>


-- 
----
Andrew Jorgensen
@ajorgensen

ajorgensen:
@kramasamy I added the spi jar to the heron-api-install.sh flow and tested it out locally. I added the spi jar to the list returned by `heron_api_lib_files`, let me know if thats not the appropriate place to put it. It seemed logical to group them together and as far as I could tell the only place that method was used was in the generation of `heron-api.tar.gz` and the install script.

kramasamy:
👍

kramasamy:
Alternatively, if you check in whatever you have, I add a separate package called 'spi' in the next iteration.

PR #869, opened 2016-06-06
#816 compile heron with bazel 0.2.3

caofangkun:
Since Bazel Release 0.1.5 (2016-02-05)	

"Repository rules must use names that are valid workspace names.", see: https://github.com/bazelbuild/bazel/issues/874

```
➜  heron git:(master) ✗ bazel version 
Build label: 0.2.3
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue May 17 14:21:13 2016 (1463494873)
Build timestamp: 1463494873
Build timestamp as int: 1463494873
```

PR #938, opened 2016-06-16
Java unit test name changes in BUILD files

taishi8117:
- Changed Java unit test names in BUILD files, for compatibility with later versions of Bazel.
- This is needed for moving to bazel 0.2.3


kramasamy:
👍

billonahill:
Is this because newer bazel not allow `_`? The rest of the targets in the repo seem to use a lowercased-dash-delimited convention instead of CamelCase. Could we also follow that here?

taishi8117:
@billonahill This is because the newer versions of Bazel require that for `java_test` fields, `name` needs to match with Java class name that is tested.
This PR should not conflict with the current master (built with bazel 0.1.2), and we are working on other necessary changes that might conflict with the current master in #869

billonahill:
👍 

(not so sure how I feel about bazel's design decision there :) )

Issue #915, opened 2016-06-14
undefined reference to symbol 'log@@GLIBC_2.2.5'


jiandongjia:
hi guys:

I use the CentOS system.
When I build the IDEA project, the error  as 
```
INFO: From Linking heron/common/tests/cpp/basics/fileutils_unittest:
/usr/bin/ld: bazel-out/local_linux-fastbuild/genfiles/3rdparty/gperftools/lib/libtcmalloc.a(libtcmalloc_internal_la-sampler.o): undefined reference to symbol 'log@@GLIBC_2.2.5'
/usr/bin/ld: note: 'log@@GLIBC_2.2.5' is defined in DSO /lib64/libm.so.6 so try adding it to the linker command line
/lib64/libm.so.6: could not read symbols: Invalid operation
collect2: error: ld returned 1 exit status
ERROR: /root/heron/heron/common/tests/cpp/basics/BUILD:3:1: Linking of rule '//heron/common/tests/cpp/basics:fileutils_unittest' failed: gcc failed: error executing command /usr/bin/gcc -o bazel-out/local_linux-fastbuild/bin/heron/common/tests/cpp/basics/fileutils_unittest -B/usr/bin/ -Wl,-z,relro,-z,now -no-canonical-prefixes -pass-exit-codes '-Wl,--build-id=md5' ... (remaining 3 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
INFO: Elapsed time: 50.939s, Critical Path: 41.63s.
WARNING!!! - bazel build failed - intellij setup may not be consistent
Path is /root/heron
Generating IDEA project...
scripts/get_all_heron_paths.sh: line 70: JAVA_PATHS: unbound variable
Done. IDEA module file: heron.iml

Could not locate IntelliJ IDEA. Manually open /root/heron as a project in IDEA.
```
I try to modify the config/configure.ac, add the following variables, but no effect.
```
#export LDFLAGS="$LDFLAGS -lm"
#LD="$LD -lm"
```

jiandongjia:
I modified get_all_heron_paths.sh file, add a parameter --config = centos, now works correctly.

![default](https://cloud.githubusercontent.com/assets/19700922/16071301/2feb9f68-330d-11e6-83e2-da2cd7e2d49d.jpg)

kramasamy:
When we build manually we provide the config=<platform> but in IDEA you need to query the platform and accordingly pass the platform. 

[...]

PR #941, opened 2016-06-17
#915 pass '--config=[os]' to bazel in get_all_heron_paths.sh

caofangkun:


kramasamy:
@billonahill - can you please check this as well?

scripts/get_all_heron_paths.sh
```diff
@@ -23,7 +23,19 @@ function query() {
 
 set +e
 # Build everything
-bazel build {heron,integration-test,tools/java}/...
+os_type=darwin
+case "$OSTYPE" in
```
billonahill: instead of requiring the user to pass this, could we instead try to auto-detect it using `uname`?

kramasamy: OSTYPE is a predefined environment variable. For mac

```
echo $OSTYPE
darwin14
```

for Centos
```
echo $OSTYPE
linux-gnu
```

scripts/get_all_heron_paths.sh
```diff
@@ -23,7 +23,19 @@ function query() {
...
+  darwin*)
+    os_type=darwin
+    ;;
+  linux*)
+    os_type=ubuntu
```
kramasamy: how could you distinguish between different flavors of linux? We support both centos and ubuntu.

scripts/get_all_heron_paths.sh
kramasamy: One issue with $OSTYPE might be the return value 'linux-gnu' for both Centos and Ubuntu and we cannot distinguish between the two.

scripts/get_all_heron_paths.sh
billonahill:
```
$ docker run -i -t heron-ubuntu14.04 /bin/bash
root@45c2337fdea9:/# ls -l /etc/*-release
-rw-r--r-- 1 root root 105 Feb 16 16:29 /etc/lsb-release
-rw-r--r-- 1 root root 249 Feb 16 16:43 /etc/os-release
root@45c2337fdea9:/# cat /etc/os-release
NAME="Ubuntu"
VERSION="14.04.4 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.4 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"

$ docker run -i -t heron-centos7 /bin/bash
[root@ea2d85dbf2af /]# cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

On our internal centos hosts though, `/etc/os-release` doesn't exist but `/etc/centos-release` does, so we'd need to also check for that. Distro-detection would be a good thing to put in it's own scripts, since I suspect it might be useful to others as well.
```
$ cat /etc/centos-release
Nest Enterprise Linux release 5.11 (Final) (Based on CentOS release 5.11)
```

kramasamy: +1 - we should make this as a generic script. @caofangkun - would appreciate if you could detect the OS and include it in get_all_heron_paths()

scripts/get_all_heron_paths.sh
```diff
@@ -23,7 +23,19 @@ function query() {
...
+    if [ -f /etc/redhat-release ] ; then
+      os_type=centos
```
caofangkun: @kramasamy If file ```/etc/redhat-release``` exits then It should be CentOS .

website/scripts/setup.sh
```diff
@@ -1,7 +1,28 @@
 #!/bin/bash
 
-brew list hugo || brew install hugo
-which wget || brew install wget
+source ../scripts/detect_os_type.sh
+
+PLAT_FORM=`platform`
```
kramasamy: @caofangkun - recommend using $PLATFORM rather than $PLAT_FORM - easier to read?

caofangkun: fixed.

caofangkun:
Recreated PR at #985 and  close this PR.

billonahill:
@caofangkun please don't close and re-open new PRs for small revisions, since it makes it harder to track things like discussions and deps across the multiple PRs with the same scope.

caofangkun:
Hi @billonahill 
sorry to have bothered you, it's my fault. I deleted the fork  several days ago, and the pull request came from an ```unknown repository```.
I don't know how to recover this , so I recreated new PRs.
 will never happen again.

billonahill:
No problem! I didn't realize the repo was messed up, so opening a new PR makes sense now. Some folks have a habit of closing and re-opening for minor changes, so I wanted to make sure that's not what was happening.

Issue #770, opened 2016-05-25
Linkchecker fails to check some links


billonahill:
Go to `website/content/docs/developers/data-model.md` and change `/api/com/twitter/heron/api/tuple/Tuple.html` to `/api/foobar/com/twitter/heron/api/tuple/Tuple.html` and run 'make site'.

Linkchecker passes. It doesn't test the link. WHen I add `-v` to the linchecker command and comment out `rm -f $OUT_FILE` from `website/scripts/linckchecher.sh` you see that the link is not checked because it appears external:

```
http://twitter.github.io/heron/api/foobar/com/twitter/heron/api/tuple/Tuple.html;
file:///Users/billg/ws/git/heron/website/public/docs/developers/data-model/index.html;;
filtered;;
The URL is outside of the domain filter, checked only syntax.;
True;
http://twitter.github.io/heron/api/foobar/com/twitter/heron/api/tuple/Tuple.html;
14;82;"Javadoc
```

Ideally we could map `twitter.github.io/heron` in linkchecker.

cc/ @lucperkins

objmagic:
close. #797 and #841 will fix this.

Issue #766, opened 2016-05-25
Re-enable fail on javadoc error


billonahill:
We had to disable fail-on-javadoc-error since it failed on travis although passing locally. Figure out why and re-enable it.

kramasamy:
@billonahill - this is completed I believe. Can we close this?

billonahill:
No, it's still disabled:
https://github.com/twitter/heron/blob/master/website/scripts/javadocs.sh#L2

objmagic:
I added ``set -e`` in commit https://github.com/twitter/heron/commit/7fb03ebc352cacfaac0ccaaae6006fccf0f3b449 and the CI worked fine. However, I failed to realize the error happens nondeterministically. The actual problem may be related to parallel building of protobuf Java files.

This means some PRs could be blocked by this problem. An immediate fix could be to remove ``set -e`` again.

kramasamy:
can you explain when the java docs are built? If you could introduce a bazel dependency it will be nice.

taishi8117:
Not using `tools/travis-ci/bazel.rc` for `javadoc.sh` might be the root cause of this issue -- that case, #968 should resolve this.

objmagic:
I believe this issue was fixed via #968. Feel free to close this.

PR #943, opened 2016-06-17
wget as linkchecker

objmagic:
This PR comes with the following changes:

1. Fix Java documentation character escaping errors introduced in https://github.com/twitter/heron/commit/9f8b0b04b2c9bd767b74c57451759aebbd28400e

2. Copy ``public/api`` to ``static/api`` so that the following link checking mechanism can work well. 
See related discussion at #798.

2. Use ``wget`` as link checker instead of 3rd party's ``linkchecker``. 
This change has two benefits.
    1. Less dependencies. No Python's ``request`` library is needed. 
        We can reuse ``wget`` which is [easily installed by Travis CI](https://github.com/twitter/heron/blob/master/.travis.yml#L16).
    2. Working better with static site generator ``hugo``. The current ``linkchecker`` does not fully understand the website generated by static site generator. It considers Java documentation pages as external links and even ``--check-extern`` flag is set, it fails to find the actual location of those documentation pages. This PR runs ``hugo`` in the background to serve the website and then runs ``wget`` to recursively check all links. This is almost the same as using ``wget`` directly on ``http://twitter.github.io/heron/``.

This PR addresses issues #766, #770, #797, #798.

/cc @billonahill @lucperkins


lucperkins:
Because `make linkcheck` is called whenever you make the site, users will get errors if they don't have `wget` installed. I think that `wget` installation should be added to `setup.sh`.

website/scripts/javadocs.sh
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
-#set -e TODO: figure out why this breaks CI and re-enable (https://github.com/twitter/heron/issues/766)
-
+set -e
```
billonahill: Looks like CI is still failing.

objmagic: yes. let me check again...

objmagic:
Thanks, @lucperkins! I updated script so that ``wget`` will be installed if there is no ``wget``. 

I'll also deliberately make some errors in Java documentation and break some links to see if script can catch it during CI

website/scripts/javadocs.sh
```diff
@@ -31,5 +31,7 @@ export CLASSPATH=$BIN_JARS:$GEN_JARS:$SCRIBE_JARS:$PROTO_JARS:$CLOSURE_CLASSES
 
 $JAVADOC $FLAGS -d $JAVADOC_OUTPUT_DIR $GEN_FILES $HERON_SRC_FILES $BACKTYPE_SRC_FILES $APACHE_SRC_FILES
 
+cp -r $JAVADOC_OUTPUT_DIR $JAVADOC_OUTPUT_LOCAL_DIR
```
billonahill: Let's add comments here explaining why we need two copies of the javadocs and what each should be used for.

website/scripts/javadocs.sh
objmagic: I'm thinking maybe it's because we should first build and then ``make site``? see [ci.sh](https://github.com/twitter/heron/blob/master/scripts/travis/ci.sh)

billonahill: There's someone going on with the way that docs generation script is trying to build with bazel here that is making the protobuf gen fail. I suspect we need to change the bazel command that we execute somehow.

objmagic:
@billonahill The CI passed now.

objmagic:
CI "successfully" exited after I introduced documentation syntax error. See [CI log](https://travis-ci.org/twitter/heron/builds/138483576). I will try breaking links next.

objmagic:
One downside of this PR is that there is no good way to limit the depth of checking if we want to check existence of external webpages.

Say we limit depth to 2 to prevent ``wget`` from getting sucked into infinite levels down when checking external webpages. This means that for local webpages, we also search only two levels down, but I believe if we want to do an **exhaustive** checking on local webpages, the actual depth has to be much larger than 2.

billonahill:
Seems better to exclude checking public links if it meant we could not specify a depth for internal links. Not sure if that's possible with wget.

objmagic:
Yes, currently external checking option is off. Also, after some study, I am pretty sure we cannot only limit depth for external checking.

objmagic:
CI script successfully displayed non-existing links and exited with 1. See the end of [CI log](https://travis-ci.org/twitter/heron/builds/138495013).

objmagic:
If everyone is happy with this PR, I'll reset head of this branch to https://github.com/twitter/heron/pull/943/commits/ad1bfce5449613b2b28d5408e125d089a0d47305 so that we can merge.

billonahill:
👍 

nice!

objmagic:
CI tests done, head reset to ad1bfce, ready for merge.

kramasamy:
Please go ahead and merge it.

website/scripts/javadocs.sh
objmagic: ah, I forgot this. Let me add comment and submit a new PR.

Issue #910, opened 2016-06-13
Set correct file timestamp


maosongfu:
Referenced: #845 
----------------------

Uncompressing..tar: ./release.yaml: implausibly old time stamp 1970-01-01 08:00:00
tar: ./bin/heron: implausibly old time stamp 1970-01-01 08:00:00
tar: ./bin: implausibly old time stamp 1970-01-01 08:00:00
tar: ./conf/local/client.yaml: implausibly old time stamp 1970-01-01 08:00:00
tar: ./conf/local/heron_internals.yaml: implausibly old time stamp 1970-01-01 08:00:00
tar: ./conf/local/metrics_sinks.yaml: implausibly old time stamp 1970-01-01 08:00:00
tar: ./conf/local/packing.yaml: implausibly old time stamp 1970-01-01 08:00:00

Though it is a WARNING, it is good if we can remove it.

maosongfu:
Travis-ci building also shows this WARNING during untar:
```
[2016-06-14 02:05:18 +0000] com.twitter.heron.spi.common.ShellUtils FINE:  $> [tar, -xvf, /home/travis/.herondata/topologies/local/heron-staging/1465869225_IntegrationTest_OneBoltMultiTasks_de8c043e-7d22-4ac4-ba17-902fbba5db81/topology.tar.gz]  
integration-tests.jar
1465869225_IntegrationTest_OneBoltMultiTasks_de8c043e-7d22-4ac4-ba17-902fbba5db81.defn
./heron-conf/
./heron-conf/heron_internals.yaml
tar: ./heron-conf/heron_internals.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/packing.yaml
tar: ./heron-conf/packing.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/uploader.yaml
tar: ./heron-conf/uploader.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/client.yaml
tar: ./heron-conf/client.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/scheduler.yaml
tar: ./heron-conf/scheduler.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/statemgr.yaml
tar: ./heron-conf/statemgr.yaml: implausibly old time stamp 1970-01-01 00:00:00
./heron-conf/metrics_sinks.yaml
[...]

PR #956, opened 2016-06-20
Add comments on why Java API doc should be copied.

objmagic:
See https://github.com/twitter/heron/pull/943#discussion_r67540560


objmagic:
need more time to study why adding comments can make CI fail

objmagic:
I simply restarted CI and CI now passed. I noted some difference between the past two CI logs. The first CI log displays some message saying "implausibly old time stamp 1970-01-01 08:00:00", but there is no such message in the second CI log. Maybe it's related to #910.

PR #957, opened 2016-06-20
Removed unused code; fixed platform code for py2.x; improved chmod

insomniacslk:
* real_path and discover_include_paths were unused
* the use of platform.uname() from my previous commit was working only with Python3. Python2.7 uses a regular tuple
* modified make_executable to be more readable and portable

### DIFF
--- a/bazel_configure.py
+++ b/bazel_configure.py
@@ -15,7 +15,7 @@
 import os
 import re
 import sys
-import shutil
+import stat
 import getpass
 import datetime
 import platform
@@ -42,7 +42,7 @@ def discover_user():
 # Discover the name of the host compiling
 ######################################################################
 def discover_host():
-  return platform.uname().node
+  return platform.node()
 
 ######################################################################
 # Get the time of the setup - does not change every time you compile
@@ -54,19 +54,19 @@ def discover_timestamp():
 # Get the processor the platform is running on
 ######################################################################
 def discover_processor():
-  return platform.uname().machine
+  return platform.machine()
 
 ######################################################################
 # Get the operating system of the platform
 ######################################################################
 def discover_os():
-  return platform.uname().system
+  return platform.system()
 
 ######################################################################
 # Get the operating system version
 ######################################################################
 def discover_os_version():
-  return platform.uname().release
+  return platform.release()
 
 ######################################################################
 # Get the git sha of the branch - you are working
@@ -123,12 +123,6 @@ def real_program_path(program_name):
     return os.path.realpath(which_path)
 
   return None
-
-######################################################################
-# Get the real path of the program
-######################################################################
-def real_path(apath):
-  return os.path.realpath(apath)
 
 def fail(message):
   print("\nFAILED:  %s" % message)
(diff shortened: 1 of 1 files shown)


maosongfu:
@nlu90 
Can u take a look at the why the local integration test case "KILL_TMASTER" failed? 
Also, I noticed that:
1. If "ERROR:root:Failed to get expected and actual results", the local integration script will not kill the test topology.
2. When invoking:
```
submitTopology(
        params['cliPath'], 
        params['cluster'],
        params['testJarPath'], 
        params['topologyClassPath'],
        params['topologyName'],
        params['readFile'],
        params['outputFile']
    )
```
the local integration script consider the submission fails only when it throws exceptions, without taking care the case return code is -1.

Both of them seem are kind of bugs.

insomniacslk:
The tests seem to be passing now, I guess it was an issue unrelated to my PR

maosongfu:
Yes, it was caused by some flaky tests and non-related to your pull request.

nlu90:
@maosongfu I relaunched the test yesterday and it passed. This is a flaky test due to python file open and read operations. It's not caused by this PR. And for your suggestion, we can refine the test in a separate pr

bazel_configure.py
```diff
@@ -42,7 +42,7 @@ def discover_user():
 # Discover the name of the host compiling
 ######################################################################
 def discover_host():
-  return platform.uname().node
+  return platform.uname()[1]
```
nlu90: call `platform.node()` instead

bazel_configure.py
```diff
@@ -54,19 +54,19 @@ def discover_timestamp():
 # Get the processor the platform is running on
 ######################################################################
 def discover_processor():
-  return platform.uname().machine
+  return platform.uname()[4]
```
nlu90: call `platform.machine()` instead

bazel_configure.py
```diff
@@ -54,19 +54,19 @@ def discover_timestamp():
...
 
 ######################################################################
 # Get the operating system of the platform
 ######################################################################
 def discover_os():
-  return platform.uname().system
+  return platform.uname()[0]
```
nlu90: call `platform.system()` instead

bazel_configure.py
```diff
@@ -54,19 +54,19 @@ def discover_timestamp():
...
 
 ######################################################################
 # Get the operating system version
 ######################################################################
 def discover_os_version():
-  return platform.uname().release
+  return platform.uname()[2]
```
nlu90: call `platform.release()` instead

bazel_configure.py
kramasamy: +1 - makes it more readable

bazel_configure.py
kramasamy: +1 - makes it more readable

bazel_configure.py
kramasamy: +1

bazel_configure.py
kramasamy: +1

insomniacslk: you're really doing +1 on all the notes from @nlu90? :D

insomniacslk: However good point, I'm replacing the calls as @nlu90 suggest, except that these are function calls, not attributes.

nlu90:
LGTM 👍

PR #960, opened 2016-06-21
Limit the maximum size of heron tuple set

maosongfu:
Currently heron can only limit # of tuples inside heron tuple set, but can not limit the size in bytes.
Sometimes misconfig can lead to constructing protobuf bigger than default limitation to parse (64M).

This pull request limits the maximum size of heron tuple set. It is a soft bound, calculating the size of
tuple set. And once the size exceeds the limitation, we will create a new heron tuple set.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
+++ b/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
@@ -131,6 +131,12 @@ public class SystemConfig {
       = "heron.instance.set.data.tuple.capacity";
 
   /**
+   * The maximum size in bytes of data tuple to batch in a HeronDataTupleSet protobuf
+   */
+  public static final String INSTANCE_SET_DATA_TUPLE_SIZE_BYTES
+      = "heron.instance.set.data.tuple.size.bytes";
+
+  /**
    * The size of packets read from stream manager will be determined by the minimal of
    * (a) time based (b) size based
    */
@@ -341,6 +347,11 @@ public class SystemConfig {
   public int getInstanceSetDataTupleCapacity() {
     return TypeUtils.getInteger(this.config.get(
         SystemConfig.INSTANCE_SET_DATA_TUPLE_CAPACITY));
+  }
+
+  public long getInstanceSetDataTupleSizeBytes() {
+    return TypeUtils.getInteger(this.config.get(
+        SystemConfig.INSTANCE_SET_DATA_TUPLE_SIZE_BYTES));
   }
 
   public int getInstanceSetControlTupleCapacity() {
--- a/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
+++ b/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
@@ -272,7 +272,7 @@ public class SystemConfig {
   /**
    * The size of packets read from socket will be determined by the minimal of:
    * (a) time based (b) size based
-   *
+   * <p>
    * Time based, the maximum batch time in ms for instance to read from socket per attempt
    */
   public static final String METRICSMGR_NETWORK_READ_BATCH_TIME_MS
@@ -287,7 +287,7 @@ public class SystemConfig {
   /**
    * The size of packets written to socket will be determined by the minimal of
    * (a) time based (b) size based
-   *
+   * <p>
    * Time based, the maximum batch time in ms to write to socket
    */
   public static final String METRICSMGR_NETWORK_WRITE_BATCH_TIME_MS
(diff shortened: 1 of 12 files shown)


maosongfu:
Waiting for the fix of flaky unit tests: #961

ashvina:
@maosongfu Pull request #967 fixes #961. Please take a look

heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
```diff
@@ -131,6 +131,12 @@
       = "heron.instance.set.data.tuple.capacity";
 
   /**
+   * The maximum size in bytes of data tuple to batch in a HeronDataTupleSet protobuf
+   */
+  public static final String INSTANCE_SET_DATA_TUPLE_SIZE_BYTES
+      = "heron.instance.set.data.tuple.size.bytes";
```
ajorgensen: Would the word `batch` in here help clarify at all? `heron.instance.set.data.tuple.batch.size.bytes` for instance?

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
```diff
@@ -51,7 +56,10 @@ public OutgoingTupleCollection(
     this.helper = helper;
     this.systemConfig =
         (SystemConfig) SingletonRegistry.INSTANCE.getSingleton(SystemConfig.HERON_SYSTEM_CONFIG);
+
+    // Read the config values
     this.dataTupleSetCapacity = systemConfig.getInstanceSetDataTupleCapacity();
+    this.maxDataTupleSizeInBytes = systemConfig.getInstanceSetDataTupleSizeBytes();
```
ajorgensen: If this value does not exist in heron_internal.yaml for example if someone has bundled an older version of heron_internals.yaml with their topology for overriding purposes what happens?

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
```diff
@@ -65,11 +73,13 @@ public void addDataTuple(
       long tupleSizeInBytes) {
     if (currentDataTuple == null
         || !currentDataTuple.getStream().getId().equals(streamId)
-        || currentDataTuple.getTuplesCount() > dataTupleSetCapacity) {
+        || currentDataTuple.getTuplesCount() > dataTupleSetCapacity
+        || currentDataTupleSizeInBytes > maxDataTupleSizeInBytes) {
```
ajorgensen: Should this be `>=`

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
```diff
@@ -41,8 +41,13 @@
   // Total data emitted in bytes for the entire life
   private long totalDataEmittedInBytes = 0;
 
-  private int dataTupleSetCapacity;
-  private int controlTupleSetCapacity;
+  // Current size in bytes for data types to pack into the HeronTupleSet
+  private long currentDataTupleSizeInBytes = 0;
+  // Maximum data tuple size in bytes we can put in one HeronTupleSet
+  private long maxDataTupleSizeInBytes = Long.MAX_VALUE;
```
ajorgensen: Wont these always get overwritten with the config values in the constructor? Why set them here to such a large value?

heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
maosongfu: ```data tuple set``` inherently means it is a batch since it is defined in proto as a batch of tuples with meta data.

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
maosongfu: Yes it throws a NPE. Will add a default value for it.

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
maosongfu: LOL. Good catch. Will change it.

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
maosongfu: Yes they are always get overwritten in the constructor. Will uniform the initialization in construct in the new commit.

heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
ajorgensen: awesome, thanks for the clarification!

heron/instance/src/java/com/twitter/heron/instance/OutgoingTupleCollection.java
ajorgensen: Sounds awesome!

ajorgensen:
This looks good to me. Thanks for the quick turn around!

PR #969, opened 2016-06-21
clean spi/common dir

nlu90:
1. Move NetworkUtils and ShellUtils into utils/ dir

2. Move PackingPlan into packing/ dir

3. Fix import and dependency problems introduced by the above two changes


heron/spi/src/java/BUILD
```diff
@@ -59,6 +59,7 @@ utils_deps_files = \
 scheduler_deps_files = \
     heron_java_proto_files() + [
         ":common-spi-java",
+        ":packing-spi-java",
```
maosongfu: Move this to packing_deps_files below?

heron/statemgrs/src/java/BUILD
```diff
@@ -6,6 +6,7 @@ common_deps_files = \
     heron_java_proto_files() + [
         "//heron/spi/src/java:common-spi-java",
         "//heron/spi/src/java:statemgr-spi-java",
+        "//heron/spi/src/java:utils-spi-java",
```
maosongfu: Let us not directly depend on utils-spi-java; it contains too many deep dependencies. 
Two options can be considered:
1. Add more specific building target in spi/src/java:BUILD
2. Re-org classes in utils, for instance, put SchedulerUtils into spi/src/java/scheduler or something similar.

heron/statemgrs/tests/java/BUILD
```diff
@@ -12,6 +12,7 @@ spi_deps_files = [
     "//heron/common/src/java:basics-java", 
     "//heron/spi/src/java:common-spi-java", 
     "//heron/spi/src/java:statemgr-spi-java",
+    "//heron/spi/src/java:utils-spi-java",
```
maosongfu: Let us not directly depend on utils-spi-java; it contains too many deep dependencies. 
Two options can be considered:
1. Add more specific building target in spi/src/java:BUILD
2. Re-org classes in utils, for instance, put SchedulerUtils into spi/src/java/scheduler or something similar.

heron/uploaders/src/java/BUILD
```diff
@@ -4,6 +4,7 @@ uploader_spi_files = [
     "//heron/common/src/java:basics-java",
     "//heron/spi/src/java:common-spi-java",
     "//heron/spi/src/java:uploader-spi-java",
+    "//heron/spi/src/java:utils-spi-java",
```
maosongfu: Let us not directly depend on utils-spi-java; it contains too many deep dependencies. 
Two options can be considered:
1. Add more specific building target in spi/src/java:BUILD
2. Re-org classes in utils, for instance, put SchedulerUtils into spi/src/java/scheduler or something similar.

heron/spi/src/java/BUILD
nlu90: The rule `scheduler-spi-java` depends on the PackingPlan.java class and the scheduler_deps_files is used to indicate the dependency for this rule.

nlu90:
@maosongfu can you have a look?

maosongfu:
looks good to me. 👍

PR #972, opened 2016-06-22
heron-explorer

objmagic:
``heron-explorer`` is a command-line tool, which has functionalities equivalent to ``heron-ui``'s.

#### Background

Heron UI is designed for users to allow them have a good understanding the status of their topologies. It provides a wide range of information in an organized way. However, we believe there is also need for a command-line tool that displays topology information on terminal. The CLI tool ``heron-explorer`` has the following advantages over ``heron-ui``:

* One can get desired information instantly on terminal, instead of opening web browser, clicking several buttons, and searching on the webpage

* Auto-complete feature via [Bash Programmable Completion](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html)

#### Design

The design is quite simple. 

``heron-explorer`` parses user input using Python’s ``argparse`` library. It reuses API call wrapper in ``heron.ui.src.python.handlers.access`` to talk to heron-tracker. It then converts API call result from JSON structure to tabulate and prints it on the terminal.

``heron-explorer`` assumes ``heron-tracker`` locates at ``127.0.0.1:8888``, but tracker base URL can be optionally set using ``--tracker_url``. 

**Note**: ``heron-tracker`` is extended so that its API supports ``role`` parameter. This means that we can make API call that queries topologies submitted only by specified user. It is extended in a way such that no function using its current APIs will break.

#### Functionalites

``heron-explorer`` currently supports three main categories of subcommands.

##### Display running topologies

* ``heron-explorer show cluster/[role]/[env]``  
The ``show`` subcommand displays running topologies. Cluster name is required, while role and environment name are optional. 

##### Display logical plan info of a particular topology

* ``heron-explorer bolts cluster/role/env/topology``  
The ``bolts`` subcommand displays bolts of a running topology. 

* ``heron-explorer spouts cluster/role/env/topology``  
The ``spouts`` subcommand displays spouts of a running topology. 

* ``heron-explorer components cluster/role/env/topology``
The ``components`` subcommand displays spouts and bolts of a running topology. 

##### Display physical plan info of a particular topology

* ``heron-explorer bolts-metric cluster/role/env/topology``
The ``bolts-metric`` subcommand displays metrics of bolts in all the physical containers of a topology.

* ``heron-explorer spouts-metric cluster/role/env/topology``
The ``spouts-metric`` subcommand displays metrics of spouts in all the physical containers of a topology.

* ``heron-explorer containers cluster/role/env/topology``
The ``containers`` subcommand shows all the physical containers of a running topology.


maosongfu:
Comments on the demo:

1. Remove the unncessary logs printed
2. Separate heron-explorer's dependencies on heron-ui
3. Some results can be printed in a better format, for instance, print the list of topologies running on a specific cluster in a table.

nlu90:
1. Please separate your changes for heron-tracker in another PR
2. fix the problem when a non-existing role provided the program throws exceptions, and checks if any other information missing will also cause the exception

objmagic:
Thanks. I'll fix all the problems, and submit two PRs separately.

objmagic:
close this because we are going to split this PR

PR #979, opened 2016-06-23
Tracker with role

objmagic:
1. Extend tracker APIs with ``role`` parameter so that tracker can provide APIs that queries topologies with role/submission_user matching specified ``role`` parameter

2. Move Tracker API call wrappers to ``heron/common`` since both ``heron-ui`` and ``heron-explorer`` have them as dependency.

Related to #972


objmagic:
Just pushed a commit that extends ``/topologies`` API JSON result with ``role``

heron/ui/src/python/BUILD
```diff
@@ -13,6 +13,9 @@ pex_library(
         "requests==2.3.0",
         "tornado==4.0.2",
     ],
+    deps = [
+        "//heron/common/src/python:common",
```
nlu90: You should create a new pex_library rule for these code moved from heron-ui and depend on that new rule only.

heron/common/src/python/handler/access/heron.py
```diff
@@ -122,36 +122,64 @@ def get_cluster_topologies(cluster):
...
 @tornado.gen.coroutine
-def get_execution_state(cluster, environ, topology):
+def get_execution_state(cluster, environ, topology, role=None):
   request_url = tornado.httputil.url_concat(
       create_url(EXECUTION_STATE_URL_FMT),
       dict(cluster = cluster, environ = environ, topology = topology)
   )
+  if role:
+    request_url = tornado.httputil.url_concat(request_url, dict(role=role))
+
```
nlu90:
```python
params = dict(cluster = cluster, environ = environ, topology = topology)
if role is not None:
  params['role'] = role
resuest_url = tornado.httputil.url_concat(create_url(EXECUTION_STATE_URL_FMT), params)
)
```

heron/tracker/src/python/handlers/containerfilehandler.py
```diff
@@ -33,11 +34,12 @@ def get(self):
       cluster = self.get_argument_cluster()
       environ = self.get_argument_environ()
       topology_name = self.get_argument_topology()
+      role = self.get_argument(constants.PARAM_ROLE, default=None)
```
nlu90: why other parameters have very specific method but not this one?

is there any method like `self.get_argumemt_role()`?

heron/tracker/src/python/handlers/containerfilehandler.py
```diff
@@ -60,6 +62,7 @@ class ContainerFileStatsHandler(BaseHandler):
   Parameters:
    - cluster - Name of cluster.
    - environ - Running environment.
+   - role - Name of person who submits the topology (optional)
```
nlu90: not limited to person, sometimes people may use their service account to submit a topology.

`role used to submit the topology (optional)` is good enough

heron/tracker/src/python/handlers/containerfilehandler.py
objmagic: understood. I can add ``self.get_argumemt_role()`` and use it

heron/tracker/src/python/handlers/containerfilehandler.py
```diff
@@ -77,9 +80,10 @@ def get(self):
       cluster = self.get_argument_cluster()
       environ = self.get_argument_environ()
       topology_name = self.get_argument_topology()
+      role = self.get_argument(constants.PARAM_ROLE, default=None)
```
nlu90: same question as before

heron/tracker/src/python/handlers/exceptionhandler.py
```diff
@@ -31,6 +31,7 @@ class ExceptionHandler(BaseHandler):
   Parameters:
    - cluster - Name of cluster.
    - environ - Running environment.
+   - role - Name of person who submits the topology (optional)
```
nlu90: same suggestion as before

heron/tracker/src/python/handlers/exceptionhandler.py
```diff
@@ -46,9 +47,10 @@ def get(self):
     try:
       cluster = self.get_argument_cluster()
       environ = self.get_argument_environ()
+      role = self.get_argument(constants.PARAM_ROLE, default=None)
```
nlu90: ditto

heron/tracker/src/python/handlers/exceptionsummaryhandler.py
```diff
@@ -31,6 +31,7 @@ class ExceptionSummaryHandler(BaseHandler):
   Parameters:
    - cluster - Name of cluster.
    - environ - Running environment.
+   - role - Name of person who submits the topology (optional)
```
nlu90: ditto

heron/tracker/src/python/handlers/metricshandler.py
```diff
@@ -29,6 +29,7 @@ class MetricsHandler(BaseHandler):
   URL - /topologies/metrics
   Parameters:
    - cluster (required)
+   - role - Name of person who submits the topology (optional)
```
nlu90: plese follow the comment format here.

`- parameter_name (required) explainations`

heron/common/src/python/handler/access/heron.py
objmagic: This was the first way of refactoring I tried. I believe this breaks consistency because these pieces of code use ``url_concat`` on optional parameters: https://github.com/twitter/heron/blob/master/heron/ui/src/python/handlers/access/heron.py#L324

kramasamy:
@objmagic - can you investigate the possibility of making role as a mandatory argument?

heron/common/src/python/handler/access/heron.py
cckellogg: We should not be afraid to break consistency if we think it makes the code better and we can refactor that file going forward. I like Neng's suggestion and your first thought when you did the refactor.

heron/common/src/python/handler/access/heron.py
```diff
@@ -122,36 +122,64 @@ def get_cluster_topologies(cluster):
   raise tornado.gen.Return((yield fetch_url_as_json(request_url)))
```
cckellogg: do we need a way to get topologies by cluster and environment too? Should we have an internal method that each of the public ones call?

_get_topologies(cluster, role = None, env = None)

heron/ui/src/python/handlers/api/metrics.py
```diff
@@ -18,7 +18,7 @@
 import json
 
 from heron.ui.src.python.handlers import base
-from heron.ui.src.python.handlers import access
+from heron.common.src.python.handler import access
```
cckellogg: Out of curiosity why are we changing the module naming.

heron/common/src/python/handler/access/heron.py
objmagic: I think this is a better design. I'll make the change.

heron/common/src/python/handler/access/heron.py
objmagic: agree. I'll make the change.

heron/ui/src/python/handlers/api/metrics.py
objmagic: It is because not only `heron/ui` depends on `handler`, `heron/explorer` will also depend on `handler`. We think it would be better if we move `handler` to `heron/common`. It might be confusing that `heron/explorer` depending on part of the code in `heron/ui`.

heron/tracker/src/python/handlers/containerfilehandler.py
objmagic: this is solved.

heron/tracker/src/python/handlers/metricshandler.py
objmagic: fixed via https://github.com/twitter/heron/pull/979/commits/c8d4d5038959bcfaff35598fb35a51045bd4341c

heron/tracker/src/python/handlers/containerfilehandler.py
objmagic: fixed via https://github.com/twitter/heron/pull/979/commits/757878ab582d3ed57eff02fa850939f006cbad81

heron/tracker/src/python/handlers/metricshandler.py
objmagic: solved.

heron/tracker/src/python/handlers/exceptionsummaryhandler.py
objmagic: solved.

heron/ui/src/python/BUILD
objmagic: fixed via https://github.com/twitter/heron/pull/979/commits/e134ab94e209ee73abbe784ce9b802af1e6c9721

objmagic:
all problems fixed. Please review again so that we can merge.

cckellogg:
👍

heron/common/src/python/handler/access/BUILD
```diff
@@ -0,0 +1,9 @@
+package(default_visibility = ["//visibility:public"])
+
+load("/tools/rules/pex_rules", "pex_library")
+
+pex_library(
+  name = 'tracker-access',
+  srcs = glob(['**/*.py']),
+  deps = [],
+)
```
nlu90: Instead of creating a new BUILD file, please add a new rule in the existing common/src/python/BUILD file.

Please refer the BUILD file for spi: https://github.com/twitter/heron/blob/master/heron/spi/src/java/BUILD#L163

heron/common/src/python/handler/access/heron.py
```diff
@@ -112,117 +114,161 @@ def get_topologies_states():
...
+
 
 ################################################################################
 # Get the execution state of a topology in a cluster
 ################################################################################
 @tornado.gen.coroutine
-def get_execution_state(cluster, environ, topology):
+def get_execution_state(cluster, environ, topology, role=None):
+  params = dict(cluster=cluster, environ=environ, topology=topology)
+  if role:
```
nlu90: `if role` and `if role is not None` are different, can you check and see which one is better here?

objmagic: thanks for the suggestion. fixed in https://github.com/twitter/heron/pull/979/commits/882ac3fbfe21c1f4e2dcc2b32ccd6dd4c7460543

heron/ui/src/python/BUILD
objmagic: thanks. fix via https://github.com/twitter/heron/pull/979/commits/fe75a271eac45a53195045663c9161319c264295

nlu90:
👍

PR #980, opened 2016-06-24
Apply override config on scheduler

maosongfu:
Currently any override config passed to heron-cli would apply on client side, i.e. SubmitterMain or RuntimeManagerMain only.
This pull request would apply the override config on server side, for instance, SchedulerMain, too.


kramasamy:
looks good!

kramasamy:
👍

Issue #963, opened 2016-06-21
Travis CI stdout too long log now, would be better capture the test results and only display necessary info 


caofangkun:
Ref: https://github.com/apache/storm/pull/539/

PR #982, opened 2016-06-24
#963 capture bazel build info into files, print info when return code is not zero

caofangkun:


nlu90:
👍

kramasamy:
@caofangkun - thanks for the contribution.

Issue #912, opened 2016-06-13
Upgrade the thrift version to latest open source one


maosongfu:
Currently heron uses a thrift 0.5.1 only available in at maven.twttr.com. We should upgrade it to the latest open source one available from Maven Central Repository.
Referenced: #832

nlu90:
Twitter's environment is based on old version of thrift, please make sure the upgrade doesn't break our internal users before merge it.

PR #986, opened 2016-06-24
#912 update to org.apache.thrift:libthrift:0.9.3

caofangkun:


maosongfu:
This will not work.
Currently the workflow is:
1. Use thrift executable in third_patry (https://github.com/twitter/heron/tree/master/third_party/thrift) to generate the java source code from thrift files. And those thrift executables are version 0.5.1
2. Then add thrift-java dependencies (which is changed in this pull request) to compile the scribe-sink.

So chaning the thrift-java dependencies alone will not solve the problem. This is traced by issue: #912 . You may re-assign the issue to yourself if you want.
@kramasamy made this workflow. He may have better ideas on it.

heron/metricsmgr/src/thrift/scribe.thrift
```diff
@@ -25,8 +25,8 @@ enum ResultCode
 
 struct LogEntry
 {
-  1:  string category,
-  2:  string message
+  1:  string category;
```
maosongfu: 1.
https://github.com/facebookarchive/scribe/blob/master/if/scribe.thrift is using "," rather than ";".

2. Good to keep consistent for all: enum ResultCode above is still using ","

caofangkun:
Hi @maosongfu 
Thank you so much for your clear and helpful explanation.
 I'll  look into this issue and try to find out how to solve it.

kramasamy:
@caofangkun - another option is to compile entire thrift 0.9.3 similar to protobuf. thrift compilation takes a long time time though. we will be better off with this approach, I think.

nlu90:
@kramasamy @maosongfu This change may break Twitter internal heron usage. Please do not merge it until we verify it also works with Twitter environment.

kramasamy:
@nlu90 - sounds good!

caofangkun:
HI @kramasamy 

* 1. compile and install thrift 0.9.3 and use `which thrift` eg ```/usr/local/bin/thrift``` to generate java code , 
but this consumes too much time on compiling thrift;
* 2.use thrift executables like ```third_party/thrift/thrift-linux-x86_64.exe``` to generate code at runtime , but this may  not proper  run on all linux-dist platforms;

So why don't we :
* 3. do not generate code at compile time , just use thrift 0.9.3 to generate java code locally, and use generated java code directlly?

caofangkun:
Hi @nlu90 
This PR is ready for reivew now.
I would be really grateful if you could spend some time to review.

billonahill:
@nlu90 when verifying we need to assert that we can build and deploy a topology in source, which still requires thrift 0.5.0. I suspect there could be difficulties there.

nicknezis:
We are closing old pull requests. Please let us know if the close is premature and which aspects of this pull request are still needed.

PR #989, opened 2016-06-24
heron-explorer

objmagic:
``heron-explorer`` is a command-line tool, which has functionalities equivalent to ``heron-ui``'s.

#### Background

Heron UI is designed for users to allow them have a good understanding the status of their topologies. It provides a wide range of information in an organized way. However, we believe there is also need for a command-line tool that displays topology information on terminal. The CLI tool ``heron-explorer`` has the following advantages over ``heron-ui``:

* One can get desired information instantly on terminal, instead of opening web browser, clicking several buttons, and searching on the webpage

* Auto-complete feature via [Bash Programmable Completion](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html) (which is possibly going to be implemented in the next cycle)

#### Design

The design is quite simple. 

``heron-explorer`` parses user input using Python’s ``argparse`` library. It reuses API call wrapper in ``heron.ui.src.python.handlers.access`` to talk to heron-tracker. It then converts API call result from JSON structure to tabulate and prints it on the terminal.

``heron-explorer`` assumes ``heron-tracker`` locates at ``127.0.0.1:8888``, but tracker base URL can be optionally set using ``--tracker_url``. 

As with ``heron``, ``heron-explorer`` will attempt to find the config for the cluster that user specifies. If the desired config file exists, command line arguments role and environment can be omitted.

**Note**: ``heron-tracker`` is extended so that its API supports ``role`` parameter. This means that we can make API call that queries topologies submitted only by specified user. It is extended in a way such that no function using its current APIs will break. See #979 for more details. 

#### Functionalites

``heron-explorer`` currently supports three main categories of subcommands.

##### Display existing clusters
* ``heron-explorer clusters``  
The ``clusters`` subcommand displays existing clusters.

##### Display running topologies

* ``heron-explorer topologies cluster/[role]/[env]``  
The ``show`` subcommand displays running topologies. Cluster name is required, while role and environment name are optional. 

##### Display logical plan info of a particular topology

* ``heron-explorer bolts cluster/[role]/[env] topology``  
The ``bolts`` subcommand displays bolts of a running topology. 

* ``heron-explorer spouts cluster/[role]/[env] topology``  
The ``spouts`` subcommand displays spouts of a running topology. 

* ``heron-explorer components cluster/[role]/[env] topology``  
The ``components`` subcommand displays spouts and bolts of a running topology. 

##### Display physical plan info of a particular topology

* ``heron-explorer metrics cluster/[role]/[env] topology [--component COMP]``  
The ``metrics`` subcommand displays metrics of components in all the physical containers of a topology. Component name can be optionally provided so that only metrics of selected component will be displayed.

Display physical plan info of a particular topology

* ``heron-explorer containers cluster/[role]/[env] topology [--cid ID]``  
The ``containers`` subcommand shows all the physical containers of a running topology. “--cid” can be optionally provided so that only information of selected container will be displayed.

Other commands

* ``heron-explorer help [subcommand]``  
The ``help`` subcommand displays usage of a subcommand.

* ``heron-explorer version``  
The ``version`` subcommand shows heron-explorer’s version.

Thanks for all the Heron team members who provided comments and suggestions.


objmagic:
* fixed lots of bugs, styles, etc...
* added tests using Mock
* added ``cluster`` command
* rebased against branched ``tracker-with-role``

objmagic:
rebased against https://github.com/twitter/heron/pull/979 again

objmagic:
fixed bugs, add ``version`` subcommand, various tiny improvements, and re-rebased against master.

heron/explorer/src/python/args.py
```diff
@@ -0,0 +1,207 @@
...
+    # but better save than sorry
+    for subparsers_action in subparsers_actions:
+      # get all subparsers and print help
+      for choice, subparser in subparsers_action.choices.items():
+        print("Subparser '{}'".format(choice))
+        print(subparser.format_help())
+        return
+
+
+class SubcommandHelpFormatter(argparse.RawDescriptionHelpFormatter):
```
kramasamy: This seems to be repeating code in main.py also?

heron/explorer/src/python/args.py
```diff
@@ -0,0 +1,207 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import heron.explorer.src.python.utils as utils
+import os
+
```
kramasamy: There is not much comment in the code. Can you add function level comments? and code comments where it is not intuitive.

heron/explorer/src/python/args.py
objmagic: fixed via https://github.com/twitter/heron/pull/989/commits/86963e51094e627dcd80710ceb9d1f38e140a5b8

heron/explorer/src/python/args.py
objmagic: resolved via https://github.com/twitter/heron/pull/989/commits/a8d40cdff47a85e9013328154ffda90d9b5ebd41

heron/explorer/src/python/utils.py
```diff
@@ -0,0 +1,423 @@
+# Copyright 2016 Twitter. All rights reserved.
```
objmagic: that's a good suggestion.

objmagic: moved ``utils.py`` to ``common/`` so that cli and explorer can share it. See commit https://github.com/twitter/heron/pull/989/commits/0d5896073f28f27a8336cdbe428d549aceb08c2f

objmagic:
change committed this morning is lost when rebasing..

kramasamy:
👍

PR #991, opened 2016-06-28
fix SpoutInstance OutQueueFullCount metric

nlu90:
The `isProduceTuple()` returns false in either of the following two cases:

1. the OutQueue is not available
2. the topology is not in RUNNING state

`OutQueueFullCount` metric should only be increased when the out queue is not available and unchanged when the topology is not in RUNNING state


nlu90:
@maosongfu Can you have a look at this PR?

maosongfu:
looks good to me. 👍

maosongfu:
@nlu90 Need we also change the one in simulator: https://github.com/twitter/heron/blob/master/heron/simulator/src/java/com/twitter/heron/simulator/instance/SpoutInstance.java#L181 ?

PR #997, opened 2016-06-28
fix build file

nlu90:
#937 exported the heron-spi jar, the location for heron-spi rule is wrong for scripts/centos5 BUILD file


billonahill:
👍

PR #1000, opened 2016-06-29
Remove ArrayList boxing

ajorgensen:
`Arrays.asList()` returns a private inner class implementation of ArrayList [see here](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/Arrays.java#3355). This class is not register with kryo by default and it is a tedious to
do so. The fallout of this is that anyone who uses the RawScheme will receive the following exception:

`java.lang.RuntimeException: java.lang.IllegalArgumentException: Class
is not registered: java.util.Arrays$ArrayList`


billonahill:
Thanks @ajorgensen, looks good!

Issue #1004, opened 2016-06-29
use Exception as ...


objmagic:
See: https://github.com/twitter/heron/blob/master/heron/cli/src/python/utils.py#L249

````Python
except e:
  raise Exception("Failed to parse override config: %s" % str(e))
````

should be be better changed to

````Python
except Exception as e:
  raise Exception("Failed to parse override config: %s" % str(e))
````

objmagic:
close via #1006

PR #1006, opened 2016-06-29
Fix undefined name, remove semicolon.

### LINKED ISSUE #1003: Undefined name ``argstr``
See: https://github.com/twitter/heron/blob/master/heron/cli/src/python/utils.py#L232

objmagic:
resolves #1003 #1004


billonahill:
👍

PR #1008, opened 2016-06-30
fix executor instance_cmd

nlu90:
There's one extra directory separator before bin/java in the current executor code. One example as follows:

`2016-06-28 23:01:36: Running container_6_kvs_sink_bolt_216 process as /usr/lib/jvm/java-1.8.0-twitter//bin/java -Xmx4672M -Xms4672M .... 31029 31806 heron_internals.yaml`

This PR fixes it and generates java path with exactly one directory separator along the path


kramasamy:
:+1:

billonahill:
👍

PR #1014, opened 2016-06-30
Make taskIndex never be negative

windie:
As `Math.abs(Integer.MIN_VALUE)` returns a negative value, taskIndex may be negative. This PR fixed it to avoid IndexOutOfBoundsException.


kramasamy:
@nlu90 - can you take a look at this?

nlu90:
👍

Issue #1010, opened 2016-06-30
Not successfully start the heron-executor because of some system dependent error


zhangzhonglai:
These days I'm trying to make heron running on cluster mode with yarn scheduler. After fixing several compiling error and platform dependent error. It can start successfully and without error in log. But It still has problem.

I track the log and see following: 
tmaster:
```
Jun 30, 2016 5:04:56 PM com.twitter.heron.scheduler.yarn.HeronExecutorTask call
INFO: Preparing evaluator for running executor-id: 0
Jun 30, 2016 5:04:56 PM com.twitter.heron.scheduler.yarn.HeronExecutorTask getExecutorCommand
INFO: Executor command line: [./heron-core/bin/heron-executor, 0, AckingTopology, AckingTopology8e27cd3f-b90e-4c18-a59e-32a6dd05da43, AckingTopology.defn, 1:exclaim1:2:1:word:4:1:word:3:0:exclaim1:1:0, 28x003.xd.com:2181,28x004.xd.com:2181,28x005.xd.com:2181, /heron, ./heron-core/bin/heron-tmaster, ./heron-core/bin/heron-stmgr, ./heron-core/lib/metricsmgr/*, "LVhYOitIZWFwRHVtcE9uT3V0T2ZNZW1vcnlFcnJvcg&equals;&equals;", heron-examples.jar, 46231, 42304, 50666, ./heron-conf/heron_internals.yaml, exclaim1:1073741824,word:1073741824, "", jar, heron-examples.jar, /usr/java/jdk1.7.0_67-cloudera, 39990, ./heron-core/bin/heron-shell, 59070, yarn, yarn, default, ./heron-core/lib/instance/*, ./heron-conf/metrics_sinks.yaml, ./heron-core/lib/scheduler/*:./heron-core/lib/packing/*:./heron-core/lib/statemgr/*, 34119]
// added by myself
Jun 30, 2016 5:04:56 PM com.twitter.heron.spi.utils.ShellUtils runASyncProcess
INFO: working dir: /data/9/yarn/nm/usercache/yarn/appcache/application_1466049056624_1522/container_e14_1466049056624_1522_01_000002/.
Jun 30, 2016 5:04:56 PM com.twitter.heron.scheduler.yarn.HeronExecutorTask call
INFO: Started heron executor-id: 0
```
stmgr:
```
Jun 30, 2016 5:04:59 PM com.twitter.heron.scheduler.yarn.HeronExecutorTask call
[...]

PR #1017, opened 2016-07-01
add some useful information when run async process

zhangzhonglai:
in runASyncProcess method:

- add a log for working directory which can help people locate the working directory to debug in some scheduler like yarn scheduler
- add a log file which can help people to find out what happened between pb.start() and the async process started (maybe some system error occurred)

These information will be very useful for people to solve the problem when deploying the multi-node heron cluster.

Extra information: #1010

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.StringTokenizer;
+import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -127,10 +128,20 @@ public final class ShellUtils {
     // Log the command for debugging
     LOG.log(Level.FINE, "$> {0}", Arrays.toString(command));
 
+    // the log file can help people to find out what happened between pb.start()
+    // and the async process started
+    String commandFileName = Paths.get(command[0]).getFileName().toString();
+    String uuid = UUID.randomUUID().toString().substring(0, 8);
+    String logFilePath = workingDirectory + "/" + commandFileName + "-" + uuid + "-started.stderr";
+    File logFile = new File(logFilePath);
+
     // For AsyncProcess, we will never inherit IO, since parent process will not
     // be guaranteed alive when children processing trying to flush to
     // parent processes's IO.
     ProcessBuilder pb = getProcessBuilder(false, command, workingDirectory, envs);
+    pb.redirectErrorStream(true);
+    pb.redirectOutput(ProcessBuilder.Redirect.appendTo(logFile));
+
     Process process = null;
     try {
       process = pb.start();
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
@@ -109,7 +109,12 @@ public class HeronExecutorTask implements Task {
 
     processTarget = Thread.currentThread();
 
-    final Process regularExecutor = ShellUtils.runASyncProcess(true, executorCmd, new File("."));
+    // Log the working directory, this will make people fast locate the
+    // directory to find the log files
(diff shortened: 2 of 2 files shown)


ashvina:
@zhangzhonglai Hi. I think #884 addresses a similar need. Could you please take a look.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -111,11 +112,24 @@ public static Process runASyncProcess(
       boolean verbose, String[] command, File workingDirectory) {
     // Log the command for debugging
     LOG.log(Level.FINE, "$> {0}", Arrays.toString(command));
+    // Log the working directory, this will make people fast locate the
+    // directory to find the log files
+    LOG.log(Level.INFO, "working dir: {0}", workingDirectory.getAbsolutePath());
```
maosongfu: We have already logged the location of working directory when a topology is submitted successfully via LocalScheduler: https://github.com/twitter/heron/blob/master/heron/schedulers/src/java/com/twitter/heron/scheduler/local/LocalLauncher.java#L83

Also, working directory is just a concept existing only in LocalScheduler, while ShellUtils is a common utils method used in a lot of places.

A better alternative can be log the working directory no matter a topology is submitted successfully or not at the start for LocalScheduler.

zhangzhonglai: @maosongfu Actually, this log is for people who use YARN scheduler, quickly and easily to locate the working directory. For now, only the yarn working directory is too difficult to find in executor machine. So I think this code can move to another class `HeronExecutorTask` which is only for yarn scheduler.
On the other hand, is there a way to easily adjust the log level. If not, then the log level is set to `Level.INFO` will be better.

maosongfu: @ashvina can provide better ideas when this logging should be.

Is there a way to easily adjust the log level.
People could submit topology with verbose flag, and then expect the logging level be Level.ALL (https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java#L275). 
Admittedly this is not fully done, for instance, not yet done in SchedulerMain or HeronExecutorTask. Pull request for them are welcome.

kramasamy: @ashvina - Do you have any comments on this?

ashvina:
Looks good to me. A user can reach initial app logs from RM UI and find out working directory from that. Capturing Java process logs in a file seems reasonable too.
:+1:

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -111,11 +112,24 @@ public static Process runASyncProcess(
...
+    String uuid = UUID.randomUUID().toString().substring(0, 8);
+    String logFilePath = workingDirectory + "/" + commandFileName + "-" + uuid + "-started.stderr";
+    File logFile = new File(logFilePath);
 
     // For AsyncProcess, we will never inherit IO, since parent process will not
     // be guaranteed alive when children processing trying to flush to
     // parent processes's IO.
     ProcessBuilder pb = getProcessBuilder(false, command, workingDirectory);
+    pb.redirectErrorStream(true);
+    pb.redirectOutput(ProcessBuilder.Redirect.appendTo(logFile));
```
zhangzhonglai: Hi @ashvina.
I find a simple way to do this, as long as we capture the output to the yarn's logging system and the log file is no longer necessary.
```
LOG.log(Level.FINE, "\tSTDOUT: {0}", inputstreamToString(process.getInputStream()));
LOG.log(Level.SEVERE, "\tSTDERR: {0}", inputstreamToString(process.getErrorStream()));
```
This works fine too. What do you think? If you agree with me, I'll put a new commit.

kramasamy:
@ashvina - with the new commit @zhangzhonglai pushed - can you check?

ashvina:
👍

kramasamy:
@zhangzhonglai - can you update the PR since there are some conflicts? Then I can merge it.

zhangzhonglai:
@kramasamy Done.

PR #1021, opened 2016-07-01
Fix corner cases for FileSink.flush

windie:
There are two corner cases for `flush`:

- No record has been processed before the first `flush`: it will output "]" to the stdout and close it.
- No record has been processed between two `flush`s: it will write "]" to a closed writer. Currently it's silent as PrintStream always swallows IOException.

This PR fixed `flush` by creating a new file for these corner cases.

### DIFF
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/FileSink.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/FileSink.java
@@ -55,7 +55,7 @@ public class FileSink implements IMetricsSink {
 
   // We would convert a file's metrics into a JSON object, i.e. array
   // So we need to add "[" at the start and "]" at the end
-  private static boolean isFileStart = true;
+  private boolean isFileStart = true;
   private PrintStream writer;
   private String filenamePrefix;
   private int fileMaximum = 1;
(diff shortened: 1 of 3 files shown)


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/FileSink.java
```diff
@@ -64,25 +64,20 @@
 
   @Override
   public void init(Map<String, Object> conf, SinkContext context) {
-    filenamePrefix = (String) conf.get(FILENAME_KEY) + "." + context.getMetricsMgrId();
+    filenamePrefix = conf.get(FILENAME_KEY) + "." + context.getMetricsMgrId();
```
maosongfu: Can we check and throw exceptions here if config does not contain "filename-output" or "file-maximum"? Or provide some defaults values?

windie: What's the contract for conf? I saw most of places don't check the existence. E.g.,  https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java#L468

maosongfu: Yes those places should have been checked.

maosongfu:
Can a also add a unit test to verify the corner cases trying to solve in https://github.com/twitter/heron/tree/master/heron/metricsmgr/tests/java/com/twitter/heron/metricsmgr ?

windie:
Updated. PTAL.

maosongfu:
Thanks for the contribution. Looks good to me. :+1:

PR #1029, opened 2016-07-06
Add API doc link, fix broken link.

objmagic:
* Fix broken links introduced by https://github.com/qiuyij/heron/commit/83a50a70c73a6d7e71f46ffbab3c504dec8a0f24.  

  Currently there are two broken links on http://twitter.github.io/heron/docs/getting-started/  
image

 The reason link checker does not catch these is that they are external links.
  **The correct way of adding images to website should be putting images under `website/assets/img` instead of linking an external link.**

* Do not format ``127.0.0.1:8888`` as link  
  Since this will make link checker fail. And it is not always true that when people are reading the website, the tracker is running on their machine, which means there is no point to put a link to ``127.0.0.1:8888`` there.

* Add link to API doc
image

Change of website after this PR can be previewed at: https://objmagic.ml/heron/


website/content/docs/getting-started.md
```diff
@@ -156,9 +156,9 @@ $ heron-tracker
 ... Using config file: $HOME/.herontools/conf/heron_tracker.yaml
 ```
 
-You can reach Heron Tracker in your browser at [http://localhost:8888](http://localhost:8888)
-and see something like the following upon successful submission of the topology: 
-![alt tag](http://twitter.github.io/heron/img/heron-tracker.png)
+You can reach Heron Tracker in your browser at ``http://localhost:8888``
```
billonahill: how come you removed the localhost link here but keep it below.. it's convenient to be able to click that when getting started.

objmagic: yes, but link checker will complain it is invalid.

kramasamy: Adding API docs - along side 'Get Started' is not a good idea. Instead, it might be better if you could add it at the top next to 'Docs'. Instead of calling it Docs, call it just an 'API'

objmagic: ok. this is better.

billonahill: can we instead patch linkchecker to ignore localhost URIs?

objmagic: I'll try.

objmagic: also I need to figure out why link checker did not complain ``http://localhost:8888`` during CI.

objmagic:
@billonahill 

I patched the script. See: https://github.com/twitter/heron/pull/1029/commits/fd832aca4bd61183a5c7b57e35c1c25bf2a9ed09

@kramasamy 

I updated the website design as you suggested. Please see: https://objmagic.ml/heron/

website/scripts/linkchecker.sh
```diff
@@ -31,13 +32,40 @@ GREP_STATUS=$?
...
-  BROKEN_MSG="broken link\."
+  BROKEN_MSG="^Found.*broken link"
   BROKEN_MSG_LINE=$(grep -n "${BROKEN_MSG}" ${WGET_LOG})
   LN=$(echo "${BROKEN_MSG_LINE}" | cut -f1 -d:)
-  tail "+${LN}" "$WGET_LOG"
-  exit 1
+  LINES=$(tail "+${LN}" "${WGET_LOG}")
+  BAD_LINKS=""
+  COUNT=0
+  # filter out broken links with prefix ``localhost:15532``
```
billonahill: this seems like it would filter valid links to pages we want to check. Don't we want to filter out links to the tracker and heron ui on localhost?

objmagic: sorry, my word usage is wrong! The code below has the right logic.

objmagic: fixed in https://github.com/twitter/heron/pull/1029/commits/464d811fe90fe26878e024a44ce9bac26a8ad68a!

kramasamy:
looks good to me. @billonahill - does this PR looks ok?

billonahill:
👍

PR #1033, opened 2016-07-06
Fix org/apache/storm/clojure/lang/Atom.java's package name

windie:


maosongfu:
@nlu90 Can u take a look at this?

kramasamy:
@windie - Is this for Storm 1.0.0+?

windie:
@kramasamy - I just found the package name doesn't match the file path. However, I don't understand the hack here. @nlu90 could you take a look since you added this file?

nlu90:
👍  

@windie This seems a typo when we added the new package namespace for storm compatibility

PR #1037, opened 2016-07-06
Update Heron version to 0.14.1 in the website's config

lucperkins:
Just one teensy change that I must unfortunately include in its own PR


billonahill:
@kramasamy we should also update the release process to include making this change before tagging the release.

PR #1038, opened 2016-07-06
Website build for Heron 0.14.1

lucperkins:
See http://lucperkins.github.io/heron/ for a running version


kramasamy:
looks good to me!

billonahill:
Looks good, will merge. Totally unrelated, but I think it's too hard for users to find the mailing list currently. I think it's only hidden in the nav bar. As a result we get tons of questions as github issues and very little mailing list traffic. We should more prominently steer people to the list for help somewhere.

PR #1040, opened 2016-07-07
Fix HeronMasterDriverTest.createsContextConfigForExecutorId

windie:
`Assert.assertEquals(4, config.getNamedParameter(namedParameterNode));` compares the int `4` with a String, but it doesn't fail because this line has never been called.

This PR fixes this issue and also makes sure the test will fail if the above assertion is not called.


kramasamy:
@ashvina - can you take a look at this?

ashvina:
Good catch. Looks good to me.
:+1:

PR #1042, opened 2016-07-07
Update aws dependency version

ajorgensen:
This is to address a 400 Bad Request exception coming from the s3 client
in some cases.

This closes: https://github.com/twitter/heron/issues/1035


kramasamy:
@ajorgensen - looks like the unit tests failed.

ajorgensen:
@kramasamy yeah it looks like they updated their dependency on http_core, I updated the PR with that change, tests should pass now.

PR #1046, opened 2016-07-07
Update version numbers to 0.14.1

objmagic:
would be good to use ``grep`` to check version number every time when one wants to update website before releasing.

### DIFF
--- a/website/content/docs/getting-started.md
+++ b/website/content/docs/getting-started.md
@@ -66,7 +66,7 @@ To check Heron is successfully installed, run:
 
 ```bash
 $ heron version
-heron.build.version : 0.14.0
+heron.build.version : 0.14.1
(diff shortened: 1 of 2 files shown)


caofangkun:
Hi @objmagic 

does ```release/release-process.md ```  should also change ```0.14.0``` to ```0.14.1```?

and should we provide ```update-version.sh {version_number}``` shell script to update version?

objmagic:
@caofangkun 

Hi, thanks for the review. 

> does release/release-process.md should also change 0.14.0 to 0.14.1?

I don't think we need to because we are talking about an example in ``release/release-process.md``, which is independent from the version number we are using.

> and should we provide update-version.sh {version_number} shell script to update version?

see https://github.com/twitter/heron/pull/827

PR #1055, opened 2016-07-09
Add extension when generating filename

maosongfu:
Currently UploaderUtils.generateFilename() returns filename without extension.
This can lead to file-type mis-parsing, for instance, MesosFetcher.

This pull request adds filename extension.


kramasamy:
Looks good to me!

PR #1059, opened 2016-07-11
Ffdpacking

### LINKED ISSUE #934: Add a new packing algorithm based on binpacking

avflor:
fixes #934 

The branch contains a new packing algorithm based on binpacking that can support heterogeneous containers and can be used with schedulers like YARN.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -37,13 +37,10 @@ import com.twitter.heron.common.basics.TypeUtils;
  * Spouts.
  */
 public class Config extends HashMap<String, Object> {
-  private static final long serialVersionUID = 2550967708478837032L;
-
   /**
    * Topology-specific options for the worker child process. This is used in addition to WORKER_CHILDOPTS.
    */
   public static final String TOPOLOGY_WORKER_CHILDOPTS = "topology.worker.childopts";
-
   /**
    * Per component jvm options.  The format of this flag is something like
    * spout0:jvmopt_for_spout0,spout1:jvmopt_for_spout1. Mostly should be used
@@ -52,30 +49,25 @@ public class Config extends HashMap<String, Object> {
    * all components, this is per component
    */
   public static final String TOPOLOGY_COMPONENT_JVMOPTS = "topology.component.jvmopts";
-
   /**
    * How often a tick tuple from the "__system" component and "__tick" stream should be sent
    * to tasks. Meant to be used as a component-specific configuration.
    */
   public static final String TOPOLOGY_TICK_TUPLE_FREQ_SECS = "topology.tick.tuple.freq.secs";
-
   /**
    * True if Heron should timeout messages or not. Defaults to true. This is meant to be used
    * in unit tests to prevent tuples from being accidentally timed out during the test.
    */
   public static final String TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS = "topology.enable.message.timeouts";
-
   /**
    * When set to true, Heron will log every message that's emitted.
    */
   public static final String TOPOLOGY_DEBUG = "topology.debug";
-
   /**
    * The number of stmgr instances that should spin up to service this
    * topology. All the executors will be evenly shared by these stmgrs.
    */
   public static final String TOPOLOGY_STMGRS = "topology.stmgrs";
-
   /**
    * The maximum amount of time given to the topology to fully process a message
    * emitted by a spout. If the message is not acked within this time frame, Heron
@@ -83,13 +75,11 @@ public class Config extends HashMap<String, Object> {
    * the message at a later time.
    */
   public static final String TOPOLOGY_MESSAGE_TIMEOUT_SECS = "topology.message.timeout.secs";
-
   /**
    * The per componentparallelism for a component in this topology.
    * Note:- If you are changing this, please change the utils.h as well
    */
   public static final String TOPOLOGY_COMPONENT_PARALLELISM = "topology.component.parallelism";
-
   /**
    * The maximum number of tuples that can be pending on a spout task at any given time.
    * This config applies to individual tasks, not to spouts or topologies as a whole.
@@ -99,19 +89,16 @@ public class Config extends HashMap<String, Object> {
    * their tuples with a message id.
    */
   public static final String TOPOLOGY_MAX_SPOUT_PENDING = "topology.max.spout.pending";
-
   /**
    * A list of task hooks that are automatically added to every spout and bolt in the topology. An example
    * of when you'd do this is to add a hook that integrates with your internal
    * monitoring system. These hooks are instantiated using the zero-arg constructor.
    */
   public static final String TOPOLOGY_AUTO_TASK_HOOKS = "topology.auto.task.hooks";
-
   /**
    * The serialization class that is used to serialize/deserialize tuples
    */
   public static final String TOPOLOGY_SERIALIZER_CLASSNAME = "topology.serializer.classname";
-
   /**
    * How many executors to spawn for ackers.
    * <p>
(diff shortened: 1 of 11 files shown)


objmagic:
Thx. Please sign the [CLA](https://engineering.twitter.com/opensource/cla) first.

avflor:
I see there are some checkstyle errors which I don't get when building locally and running the tests. Is there any way to run chcekstyle alone and figure out what is wrong?

maosongfu:
@avflor 
Try with: bazel build --config=darwin --experimental_action_listener=tools/java:compile_java heron/...

avflor:
@maosongfu Thanks! I tried this command and it still works locally. I'm working on Ubuntu but we also tried on a mac and we don't get any checkstyle errors there as well.

objmagic:
I checked out your branch and ran

``bazel build --config=darwin --experimental_action_listener=tools/java:compile_java heron/packing/...``

I can see lots of style errors.

maosongfu:
@avflor Maybe you need "bazel clean" first? @billonahill may have better ideas on this.

ashvina:
Yes. I can see the errors after "bazel clean".

kramasamy:
@objmagic - can you cut and paste the checkstyle errors for @avflor?

objmagic:
@ashvina are you the same guy as @avflor? and who is @avrilia? what is going on here?

kramasamy:
@avflor - one broader comment is to write a documentation about this packing algorithm.

- What does the algorithm do?
- When is it useful and its pros and cons?
- How to enable it?

avflor:
@kramasamy I fixed the checkstyle errors. After bazel clean I was able to see them and I checked in the new code. I will write some extra documentation. I have some comments about the algo in the code but I can expand.

@objmagic Thanks for noticing that. There has been some weird configuration issue and apparently GitHub linked my email with an old GitHub account (avrilia). I will fix this and from now on only avflor will be used.

objmagic:
@avflor please fix two Java Docs error [here](https://travis-ci.org/twitter/heron/builds/144013684#L1665). You should use ``&lt;`` and ``&gt;``.

kramasamy:
@ajorgensen - can you review this PR since you are the closest to write a new packing algorithm?

heron/packing/src/java/com/twitter/heron/packing/binpacking/FFDPacking.java
```diff
@@ -0,0 +1,384 @@
...
+ * ((# containers used by FFD)* size of each container
+ * + 2 * (size of a container that includes one instance with default resource requirements)).
+ * The extra 2 is considered for Heron internal container (i.e. the one containing Scheduler
+ * and TMaster) and for the YARN application master,
+ * <p>
+ * 7. The pack() return null if PackingPlan fails to pass the safe check, for instance,
+ * the size of ram for an instance is less than the minimal required value.
+ */
+
+public class FFDPacking implements IPacking {
```
ajorgensen: `FFD` is a pretty esoteric term. What if we expanded it to `FirstFitDecreasing` or `BinPacking`? I feel like expanding out the name would make it more discoverable.

heron/packing/src/java/com/twitter/heron/packing/binpacking/FFDPacking.java
```diff
@@ -0,0 +1,384 @@
...
+
+      containerPlanMap.put(containerId, containerPlan);
+      topologyRam += containerRam;
+      topologyCpu += containerCpu;
+      topologyDisk += containerDiskInBytes;
+    }
+
+    // Take the heron internal container into account and the application master for YARN
+    // scheduler
+    topologyRam += 2 * (instanceRamDefault + DEFAULT_RAM_PADDING_PER_CONTAINER);
```
ajorgensen: So this calculation is not generalizable to other schedulers correct? For example Aurora only has 1 extra container, I am not familiar with YARN but it looks like in this case there are 2 extra containers correct?.

ashvina: Good catch. Each topology running on YARN needs one extra container for application master. However the resources allocated to AM container and TMaster container are constant and independent of instanceRam and default padding. Which means the total resources allocated to a topology is a `constant + cost-of-deploying-instances` for both `yarn` and `aurora` clusters.

heron/packing/src/java/com/twitter/heron/packing/binpacking/FFDPacking.java
avflor: Sure, I can do that. Thanks!

heron/packing/src/java/com/twitter/heron/packing/binpacking/FFDPacking.java
ajorgensen: I've been thinking about this a little more recently and would be curious what your thoughts are about generic schedulers for all platforms and whether or not we should start splitting some of the at least shell implementations up into separate scheduling technologies (yarn, aurora, mesos, etc). I think that trying to generalize the requirements of all of those platforms into one algorithm might actually be making this more complex than it needs to be and certainly some schedulers will have different trade-offs and be able to make certain optimizations based on how they work. I dont necessarily think we need to do that in this case but I think its worth thinking about how that would work and whether the level effort/complexity that splitting them out into separate scheduler implementations might introduce. @kramasamy I'd be curious of your thoughts as well.

avflor: @ajorgensen We had a discussion with @ashvina about the same topic. The main limitation now is the fact that some schedulers support heterogeneous containers whereas others do not. However, the algorithm should be able to operate on both types of schedulers. The algorithm mainly decides how to pack instances in containers and should provide this packing to the scheduler. The scheduler can then decide to use homogeneous sor heterogeneous containers to represent this packing. I think @ashvina will open an issue on that. In this way, we would be able to use RoundRobin and binpacking on both Aurora and Yarn.

kramasamy:
@avflor @ajorgensen @ashvina - I think the consensus to keep the packing algorithms and schedulers orthogonal - the PR for this is here

https://github.com/twitter/heron/pull/1079

kramasamy:
Is FFD packing algorithm is good enough to merge?

avflor:
I made some updates to address some comments and also changed some configuration parameters. I will check it in tomorrow.

avflor:
@ajorgensen Let me know if you have other comments and I will address them tomorrow before I push the current version.

ajorgensen:
@avflor I think aside from my initial comments it looks good to me. I have other aspirations to have those default padding values be configurable as well but I don't believe that needs to happen in this pass.

kramasamy:
Can we open an issue for tracking?

ajorgensen:
Opened one here: https://github.com/twitter/heron/issues/1084

avflor:
@ajorgensen @kramasamy @maosongfu I submitted a revised version of the binpacking algorithm. I noticed when I run in local mode that the number of stream managers is picked based on the topology config. However, the binpacking algorithm decides on its own how many containers to use. The number of stream managers launched should be based on the result of packing algorithm. I enhanced the SchedulerMain.java code to take the value from the PackedPlan. I made the necessary change to the junit tests as well. I also changed the build files so that the Heron client can use the binpacking algorithm. Finally, I made the padding configurable and changed the binpacking algorithm to accommodate it. Please let me know if you have any further comments.

kramasamy:
@avflor - excellent. @ajorgensen @maosongfu - if you are fine with the PR, we can merge it.

@avflor - should we check in the documentation as a part of this PR as we discussed earlier or it will go in a separate PR?

avflor:
@kramasamy Karthik, what do you think about me submitting the documentation PR as a separate one next week?  I need some more time to look into the tools and the setup. In the meantime, I have comments on how to use the algorithm in the FirstFitDecreasing.java class.

kramasamy:
Ok. Please create a github issue and reference it here.

avflor:
First fit decreasing documentation: #1098

ajorgensen:
PR looks good to me!

maosongfu:
looks good to me. 👍

PR #1060, opened 2016-07-11
Add utils method schedulerCommandArgs

maosongfu:
Add utils method schedulerCommandArgs
    
Add utils method schedulerCommandArgs, allowing people to get only the args
of scheduler command. This could provide more flexibility when trying to invoke SchedulerMain.
Add unit tests.


kramasamy:
👍

PR #1062, opened 2016-07-11
Allow passing properties to SchedulerMain

maosongfu:
Currently SchedulerMain accepts only pre-defined command line arguments.
And there are no easy ways to pass custimzed arguments or properties to SchedulerMain,
while IScheduler implementation needs runtime properties generated on client side, i.e. ILauncher.
(Environment variables can be a workaround, though ugly.)

This pull request allows people to pass properties to SchedulerMain via arguments like "-Pproperty=value",
and IScheduler implementation can fetch them from Runtime.schedulerProperties(..).

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/common/Keys.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/common/Keys.java
@@ -15,6 +15,8 @@
 package com.twitter.heron.spi.common;
 
 public class Keys {
+  public static final String SCHEDULER_PROPERTIES = "heron.scheduler.properties";
+  public static final String SCHEDULER_COMMAND_LINE_PROPERTIES_OVERRIDE_OPTION = "P";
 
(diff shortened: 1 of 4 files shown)


kramasamy:
👍

heron/spi/src/java/com/twitter/heron/spi/common/Keys.java
```diff
@@ -15,6 +15,8 @@
 package com.twitter.heron.spi.common;
 
 public class Keys {
+  public static final String SCHEDULER_PROPERTIES = "heron.scheduler.properties";
+  public static final String SCHEDULER_COMMAND_LINE_PROPERTIES_OVERRIDE_OPTION = "P";
```
nlu90: These two new keys are not consistent keys used below. Is there any particular reason for this inconsistency?

nlu90:
👍

PR #1065, opened 2016-07-12
Change AuroraController.appendAuroraCommandOptions to private

billonahill:
This method is not used outside of this class. Changing to private.


maosongfu:
Should we change it into protected so later we can use it in Unit test?

billonahill:
I prefer making methods as closed-scoped as possible and opening scope when needed. So once we write a test we can change it to protected and annotate it `@VisibleForTesting`.

kramasamy:
👍

maosongfu:
:+1:

PR #1066, opened 2016-07-12
Tracker to set correct HTTP response Content-Type header

billonahill:
Without this the content type was text/html. This causes json browser plugins to not work always.


kramasamy:
Looks good to me!

objmagic:
LGTM

PR #1067, opened 2016-07-13
Heron native mesos scheduler

maosongfu:
Add heron native mesos scheduler and unit tests.
This scheduler aims to help people get started with heron on mesos; it is not for production.
It has a lot of concerns, for instance, not fully fault tolerant, etc.

In this implementation, heron would start the MesosFramework as a background process on the host doing submission, communicate
with mesos master directly and do the actual scheduling. So:
1. It assumes the host for submission is in the same network as mesos cluster.
2. It assumes the heron mesos scheduler is long running.

The heron mesos scheduler would re-schedule any failed executors. But if the heron mesos scheduler fails due to any
unexpected reasons, it would not get recovered and the whole topology would halt.


heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosLauncher.java
```diff
@@ -0,0 +1,147 @@
...
+  private Config runtime;
+
+  private String schedulerWorkingDirectory;
+
+  @Override
+  public void initialize(Config mConfig, Config mRuntime) {
+    this.config = mConfig;
+    this.runtime = mRuntime;
+
+    // get the topology working directory
```
nlu90: get the scheduler working directory?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,160 @@
...
+
+    // start the driver, non-blocking
+    driver.start();
+
+    return mesosFramework.createJob(jobDefinition);
+  }
+
+  @Override
+  public List<String> getJobLinks() {
+    return new ArrayList<>();
```
nlu90: If an empty list is returned, user is not able to go to the job page through heron ui.

In the Aurora scheduler, we constructed this job link on the fly.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/BaseContainer.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @JsonProperty
+  public String description;
+
+  @JsonProperty
+  public double cpu;
+
+  @JsonProperty
+  // in mb
+  public double disk;
```
nlu90: should we put the unit into variable name? like diskInMB?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/TaskResources.java
```diff
@@ -0,0 +1,151 @@
...
+  public double getDisk() {
+    return disk;
+  }
+
+  public int getPorts() {
+    return ports;
+  }
+
+  public String toString() {
+    return String.format("cpu: %s; mem: %s; disk: %s; ports: %d.",
```
nlu90: %s is for string, cpu, mem and disk are doubles

kramasamy:
@maosongfu - why the claim that it is not scalable? I understand the fault tolerance aspect of it but not sure the scalability part of it.

maosongfu:
@kramasamy 
Scalability can be considered from different point of views. For instance, this scheduler would not cache offer, which means it may take a long time to schedule a big topology, since the tasks need to be scheduled one by one.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosLauncher.java
maosongfu: Fixed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/TaskResources.java
maosongfu: Fixed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/BaseContainer.java
maosongfu: Fixed

kramasamy:
@maosongfu - however when the job is scheduled, it knows upfront the number of containers and the resources it requires. As Mesos keeps giving those offers, we could grab those resources and schedule containers.

maosongfu:
@kramasamy 
Consider Mesos keeps giving offers slowly, it will take a long time to receive enough offers to fully schedule a big topology. Also, it is just one case. There are still a lot of potential improvement for scalability.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
maosongfu: Added the job link.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,179 @@
...
+  @Override
+  public void initialize(Config mConfig, Config mRuntime) {
+    this.config = mConfig;
+    this.runtime = mRuntime;
+    this.mesosFramework = getMesosFramework();
+  }
+
+  @Override
+  public void close() {
+    // Stop the SchedulerDriver
```
nlu90: upon close, do we need to kill all the running jobs and close the mesosFrameWork also?

maosongfu: No. Close() can be invoked when the jvm process SchedulerMain exits. There can be cases people restart SchedulerMain process to pick up earlier running topology (for fault tolerance). So the close of MesosScheduler should be independent with the killing of a topology.

maosongfu: Comments added in the source code: The lifecycle of Topology is independent from Scheduler, for instance, Scheduler may crash, restart and pick up already running topology (for fault tolerance).

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,232 @@
...
+         containerIndex < Runtime.numContainers(runtime);
+         containerIndex++) {
+      jobDefinition.put(containerIndex, getBaseContainer(containerIndex, packing));
+    }
+
+    String masterURI = MesosContext.getHeronMesosMasterUri(config);
+    this.driver = getSchedulerDriver(masterURI, mesosFramework);
+
+    // start the driver, non-blocking
+    driver.start();
```
nlu90: Why we start the driver inside onSchedule() instead of inside initialize()? Put it in initialize seems more reasonable  and all the information needed is also available in initialize()

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,541 @@
...
+    // Convert the offers into TasksResources, an utils class.
+    for (Protos.Offer offer : offers) {
+      offerResources.put(offer, TaskResources.apply(offer, Context.role(heronConfig)));
+    }
+
+    // Generate launchable tasks
+    List<LaunchableTask> tasksToLaunch = generateLaunchableTasks(offerResources);
+
+    // Launch tasks
+    launchTasks(tasksToLaunch);
```
nlu90: If a task to be launched fails, what should we do with the resource offers? Should we decline the offer explicitly or we retry until the task is launched successfully?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,541 @@
...
+
+  @Override
+  public void disconnected(SchedulerDriver schedulerDriver) {
+    // TODO(mfu): TO handle this failure
+    LOG.info("Disconnected!");
+  }
+
+  @Override
+  public void slaveLost(SchedulerDriver schedulerDriver, Protos.SlaveID slaveID) {
+    // TODO(mfu): TO handle this failure
```
nlu90: If a slave is lost, that means the tasks on that slave are also lost. Do we need some kind of recovering mechanism? Otherwise, the topology is not working correctly.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,541 @@
...
+          .iterator();
+      while (it.hasNext()) {
+        Map.Entry<Protos.Offer, TaskResources> kv = it.next();
+        Protos.Offer offer = kv.getKey();
+        TaskResources resources = kv.getValue();
+
+        if (resources.canSatisfy(neededResources)) {
+          resources.consume(neededResources);
+
+          // Add the tasks
```
nlu90: dangling comment?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,541 @@
...
+          break;
+        } else {
+          LOG.info(String.format("Insufficient resources remaining for baseContainer: %s, "
+                  + "will append to queue. Need: [%s], Found: [%s]",
+              taskId, neededResources.toString(), resources.toString()));
+
+          toScheduleTasks.add(taskId);
+
+          // No enough Resources remaining. Return the tasks generated.
+          return tasks;
```
nlu90: The "else" case means the current resource offer cannot meet the request. But we can try remaining resource offers  using iterator instead of returning it directly.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,541 @@
...
+      TaskResources neededResources =
+          new TaskResources(
+              baseContainer.cpu,
+              baseContainer.memInMB,
+              baseContainer.diskInMB,
+              baseContainer.ports);
+
+      Iterator<Map.Entry<Protos.Offer, TaskResources>> it = offerResources.entrySet()
+          .iterator();
+      while (it.hasNext()) {
```
nlu90: The resource assignment strategy here is "first-fit", which may not be a very good strategy in practise. Should we abstract this assignment logic here out and refine it further?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
maosongfu: sounds good.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: The mesos master will reclaim the offer back if we fail to launch a task with given offer.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: True the topology will not work normally if a slave lost. The recovering mechanism is complicated and reconcile() is needed. Prefer not to handle this in this iteration consider this scheduler implementation does not aim for production.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Fixed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Fixed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: First-fit is a reasonable strategy in practice. Method generateLaunchableTasks(..) is already separated to allow people customize their own assigning strategy. We could evolve it later if any practical concerns raised.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
kramasamy: Can you open an issue to track this?

maosongfu: https://github.com/twitter/heron/issues/1077

kramasamy:
Sounds good to me. Let us wait until @ashvina takes a look at it.

nlu90:
LGTM 👍

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosLauncher.java
```diff
@@ -0,0 +1,147 @@
...
+  // Utils methods leveraging unit tests
+  ///////////////////////////////////////////////////////////////////////////////
+  protected boolean setupWorkingDirectory() {
+    if (!FileUtils.isDirectoryExists(schedulerWorkingDirectory)) {
+      FileUtils.createDirectory(schedulerWorkingDirectory);
+    }
+
+    String topologyPackageURL = String.format("file://%s", Context.topologyPackageFile(config));
+    String topologyPackageDestination = Paths.get(
+        schedulerWorkingDirectory, "topology.tar.gz").toString();
```
ashvina: Can the topology package name be picked up from config? For e.g. `Context.topologyPackageFile(config)`

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
```diff
@@ -0,0 +1,237 @@
...
+
+    container.name = TaskUtils.getTaskNameForContainerIndex(containerIndex);
+
+    container.runAsUser = Context.role(config);
+    container.description = String.format("Container %d for topology %s",
+        containerIndex, Context.topologyName(config));
+
+    // We assume every container is homogeneous
+    PackingPlan.Resource containerResource =
+        packing.containers.values().iterator().next().resource;
```
ashvina: Instead of assuming that all packing plans will always return homogeneous containers, this method could find max container size (one time computation) and use it?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+      return false;
+    }
+    isTerminated = true;
+
+    LOG.info(String.format("Kill job: %s", Context.topologyName(heronConfig)));
+    LOG.info("Remove all tasks to schedule");
+    toScheduleTasks.clear();
+
+    // Kill all the tasks
+    for (String taskId : tasksId.values()) {
```
ashvina: There could be a race condition here. A task might get scheduler by `scheduleNewTask` after this for loop completes?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+      List<Protos.TaskInfo> mesosTasks = new LinkedList<>();
+      for (LaunchableTask task : subTasks) {
+        Protos.TaskInfo pTask = task.constructMesosTaskInfo(heronConfig, heronRuntime);
+        mesosTasks.add(pTask);
+      }
+
+      LOG.info("Launching tasks from offer: " + offer + " with tasks: " + mesosTasks);
+
+      // Launch the tasks for the same offer
+      Protos.Status status =
```
ashvina: Test if the topology is killed/terminated before submitting any new tasks?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+          driver.launchTasks(Arrays.asList(new Protos.OfferID[]{offer.getId()}),
+              mesosTasks);
+
+      if (status == Protos.Status.DRIVER_RUNNING) {
+        LOG.info(String.format("Tasks launched, status: '%s'", status));
+
+      } else {
+        LOG.info("Other status returned: " + status);
```
ashvina: Is any other status ok? Also in method `generateLaunchableTasks`, taskId is removed from `toScheduleTasks` if it has a matching offer. However if the status is not running, does that still mean that task is running. Otherwise the task should not be removed from `toScheduleTasks` till it is successfully scheduled.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+      Map<Protos.Offer, TaskResources> offerResources) {
+    List<LaunchableTask> tasks = new LinkedList<>();
+
+    if (isTerminated) {
+      LOG.info("Job has been killed");
+      return tasks;
+    }
+
+    while (!toScheduleTasks.isEmpty()) {
+      String taskId = toScheduleTasks.poll();
```
ashvina: taskId can still be null. Check for null?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+  @Override
+  public void offerRescinded(SchedulerDriver schedulerDriver, Protos.OfferID offerID) {
+    LOG.info("Offer Rescinded!");
+  }
+
+  @Override
+  public void statusUpdate(SchedulerDriver schedulerDriver, Protos.TaskStatus taskStatus) {
+    LOG.info(String.format("Received status update [%s]", taskStatus));
+    if (isTerminated) {
+      LOG.info("The framework terminated. Ignoring any status update.");
```
ashvina: return to ignore the message?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    String taskId = taskStatus.getTaskId().getValue();
+
+    BaseContainer container = containersInfo.get(TaskUtils.getContainerIndexForTaskId(taskId));
+
+    if (container == null) {
+      LOG.info("Received a status update from unknown container. ");
```
ashvina: Is it better to kill the task instead of ignoring the message? Should such conditions be logged at WARNING level?

kramasamy:
@maosongfu - can we move this scheduler to contrib directory since it is experimental and also does not cover fault tolerance?

maosongfu:
@kramasamy It is not experimental; it is a scheduler aiming at getting started, similar to LocalScheduler. Should we also move all schedulers to contrib/ directory? Or we may create a new repo called: heron-schedulers, similar to storm or presto? since they are all on top of heron.

kramasamy:
one of the model that we discussed was the possibility of using a contrib directory - the code is allowed to check into the directory especially those are experimental / getting started. once they attain production quality, we move that code from contrib to main directories. This will allow for faster iteration in contrib rather than the PR waiting to be merged for a long time. This is a model that is followed in other open source projects such as Hadoop. 

Looks like a good idea IMO.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/MesosScheduler.java
maosongfu: Sounds good. Updated.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Good catch. Added synchronized block: 
```
synchronized (toScheduleTasks) {}
```
in critical blocks inside some public methods to avoid concurrent issues.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Added synchronized block: 
```
synchronized (toScheduleTasks) {}
```
outside to guarantee toScheduleTasks will not be changed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Sounds good. Addressed by invoking handlesMesosFailure(TaskId)

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Added synchronized block: 
```
synchronized (toScheduleTasks) {}
```
outside to guarantee toScheduleTasks will not be changed in other threads. So taskId will not be null now.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Fixed.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Updated to log at WARNING level. 
Added a comment in the source code:
```
// TODO(mfu): Don't know what should be done here. Simply ignore this status update for now.
```
I am hesitate to kill the task here since it is also possible to kill tasks belonging to other jobs.
This case should rarely happen.

maosongfu:
@kramasamy 
On the one hand, Hadoop may not be a good example to follow; people always complain its huge mono-repo. For more current github projects, for instance, presto, they would like to split them into small sub repos, making each one clear.

On the other hand, looking at some popular open source projects, I could not see “contrib” folder: jQuery, MongoDB, Redis, Puppet, Ruby on Rails, Jenkins. Django has a contrib folder and the role of folder is already explained here: https://docs.djangoproject.com/en/dev/ref/contrib/
```
Django aims to follow Python’s “batteries included” philosophy. It ships with a variety of extra, optional tools that solve common Web-development problems.

This code lives in django/contrib in the Django distribution. This document gives a rundown of the packages in contrib, along with any dependencies those packages have.
```
So, I am not convinced by the point that contrib folder is for experimental / getting started. I think it is more for tools building on top of the core part.

To summarize, it can be a good idea to move all scheduler-implementations and similar things to an individual repo (to an individual folder is debatable). But whether to move them bases on the nature of those projects, rather than the quality of those projects.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,575 @@
...
+  }
+
+  // Kill a topology
+  public boolean killJob() {
+    synchronized (toScheduleTasks) {
+      if (isTerminated) {
+        LOG.info("Job has been killed");
+        return false;
+      }
+      isTerminated = true;
```
ashvina: I think `isTerminated` should be outside the critical section. Setting `isTerminated` or killJob need not wait for any other operation to complete.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,575 @@
...
+   *
+   * @param taskId the taskId of container to handle
+   */
+  protected boolean scheduleNewTask(String taskId) {
+    // Put it to the pending start queue
+    LOG.info(String.format("We are to schedule task: [%s]", taskId));
+
+    // Update the tasksId
+    int containerIndex = TaskUtils.getContainerIndexForTaskId(taskId);
+    tasksId.put(containerIndex, taskId);
```
ashvina: taskId is not protected if `handleMesosFailure` and `killJob` are concurrently executed?

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
```diff
@@ -0,0 +1,575 @@
...
+  private static final Logger LOG = Logger.getLogger(MesosFramework.class.getName());
+
+  private final Config heronConfig;
+  private final Config heronRuntime;
+
+  // ContainerIndex -> BaseContainer
+  private final Map<Integer, BaseContainer> containersInfo;
+
+  // ContainerIndex -> TaskId
+  private final Map<Integer, String> tasksId;
```
ashvina: TaskId may need a special data structure for thread safety. I am thinking of a hashMap in which new tasks cannot be added once killjob is initiated.

ashvina:
@kramasamy @maosongfu 
I agree with @maosongfu now. I had a couple of conversations with some OSS developers here. A `contrib` directory is primarily used for "add-on" features. For e.g. scripts and tools for debugging. There are a few projects where the `contrib` directory is used for un-important and un-supported code. However that may result in hard to manage "trash" and may be discouraging for new contributors. A scheduler implementation is core to Heron. So I think the Mesos scheduler should be added to core. 

However, I am not sure how to manage "not-for-production" contributions. A experimental feature may starve for users if left in its own branch. So I would prefer merging them in master. But how will a user know if he is using a experimental feature? Should we print a warning on the console.

With respect to this patch, I think a few minor issues are pending in addition to fault tolerance cases. Also, I may not be able to deploy and test this on a mesos cluster.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: `isTerminated` needs to be within the synchronized block; otherwise the check of isTerminated and the update of Terminated could lead to concurrent bugs. It may be a little confusing to `synchronized (toScheduleTasks)`; changed it to `synchronized (this)` to make it clear.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: handleMesosFailure is a protected method invoked in a public method, which is within the synchronized block; so `handleMesosFailure` and `killJob` will not concurrently executed.
It may not be a good idea to add synchronized block in an internal method, i.e. private or protected.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: Not sure how necessary to do that. Currently the critical parts are within synchronized block and can not execute concurrently; we also have `isTerminated` checking at the start of those critical parts. So logically, tasksId will not be updated by other threads once killJob() is initiated.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
ashvina: Ok. In most methods, the entire method body is now protected by synchronized(this). Wondering if it is better to make the methods synchronized instead.

ashvina:
The scheduler code seems fine now. There are a few TODO items. Are you planning to create issues to help track them?

ashvina:
@kramasamy @maosongfu 

I have one orthogonal question related to the Mesos scheduler. Apache REEF seems to support Mesos also. The YARN scheduler is based on REEF. I think a lot of code can be reused if we developed a Mesos Scheduler based on REEF. What do you think about it? I can give it a shot if you'd like.

heron/schedulers/src/java/com/twitter/heron/scheduler/mesos/framework/MesosFramework.java
maosongfu: I would prefer using synchronized blocks over synchronized methods, since:
1. It clearly shows what object is used as lock
2. It could provide more flexibility because it is kind of an internal implementation people can use other object as lock or narrow the block scope in future, whereas syncing a public method would is more kind of determining an interface, not easy to evolve.

maosongfu:
@ashvina 
Issue created: #1077 

I also heard about that Apache REEF supports Mesos also. But I could not find enough documentation or status on that. It would be great if we can make heron run on Mesos basing on REEF at production quality. Just don't know how hard it is or how long it will take.

maosongfu:
@ashvina @kramasamy @nlu90 
Any more concerns I need to handle beforing merging this pull request?

kramasamy:
@maosongfu @ashvina - I am fine with getting Mesos to the core rather than being contrib. As a policy we can adapt the following (LMK if this is ok)

- contrib directory will contains features/debugging tools/other software that are addons to heron
- core software will continue to be developed in the master and can reside on their own branches/forks until it attains quality before merging into master

Any other items that is not covered by these two?

kramasamy:
@ashvina - Regarding Mesos on Apache REEF, we are good if the project supports it and to some extend proved in production. The key aspect with Mesos is the fault tolerance - Mesos has one way of doing this.

PR #1070, opened 2016-07-13
Move duplicate get/delete code up into FileSystemStateManager

billonahill:
This diff is rendering messy but basically all get/delete/exists interface methods do the same thing across curator and local filesystem if we abstract out the following:
```
abstract protected ListenableFuture<Boolean> nodeExists(String path);
abstract protected ListenableFuture<Boolean> deleteNode(String path);
abstract protected <M extends Message> ListenableFuture<M> getNodeData(WatchCallback watcher,
                                                                         String path,
                                                                         Message.Builder builder);
```

This removes duplicate code and makes it easier to add new state manager methods. Creates are handled differently since ZK has the notion of ephemeral nodes and local FS has the notion of overwrites.


maosongfu:
looks good to me. 👍

PR #1079, opened 2016-07-14
Move scheduler specific computation outside packing

### LINKED ISSUE #1073: RoundRobinPacking over allocate resources
Currently `RoundRobinPacking` executes the following steps:

1. creates a packing plan and computes resources required by each container
2. finds the biggest container, i.e. container with the largest resource requirement, say R
3. assigns the largest resource (R) to all the containers

It seems the last step is required by clusters where heterogenous containers cannot be allocated, for .e.g. `Aurora`. As a result this `RoundRobinPacking` is not leveraging `yarn scheduler`'s ability to allocate heterogeneous container. 

Ideally the packing algorithm should create a packing plan independent of the runtime cluster. Hence the max resource computation logic should be moved to the scheduler or util classes.

ashvina:
Fixes #1073 

This pull request address two issues
1. Currently Aurora scheduler assumes that the Packing algorithm has homogenized all containers and uses the size of first container for requesting containers. It should pick the largest container and use it instead.
2. RoundRobinAlgorithm need not spread the resource requirement of largest container over smaller containers.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
@@ -107,18 +107,16 @@ public class RoundRobinPacking implements IPacking {
     long containerDiskInBytes = getContainerDiskHint(roundRobinAllocation);
     double containerCpu = getContainerCpuHint(roundRobinAllocation);
 
-    // Align the ram to the maximal one
-    long containerRam = getLargestContainerRam(instancesRamMap);
-
     // Construct the PackingPlan
     Map<String, PackingPlan.ContainerPlan> containerPlanMap = new HashMap<>();
-
+    long topologyRam = 0;
     for (Map.Entry<String, List<String>> entry : roundRobinAllocation.entrySet()) {
       String containerId = entry.getKey();
       List<String> instanceList = entry.getValue();
 
       // Calculate the resource required for single instance
       Map<String, PackingPlan.InstancePlan> instancePlanMap = new HashMap<>();
+      long containerRam = DEFAULT_RAM_PADDING_PER_CONTAINER;
       for (String instanceId : instanceList) {
         long instanceRam = instancesRamMap.get(containerId).get(instanceId);
 
@@ -136,6 +134,7 @@ public class RoundRobinPacking implements IPacking {
                 resource);
         // Insert it into the map
         instancePlanMap.put(instanceId, instancePlan);
+        containerRam += instanceRam;
       }
 
       PackingPlan.Resource resource =
@@ -144,11 +143,11 @@ public class RoundRobinPacking implements IPacking {
           new PackingPlan.ContainerPlan(containerId, instancePlanMap, resource);
 
       containerPlanMap.put(containerId, containerPlan);
+      topologyRam += containerRam;
     }
 
     // Take the heron internal container into account
     int totalContainer = containerPlanMap.size() + 1;
-    long topologyRam = totalContainer * containerRam;
     long topologyDisk = totalContainer * containerDiskInBytes;
     double topologyCpu = totalContainer * containerCpu;
 
@@ -182,6 +181,8 @@ public class RoundRobinPacking implements IPacking {
 
     Map<String, Map<String, Long>> instancesRamMapInContainer = new HashMap<>();
 
+    List<String> instancesToBeAccounted = new ArrayList<>();
+
     for (Map.Entry<String, List<String>> entry : allocation.entrySet()) {
       String containerId = entry.getKey();
       Map<String, Long> ramInsideContainer = new HashMap<>();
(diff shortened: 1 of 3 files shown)


maosongfu:
@nlu90 Can u take a look at this?

billonahill:
@ashvina looks like the packing plan unit test is failing in CI.

ashvina:
@billonahill Thanks. I will update the patch in a bit.

kramasamy:
@maosongfu @ajorgensen @billonahill - can you review this PR?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -144,11 +143,11 @@ public PackingPlan pack() {
           new PackingPlan.ContainerPlan(containerId, instancePlanMap, resource);
 
       containerPlanMap.put(containerId, containerPlan);
+      topologyRam += containerRam;
```
nlu90: topologyRam needs to add the ram for Container 0

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -190,6 +194,22 @@ protected String getHeronAuroraPath() {
     return auroraProperties;
   }
 
+  /**
+   * This method finds the container with highest resource requirement and returns the resource.
+   * Currently only RAM is used for max identification.
+   */
+  private Resource getMaxRequiredResource(PackingPlan packingPlan) {
```
nlu90: We can add a `PackingPlan getHomogeneousResource(PackingPlan pp)`  in `SchedulerUtils` to make the logic resuable.

And also, here only ram is considered, should we also consider cpu and disk? It's possible that containers have different requirement for cpus if containers have different # of instances.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
ashvina: Currently packing algorithm has no knowledge about TMaster ram requirement. As `AuroraScheduler` is requesting a container for container-0, it increments the topologyRam after receiving a packing plan. See new `AuroraScheduler.java`#L136 in the diff

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
ashvina: I like the first suggestion, some helper method like `getHomogeneousResource`. As I understand, the helper method will be used by a scheduler to update the `PackingPlan`, so that downstream components are up-to-date with the actual allocation. 

Regarding your second point, CPU and Disk need similar treatment. However a couple of improvements required in RoundRobinPacking first. CPU and disk do not have the same api and config support as RAM. It could have been done to reduce complexity. I would suggest that CPU and disk be discussed in a separate thread.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -190,6 +194,22 @@ protected String getHeronAuroraPath() {
...
+    Resource maxResource = packingPlan.containers.values().iterator().next().resource;
+    for (ContainerPlan entry : packingPlan.containers.values()) {
```
billonahill: nit: s/entry/plan/g

billonahill:
Just left a small naming nit, but changes LGTM if @nlu90 and @maosongfu don't have add'l comments.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -128,8 +130,10 @@ protected String getHeronAuroraPath() {
 
     TopologyAPI.Topology topology = Runtime.topology(runtime);
 
-    PackingPlan.Resource containerResource =
-        packing.containers.values().iterator().next().resource;
+    // Align the ram to the maximal one
+    PackingPlan.Resource containerResource = getMaxRequiredResource(packing);
+    // Update total topology resource requirement on Aurora clusters
+    packing.resource.ram = containerResource.ram * (packing.containers.size() + 1);
```
nlu90: can you add some more comments about the "+1", which is for container 0?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
nlu90: KK, found the line which requesting one more container for TMaster. Could you add one line comment for it?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
nlu90: Could you file an issue regarding the cpu and disk update so that we can track it?

nlu90:
👍  once the cpu and disk issue is filed and comments are added

kramasamy:
@nlu90 and @maosongfu is this good to go?

nlu90:
@kramasamy LGTM

Issue #1072, opened 2016-07-13
Follow-up of Pylint (#849)


objmagic:
keeping track of related issues after merging #849. Namely:
- [ ] fix https://github.com/twitter/heron/pull/849#issuecomment-231932982
- [x] fix https://github.com/twitter/heron/pull/849#issuecomment-231834711 ("errors that I don't have enough context to fix (labelled with FIXME)")
-  [x] #578 #579 #580 #581 #582 #583 #584 #1091

objmagic:
cf. #587

objmagic:
small issue: https://github.com/twitter/heron/pull/1081#discussion_r70904671

objmagic:
interesting warning style error "dangerous default value"
https://github.com/twitter/heron/blob/be87b09f348e0ed05f45503340a2245a4ef68a35/heron/tracker/src/python/handlers/exceptionhandler.py#L60

objmagic:
now I think it is not necessary to make this change: https://github.com/twitter/heron/pull/1081#discussion_r70904671

objmagic:
#578 #579 #580 #581 #582 #583 #584 #1091 are all fixed.

waiting #1096 CI to finish and I'll PR to enable Python style checking tomorrow morning.

PR #1081, opened 2016-07-15
Fix Python code style under heron/ui

### LINKED ISSUE #578: Enforce python check style in UI

objmagic:
resolves #578


heron/ui/src/python/main.py
```diff
@@ -36,49 +35,68 @@ class Application(tornado.web.Application):
...
-    # pylint: disable=line-too-long,bad-whitespace
     callbacks = [
         (r"/", handlers.MainHandler),
 
         (r"/topologies", handlers.ListTopologiesHandler),
-        (r"/topologies/filestats/([^\/]+)/([^\/]+)/([^\/]+)/([^\/]+)", handlers.ContainerFileStatsHandler),
-        (r"/topologies/([^\/]+)/([^\/]+)/([^\/]+)/config", handlers.TopologyConfigHandler),
-        (r"/topologies/([^\/]+)/([^\/]+)/([^\/]+)/([^\/]+)/([^\/]+)/exceptions", handlers.TopologyExceptionsPageHandler),
-        (r"/topologies/([^\/]+)/([^\/]+)/([^\/]+)", handlers.TopologyPlanHandler),
+        (r"/topologies/filestats/([^\/]+)/([^\/]+)/([^\/]+)/([^\/]+)",
```
kramasamy: do you think callbacks can be defined as a list outside and assign this outside list to callbacks = listofcallbacks. This will provide space per line and you don't need to introduce new lines.

billonahill: I would argue that we shouldn't change logic in patches that are just style changes. I like the suggestion but think it should be done separately.

kramasamy: Sounds good to me. @objmagic - can you create an issue for tracking?

objmagic: I think it's trivial so I put it here: https://github.com/twitter/heron/issues/1072#issuecomment-232838919

PR #1086, opened 2016-07-15
Remove cli utils, error invalid cli subcommand

objmagic:
* remove ``utils.py`` under ``heron/cli`` since it is no longer being used. We are using ``utils.py`` under ``heron/common`` now.

* When invalid subcommand is provided to ``heron``, the CLI program exits with 0. Users do not like this. Problem is fixed so that CLI program errors and exits with 1 in such scenario.


nlu90:
👍

PR #1090, opened 2016-07-15
Fix Python style under heron/shell.

### LINKED ISSUE #583: Enforce python check style in executor

objmagic:
resolves #583


kramasamy:
👍

PR #1097, opened 2016-07-17
Enable Python style checking during build and CI.

objmagic:
This PR enables Python style checking during build and CI.

It also reduces noise in Pylint's output (no chart displayed, no locally disabled warning displayed)

cross-reference: #587 #1072


kramasamy:
👍

PR #1105, opened 2016-07-18
Sort containers by id in the UI so they are easier to find

ajorgensen:
Sorting the containers by the id makes it easier to find a specific container in the UI without having to hunt around by hovering over the boxes. If you know that there is a problem with a component in container 5 you will now be able to easy go straight to it.

Before:
screen shot 2016-07-18 at 2 32 51 pm

After:
screen shot 2016-07-18 at 2 33 59 pm

### DIFF
--- a/heron/ui/resources/static/js/physical-plan.js
+++ b/heron/ui/resources/static/js/physical-plan.js
@@ -42,6 +42,12 @@
       // make ordering of instances the same in each container
       container.children = _.sortBy(container.children, 'name');
     });
+
+    // Sort the containers by their id so they are easier to find in the UI.
+    auroraContainers = _.sortBy(auroraContainers, function(container) {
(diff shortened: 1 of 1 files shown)


kramasamy:
looks good 👍

kramasamy:
@ajorgensen - is there a way to display the host information in addition to the container id? This information is lacking.

ajorgensen:
@kramasamy that information is available. I'm not sure from a UI perspective where the best place to put that information would be, my feeling is it would look too cluttered in the hover over. We could put that information under the `Instance Metrics` when you click on a specific instance in a container. It feels a bit weird to get container level information (host in this case) by clicking on a specific instance but I'm not sure where else in the UI it should go.

ajorgensen:
I'll leave pulling out the host information for another PR, I think some thought needs to go into where that information would be best located.

PR #1114, opened 2016-07-20
Switch to using external pex skylark rules

### LINKED ISSUE #1032: Split out pex_rules.bzl to a separate project?
The pex python rules found in `//tools/rules:pex_rules.bzl` would be useful to numerous projects beyond Heron.  I have started maintaining an enhanced variant of those rules in a new repository: https://github.com/benley/bazel_rules_pex.

Some of the changes I made:

* Everything works with bazel's sandboxing enabled now(!)
* `setuptools` and `wheel` are fetched as external `http_file` resources, not vendored into the project tree.
* The pex wrapper can now correctly (but optionally) use .whl dependencies along with eggs.
* The rules now work when loaded from a remote skylark repository.
* There is full documentation, generated from the source docstrings via skydoc.
* Added an `entrypoint` attribute to `pex_binary`, allowing a python module name to be used instead of a `main` file.  Useful for building pexes for external egg/whl dependencies.
* pex_library targets can now be passed to the native `py_library`, `py_binary`, and `py_test` rules as a dependency, and vice versa.
* `main` module resolution now works correctly for generated source files.
* `pex_test` no longer invokes the py.test runner by default. Instead it works more like bazel's built-in py_test rule.  The previous py.test behaviour is now available as a `pex_pytest` macro. _(incompatible change! but trivially reconciled)_
* `pex_test` is now a trivial special case of `pex_binary`.
* `pex_pytest` itself is now a fairly simple wrapper around `pex_binary`.
* `pex_pytest` now automatically adds a dependency on py.test
* The `resources` attribute has been renamed to `data` to be consistent with most other Bazel rule types. Its behaviour has not changed, just the name. _(incompatible change! but trivially reconciled.)_
* There is a lot less duplicated code.

Would there be any interest in combining efforts here?  My goal is to get rules_pex published as a semi-official project in the bazelbuild github organization, alongside the various other language-specific Bazel rules.  I would also be happy to contribute some or all of these improvements back to Heron, whether or not Heron ends up importing the rules from a separate project.

Thanks! Looking forward to any feedback.

benley:
Fixes #1032 

Hopefully this doesn't break compatibility with Bazel 0.2.3


tools/build_rules/prelude_bazel
```diff
@@ -0,0 +1,11 @@
+load(
+    "@io_bazel_rules_pex//pex:pex_rules.bzl",
+    "pex_binary",
+    "pex_library",
+    "pex_test",
+    "pex_pytest",
+)
+load(
+    "@bazel_tools//tools/build_defs/pkg:pkg.bzl",
```
billonahill: Is @bazel_tools defined globally somewhere and is this the proper way to reference the pkg targets we need? I see where @io_bazel_rules_pex is defined, but not @bazel_tools.

WORKSPACE
```diff
@@ -286,3 +288,11 @@ maven_jar(
   name = "org_mockito_mockito_all",
   artifact = "org.mockito:mockito-all:1.10.19",
 )
+
+git_repository(
+  name = "io_bazel_rules_pex",
+  remote = "https://github.com/benley/bazel_rules_pex.git",
```
billonahill: Can these rules get a better home either with the bazel or pex projects? This is fine for testing but we don't want heron to depend on an individual's git project.

Also this will fail in Twitters internal CI build since it's a closed environment without access to github, only a maven repo mirror and a pipy mirror. Could this code be distributed by either of those means somehow?

tools/build_rules/prelude_bazel
benley: I believe `@bazel_tools` is defined internally by Bazel; it replaces the former implicit merging of `//tools` into the user workspace.

WORKSPACE
benley: Better home: yes, I hope so.  I haven't directly requested a github project from bazelbuild, but moving it out of my personal account does seem appropriate before merging this change.

Internal CI: it would work equally well to get these rules from a .zip or .tgz archive using a [http_archive rule](http://www.bazel.io/docs/be/workspace.html#http_archive), or even from a path on disk using [local_repository](http://www.bazel.io/docs/be/workspace.html#local_repository) if that turns out to be easier.  In theory we could publish a pypi project containing the bazel rules, though it would be a bit awkward.  I'm not very familiar with Maven - can it be used to deliver arbitrary files?  If so, I imagine that could be made to work as well.

If none of that is practical, we could just vendor things back into this repo rather than fetching it externally as part of the build.

tools/build_rules/prelude_bazel
kramasamy: with 0.3.0 ```@bazel_tools``` is not supported I believe. Our repo can compile both in bazel 0.2.3 and bazel 0.3.0. Can you please ensure that it compiles with bazel 0.3.0 as well?

WORKSPACE
billonahill: Agreed pypi would be awkward, but maven does have support for tar.gz artifacts. See http://stackoverflow.com/questions/32088885/install-tar-gz-file-in-maven-repository

tools/build_rules/prelude_bazel
benley: I'm doing all my development and testing on 0.3.0 so far.

kramasamy: Cool - travis ci works with bazel 0.2.3

kramasamy:
@benley - any updates on this?

benley:
Sorry for disappearing; I haven't had time to work on this in the past week, but I intend to get back to it fairly soon.

objmagic:
hi @benley, any update on this PR?

benley:
Not much news, but I think I know of a reasonably sane way of publishing a usable source archive via pypi for the pex rules, which should remove the main obstacle.  I'll try to make progress on this issue this week.

kramasamy:
@benley - any progress on this?

kramasamy:
@benley - wondering if you would like to merge these rules into Bazel?

benley:
Hi there, sorry I have gone silent for so long.  I changed jobs and have not had much time to work on bazel-related things recently.

It would be cool to get the pex rules hosted under the bazelbuild project, but I don't think they will be interested because it more or less duplicates the functionality of https://github.com/google/subpar, along with bazel's built-in Python rules.

I did work out a way of hosting skylark rules on pypi, though! https://pypi.python.org/pypi/bazel-rules-pex

kramasamy:
Thanks @benley - we could take care of maintaining the rules. The only issue that I am facing with it - when I run it the first I get some errors. However, if we run bazel again, it seems to work. Not sure what is the issue - if you could help, it will be great.

https://github.com/benley/bazel_rules_pex/issues/37

PR #1121, opened 2016-07-20
Improve exception handling of heron-explorer.

### LINKED ISSUE #1117: suppress explorer's traceback when tracker is not running

objmagic:
1. unify format of error message when exception is raised
2. fix two places where exception will not be caught and traceback will be printed out.  
    This is ugly so I fixed it. (resolves #1117)

### DIFF
--- a/heron/common/src/python/utils.py
+++ b/heron/common/src/python/utils.py
@@ -20,6 +20,7 @@ import sys
 import subprocess
 import tarfile
 import tempfile
+import traceback
 import tornado.gen
 import tornado.ioloop
 import yaml
@@ -352,9 +353,8 @@ def get_clusters():
   # pylint: disable=unnecessary-lambda
   try:
     return instance.run_sync(lambda: API.get_clusters())
-  except Exception as ex:
-    Log.error(str(ex))
-    Log.error('Failed to retrive clusters')
+  except Exception:
+    Log.info(traceback.format_exc())
     raise
 
 
@@ -363,10 +363,8 @@ def get_logical_plan(cluster, env, topology, role):
   instance = tornado.ioloop.IOLoop.instance()
   try:
     return instance.run_sync(lambda: API.get_logical_plan(cluster, env, topology, role))
-  except Exception as ex:
-    Log.error(str(ex))
-    Log.error('Failed to retrive logical plan info of topology \'%s\''
-              % ('/'.join([cluster, role, env, topology])))
+  except Exception:
+    Log.info(traceback.format_exc())
     raise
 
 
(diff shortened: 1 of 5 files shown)


heron/common/src/python/utils.py
```diff
@@ -353,7 +353,7 @@ def get_clusters():
   try:
     return instance.run_sync(lambda: API.get_clusters())
   except Exception as ex:
-    Log.error('Error: %s' % str(ex))
+    Log.error(str(ex))
```
kramasamy: this will print the exception stack.

heron/common/src/python/utils.py
```diff
@@ -364,7 +364,7 @@ def get_logical_plan(cluster, env, topology, role):
   try:
     return instance.run_sync(lambda: API.get_logical_plan(cluster, env, topology, role))
   except Exception as ex:
-    Log.error('Error: %s' % str(ex))
+    Log.error(str(ex))
```
kramasamy: same as before.

heron/common/src/python/utils.py
```diff
@@ -377,6 +377,7 @@ def get_topology_info(*args):
     return instance.run_sync(lambda: API.get_topology_info(*args))
   except Exception as ex:
     Log.error(str(ex))
+    Log.error('Failed to get topology info')
```
kramasamy: same as before.

heron/common/src/python/utils.py
```diff
@@ -387,7 +388,7 @@ def get_topology_metrics(*args):
     return instance.run_sync(lambda: API.get_comp_metrics(*args))
   except Exception as ex:
     Log.error(str(ex))
-    Log.error("Failed to retrive metrics of component \'%s\'" % args[3])
+    Log.error("Failed to retrive metrics of topology \'%s\'" % args[3])
```
kramasamy: same as before.

heron/common/src/python/utils.py
```diff
@@ -398,7 +399,10 @@ def get_component_metrics(component, cluster, env, topology, role):
     result = get_topology_metrics(
         cluster, env, topology, component, [], all_queries, [0, -1], role)
     return result["metrics"]
-  except:
+  except Exception as ex:
+    Log.error(str(ex))
```
kramasamy: same as before.

heron/common/src/python/utils.py
objmagic: For a Python program ``test.py``:
````
def a():
  return 1 / 0

def b():
  return a()

def c():
  return b()

try:
  c()
except Exception as ex:
  print(str(ex))
  pass
````

``python test.py`` will only print out "division by zero"

An informative error will be printed out after ``Log.error(str(ex))`` (I can add cluster name, though)

objmagic: If you print out exception using ``print(ex)``, you will see stack trace.

kramasamy:
@objmagic - are we adding a verbose option? or it is going to be in a later PR?

objmagic:
It already has a verbose option.

kramasamy:
👍

PR #1123, opened 2016-07-21
Add double quote to support spaces in envionment variables

windie:
This PR adds double quote for environment variables to avoid crash when some existing variable (e.g., CFLAGS) contains spaces.

### DIFF
--- a/bazel_configure.py
+++ b/bazel_configure.py
@@ -289,7 +289,7 @@ def discover_tool_default(program, msg, envvar, defvalue):
 
 def export_env_to_file(out_file, env):
   if env in os.environ:
-    out_file.write('export %s=%s\n' % (env, os.environ[env]))
+    out_file.write('export %s="%s"\n' % (env, os.environ[env]))
 
 ######################################################################
 # Generate the shell script that recreates the environment


kramasamy:
👍

kramasamy:
@windie - curious to know what caused this issue.

windie:
@kramasamy - E.g., the following command fails
```
$ export CFLAGS=-I /usr/local/include/
-bash: export: `/usr/local/include/': not a valid identifier
```

I found this one when building heron in homebrew environment which sets CFLAGS and other similar environment variables.

In addition, right now I could not build heron using homebrew's bazel. Bazel just crashes with a useless error info: `ERROR: Process exited with status 1.`. Still investigating the cause.

kramasamy:
@windie - we have not tried with homebrew environment actually. Since the bazel files are self sufficient, there is no need to use CFLAGS etc. In fact, I would advice to try it out without setting any of the CFLAGS, CXXFLAGS, etc

kramasamy:
our current master works with both bazel 0.3.0 and bazel 0.2.3.

kramasamy:
@windie - one more suggestion would be to use the binaries from the release page to construct a brew formula. With this approach, you won't have the need to compile any code at all when you do brew install - it will be fast and easy and no need to even install bazel.

windie:
@kramasamy - [homebrew-core](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Acceptable-Formulae.md#we-dont-like-binary-formulae) requires Heron built from source. Otherwise, it needs to go into [homebrew-cask](https://github.com/caskroom/homebrew-cask). However, homebrew-cask only supports downloading a compressed package instead of an install binary like Heron does. It requires Heron releases a bundled tar that contains all install scripts.

I was thinking to submit a PR to https://github.com/Homebrew/homebrew-binary but it's just deprecated recently :(.

Issue #1126, opened 2016-07-21
Execution state fails to be written on deploy


ajorgensen:
I am not sure what the exact triggering case is but we've had some deploy where the executionstate has failed to make it to zookeeper. When we go to kill the topology we get the following error:

```
[2016-07-21 13:42:23 +0000] com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor SEVERE:  Exception processing future
java.util.concurrent.ExecutionException: java.lang.RuntimeException: Failed to fetch data from path: /heron/executionstate/foo
        at com.google.common.util.concurrent.AbstractFuture$Sync.getValue(AbstractFuture.java:299)
        at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:272)
        at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:96)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.awaitResult(SchedulerStateManagerAdaptor.java:71)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.awaitResult(SchedulerStateManagerAdaptor.java:63)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.getExecutionState(SchedulerStateManagerAdaptor.java:198)
        at com.twitter.heron.scheduler.RuntimeManagerMain.validateRuntimeManage(RuntimeManagerMain.java:351)
        at com.twitter.heron.scheduler.RuntimeManagerMain.manageTopology(RuntimeManagerMain.java:306)
        at com.twitter.heron.scheduler.RuntimeManagerMain.main(RuntimeManagerMain.java:251)
Caused by: java.lang.RuntimeException: Failed to fetch data from path: /heron/executionstate/foo
        at com.twitter.heron.statemgr.zookeeper.curator.CuratorStateManager$1.processResult(CuratorStateManager.java:236)
        at org.apache.curator.framework.imps.CuratorFrameworkImpl.sendToBackgroundCallback(CuratorFrameworkImpl.java:743)
        at org.apache.curator.framework.imps.CuratorFrameworkImpl.processBackgroundOperation(CuratorFrameworkImpl.java:520)
        at org.apache.curator.framework.imps.GetDataBuilderImpl$3.processResult(GetDataBuilderImpl.java:254)
        at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:564)
[...]

PR #1128, opened 2016-07-21
Protect against failures when killing a running topology.

ajorgensen:
I believe that this is functionally complete, I need to write some tests but wanted to get some feedback on the approach first.

If any of the state in zookeeper is deleted while others are left in
tact the topology will become undeployable because the cli will fail to
clean up the leftover state. The exception looks like:

```
[2016-07-21 13:42:23 +0000] com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor SEVERE:  Exception processing future
java.util.concurrent.ExecutionException: java.lang.RuntimeException: Failed to fetch data from path: /heron/executionstate/foo
        at com.google.common.util.concurrent.AbstractFuture$Sync.getValue(AbstractFuture.java:299)
        at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:272)
        at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:96)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.awaitResult(SchedulerStateManagerAdaptor.java:71)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.awaitResult(SchedulerStateManagerAdaptor.java:63)
        at com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor.getExecutionState(SchedulerStateManagerAdaptor.java:198)
        at com.twitter.heron.scheduler.RuntimeManagerMain.validateRuntimeManage(RuntimeManagerMain.java:351)
        at com.twitter.heron.scheduler.RuntimeManagerMain.manageTopology(RuntimeManagerMain.java:306)
        at com.twitter.heron.scheduler.RuntimeManagerMain.main(RuntimeManagerMain.java:251)
```

This change makes it so the kill command is more robust and will treat
instances where the state does not exist as a no-op and continue its
clean up.

This fixes: https://github.com/twitter/heron/issues/1126

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
@@ -160,9 +160,10 @@ public class LaunchRunner implements Callable<Boolean> {
 
     // launch the topology, clear the state if it fails
     if (!launcher.launch(packedPlan)) {
-      statemgr.deleteExecutionState(topologyName);
-      statemgr.deletePackingPlan(topologyName);
-      statemgr.deleteTopology(topologyName);
+      if (!statemgr.cleanState(topologyName)) {
+        LOG.severe("Failed to clean topology state");
+      }
+
       LOG.log(Level.SEVERE, "Failed to launch topology");
       return false;
     }
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
@@ -160,10 +160,7 @@ public class LaunchRunner implements Callable<Boolean> {
 
     // launch the topology, clear the state if it fails
     if (!launcher.launch(packedPlan)) {
-      if (!statemgr.cleanState(topologyName)) {
-        LOG.severe("Failed to clean topology state");
-      }
-
+      statemgr.cleanState(topologyName);
       LOG.log(Level.SEVERE, "Failed to launch topology");
       return false;
     }
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -347,18 +347,14 @@ public class RuntimeManagerMain {
       return false;
     }
 
-    // Check whether cluster/role/environ matched. It is possible to get into a state where some of
-    // the data does not actually exist in zookeeper. If we are in this state them assume the
-    // runtime is valid.
-    if (adaptor.doesExecutionStateExist(topologyName)) {
-      ExecutionEnvironment.ExecutionState executionState = adaptor.getExecutionState(topologyName);
-      if (executionState == null
-          || !executionState.getCluster().equals(Context.cluster(config))
-          || !executionState.getRole().equals(Context.role(config))
-          || !executionState.getEnviron().equals(Context.environ(config))) {
-        LOG.severe("cluster/role/environ not matched");
-        return false;
-      }
+    // Check whether cluster/role/environ matched
+    ExecutionEnvironment.ExecutionState executionState = adaptor.getExecutionState(topologyName);
+    if (executionState == null
+        || !executionState.getCluster().equals(Context.cluster(config))
+        || !executionState.getRole().equals(Context.role(config))
+        || !executionState.getEnviron().equals(Context.environ(config))) {
+      LOG.severe("cluster/role/environ not matched");
+      return false;
     }
 
     return true;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -303,7 +303,8 @@ public class RuntimeManagerMain {
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      boolean isValid = (Command.KILL.equals(command)) || validateRuntimeManage(adaptor, topologyName);
+      boolean isValid = (Command.KILL.equals(command))
+          || validateRuntimeManage(adaptor, topologyName);
 
       // 2. Try to manage topology if valid
       if (isValid) {
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -347,14 +347,18 @@ public class RuntimeManagerMain {
       return false;
     }
 
-    // Check whether cluster/role/environ matched
-    ExecutionEnvironment.ExecutionState executionState = adaptor.getExecutionState(topologyName);
-    if (executionState == null
-        || !executionState.getCluster().equals(Context.cluster(config))
-        || !executionState.getRole().equals(Context.role(config))
-        || !executionState.getEnviron().equals(Context.environ(config))) {
-      LOG.severe("cluster/role/environ not matched");
-      return false;
+    // Check whether cluster/role/environ matched. It is possible to get into a state where some of
+    // the data does not actually exist in zookeeper. If we are in this state them assume the
+    // runtime is valid.
+    if (adaptor.doesExecutionStateExist(topologyName)) {
+      ExecutionEnvironment.ExecutionState executionState = adaptor.getExecutionState(topologyName);
+      if (executionState == null
+          || !executionState.getCluster().equals(Context.cluster(config))
+          || !executionState.getRole().equals(Context.role(config))
+          || !executionState.getEnviron().equals(Context.environ(config))) {
+        LOG.severe("cluster/role/environ not matched");
+        return false;
+      }
     }
 
     return true;
@@ -365,7 +369,7 @@ public class RuntimeManagerMain {
     RuntimeManagerRunner runtimeManagerRunner =
         new RuntimeManagerRunner(config, runtime, command, schedulerClient);
 
-    // invoke the appropriate handlers based on command
+    S// invoke the appropriate handlers based on command
     boolean ret = runtimeManagerRunner.call();
 
     return ret;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -369,7 +369,7 @@ public class RuntimeManagerMain {
     RuntimeManagerRunner runtimeManagerRunner =
         new RuntimeManagerRunner(config, runtime, command, schedulerClient);
 
-    S// invoke the appropriate handlers based on command
+    // invoke the appropriate handlers based on command
     boolean ret = runtimeManagerRunner.call();
 
     return ret;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -303,7 +303,7 @@ public class RuntimeManagerMain {
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      boolean isValid = validateRuntimeManage(adaptor, topologyName);
+      boolean isValid = (Command.KILL.equals(command)) || validateRuntimeManage(adaptor, topologyName);
 
       // 2. Try to manage topology if valid
       if (isValid) {
--- a/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
@@ -79,6 +79,12 @@ public interface IStateManager extends AutoCloseable {
       return concatPath(getDirectory(root), topology);
     }
 
+    public static void deleteAll(IStateManager stateManager, String topology) {
+      for (IStateManager.StateLocation stateLocation : IStateManager.StateLocation.values()) {
+        stateLocation.delete(stateManager, topology);
+      }
+    }
+
     public void delete(IStateManager stateManager, String topology) {
       switch(this) {
         case TMASTER_LOCATION:
--- a/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
@@ -51,6 +51,64 @@ import com.twitter.heron.spi.common.Config;
  */
 
 public interface IStateManager extends AutoCloseable {
+  enum StateLocation {
+    TMASTER_LOCATION("tmasters", "TMaster location"),
+    TOPOLOGY("topologies", "Topologies"),
+    PACKING_PLAN("packingplans", "Packing plan"),
+    PHYSICAL_PLAN("pplans", "Physical plan"),
+    EXECUTION_STATE("executionstate", "Execution state"),
+    SCHEDULER_LOCATION("schedulers", "Scheduler location");
+
+    private final String dir;
+    private final String name;
+
+    StateLocation(String dir, String name) {
+      this.dir = dir;
+      this.name = name;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public String getDirectory(String root) {
+      return concatPath(root, dir);
+    }
+
+    public String getNodePath(String root, String topology) {
+      return concatPath(getDirectory(root), topology);
+    }
+
+    public void delete(IStateManager stateManager, String topology) {
+      switch(this) {
+        case TMASTER_LOCATION:
+          stateManager.deleteTMasterLocation(topology);
+          break;
+        case TOPOLOGY:
+          stateManager.deleteTopology(topology);
+          break;
+        case PACKING_PLAN:
+          stateManager.deletePackingPlan(topology);
+          break;
+        case PHYSICAL_PLAN:
+          stateManager.deletePhysicalPlan(topology);
+          break;
+        case EXECUTION_STATE:
+          stateManager.deleteExecutionState(topology);
+          break;
+        case SCHEDULER_LOCATION:
+          stateManager.deleteSchedulerLocation(topology);
+          break;
+        default:
+          // This should never be reached, nothing to do.
+      }
+    }
+
+    private static String concatPath(String basePath, String appendPath) {
+      return String.format("%s/%s", basePath, appendPath);
+    }
+  }
+
   /**
    * Initialize StateManager with the incoming context.
    */
@@ -138,6 +196,8 @@ public interface IStateManager extends AutoCloseable {
    * @return Boolean - Success or Failure
    */
   ListenableFuture<Boolean> deleteSchedulerLocation(String topologyName);
+
+
 
   /**
    * Get the tmaster location for the given topology
--- a/heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
@@ -226,39 +226,20 @@ public class SchedulerStateManagerAdaptor {
   public void cleanState(String topologyName) {
     LOG.fine("Cleaning up topology state");
 
-    Boolean result;
-
-    result = deleteTMasterLocation(topologyName);
-    if (result == null || !result) {
-      LOG.warning("Failed to clear TMaster location.");
-    }
-
-    result = deletePackingPlan(topologyName);
-    if (result == null || !result) {
-      LOG.warning("Failed to clear packing plan.");
-    }
-
-    result = deletePhysicalPlan(topologyName);
-    if (result == null || !result) {
-      LOG.warning("Failed to clear physical plan.");
-    }
-
-    result = deleteSchedulerLocation(topologyName);
-    if (result == null || !result) {
-      LOG.warning("Failed to clear scheduler location.");
-    }
-
-    result = deleteExecutionState(topologyName);
-    if (result == null || !result) {
-      LOG.severe("Failed to clear execution state");
-    }
-
-    result = deleteTopology(topologyName);
-    if (result == null || !result) {
-      LOG.severe("Failed to clear topology definition");
-    }
+    logDeleteStatus(deleteTMasterLocation(topologyName), "topology master");
+    logDeleteStatus(deletePackingPlan(topologyName), "packing plan");
+    logDeleteStatus(deletePhysicalPlan(topologyName), "physical plan");
+    logDeleteStatus(deleteSchedulerLocation(topologyName), "scheduler");
+    logDeleteStatus(deleteExecutionState(topologyName), "execution state");
+    logDeleteStatus(deleteTopology(topologyName), "topology definition");
 
     LOG.fine("Cleaned up topology state");
+  }
+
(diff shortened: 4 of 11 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -347,14 +347,18 @@ protected boolean validateRuntimeManage(
...
-    if (executionState == null
-        || !executionState.getCluster().equals(Context.cluster(config))
-        || !executionState.getRole().equals(Context.role(config))
-        || !executionState.getEnviron().equals(Context.environ(config))) {
-      LOG.severe("cluster/role/environ not matched");
-      return false;
+    // Check whether cluster/role/environ matched. It is possible to get into a state where some of
+    // the data does not actually exist in zookeeper. If we are in this state them assume the
+    // runtime is valid.
+    if (adaptor.doesExecutionStateExist(topologyName)) {
```
billonahill: if execution state doesn't exist we return true, which doesn't seem like what we want here no?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -365,7 +369,7 @@ protected boolean callRuntimeManagerRunner(Config runtime, ISchedulerClient sche
     RuntimeManagerRunner runtimeManagerRunner =
         new RuntimeManagerRunner(config, runtime, command, schedulerClient);
 
-    // invoke the appropriate handlers based on command
+    S// invoke the appropriate handlers based on command
```
billonahill: typo

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -179,6 +189,63 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+    // Just log but don't consider them failure
+    result = deleteTMasterLocation(topologyName);
+    if (result == null || !result) {
+      // We would not return false since it is possible that TMaster didn't write physical plan
+      LOG.warning("Failed to clear TMaster location. Check whether TMaster set it correctly.");
+    }
+
+    result = deletePhysicalPlan(topologyName);
+    if (result == null || !result) {
+      // We would not return false since it is possible that TMaster didn't write physical plan
```
billonahill: please remove this and the 2 other (previously) copy/pasted comments that are erroneous.

heron/statemgrs/src/java/com/twitter/heron/statemgr/FileSystemStateManager.java
```diff
@@ -154,10 +154,40 @@ public void initialize(Config config) {
...
+  }
+
+  @Override
+  public ListenableFuture<Boolean> doesPhysicalPlanExists(
+      String topologyName) {
+    return nodeExists(getPhysicalPlanPath(topologyName));
+  }
+
+  @Override
+  public ListenableFuture<Boolean> isTopologyRunning(String topologyName) {
```
billonahill: The semantics here are a bit loose, since we could be in a bad state where processes are running bet the state manager has state. Could we remove this method an use doesTopologyExists instead?

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -191,4 +191,39 @@
...
+   * @return Boolean
+   */
+  ListenableFuture<Boolean> doesTMasterLocationExists(String topologyName);
+
+  /**
+   * Checks whether the scheduler location for the given topology exists
+   *
+   * @return Boolean
+   */
+  ListenableFuture<Boolean> doesSchedulerLocationExists(String topologyName);
```
billonahill: open question, but do we need 'does'? in these methods (e.g., `boolean schedulerLocationExists(..)`).

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
ajorgensen: I think it is, if the execution state doesn't exist we cant properly validate the runtime so my proposal is that we default to true. If we return false the command will fail and we'd be in the same situation where the command will fail to run (see: https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java#L309).

The manifestation of the bug is that the execution state got deleted from zookeeper, since that happens we cannot really do this validation step. Do you know what this validation was meant to protect against? It seems to be that its meant to make it so you cannot deploy a topology twice to two different cluster, role, environments correct? Can we always assume that the user does not want to do that?

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
ajorgensen: No I think you're right, `schedulerLocationExists` reads better.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
billonahill: Looking closer at the code I think this method would be better named ` validatePrecommandState` or something along those lines. It's basically asserting that the command is ok to be executed based on alignment of client state with zookeeper. If executionState doesn't exist in ZK most commands probably should not be run (active, deactivate, restart). If we want to permit `kill` to run, then I suggest we change this method to also take the command and use that is an input.

cc/ @maosongfu for his thoughts as well.

ajorgensen: I see what you're saying now. So the idea would be that if for some reason the execution state was gone then some commands (restart, active, deactivate) would just simply fail and would require the topology to be killed and then submitted it order to use those commands?

 A larger change that could be made would be to heron itself write the execution state. The topology master could write the execution state as an ephemeral node in zookeeper so that its always there if the topology is running. This would fix the edge case around deleting that piece of state from zookeeper and might simplify things a bit.

billonahill: +1 to your first paragraph. We could limit this change to say that unless the command is a `kill` we should require valid system state. Kill should hard-kill everything and clean out all state.

I don't think the 2nd paragraph suggestion would work, since AFAIK execution state needs to persist across process failures so ephemeral nodes wouldn't work. Those are better suited for state that truly should disappear when the process dies, like tmaster location.

ajorgensen: Yeah. It doesn't feel great to tell people they have to restart their topology in order to do some commands. In the end its probably not the end of the world but feels like there might be a different change to automatically recover from this scenario. The nice part about possibly having the tmaster manager that state is that it can be restored because its encoded in the arguments or in a protobuf object somewhere that could be re-used to automatically recover and restore the missing state.

That being said I think we can make an incremental improvement here and think more about some sort of recovery mechanism.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -179,6 +189,60 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+  public boolean doesExecutionStateExist(String topologyName) {
+    return awaitResult(delegate.executionStateExists(topologyName));
+  }
+
+  /**
+   * Clean all states of a heron topology
+   * 1. Topology def and ExecutionState are required to exist to delete
+   * 2. TMasterLocation, SchedulerLocation and PhysicalPlan may not exist to delete
+   */
+  public Boolean cleanState(String topologyName) {
```
maosongfu: Not sure whether to put `cleanState(...)` here or in the Utils class since it is more like an utils method, providing an aggregation of primitive methods.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
```diff
@@ -157,8 +157,10 @@ public Boolean call() {
 
     // launch the topology, clear the state if it fails
     if (!launcher.launch(packedPlan)) {
-      statemgr.deleteExecutionState(topologyName);
-      statemgr.deleteTopology(topologyName);
+      if (!statemgr.cleanState(topologyName)) {
```
maosongfu: Need to clean all state? Or only explicitly clean execution_state and topology?

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -122,7 +122,9 @@ public Boolean setSchedulerLocation(
    * @return Boolean - Success or Failure
    */
   public Boolean deleteTMasterLocation(String topologyName) {
-    return awaitResult(delegate.deleteTMasterLocation(topologyName));
+    return awaitResult(delegate.tmasterLocationExists(topologyName))
+        ? awaitResult(delegate.deleteTMasterLocation(topologyName))
+        : true;
```
maosongfu: This style of implementation could lead to one more remote call. Could we figure out a more-efficient way?
Similar to below.

kramasamy:
@ajorgensen - some syntax errors in travis-ci. please take a look.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
billonahill: I'm not a fan of util methods for things that do statefull operations on the system. IMO utils should be simple static helpers, like string or date helpers.

It's also good IMO to keep this as an instance method here to keep the logic together. If a new state method gets added to this class, it should be clear that this method also in this class needs to be updated. If cleanState were in some utility that could be lost.

maosongfu: sounds reasonable.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/LaunchRunner.java
ajorgensen: It's all state.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
ajorgensen: The other way I could do this is surround these all with try/catch blocks and specifically catch zookeeper exceptions. I dont know if there is a "safe" version of the zookeeper interaction where if the node does not exist it doesnt throw an exception but I can look.

heron/statemgrs/src/java/com/twitter/heron/statemgr/zookeeper/curator/CuratorStateManager.java
```diff
@@ -208,6 +209,10 @@ public String getConnectionString() {
       LOG.info("Deleted node for path: " + path);
       result.set(true);
 
+    } catch (KeeperException.NoNodeException e) {
+      LOG.info(path + " does not exist so there's nothing to delete");
```
billonahill: I think LOG.warn would be justified, since code shouldn't be calling delete on something that they don't expect to be set.

billonahill:
👍

kramasamy:
@ajorgensen - if everything is fine, can you merge it?

ajorgensen:
Looks like it, I missed the 👍 comment.

ajorgensen:
This implementation should be working now. I would like to get some feedback on the last commit (https://github.com/twitter/heron/pull/1128/commits/12258d10bb1038adc280a17f6b43bcafe1877836). We need to assume that the topology is valid in the kill case in order for the command not to fail. This was the quickest way of solving that I could find without doing a larger refactoring.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,68 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+   */
+  public boolean doesExecutionStateExist(String topologyName) {
+    return awaitResult(delegate.executionStateExists(topologyName));
+  }
+
+  /**
+   * Clean all states of a heron topology.
+   * TMasterLocation, PackingPlan, PhysicalPlan, SchedulerLocation, ExecutionState, and Topology
+   */
+  public Boolean cleanState(String topologyName) {
```
billonahill: The meaning of the return Boolean in ambiguous. Please define it in the javadocs.

Also most of the javadocs within this method are previously copy/pasted and invalid and could even be removed. I'd say we remove them all since they don't seem helpful.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,68 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+    result = deleteSchedulerLocation(topologyName);
+    if (result == null || !result) {
+      // We would not return false since it is possible that TMaster didn't write physical plan
+      LOG.warning("Failed to clear scheduler location.");
+    }
+
+    result = deleteExecutionState(topologyName);
+    if (result == null || !result) {
+      LOG.severe("Failed to clear execution state");
+      return false;
```
billonahill: how come we return false for execution state and topology but not the others?

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,68 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+    }
+
+    // Set topology def at last since we determine whether a topology is running
```
billonahill: What is meant by "Set topology def at last" here?

heron/statemgrs/src/java/com/twitter/heron/statemgr/zookeeper/curator/CuratorStateManager.java
```diff
@@ -211,6 +212,10 @@ public String getConnectionString() {
       LOG.info("Deleted node for path: " + path);
       result.set(true);
 
+    } catch (KeeperException.NoNodeException e) {
+      LOG.warning(path + " does not exist so there's nothing to delete");
```
billonahill: Let's include more context here. Like "deleteNode called on <path> but..."

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,68 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+   * TMasterLocation, PackingPlan, PhysicalPlan, SchedulerLocation, ExecutionState, and Topology
+   */
+  public Boolean cleanState(String topologyName) {
+    LOG.fine("Cleaning up topology state");
+
+    Boolean result;
+
+    // It is possible that  TMasterLocation, PhysicalPlan and SchedulerLocation are not set
+    // Just log but don't consider them failure
+    result = deleteTMasterLocation(topologyName);
```
billonahill: We also now have PackgingPlan in the state manager so please clean that up too.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -303,7 +303,7 @@ public boolean manageTopology() {
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      boolean isValid = validateRuntimeManage(adaptor, topologyName);
+      boolean isValid = (Command.KILL.equals(command)) || validateRuntimeManage(adaptor, topologyName);
```
billonahill: I think this is fine for KILL. It should be a blunt tool that cleans out any possible state it finds of the topology. It shouldn't fail when one part is missing and then leave behind another.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
ajorgensen: This might have been a poor merge conflict resolution, good catch.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
ajorgensen: I agree. This will actually still fail if the scheduler state is missing which isn't great but that seemed to be a bit more to unravel but I can see if theres a way to succeed in that case as well.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
ajorgensen: It is deleted a few lines below this (https://github.com/twitter/heron/pull/1128/files#diff-6564229e92c7d6359355807e020d0972R237)

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,59 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+   * to be cleaned up and will log out a warning if it could not be cleaned up properly.
+   * TMasterLocation, PackingPlan, PhysicalPlan, SchedulerLocation, ExecutionState, and Topology
+   * @param topologyName the name of the topology that we should clean the state for
+   */
+  public void cleanState(String topologyName) {
+    LOG.fine("Cleaning up topology state");
+
+    Boolean result;
+
+    result = deleteTMasterLocation(topologyName);
```
billonahill: We can eliminate a lot of dup lines of with a method like this:
```
private void logDeleteStatus(Boolean success, String name) {
  if (result == null || !result) {
      LOG.warning("Failed to clear " + name);
  }
}
...
logDeleteStatus(deleteTMasterLocation(topologyName), "TMaster location");
logDeleteStatus(...
```

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,31 @@ public Boolean deleteSchedulerLocation(String topologyName) {
   }
 
   /**
+   * Clean all states of a heron topology. This goes through each piece of state that needs
+   * to be cleaned up and will log out a warning if it could not be cleaned up properly.
+   * TMasterLocation, PackingPlan, PhysicalPlan, SchedulerLocation, ExecutionState, and Topology
+   * @param topologyName the name of the topology that we should clean the state for
+   */
+  public void cleanState(String topologyName) {
```
billonahill: We want to guard against new locations being added that are not included in this block. Now that we have the enum [FileSystemStateManager.StateLocation](https://github.com/twitter/heron/blob/f0472c3932f839657fd98e6d3f85c7e87f7b0982/heron/statemgrs/src/java/com/twitter/heron/statemgr/FileSystemStateManager.java#L42) we could iterate over all enums with a case statement for deleting each. If the case statement misses we could throw an exception, which would let us know that a new one was added but not included.

If we do that though, we might want to push this method into the `IStateManager` interface though. Thoughts?

billonahill: Actually, we could add delete to the enum so it really won't get missed:
```
    public void delete(IStateManager stateManager, String topology) {
      switch(this) {
        case TMASTER_LOCATION: stateManager.deleteTMasterLocation(topology);
        case ....
      }
    }
```

then we could do things like this:
```
for (StateLocation location : StateLocation.values()) {
  location.delete(stateManager, topologyName)
}
```

ajorgensen: I like that. I didn't see that new data structure, let me refactor this based on your suggestions.

ajorgensen:
@billonahill Made the changes you suggested and fixed all the tests. Let me know if it looks good to you.

kramasamy:
@ajorgensen @billonahill - is this good to go? I am trying to get 0.14.3 release quickly.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -51,6 +51,64 @@
...
+          stateManager.deletePhysicalPlan(topology);
+          break;
+        case EXECUTION_STATE:
+          stateManager.deleteExecutionState(topology);
+          break;
+        case SCHEDULER_LOCATION:
+          stateManager.deleteSchedulerLocation(topology);
+          break;
+        default:
+          // This should never be reached, nothing to do.
```
billonahill: This will be reached if someone adds a new Enum without updating the case statement. Please throw an `IllegalArguementException` (?) here. Bonus points for a way to catch such a condition at compile time, which I can't think of atm.

heron/spi/src/java/com/twitter/heron/spi/statemgr/SchedulerStateManagerAdaptor.java
```diff
@@ -209,6 +209,22 @@ public Boolean deleteSchedulerLocation(String topologyName) {
...
+    LOG.fine("Cleaning up topology state");
+
+    for (IStateManager.StateLocation stateLocation : IStateManager.StateLocation.values()) {
```
billonahill: My suggestion was to also move the loop into the enum in a `deleteAll()` method that then delegates to `delete()`.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
ajorgensen: It *might* also be easier if there was a `deleteState` interface on `IStateManager` that took this enum. All of the state lives in a single location so there really isnt a huge need to define all of the different delete methods for the different locations `deleteTMasterState`, `deleteSchedulerLocation`, etc.

kramasamy:
@ajorgensen @billonahill - are we getting close to complete this PR?

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
billonahill: If we have delete state that takes an enum then each implementer will need a switch on the enum, and will need to updated when a new location is introduced. By putting the switch inside the enum as a helper, than only that enum needs to be changes when the a new deleteFoo is added to the interface/enum, so that seems more ideal to me.

The idea is to minimize the number of not-always-known places that need to be updated when a new IStateManager method is introduced.

billonahill:
@kramasamy this PR isn't critical for 0.14.3 so no need for you to block on it.

kramasamy:
thanks @billonahill

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
ajorgensen: My thought was that we could remove the need for the switch by just
deleting based on the 'dir' property on enum. Let's just table that idea
for now.

-- 
Andrew Jorgensen
@ajorgensen

joelanford:
@ajorgensen - Did this (or another similar PR) get merged? I think I'm running into this problem in 0.14.5.

ajorgensen:
@joelanford as far as I know this is likely still an issue, the way the scheduler code was architected changed a bit and I didnt have time to update this PR so I closed it. The issue is still open https://github.com/twitter/heron/issues/1126 if you want to take a stab at it.

PR #1132, opened 2016-07-22
Increase count of python method lines requires for a docstring

billonahill:
Python doctsrings were required on all methods, which was strict and encouraged unnecessary 1-liner docstrings that only echo the method name. Increase the threshold to not require it for short methods.

Also cleaning up the overly verbose `statemanager.py`, which had more lines of comments than code, many of which were echoing well-named method names, or copy/pasted and incorrect.


objmagic:
👍

Issue #1130, opened 2016-07-22
IUploader is not included in heron-spi.jar


chris-pardy:
The heron spi jar that is built and distributed to Maven Central does not include the com.twitter.heron.spi.uploader package and the IUploader file. These are needed to extend Heron by adding a uploader.

lewiskan:
@chris-pardy @kramasamy Thanks for heads up, the 'heron-spi.jar' artifact uploaded to Maven Central is generated by the Heron release script.  We will need to check the release process to include the IUploader file.  I'm discussing with Karthik to resolve, thanks.

nlu90:
I just looked into the spi-jar genrule, only files in `utils` and `common` are included in the spi jar.

```
java_binary(
    name = "spi-unshaded",
    srcs = glob([
        "**/spi/utils/**/*.java",
        "**/spi/common/**/*.java",
    ]),
    deps = utils_deps_files,
)

genrule(
    name = "heron-spi-jar",
    srcs = [":spi-unshaded_deploy.jar"],
    outs = ["heron-spi.jar"],
    cmd  = "cp $< $@",
)
```

kramasamy:
@chris-pardy - this is fixed in the PR https://github.com/twitter/heron/pull/1133 and it will be available in the next release 0.14.2

PR #1133, opened 2016-07-22
package all files in spi directory for publishing

nlu90:
As mentioned in #1130, currently only java files under `common` and `utils` are published in spi.jar. We need to publish all files under `spi` directory.

### DIFF
--- a/heron/spi/src/java/BUILD
+++ b/heron/spi/src/java/BUILD
@@ -171,7 +171,13 @@ java_library(
 java_binary(
     name = "spi-unshaded",
     srcs = glob([
-        "**/spi/**/*.java",
+        "**/spi/common/**/*.java",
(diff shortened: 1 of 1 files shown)


nlu90:
The original changes is made in #937

heron/spi/src/java/BUILD
```diff
@@ -171,8 +171,7 @@ java_library(
 java_binary(
     name = "spi-unshaded",
     srcs = glob([
-        "**/spi/utils/**/*.java",
-        "**/spi/common/**/*.java",
+        "**/spi/**/*.java",
```
kramasamy: Instead of specifying everything under spi - I would suggest module by module - helps to keep track of it in a cleaner fashion.

kramasamy:
👍

PR #1134, opened 2016-07-22
heron executor refactor and unit tests

billonahill:
This review includes refactorings with the following changes:

- Refactor methods to make `heron_executor.py` easier to unit test
- Adds unit tests!
- Adds functionality to allow the instance distribution to be dynamically updated, although currently it is still statically defined at startup.
- Some cleanup for pylint happiness

This change will put us in a much better place to add state manager watches on the instance distribution, once we're ready to do so.

### DIFF
--- a/heron/executor/src/python/BUILD
+++ b/heron/executor/src/python/BUILD
@@ -2,8 +2,15 @@ package(default_visibility = ["//visibility:public"])
 
 load("/tools/rules/pex_rules", "pex_library", "pex_binary")
 
+pex_library(
+    name = "heron-executor-lib",
+    srcs = ["heron_executor.py"],
+    deps = ["//heron/statemgrs/src/python:zk-statemgr-py"],
+    reqs = ["pyyaml==3.10"],
+)
+
 pex_binary(
     name = "heron-executor",
-    srcs = ["heron-executor.py"],
-    reqs = ["pyyaml==3.10"],
+    srcs = ["heron_executor.py"],
+    deps = [":heron-executor-lib"],
 )
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -18,17 +18,16 @@ import atexit
 import datetime
 import os
 import sys
+import signal
 import subprocess
 import time
 import json
 import base64
-import signal
 import string
 import random
 import yaml
 
 def print_usage():
-  '''print usage'''
   print (
       "./heron-executor <shardid> <topname> <topid> <topdefnfile> "
       " <instance_distribution> <zknode> <zkroot> <tmaster_binary> <stmgr_binary> "
@@ -40,7 +39,6 @@ def print_usage():
       " <scheduler_classpath> <scheduler_port>")
 
 def do_print(statement):
-  """ print msesage """
   timestr = datetime.datetime.fromtimestamp(time.time()).strftime('%Y-%m-%d %H:%M:%S')
   print "%s: %s" % (timestr, statement)
   sys.stdout.flush()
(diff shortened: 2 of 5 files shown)


heron/executor/src/python/BUILD
```diff
@@ -2,8 +2,15 @@ package(default_visibility = ["//visibility:public"])
...
+    deps = ["//heron/statemgrs/src/python:zk-statemgr-py"],
+    reqs = ["pyyaml==3.10"],
+)
+
 pex_binary(
     name = "heron-executor",
-    srcs = ["heron-executor.py"],
-    reqs = ["pyyaml==3.10"],
+    srcs = ["heron_executor.py"],
+    deps = [":heron-executor-lib"],
```
kramasamy: Do we need this deps since the lib refers to the same file?

billonahill: yeah, this is a little awkward but we need `pex_library` since `pex_test` takes that as a dep, but can't test against a `pex_binary`. And both `pex_library` and `pex_binary` require `srcs` so we have to declare `heron_executor.py` in both.

heron/executor/src/python/heron_executor.py
```diff
@@ -174,17 +157,48 @@ def __init__(self, args):
...
+    # Read the heron_internals.yaml for logging dir
+    self.log_dir = self.load_logging_dir(self.heron_internals_config_file)
+
+    # these get set when we call update_instance_distribution
+    self.instance_distribution = {}
+    self.stmgr_ids = []
+    self.metricsmgr_ids = []
+    self.heron_shell_ids = []
+
+    # this will soon be refactored to get this from the state manager instead of args
```
maosongfu: Not directly related to this pull request. But I think it is better "this will soon be refactored to get this from the TMaster instead of args"

heron/executor/src/python/BUILD
kramasamy: alternatively - you could separate the main into main.py and keep the remaining code as heron-executor.py. In this case, you could compile heron-executor as lib and main.py as pex-binary with heron-executor as lib for linking. Feel free to punt this suggestion if it is not worthwhile.

billonahill: That's basically what I had started with when I refactored to `heron_executor_core.py` with just the main in `heron-executor.py`. At the time I felt it complicated working with the code so I was glad to consolidate back into just `heron_executor.py` so I think one source file is preferred over removing the dup srcs in the build targets.

kramasamy: ok - at some point it might be nice to split up this file into different files from a modularization perspective.

kramasamy: especially it takes 32 arguments in the main :(  - we should augment it with options to improve readability.

heron/executor/src/python/heron_executor.py
```diff
@@ -107,61 +98,53 @@ def atomic_write_file(path, content):
...
+  container. Based on the container id and the instance distribution, it determines if the container
+  is a master node or a worker node and it starts processes accordingly."""
   def __init__(self, args):
     self.max_runs = 100
     self.interval_between_runs = 10
     self.shard = int(args[1])
     self.topology_name = args[2]
     self.topology_id = args[3]
     self.topology_defn_file = args[4]
-    # args[5] is the instance distribution encoding
```
kramasamy: @billonahill - just wondering how the args[5] is processed in the new version?

heron/executor/src/python/BUILD
billonahill: Agreed, it's not pretty passing 32 args in there. That refactor can be handled separately.

heron/executor/src/python/heron_executor.py
billonahill: See line 170 below. args[5] gets passed to a method that parses it and sets each of these:
```
self.instance_distribution = {}
self.stmgr_ids = []
self.metricsmgr_ids = []
self.heron_shell_ids = []
```

Once we add support for the state manager listener, we'd instead register the listener to call that method.

kramasamy:
👍  @maosongfu do you have any feedback on this?

maosongfu:
:+1:  Looks good to me.

PR #1135, opened 2016-07-23
Added common modules for python topology support

taishi8117:
This PR includes basic common modules necessary for adding python topology support

Related PR: #1136, #1137, #1138

This PR depends on #1150.

PR #1137, opened 2016-07-23
Added pex_loader.py for adding python topology support

taishi8117:
This script can dynamically load a given pex file and classes.

Related PR: #1135, #1136, #1138


billonahill:
Could you add some unit tests for this class please.

kramasamy:
@taishi8117 - does one unit test enough to cover sufficiently all the cases?

heron/common/src/python/pex_loader.py
```diff
@@ -0,0 +1,114 @@
...
+  if include_deps:
+    for dep in _get_deps_list(abs_path_to_pex):
+      Log.debug("Add a new dependency to the path: " + dep)
+      sys.path.insert(0, os.path.join(abs_path_to_pex, dep))
+
+  Log.debug("Python path: " + str(sys.path))
+
+def resolve_heron_suffix_issue(abs_pex_path, class_path):
+  """Resolves duplicate package suffix problems
+
```
kramasamy: I was wondering - instead of starting with the name 'heron' - can we start with some scrambled version such as "A89109120120" - a UUID and the likelihood of it clashing is very very less, isn't? Just an idea - not sure how feasible it is?

taishi8117: @kramasamy I think in order to insert UUID, we will need to modify the contents of PEX file quite a lot (like bootstrapping and entrypoints etc), so might not be the best idea.

kramasamy:
👍

billonahill:
👍

PR #1138, opened 2016-07-23
Added common network modules for adding python topology support

taishi8117:
Related PR: #1135, #1136, #1137 
This is a working branch, and more codes will be added.


heron/common/src/python/constants.py
```diff
@@ -16,6 +16,9 @@
 MB = 1024 * KB
```
kramasamy: this files seems to be repeated in multiple PRs?

heron/common/src/python/network/heron_client.py
```diff
@@ -0,0 +1,295 @@
...
+import time
+from heron.common.src.python.network import HeronProtocol, REQID, StatusCode, OutgoingPacket
+from heron.common.src.python.log import Log
+import heron.common.src.python.constants as constants
+
+# pylint: disable=too-many-instance-attributes
+# pylint: disable=fixme
+class HeronClient(asyncore.dispatcher):
+  """Python implementation of HeronClient, using asyncore module"""
+  TIMEOUT_SEC = 30.0
```
kramasamy: Can we read this value from the configuration file? I am sure this constant exists already somewhere. Please do not hard code value.

heron/common/src/python/network/heron_client.py
```diff
@@ -0,0 +1,314 @@
...
+from abc import abstractmethod
+
+import time
+from heron.common.src.python.network import HeronProtocol, REQID, StatusCode, OutgoingPacket
+from heron.common.src.python.log import Log
+import heron.common.src.python.constants as constants
+
+# pylint: disable=too-many-instance-attributes
+# pylint: disable=fixme
+class HeronClient(asyncore.dispatcher):
```
billonahill: The Heron Client is the term used to describe the heron command line client that end users use (i.e. the heron cli). This seems to be a multi-purpose networking client. Can we name this differently to avoid confusion?

taishi8117: @billonahill Just discussed on this, but this name is consistent with Java implementation, and I think it's pretty clear that heron-cli and Heron Client are different from the code structure.

billonahill:
👍

objmagic:
looks very good to me. Please use formatter as much as possible.

PR #1140, opened 2016-07-24
Set PEX_ROOT to container working directory/.pex

kramasamy:
This is needed for YARN. When yarn users run headless, pex attempts to unzip files at /home and we get permission denied.

### DIFF
--- a/heron/executor/src/python/heron-executor.py
+++ b/heron/executor/src/python/heron-executor.py
@@ -115,7 +115,7 @@ def log_pid_for_process(process_name, pid):
 # pylint: disable=too-many-instance-attributes
 class HeronExecutor(object):
   """ Heron executor """
-  def __init__(self, args):
+  def __init__(self, args, shell_env):
     self.max_runs = 100
     self.interval_between_runs = 10
     self.shard = int(args[1])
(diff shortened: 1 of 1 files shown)


heron/executor/src/python/heron-executor.py
```diff
@@ -465,7 +470,14 @@ def main():
   if len(sys.argv) != 32:
     print_usage()
     sys.exit(1)
-  executor = HeronExecutor(sys.argv)
+
+  # get the current environment and augment it with PEX_ROOT
+  shell_env = os.environ.copy()
+  shell_env["PEX_ROOT"] = os.path.join(os.path.abspath('.'), ".pex")
```
billonahill: let's explain in a comment why this is being done.

heron/executor/src/python/heron-executor.py
```diff
@@ -465,7 +470,14 @@ def main():
...
+  for k, v in shell_env.iteritems():
+    print k, v
```
billonahill: Was this print just for testing or do you want to ship this? If the latter it should be do_print and the print line should provide more context around what's being being printed.

ashvina:
Thanks for the patch. I will testing it now.

ashvina:
LGTM :+1: 
I deployed a topology on a YARN cluster. `.pex` directory is created inside container's working directory. I do not see the `permission denied` error any more.

billonahill:
👍 once CI passes.

mycFelix:
GREAT!!!

Issue #1116, opened 2016-07-20
provide Java code formatter tool


caofangkun:

kramasamy:
@caofangkun - we already have Java Style. Please check https://github.com/twitter/heron/blob/master/tools/java/src/com/twitter/bazel/checkstyle/JavaCheckstyle.java whether it satisfies your need.

maosongfu:
@caofangkun We have a formatter tool for java style for IntelliJ IDEA, will check it into the repo.

billonahill:
We already have IntelliJ java styles here:
https://github.com/twitter/heron/blob/master/scripts/resources/idea/codeStyleSettings.xml

They're used when creating a heron project here:
https://github.com/twitter/heron/blob/master/scripts/setup-intellij.sh

PR #1143, opened 2016-07-25
#1116 provide check_java_code_style.sh and JavaCodeFormatter.xml for eclipse

caofangkun:


scripts/check_java_code_style.sh
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+DIR=`dirname $0`
+source ${DIR}/detect_os_type.sh
+
+bazel build --config=`platform` heron/spi/src/java:heron-spi --experimental_action_listener=tools/java:compile_java
```
billonahill: I don't think we need a bash script just to effectively do this:
```
bazel build --config=[config] heron/spi/src/java:heron-spi
```

The `experimental_action_listener` is registered for each platform so it should always run whenever building a java target.

kramasamy:
@caofangkun - can you please check what @billonahill suggested and let us know?

objmagic:
@caofangkun any progress on this?

billonahill:
@caofangkun could you add comments to the top of this file, or a brief readme that explains how it should be used.

scripts/setup-eclipse.sh
```diff
@@ -102,3 +102,6 @@ cat >> $classpath_file << 'EOF'
   <classpathentry kind="output" path="target/classes"/>
 </classpath>
 EOF
+
+echo "Set Java code formatter file in Eclipse as following steps:"
+echo "Properties –> Java Code Style –> Formatter –> Enable project specific settings and load 'scripts/resources/eclipse/JavaCodeFormatter.xml'"
```
billonahill: Can this not be set by the script somehow, maybe in the project file. IntelliJ supports something like that.

PR #1144, opened 2016-07-25
The doc does not match paper nor the source code. Correct it correspondingly.

evenluo:


maosongfu:
Nice catch. Thanks for the contribution. Could you sign the CLA first?

evenluo:
Signed

maosongfu:
Thanks.

PR #1149, opened 2016-07-25
Modified topology.proto to support for Python topology

taishi8117:
Modified topology.proto so a component object can be defined in java or python. 
It is backward-compatible and easily extensible to add a support for another language.

This change is necessary for #1138

### DIFF
--- a/heron/proto/topology.proto
+++ b/heron/proto/topology.proto
@@ -77,7 +77,11 @@ message Config {
 message Component {
   // The name of the component. Like Tail-FlatMap
   required string name = 1;
-  // Only one of the below two are set
+  // Only one of the below two is set, whose meaning is determined
+  // by the value of spec. For example, when spec is JAVA_CLASS_NAME,
+  // `class_name` is set and contains Java's class name.
+  // When spec is JAVA_SERIALIZED_OBJECT, `serialized_object` is set and
+  // contains a serialized Java object.
   optional string class_name = 2;
   optional bytes serialized_object = 4;
 
--- a/heron/proto/topology.proto
+++ b/heron/proto/topology.proto
@@ -33,9 +33,10 @@ enum Type {
   OBJECT = 1; // generic type
 }
 
-enum Language {
-  JAVA = 1;
-  PYTHON = 2;
+enum ComponentObjectSpec {
+  JAVA_CLASS_NAME = 1;
+  JAVA_SERIALIZED_OBJECT = 2;
+  PYTHON_CLASS_NAME = 3;
 }
 
 message StreamSchema {
@@ -76,11 +77,12 @@ message Config {
 message Component {
   // The name of the component. Like Tail-FlatMap
   required string name = 1;
-  required Language lang = 5;
-  // Only one of the below three are set
+  // Only one of the below two are set
   optional string class_name = 2;
   optional bytes serialized_object = 4;
+
   required Config config = 3;
+  required ComponentObjectSpec spec = 5;
 }
 
 message Spout {
--- a/heron/proto/topology.proto
+++ b/heron/proto/topology.proto
@@ -33,6 +33,11 @@ enum Type {
   OBJECT = 1; // generic type
 }
 
+enum Language {
+  JAVA = 1;
+  PYTHON = 2;
+}
+
 message StreamSchema {
   message KeyType {
     required string key = 1;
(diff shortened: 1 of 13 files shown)


heron/proto/topology.proto
```diff
@@ -71,9 +71,10 @@ message Config {
 message Component {
   // The name of the component. Like Tail-FlatMap
   required string name = 1;
-  // Only one of the below two are set
+  // Only one of the below three are set
```
maosongfu: Could we change "jave_class_name" and "jave_object" to a more generic name, rather than adding a new filed "python_class_name", which is not extensible.

heron/proto/topology.proto
```diff
@@ -71,9 +71,10 @@ message Config {
...
   optional string java_class_name = 2;
   optional bytes java_object = 4;
+  optional string python_class_name = 5;
```
billonahill: Could we use a Union type here with the enum discriminator? That would clarify which one is set.
https://developers.google.com/protocol-buffers/docs/techniques#union

maosongfu: +1

billonahill: Or what I just saw that @maosongfu suggested, but either way we should have decent typing for what kind of object we're specifying via an enum or other means.

kramasamy:
In addition to this - we might need additional fields to indicate - 

- Indicate whether the topology is written in C++/Java/Python/Go etc
- Higher level framework used as Summingbird, SQL, etc

This will be helpful later I believe.

billonahill:
@kramasamy I suggest we limit this change to what's needed to manage the python/java processes for now and follow up with the other higher-level info after clarifying our use cases around that more.

heron/proto/topology.proto
taishi8117: @maosongfu @billonahill Ok, will do that.

kramasamy:
@billonahill - I was more worried about the migration within Twitter.  If we can consolidate the changes into one release - it makes it little less painful.

taishi8117:
@kramasamy @billonahill I added an enum to specify language, in which I added Java and Python for now, but it's easily extensible by adding new languages to the enum later.

billonahill:
I was advocating not blocking/coupling this PR with the feature you're talking about. It's best to separate features into different PRs/issue for better tracking.

We can put other things in this release if we want to, but let's deal with it separately.

heron/proto/topology.proto
```diff
@@ -71,9 +76,10 @@ message Config {
 message Component {
   // The name of the component. Like Tail-FlatMap
   required string name = 1;
-  // Only one of the below two are set
-  optional string java_class_name = 2;
-  optional bytes java_object = 4;
+  required Language lang = 5;
+  // Only one of the below three are set
```
billonahill: Just paired on this. Will change Language to ComponentObjectSpec.

kramasamy:
Fine with me.

heron/proto/topology.proto
```diff
@@ -72,9 +78,11 @@ message Component {
   // The name of the component. Like Tail-FlatMap
   required string name = 1;
   // Only one of the below two are set
```
billonahill: let's clarify this comment to describe how what's set below, and it's meaning can be determined by the value of spec.

billonahill:
👍

kramasamy:
@taishi8117 - can you look at the tracker code as well to ensure these changes do not affect? This is because tracker code use these protos as well.

maosongfu:
looks good to me. Please make sure the change does not break any existing components.

kramasamy:
👍

objmagic:
👍

heron/proto/topology.proto
```diff
@@ -71,10 +77,16 @@ message Config {
...
+  // Only one of the below two is set, whose meaning is determined
+  // by the value of spec. For example, when spec is JAVA_CLASS_NAME,
+  // `class_name` is set and contains Java's class name.
+  // When spec is JAVA_SERIALIZED_OBJECT, `serialized_object` is set and
+  // contains a serialized Java object.
+  optional string class_name = 2;
+  optional bytes serialized_object = 4;
+
   required Config config = 3;
+  required ComponentObjectSpec spec = 5;
```
nlu90: Adding required field to an existing proto file should always be prohibited. Hope we can find the problem before it's merged in the future.

taishi8117: @nlu90 Sorry for this, but #1180 fixed this problem.

nlu90: @taishi8117 thanks for the fix.

PR #1150, opened 2016-07-25
Added constants for system config and topology config

taishi8117:
This is necessary for adding python topology support.

### DIFF
--- a/heron/common/src/python/constants.py
+++ b/heron/common/src/python/constants.py
@@ -87,7 +87,7 @@ TOPOLOGY_ADDITIONAL_CLASSPATH = "topology.additional.classpath"
 
 
 ####################################################################################################
-#############################  Constants for system configuration ##################################
+#############################  Constants for System configuration ##################################
 ####################################################################################################
 
 # The relative path to the logging directory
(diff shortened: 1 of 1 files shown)


objmagic:
👍

heron/common/src/python/constants.py
```diff
@@ -11,20 +11,191 @@
...
+HERON_LOGGING_DIRECTORY = 'heron.logging.directory'
+# The maximum log file size in MB
+HERON_LOGGING_MAXIMUM_SIZE_MB = "heron.logging.maximum.size.mb"
+# The maximum number of log files
+HERON_LOGGING_MAXIMUM_FILES = "heron.logging.maximum.files"
+# The threshold level to log error
+HERON_LOGGING_ERR_THRESHOLD = "heron.logging.err.threshold"
+# The interval in seconds to get and reset the system metrics.
+HERON_METRICS_EXPORT_INTERVAL_SEC = "heron.metrics.export.interval.sec"
+
```
kramasamy: Can you add a huge separator comment between big groups of configuration? For example, before instance specific configuration - can you add a huge comment so that when we read the file we can get to specific group quickly?

heron/common/src/python/constants.py
```diff
@@ -11,20 +11,191 @@
...
+# During dynamically tuning, the weight of new sample size to calculate
+# the available capacity of queue.
+# We use Exponential moving average: En = (1-w) * En-1 + w * An
+# http://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average
+INSTANCE_TUNING_CURRENT_SAMPLE_WEIGHT = "heron.instance.tuning.current.sample.weight"
+
+# Interval in ms to tune the size of in &amp; out data queue in instance
+INSTANCE_TUNING_INTERVAL_MS = "heron.instance.tuning.interval.ms"
+
+# Time based, the maximum batch time in ms for instance to read from socket per attempt
```
kramasamy: same as before?

kramasamy:
👍

PR #1161, opened 2016-07-26
add marathon scheduler for heron

nlu90:
This PR adds Marathon Scheduler for Heron.


maosongfu:
Can u add the unit tests for the Marathon Scheduler?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,198 @@
...
+      conn.disconnect();
+    }
+
+    return true;
+  }
+
+  public boolean restartApp(int appId) {
+    if (appId == -1) {
+      // TODO (nlu): implement restart all
+      throw new RuntimeException("Restart all containers not supported yet");
```
maosongfu: Can we simply kill the whole group and resubmit the whole group in this case?

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -438,4 +438,20 @@ public static boolean curlAndExtractPackage(
 
     return true;
   }
+
+  /**
+   * This method finds the container with highest resource requirement and returns the resource.
+   * Currently only RAM is used for max identification.
+   */
+  public static PackingPlan.Resource getMaxRequiredResource(PackingPlan packingPlan) {
```
maosongfu: So only RAM is taken into account? Should we rename it to reduce confusion?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  // submit a topology as a group, containers as apps in the group
+  public boolean submitTopology(String appConf) {
+    if (this.isVerbose) {
+      LOG.log(Level.INFO, "Topology conf is: " + appConf);
+    }
+
+    // Setup Connection
+    String schedulerURI = String.format("%s/v2/groups", this.marathonURI);
```
maosongfu: Why not use or refactor the utils method already exists?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -131,7 +132,7 @@ protected String getHeronAuroraPath() {
     TopologyAPI.Topology topology = Runtime.topology(runtime);
 
     // Align the ram to the maximal one
-    PackingPlan.Resource containerResource = getMaxRequiredResource(packing);
+    PackingPlan.Resource containerResource = SchedulerUtils.getMaxRequiredResource(packing);
```
maosongfu: It is better to put this change into a separate pull request.

maosongfu:
Can we also add docs for this Marathon Scheduler?

kramasamy:
@maosongfu - can the documentation be added in a separate PR? @nlu90 - please create another issue to track this.

nlu90:
@kramasamy Created the documentation issue #1173

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: I left this unimplemented because I could not find a marathon api to restart a complete group and wondering if we can get some instructions from the Marathon people.

Yes, we can kill the group and then resubmit it. The only thing is we need to fetch the group conf first, but luckily marathon provide such api.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
nlu90: This method is moved from AuroraScheduler, maybe we should put this refactor into a separate PR.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: Those utils methods need to be refactored for usage. And to reduce the development time, I choose to write my own method for the initial commit. 

I'll refactor and use them in the following update.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
mhausenblas: Yes, see https://mesosphere.github.io/marathon/docs/generated/api.html#v2_groups_delete

nlu90: @mhausenblas, Is there any marathon api to get the application group JSON file when the group is sbumitted?

I tried `GET /v2/groups/{groupId}`, but the returned result is not the original JSON file, it modifies the "id" fields, adds some extra information that not accepted during submission.

kramasamy:
@nlu90 - you might have to resolve conflicts.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
kramasamy: can you open a PR for this?

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
```diff
@@ -142,27 +147,32 @@ public static boolean sendHttpPostRequest(HttpURLConnection connection,
...
-        LOG.log(Level.SEVERE, "Failed to close OutputStream: ", e);
+        LOG.log(Level.SEVERE, "Failed to send request: ", e);
         return false;
+      } finally {
+        try {
+          if (os != null) {
+            os.close();
+          }
+        } catch (IOException e) {
+          LOG.log(Level.SEVERE, "Failed to close OutputStream: ", e);
```
osgigeek: What is the expectation here for logging this message? It becomes very difficult to know if there was a real issue with a stream which cannot be closed or failed on housekeeping tasks to cleanup resources. Preferably we should just use something like `IOUtils.closeQuietly(os)`

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
```diff
@@ -229,9 +250,25 @@ public static boolean sendHttpGetRequest(HttpURLConnection connection) {
...
   }
 
-  public static HttpURLConnection getConnection(String endpoint) throws IOException {
-    URL url = new URL(endpoint);
-    return (HttpURLConnection) url.openConnection();
+  public static HttpURLConnection getHttpConnection(String endpoint) {
+    // construct the http url connection
+    try {
+      URL url = new URL(endpoint);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
osgigeek: We need some defensive coding here before assuming that the connection is a HTTPURLConnection

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
osgigeek: @nlu90 if you refactor the methods it would be good to keep the connection creation (preferably using a connection pool) within the class which is doing all the additional work of setting Request-type, etc. That might also mean not having the flexibility to use static methods.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+    String completeSchedulerProcessClassPath = new StringBuilder()
+        .append(Context.schedulerSandboxClassPath(config)).append(":")
+        .append(Context.packingSandboxClassPath(config)).append(":")
+        .append(Context.stateManagerSandboxClassPath(config))
+        .toString();
+    commands.add(completeSchedulerProcessClassPath);
+    commands.add(MarathonConstants.SCHEDULER_PORT);
+
+    return Joiner.on(" ").join(commands);
+  }
```
ashvina: This method duplicate most of the code from `SchedulerUtils.executorCommandArgs`. Can we reuse the original method? It may need some refactoring though.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+    if (!NetworkUtils.sendHttpPostRequest(conn, NetworkUtils.JSON_TYPE, appConf.getBytes())) {
+      LOG.log(Level.SEVERE, "Failed to ");
+      conn.disconnect();
+      return false;
+    }
+
+    // Check response
+    boolean success = NetworkUtils.checkHttpResponseCode(conn, HttpURLConnection.HTTP_CREATED);
+    // Disconnect to release resources
+    conn.disconnect();
```
ashvina: Should disconnect be put in a finally block?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+    HttpURLConnection conn = NetworkUtils.getHttpConnection(topologyURI);
+    if (conn == null) {
+      LOG.log(Level.SEVERE, "Failed to find marathon scheduler");
+      return false;
+    }
+
+    // Send kill topology request
+    if (!NetworkUtils.sendHttpDeleteRequest(conn)) {
+      LOG.log(Level.SEVERE, "Failed to set delete request");
+      return false;
```
ashvina: connection.disconnect?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    // Check response
+    boolean success = NetworkUtils.checkHttpResponseCode(conn, HttpURLConnection.HTTP_OK);
+    // Disconnect to release resources
+    conn.disconnect();
```
ashvina: I think a finally block is needed for disconnect?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+
+package com.twitter.heron.scheduler.marathon;
+
+import java.net.HttpURLConnection;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.twitter.heron.spi.utils.NetworkUtils;
+
+public class MarathonController {
```
ashvina: I am not sure I understand why this class called a `Controller`. It seems more like a "util" class to me. Moreover the http calls and related code could actually be moved to the `MarathonScheduler`. Also there is sole level of code duplication between the methods.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  public boolean restartApp(int appId) {
+    if (appId == -1) {
+      // TODO (nlu): implement restart all
+      String message = "Restarting the whole topology is not supported yet. "
+          + "Please kill and resubmit the topology.";
+      LOG.log(Level.SEVERE, message);
+      return true;
```
ashvina: Should the return value be false to let the user know that `restartAll` failed?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonScheduler.java
nlu90: Yeah, I can refactor the method to be reused

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
nlu90: Sounds good, we can make the close() be quiet.

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
```diff
@@ -0,0 +1,143 @@
...
+  ) {
+    this.marathonURI = marathonURI;
+    this.topologyName = topologyName;
+    this.isVerbose = isVerbose;
+  }
+
+  public boolean killTopology() {
+    // Setup Connection
+    String topologyURI = String.format("%s/v2/groups/%s", this.marathonURI, this.topologyName);
+    HttpURLConnection conn = NetworkUtils.getHttpConnection(topologyURI);
```
ashvina: I am not an expert. So wanted to clarify one doubt. Can Marathon service be SSL enabled? If so do we need to provide support for https?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: Do you mean, instead of do disconnect after each failure case, just `try {} finally {conn.disconnect()}`?

One concern is, all the methods called here don't throw any exception. In this case, should we still use `try` ?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: @kramasamy @osgigeek the refactor is done in this PR. For each request type, we have different util methods taking care of all the additional works

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
ashvina: I was thinking more like the first one; one try finally for each method. 

You are correct, handling IOException captures most of the error cases. A try-finally is a generally accepted pattern whenever resources are concerned. I would suggest try-finally just to be sure no return paths are executed without closing the connection.

nlu90: sounds good, I'll update to use the try-finally pattern

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: The problem with returning false here is the RuntimeManagerMain would throw exceptions claiming the operation failed. If throwing exception instead of show some log message is acceptable, we can return true

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: I'm also not very sure. @mhausenblas could you explain the SSL a little bit?

heron/schedulers/src/java/com/twitter/heron/scheduler/marathon/MarathonController.java
nlu90: The `Controller` is the one do the actual action. The reason I moved the logic out is to make the code more modularized and understandable.

ashvina:
LGTM 👍 

@nlu90 do you think it is worth creating an issue for SSL?

kramasamy:
👍 As @ashvina mentioned, please create the SSL issue - which we can resolve later.

PR #1163, opened 2016-07-26
Tracker and UI with version

### LINKED ISSUE #1155: add version for heron-tracker and heron-ui

objmagic:
This PR adds `version` subcommand for `heron-tracker` and `heron-ui`.

The following commands are supported:

````
heron-tracker version
````

````
heron-ui version
````

I made one change that needs to be explained: [``zip_safe`` is removed](https://github.com/twitter/heron/compare/master...objmagic:tracker-ui-with-version?expand=1#diff-96c1a009a7acd76ed67927c481c0ade4L32). Why? If this option is set to false, program will first be unzipped to ``~/.pex`` and then be executed. This means that [``get_heron_release_file``](https://github.com/twitter/heron/blob/master/heron/common/src/python/utils.py#L184) will fail to find release file. See [comment of ``zip_safe``](https://pex.readthedocs.io/en/stable/_modules/pex/pex_info.html) for more details. I also want to know why ``zip_safe`` is set to ``False`` at first.

resolves #1155

### CODE REVIEW

File: heron/tracker/src/python/version.py [hunk 1]
```diff
@@ -0,0 +1,31 @@
+# Copyright 2016 Twitter. All rights reserved.
```
kramasamy: can you move this common area - if everybody is interested in adding version information? heron-tracker, ui, cli, explorer, packer, etc.

heron/ui/src/python/version.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2016 Twitter. All rights reserved.
```
kramasamy: This is duplicated code - we need to find a home in common

heron/tracker/src/python/version.py
objmagic: problem is that implementation of argument parser of each CLI tool is different (because they are written by different people). You cannot reuse this ``version.py`` for `heron` and `heron-explorer`.

heron/ui/src/python/version.py
objmagic: putting them into ``utils`` module under ``heron/common`` looks strange to me. I can remove these ``version.py`` and put this simple logic of adding parser inside ``main.py``.

objmagic:
@kramasamy Refactored.

kramasamy:
👍

PR #1167, opened 2016-07-27
Pex without interpreter cache disabled

objmagic:
This lovely PR disables interpreter cache because [resolver.py L260-L263](https://github.com/twitter/heron/blob/master/third_party/pex/pex/resolver.py#L260-L263) is known to fail build due to some caching problem.

cf. https://github.com/pantsbuild/pex/issues/293

cc @kramasamy @billonahill

### DIFF
--- a/tools/rules/pex_rules.bzl
+++ b/tools/rules/pex_rules.bzl
@@ -98,7 +98,8 @@ def make_manifest(ctx, output):
     command = ('touch %s && echo "%s" > %s' % (output.path, manifest_text, output.path)))
 
 def common_pex_arguments(entry_point, deploy_pex_path, manifest_file_path):
-  arguments = ['--entry-point', entry_point]
+  arguments = ['--disable-cache']
+  arguments += ['--entry-point', entry_point]
 
   # Our internal build environment requires extra args injected and this is a brutal hack. Ideally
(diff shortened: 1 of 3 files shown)


tools/rules/pex_rules.bzl
```diff
@@ -169,12 +170,15 @@ def pex_test_impl(ctx):
   pex_test_files = pex_test_file_types.filter(pex_file_types.filter(transitive_sources))
   test_run_args = ' '.join([f.path for f in pex_test_files])
 
+  arguments = ["--disable-cache"]
```
billonahill: could we instead set this in common_pex_arguments() so we don't have to set it in two places?

kramasamy:
👍

billonahill:
👍 for code changes, but CI is failing.

objmagic:
>
W: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install gcc-4.8 g++-4.8 python2.7 wget pkg-config zip zlib1g-dev" failed and exited with 100 during .

restarted CI.

objmagic:
merged since CI is OK

objmagic:
@benley you might be interested in this..

benley:
indeed, thanks for the mention :)

PR #849, opened 2016-06-01
Alan/pylint rcfile

alanngai:
updated pr with last set of feedback from https://github.com/twitter/heron/pull/780.  Remaining potential issues

heron/ui/src/python/handler/api/topology.py: 
W:231, 0: FIXME: is there a purpose to saving the return value of this? (fixme)
W:234, 0: FIXME: estate not defined (fixme)
E:240,13: Undefined variable 'estate' (undefined-variable)
W:232, 4: Unused variable 'scheduler_location' (unused-variable)

---- pylint output for heron/cli ------------------------------------------------------------------
************* Module main
I: 42, 0: Locally disabling protected-access (W0212) (locally-disabled)
I: 64, 0: Locally disabling bad-super-call (E1003) (locally-disabled)
************* Module color
I: 18, 0: Locally disabling invalid-name (C0103) (locally-disabled)
************* Module log
I: 19, 0: Locally disabling invalid-name (C0103) (locally-disabled)
************* Module activate
I: 24, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module args
I: 24, 0: Locally disabling protected-access (W0212) (locally-disabled)
************* Module deactivate
I: 24, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module execute
W: 87, 0: TODO: We need to extract heron-core to separate folder. (fixme)
************* Module help
I: 27, 0: Locally disabling protected-access (W0212) (locally-disabled)
I: 40, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module jars
W: 21, 0: TODO, make the jars independent version free (fixme)
W: 33, 0: TODO, make the jars independent version free (fixme)
************* Module kill
I: 24, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module opts
I: 19, 0: Locally disabling invalid-name (C0103) (locally-disabled)
I: 19, 0: Locally disabling global-variable-not-assigned (W0602) (locally-disabled)
I: 19, 0: Locally disabling global-statement (W0603) (locally-disabled)
************* Module restart
I: 50, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module submit
I:234, 0: Locally disabling unused-argument (W0613) (locally-disabled)
************* Module utils
I: 86, 0: Locally disabling protected-access (W0212) (locally-disabled)
R:218, 0: Too many branches (14/12) (too-many-branches)
************* Module version
[...]

PR #1169, opened 2016-07-27
Better logging for tracker

### LINKED ISSUE #1072: Follow-up of Pylint (#849)

objmagic:
1. use Log module under `heron/common`
2. address https://github.com/twitter/heron/pull/849#issuecomment-231932982. Do not print unnecessary stack trace. Resolve #1072 completely.

### DIFF
--- a/heron/tracker/src/python/handlers/containerfilehandler.py
+++ b/heron/tracker/src/python/handlers/containerfilehandler.py
@@ -18,6 +18,7 @@ import tornado.gen
 from heron.tracker.src.python.handlers import BaseHandler
 from heron.tracker.src.python import constants
 from heron.tracker.src.python import utils
+from heron.common.src.python.color import Log
 
 
 # pylint: disable=attribute-defined-outside-init
@@ -69,7 +70,7 @@ class ContainerFileDataHandler(BaseHandler):
       self.write_success_response(json.loads(response.body))
       self.finish()
     except Exception as e:
-      traceback.print_exc()
+      Log.debug(traceback.format_exc())
       self.write_error_response(e)
 
 
@@ -116,5 +117,5 @@ class ContainerFileStatsHandler(BaseHandler):
       self.write_success_response(json.loads(response.body))
       self.finish()
     except Exception as e:
-      traceback.print_exc()
+      Log.debug(traceback.format_exc())
       self.write_error_response(e)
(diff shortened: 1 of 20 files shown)


heron/tracker/src/python/handlers/containerfilehandler.py
```diff
@@ -69,7 +70,7 @@ def get(self):
       self.write_success_response(json.loads(response.body))
       self.finish()
     except Exception as e:
-      traceback.print_exc()
+      Log.debug(traceback.format_exc())
```
kramasamy: When you log.debug - it prints only when the verbose option is enabled?

objmagic: Correct. See: [main.py#L204](https://github.com/twitter/heron/blob/master/heron/tracker/src/python/main.py#L204)

heron/tracker/src/python/metricstimeline.py
```diff
@@ -112,7 +110,7 @@ def getMetricsTimeline(tmaster,
 
   if metricResponse.status.status == common_pb2.NOTOK:
     if metricResponse.status.HasField("message"):
-      raise Exception(metricResponse.status.message)
+      Log.error(metricResponse.status.message)
```
kramasamy: What is the difference here? In case raise Exception and terminate - where as in the Log.error case, it will print the error and continue with the execution?

objmagic: In Tornado application, if an exception is raised, it's just raised (stack trace could be printed out) but program will not terminate. If we do `Log.error`, the program continue execution and [return an empty `ret` value without metrics information](https://github.com/twitter/heron/pull/1169/files/77f9e4362294273e1bd8a2e51d6f3cea9c8f2556#diff-bb948157f88754a9d9b0b27a2df041dbR142), which is ok for the UI side.

kramasamy:
👍

PR #1174, opened 2016-07-27
Don't keep STDIN open if not attached

ajorgensen:
The `-i` argument for docker keeps STDIN open even when the container is
not attached. In some ci environments this will likely by the case and
result in this error: `cannot enable tty mode on non tty input`

This also changes the scratch and output directory creation to use `-p`
so it will both create the whole path and not fail if the directory
exists.


kramasamy:
👍

PR #1175, opened 2016-07-28
Revert UI with version

objmagic:
This PR reverts UI part of the change in #1163 to make UI works on master. To make `heron-ui` has `version` subcommand, further investigation is needed because of some complicated issues which I had discussed with @kramasamy, @billonahill and @nlu90.

**I'm very sorry about merging #1163 into master which makes `heron-ui` break**


nlu90:
👍

PR #1176, opened 2016-07-28
refactor the getMaxResource method

nlu90:
As mentioned in #1161, put the getMaxResource refactor into a separate PR.

The new getMaxRequiredResource will find the max requirement for cpu, ram and disk from all the required container resources.

MesosScheduler and AuroraScheduler will use the new method to decide their resource allocation to each container.

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
@@ -438,4 +438,20 @@ public final class SchedulerUtils {
 
     return true;
   }
+
+  /**
+   * Get a resource that requires the maximum amount of ram, cpu and disk
+   */
+  public static PackingPlan.Resource getMaxRequiredResource(PackingPlan packingPlan) {
+    PackingPlan.Resource maxResource = new PackingPlan.Resource(0, 0, 0);
+
+    for (PackingPlan.ContainerPlan entry : packingPlan.containers.values()) {
+      PackingPlan.Resource resource = entry.resource;
+      maxResource.ram = Math.max(maxResource.ram, resource.ram);
+      maxResource.cpu = Math.max(maxResource.cpu, resource.cpu);
+      maxResource.disk = Math.max(maxResource.disk, resource.disk);
+    }
+
+    return maxResource;
+  }
 }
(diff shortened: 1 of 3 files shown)


objmagic:
👍

billonahill:
LGTM

@avflor and @ashvina you were also doing some similar refactors lately. Would you please review.

avflor:
@nlu90 @billonahill The change looks good to me. The only difference is that in the original code Aurora takes only memory into account when computing the max resources whereas Mesos considers all the resources. I think it is more correct to use all resources as done in this PR.

nlu90:
@avflor yeah, as we discussed in a previous refactor PR #1079 to take cpus and disks into consideration.

kramasamy:
@nlu90 - if everything looks good, can you merge it?

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -438,4 +438,20 @@ public static boolean curlAndExtractPackage(
...
+    PackingPlan.Resource maxResource = new PackingPlan.Resource(0, 0, 0);
+
+    for (PackingPlan.ContainerPlan entry : packingPlan.containers.values()) {
+      PackingPlan.Resource resource = entry.resource;
+      maxResource.ram = Math.max(maxResource.ram, resource.ram);
+      maxResource.cpu = Math.max(maxResource.cpu, resource.cpu);
+      maxResource.disk = Math.max(maxResource.disk, resource.disk);
+    }
+
+    return maxResource;
```
ashvina: Once the max resource is found, is there a need to update the `PackingPlan`? The resource requirement suggested by the`PackingPlan` could be different from actual allocation after this step. I am not sure if `PackingPlan` is used by any other component after the scheduler.

ashvina:
The change looks good to me. It would be great if someone could provide some clarity on the question I posted above.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
nlu90: Scheduler can update the PackingPlan based on its need. But here, the util method only helps to find the maximum cpu/ram/disk resource requirement.

PR #1178, opened 2016-07-28
Dedup and organize launcher 1158

### LINKED ISSUE #1158: Dedup and clean up scheduler launch and packing plan creation code
In Heron, Launcher's responsibilities includes packing plan creation and launching Scheduler. A Scheduler can either be client local (e.g. Aurora) or remote (e.g. yarn, mesos). As a result the launcher's code needs to available on the client and on the server. Currently there is some code duplication in the classes performing this job, `SchedulerMain` and `LaunchRunner`. As part of this issue the following can be provided/fixed/

1. I think a utility class `LauncherUtils` is needed which would be used in both server and client side launch paths. 
1. Some launcher related code seems to have spilled into `SchedulerUtils`. This could be moved to `LauncherUtils` too.
1. Remove scheduler related dead code from `heron-executor.py`

ashvina:
Fixes #1158 

Scheduler can operate in library or service mode. This patch organizes the launcher's util methods used to launch scheduler in both modes.


heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,185 @@
...
+    Config ytruntime = putPackingInRuntimeConfig(runtime, packing);
+
+    scheduler.initialize(config, ytruntime);
+    return scheduler;
+  }
+
+  /**
+   * This method can be mocked for unit testing.
+   */
+  public PackingPlan _createPackingPlan(final Config config, final Config runtime) {
```
billonahill: The leading _ syntax is one I have not seen. Why do we need it?

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,185 @@
...
+ * controllers; {@link com.twitter.heron.scheduler.LaunchRunner} and
+ * {@link com.twitter.heron.scheduler.SchedulerMain}
+ */
+public class LauncherUtils {
+  private static final Logger LOG = Logger.getLogger(LauncherUtils.class.getName());
+
+  private static LauncherUtils instance = new LauncherUtils();
+
+  /**
+   * Invoke the onScheduler() in IScheduler directly as a library
```
billonahill: This method has side-effects in that it modifies the config passed to it, which is not great. I understand the convenience of doing it, but I think it would be better either have the caller do it, or maybe just document this side effect in the javadocs.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,185 @@
...
+import com.twitter.heron.spi.packing.PackingPlan;
+import com.twitter.heron.spi.scheduler.IScheduler;
+import com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor;
+
+/**
+ * {@link LauncherUtils} contains helper methods used by the server and client side launch
+ * controllers; {@link com.twitter.heron.scheduler.LaunchRunner} and
+ * {@link com.twitter.heron.scheduler.SchedulerMain}
+ */
+public class LauncherUtils {
```
billonahill: can you add unit tests for this class please.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,185 @@
...
+      LOG.log(Level.SEVERE, "Failed to instantiate packing instance", e);
+      return null;
+    }
+
+    try {
+      packing.initialize(config, runtime);
+      PackingPlan packedPlan = packing.pack();
+      return packedPlan;
+    } finally {
+      SysUtils.closeIgnoringExceptions(packing);
```
billonahill: This is a side effect hidden to the caller. Could you document this effect in the relevant public javadocs.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -43,41 +42,6 @@ private SchedulerUtils() {
...
   /**
-   * Invoke the onScheduler() in IScheduler directly as a library
-   *
-   * @param config The Config to initialize IScheduler
-   * @param runtime The runtime Config to initialize IScheduler
-   * @param scheduler the IScheduler to invoke
-   * @param packing The PackingPlan to scheduler for OnSchedule()
-   * @return true if scheduling successfully
-   */
-  public static boolean onScheduleAsLibrary(
```
billonahill: Why move onScheduleAsLibrary from SchedulerUtils to LauncherUtils?

ashvina: `onSchedulerAsLibrary` is a method to launch the scheduler. A launcher performs this function. So I think this method belongs to `LauncherUtils`. Moreover most other methods in `SchedulerUtils` are used by a scheduler once launched, for e.g. set location etc.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: I am on it now. Thanks for the note!

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Good point. I am inclining towards explaining the impact in java docs. As I mentioned, there are multiple paths where config is modified. However the config modification is not consistent. The intent is to enforce consistent behavior and not really convenience. That way the consumer of the config object can be deterministic and independent of the invocation path.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: The util classes provides static methods. However, static methods are hard to test, mock or inject. So the class has these internal methods used by static methods. The user should be using static methods and not the internal ones. I thought `_` could be a good way to easily identify. However I am open to any suggestions. I could create an internal class also. What do you think?

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
billonahill: sounds good.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
billonahill: I'm generally not a fan of static methods for these reasons, as well as difficulty overriding. I would recommend using non-static methods and if they need to be visible for unit testing, making them package private and annotating with `@VisibleForTesting`.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
billonahill: Makes sense.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Fixed
I have split the method into two, one for creating the config and the other for using the new config.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Fixed

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: I am not sure if I understand the side effect here. The packing "resource" is created in this method and closed. The packing resource is not needed once the output, a packing plan, is generated.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,176 @@
...
+   */
+  public static void setInstance(LauncherUtilImpl instance) {
+    LauncherUtils.instance = instance;
+  }
+
+  /**
+   * Implementation of utility methods provided by {@link LauncherUtils}. The instance of this
+   * class are primarily used for mocking and testing.
+   */
+  public static class LauncherUtilImpl {
```
billonahill: Why create an inner static class inside another one instead of just making createPackingPlan non-static in the parent class? That would seem more straight forward, and would allow subclassing.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
billonahill: Yes, I see that now. Sorry, I misread the first time. Please disregard.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: I agree. I wasn't sure. I was trying to make this utils look like other util classes. Will revert this. Thanks!

kramasamy:
@ashvina - can you please take a look why the tests failed?

ashvina:
@kramasamy : I found an issue in SlurmLauncherTest (#1192) and have created a PR #1193 for the same. Am looking at other failures.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Fixed

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,162 @@
...
+ * {@link com.twitter.heron.scheduler.SchedulerMain}
+ */
+public class LauncherUtils {
+  private static final Logger LOG = Logger.getLogger(LauncherUtils.class.getName());
+
+  /**
+   * An instance of this utility class to avoid the need of injecting a factory when mocking utils.
+   */
+  @VisibleForTesting
+  public static LauncherUtils instance = new LauncherUtils();
```
billonahill: This is marked as `@VisibleForTesting` but used by `LaunchRunner`. Instead of a public static instance, could you change this to a static factory method without the annotation?

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * For testing only. Could be used to setup a mock
+   */
+  @VisibleForTesting
+  public static void setInstance(LauncherUtils instance) {
```
billonahill: Instead of setting a static mock object here could you instead set the mock object as an instance object in the class that needs this?

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Ahh. Good catch. Will remove it. The `instance` is used to avoid sprinkling the constructor all over in the code.

ashvina: Fixed

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
```diff
@@ -0,0 +1,161 @@
...
+ * controllers; {@link com.twitter.heron.scheduler.LaunchRunner} and
+ * {@link com.twitter.heron.scheduler.SchedulerMain}
+ */
+public class LauncherUtils {
+  private static final Logger LOG = Logger.getLogger(LauncherUtils.class.getName());
+
+  /**
+   * An instance of this utility class to avoid the need of injecting a factory when mocking utils.
+   */
+  public static LauncherUtils instance = new LauncherUtils();
```
billonahill: Can we remove this public static instance entirely and just have classes create an object with the default constructor? That seems better than sharing static instances. And for the unit tests, you could just set a mock instance on the class that needs a LauncherUtils object.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Ok. Will fix it. In some cases the utils user object is not visible. So it may not be feasible to set the mock in those objects. As of now the launchers do not have this issue.

heron/spi/src/java/com/twitter/heron/spi/utils/LauncherUtils.java
ashvina: Fixed

1. I have removed `setInstance` and `public static instance`. It has been replaced with `getInstance` which can be `spy`ed to return a mocked util instance. 
1. I am not sure how a mock can be injected if default constructor is used. Unless another layer of abstraction/indirection is used.
1. The `private static` instance is an optimization to avoid creating utils objects.

billonahill:
Changes LGTM, lets just fix the failing CI.

PR #1179, opened 2016-07-28
Made Config message in topology.proto generic

taishi8117:
Modified Config message's serialized field so it's more generic and extensible.
This PR is necessary for adding a support for python topology.

This change should be backward compatible.

### DIFF
--- a/heron/proto/topology.proto
+++ b/heron/proto/topology.proto
@@ -39,6 +39,12 @@ enum ComponentObjectSpec {
   PYTHON_CLASS_NAME = 3;
 }
 
+enum ConfigValueType {
+  STRING_VALUE = 1; //normal string
+  JAVA_SERIALIZED_VALUE = 2;
+  PYTHON_SERIALIZED_VALUE = 3;
+}
+
 message StreamSchema {
   message KeyType {
     required string key = 1;
@@ -69,7 +75,10 @@ message Config {
     required string key = 1;
     // Only one of the below are set
     optional string value = 2;
-    optional bytes java_serialized_value = 3;
+    //optional bytes java_serialized_value = 3;
+    optional bytes serialized_value = 3;
+
+    required ConfigValueType type = 4;
   }
   repeated KeyValue kvs = 1;
 }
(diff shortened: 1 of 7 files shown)


heron/proto/topology.proto
```diff
@@ -69,7 +75,10 @@ message Config {
     required string key = 1;
     // Only one of the below are set
     optional string value = 2;
-    optional bytes java_serialized_value = 3;
+    //optional bytes java_serialized_value = 3;
+    optional bytes serialized_value = 3;
+
+    required ConfigValueType type = 4;
```
billonahill: adding a new required field will make parsing of any existing protos break. Does StreamSchema get persisted anywhere across topology runs where this would break things or is it only used within a running topology?

taishi8117: @billonahill I agree that it is quite possible that adding a new required field to config would break the existing topology, as the .defn file (containing this Config and StreamSchema) for already running topologies would not have this required field. I will consider making ConfigValueType optional.

taishi8117: @billonahill Also, I added a required field to Component in PR #1149, which might have not been a great idea for the same reason. Do you think it's better to make it optional as well?

billonahill: If these objects are just created and passed around by a running topology, that's fine to do something non-backward compatible since they just update upon redeploy. If they get persisted while running and accessed by another service, like the heron UI, then they need to be backwards compatible so the service can handle both the old and the new during the transition. I'm not sure if that's the case or not for these objects though.

taishi8117: @billonahill I think these objects are created when a topology is submitted and other services (tracker/ui) reference them, so it's safe to be backward compatible.  (Also, for a similar reason, required/optioanl fields were [removed in protobuf 3.0](http://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3)) 
I just pushed a version where ConfigValueType is optional --- It shouldn't affect the existing codes, and when we add a multi-lang support later, we can possibly make it backward compatible by 1) check the existence of ConfigValueType, and if it exists, handle according to that type, 2) if not existent, handle as if the value is Java serialized.

billonahill:
👍

objmagic:
good rationale. 

👍

PR #1180, opened 2016-07-29
Made ComponentObjectSpec optional for backward compatibility

taishi8117:
Made ComponentObjectSpec field for Component message in topology.proto optional to be safe with the backward compatibility. See discussion in https://github.com/twitter/heron/pull/1179#discussion-diff-72712681.


billonahill:
👍

PR #1181, opened 2016-07-29
Made custom grouping object for protobuf more generic

taishi8117:
Modified InputStream message in topology.proto, so that it is more extensible for multi-lang support.

Related PR: #1179, #1180

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/topology/BoltDeclarer.java
+++ b/heron/api/src/java/com/twitter/heron/api/topology/BoltDeclarer.java
@@ -167,7 +167,8 @@ public class BoltDeclarer extends BaseComponentDeclarer<BoltDeclarer> {
     bldr.setStream(
         TopologyAPI.StreamId.newBuilder().setId(streamId).setComponentName(componentName));
     bldr.setGtype(TopologyAPI.Grouping.CUSTOM);
-    bldr.setCustomGroupingJavaObject(ByteString.copyFrom(Utils.serialize(grouping)));
+    bldr.setType(TopologyAPI.CustomGroupingObjectType.JAVA_OBJECT);
+    bldr.setCustomGroupingObject(ByteString.copyFrom(Utils.serialize(grouping)));
     return grouping(bldr);
   }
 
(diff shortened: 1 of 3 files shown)


objmagic:
I suggest we make the following change instead. Either both ``custom_grouping_object`` and ``CustomGroupingObjectType`` appear or neither exists.

````protobuf
message CustomGroupingObject {
    required byte custom_grouping_object = 1;
    required CustomGroupingObjectType type = 2;
}
...
{
   ...
   optional CustomGroupingObject obj = 5;
}
````
However, if this change breaks backward compatibility a lot, your change is fine.

taishi8117:
@objmagic I considered that but to be safe with the backward compatibility, changing a type for the existing field id (5) might not be a good idea.

objmagic:
@taishi8117 then it's fine. Let's wait other people's opinion though.

billonahill:
@objmagic your suggestion is the right one IMO, but since it would break backward compatibility we have to compromise a bit with the proposed solution.

👍

kramasamy:
I think it is ok to break backward compatibility - since a bunch of changes have been there for proto and further additions are coming as well.

objmagic:
btw, it would be good to add some comments like "if you change this protobuf file, you need to make change accordingly for foo.py, bar.java, wee.cpp"

billonahill:
@kramasamy I was also wondering about that. Internally we don't have any production jobs on OSS heron, by the looks of the task tracker. If we can break backward compatability we could do the right thing. @taishi8117 do you want to work on a patch to refactor our proto classes to use the pattern that @objmagic showed above?

taishi8117:
@billonahill Yes, I will work on the patch to refactor them today.

kramasamy:
I think this is a good time to change - if needed, so that we can minimize incompatible migrations.

kramasamy:
Ship it 👍

taishi8117:
We decided to leave this PR backward compatible for now, and we can later refactor them when we move to protobuf v3.0 and might need to break the backward compatibility anyways (#1183).

PR #1185, opened 2016-07-30
Updated Python's common/util modules

taishi8117:
This is a working branch to add Python topology support.

* Added TaskHook support for python topology
* Added system config to outgoing_tuple_helper


heron/common/src/python/utils/topology/topology_context.py
```diff
@@ -93,3 +153,143 @@ def _get_output_to_comp_fields(outputs):
 
       out_fields[comp_name][stream_id] = tuple(ret)
     return out_fields
+
+  ######### Task hook related ##########
+
+  def _init_task_hooks(self):
```
kramasamy: can we add a unit test for task hook?

taishi8117: @kramasamy I will add them.

objmagic:
should be good if you

1. add tests on task hooks.
2. fix CI

objmagic:
#1204 has just been merged. Please first merge with master and change to **use log module under [`heron/common/src/python/utils`](https://github.com/twitter/heron/blob/master/heron/common/src/python/utils/log.py)**. You can refer to [`heron-explorer`](https://github.com/twitter/heron/blob/master/heron/explorer/src/python/main.py#L31) to see how to config and use this log module.

taishi8117:
@objmagic Do you mean that I should use ``Log = log.Log``, instead of ``from heron.common.src.python.utils.log import Log``?

objmagic:
@taishi8117 because we are going to use [`log.set_logging_level`](https://github.com/twitter/heron/blob/master/heron/explorer/src/python/main.py#L184), I just imported `log` module. 

I agree it's kind of ugly. Feel free to give me your suggestion.

heron/common/src/python/utils/misc/communicator.py
```diff
@@ -12,12 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 '''communicator.py: module responsible for communication between Python heron modules'''
+import sys
```
kramasamy: A bunch of the files seems to be instance related. Why we are putting these in heron/common?

taishi8117: @kramasamy Only reason why they are here is because equivalent ones in Java are under ``heron/common``.

objmagic: Agree. This keeps consistency and should be fine.

objmagic:
maybe I did not explain clearly. Actually, if `log` module is not used, you can just ``from log import Log``... 

I'll merge this after removing the last commit.

taishi8117:
@objmagic oh okay, that makes sense.

PR #1187, opened 2016-07-30
Properly shutdown heron-tracker and heron-ui

### LINKED ISSUE #1147: `ctrl-c` does not stop heron tracker

objmagic:
This PR registers `SIGINT` and `SIGTERM` with callback stop function so that the program could shutdown in a proper manner.

resolves #1147

### DIFF
--- a/heron/tracker/src/python/main.py
+++ b/heron/tracker/src/python/main.py
@@ -39,8 +39,7 @@ class Application(tornado.web.Application):
   def __init__(self):
 
     config = Config(options.config_file)
-    tracker = Tracker(config)
-    self.tracker = tracker
+    self.tracker = Tracker(config)
     self.tracker.synch_topologies()
     tornadoHandlers = [
         (r"/", handlers.MainHandler),
--- a/heron/tracker/src/python/main.py
+++ b/heron/tracker/src/python/main.py
@@ -220,12 +220,12 @@ def main():
   def signal_handler(signum, frame):
     # start a new line after ^C character because this looks nice
     print '\n',
-    LOG.debug('SIGINT received. Stopping tracker')
     application.stop()
     tornado.ioloop.IOLoop.instance().stop()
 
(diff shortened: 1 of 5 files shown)


heron/tracker/src/python/main.py
```diff
@@ -205,13 +208,30 @@ def main():
...
+  define_options(namespace['port'], namespace['config_file'])
+
+  # create Tornado application
+  application = Application()
+
+  # pylint: disable=unused-argument
+  # SIGINT handler:
+  # 1. stop all the running zkstatemanager and filestatemanagers
+  # 2. stop the Tornado IO loop
+  def signal_handler(signum, frame):
```
kramasamy: signal handler is appearing two times?

heron/ui/src/python/main.py
```diff
@@ -141,6 +141,20 @@ def main():
   define_options(command_line_args['port'], command_line_args['tracker_url'])
   http_server = tornado.httpserver.HTTPServer(Application())
   http_server.listen(command_line_args['port'])
+
+  # pylint: disable=unused-argument
```
kramasamy: why use pylint and disable?

objmagic: we are not really using arguments ``signum`` and ``frame``.

heron/tracker/src/python/main.py
objmagic: do you mean `signal_handler` appears both here and a file below? It is because `heron-tracker` needs to additionally stop `filestatemanager` and `zkstatemanager`, but `heron-ui` does not need to.

objmagic: `signal_handler` in `heron-tracker` has to perform `application.stop()`, but `heron-ui` does not have to.

kramasamy: can we refactor this code?

kramasamy:
👍

PR #1193, opened 2016-07-31
Use correct method signature and fix test

### LINKED ISSUE #1192: SlurmLauncherTest.testLaunch may not be verifying the launch correctly
I think SlurmLauncherTest.testLaunch method has the following issues:

1. The signature of SchedulerUtils.onSchedulerAsLibrary is incorrect. I am not sure why PowerMockito does not catch that https://github.com/twitter/heron/blob/master/heron/schedulers/tests/java/com/twitter/heron/scheduler/slurm/SlurmLauncherTest.java#L79-L83
1. The signature of SchedulerUtils.onSchedulerAsLibrary is incorrect. I am not sure why PowerMockito does not catch that https://github.com/twitter/heron/blob/master/heron/schedulers/tests/java/com/twitter/heron/scheduler/slurm/SlurmLauncherTest.java#L89-L91
1. The `launch` method is never invoked as the previous step, `setupWorkingDirectory` still returns false: https://github.com/twitter/heron/blob/master/heron/schedulers/tests/java/com/twitter/heron/scheduler/slurm/SlurmLauncherTest.java#L73
1. `launch` method is never invoked in the end. It should have been invoked 3 times by the end of the test

I am in the process of refactoring `SchedulerUtils` class and hit these issues. Please review and let me know if you think these are valid issues.

ashvina:
Fixes #1192 

Fixes use of `SchedulerUtils.onSchedulerAsLibrary` and `launch` method invocation in the test.

### DIFF
--- a/heron/schedulers/tests/java/com/twitter/heron/scheduler/slurm/SlurmLauncherTest.java
+++ b/heron/schedulers/tests/java/com/twitter/heron/scheduler/slurm/SlurmLauncherTest.java
@@ -28,6 +28,7 @@ import org.powermock.modules.junit4.PowerMockRunner;
 import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.ConfigKeys;
 import com.twitter.heron.spi.packing.PackingPlan;
+import com.twitter.heron.spi.scheduler.IScheduler;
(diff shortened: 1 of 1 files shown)


kramasamy:
👍

Issue #574, opened 2016-05-06
Ability to write Python topologies


kramasamy:
Users should be able to write python topologies and execute them. Here stream parse is the standard. If stream parse can be cleanly on top of Heron, it will satisfy the requirement.

amontalenti:
Some notes on this:

- We have two projects, [pystorm](https://github.com/pystorm/pystorm) and [streamparse](https://github.com/Parsely/streamparse). The former is an implementation of basic APIs for bolts/spouts and implementation based on Storm's multi-lang protocol. The latter is the actual framework for managing "projects" with multiple topologies that are written using the `pystorm` API.

- As of our very-recent 3.0.0+ release, topologies themselves are specified using a Python Topology DSL, based on thriftpy to serialize the topology into a Thrift Topology struct. We serialize that by embedding a copy of `storm.thrift` inside `streamparse`, as shown [in this code](https://github.com/Parsely/streamparse/blob/master/streamparse/thrift.py).

- You can read some docs on [the way the Topology DSL works here](http://streamparse.readthedocs.io/en/stable/topologies.html#topology-dsl).

kramasamy:
Now heron support python topologies. Please try it out and let me know.

Code0x58:
This looks like it can be closed now

joshfischer1108:
Thank you @Code0x58

PR #1196, opened 2016-08-02
DO NOT MERGE: Added pyheron API + basic component for python topology support

taishi8117:
This is a working branch and these files are NOT yet ready to be merged; this PR is opened for reviewing purposes. I will need to add unit tests and codes for remaining functionalities, as well as integration tests once other python topology components are ready, before this PR is merged.

These files are Heron Instance's base components (spout/bolt/topology) in Python. Modules included in this PR serve similar functionalities as ``BoltInstance``, ``SpoutInstance``, API interfaces for Spout and Bolt, ``TopologyBuilder`` and ``HeronSubmitter`` etc. in Java implementation. 

Once Python topology is officially supported, these files will be packaged into a library (maybe called ``pyheron``) for topology development; they are also made compatible with [StreamParse API](http://streamparse.readthedocs.io/en/v3.0.0/api.html).


amontalenti:
Hey! One of the streamparse original creators here. This is looking really interesting so far! I added a comment on #574 about recent happenings in pystorm & streamparse. But I agree with the approach you've taken of building your own pyheron interop layer and then just matching the duck-typed API that streamparse's Topology DSL and that pystorm components implement.

taishi8117:
@amontalenti Thanks for your comment! As you suggested, we are making the Heron Python topology as compatible with streamparse API (v3.0) as possible. It would also be great if you could take a look at our codes and give us some suggestions.

heron/instance/src/python/basics/component.py
```diff
@@ -0,0 +1,471 @@
...
+    self.output_helper.add_control_tuple(control_tuple, tuple_size_in_bytes, is_ack)
+
+  def get_total_data_emitted_in_bytes(self):
+    return self.output_helper.total_data_emitted_in_bytes
+
+  ##################################################################
+  # The followings are to be implemented by Spout/Bolt independently
+  ##################################################################
+
+  def start(self):
```
objmagic: Please use ``@abstractmethod`` decorator.

taishi8117:
API (#1236) is separated from instance, will close this PR.

PR #1198, opened 2016-08-02
Added example topologies in python

taishi8117:
Including:
- Word count topology
- Multi stream topology
- Custom grouping topology

PR #1199, opened 2016-08-02
Modified checkstyle for python topology support

taishi8117:
This PR is necessary for python topology support, as defining attributes in ``initialize()`` is necessary when defining python components (bolt and spout, refer to #1196 and #1198), whereas defining attributes in ``prepare()`` is necessary for implementing [ITaskHook](https://github.com/twitter/heron/blob/master/heron/common/src/python/utils/topology/task_hook.py).


objmagic:
👍

PR #1195, opened 2016-08-02
Adding support for python topology to executor, spi, scheduler

taishi8117:
* Added support for python topology to executor.
* Added default config value for sandbox binary location for python heron instance (``heron-python-instance``), which will be packaged to heron-code.tar.gz, when python topology is supported.
* Renamed ``topology_jar_file`` argument in executor to ``topology_bin_file``.
* Fixed style for heron executor

PR #1201, opened 2016-08-02
Added single-thread Heron Instance in python

taishi8117:
``st_heron_instance.py`` is an implementation of Heron Instance with single-thread in Python. They server similar functionalities as ``HeronInstance``, ``Gateway``, ``Slave`` etc. in Java implementation.

[Changes](https://github.com/twitter/heron/pull/1201/files#diff-3e725585fbf57518c1e7209b1b18e401R13) to ``heron_core.bzl`` is necessary to include this heron instance to Heron's core package. Please refer to #1195 as well.

This PR depends on #1236


heron/instance/src/python/instance/BUILD
```diff
@@ -0,0 +1,17 @@
+package(default_visibility = ["//visibility:public"])
+
+load("/tools/rules/pex_rules", "pex_library", "pex_binary")
+
+pex_binary(
+  # build binary for single thread python heron instance
+  name = "heron-python-instance",
+  srcs = ["st_heron_instance.py"],
```
kramasamy: why call st_heron_instance? simply calling heron_instance will suffice?

heron/instance/src/python/instance/__init__.py
```diff
@@ -0,0 +1,4 @@
+'''single thread heron instance module'''
+__all__ = ['st_heron_instance.py', 'st_stmgr_client.py']
```
kramasamy: same as before heron_instance.py and stmgr_client.py

heron/instance/src/python/instance/st_heron_instance.py
```diff
@@ -0,0 +1,269 @@
...
+
+    self.in_stream = HeronCommunicator(producer_cb=None, consumer_cb=None)
+    self.out_stream = HeronCommunicator(producer_cb=None, consumer_cb=None)
+
+    self.socket_map = dict()
+    self.looper = GatewayLooper(self.socket_map)
+
+    # Initialize metrics related
+    self.out_metrics = HeronCommunicator()
+    self.out_metrics.\
```
kramasamy: can you break the line in the argument as compared to in the middle of invocation?

heron/instance/src/python/instance/BUILD
taishi8117: @kramasamy I just thought if we are going to add double-thread heron instance later, it's better to distinguish between them.

kramasamy: ok.

Issue #1194, opened 2016-08-01
Merge color.py and log.py under heron/common/src/python


objmagic:
`color.py` provides colorful logging on terminal, `log.py` provides normal logging and normal rotating logger. We need to improve and merge them.

Issue #1129, opened 2016-07-22
Improve Heron CLI: logging level, error when killing topology


objmagic:
1. the logging is messy. Extra info is printed out without `-verbose` option
2. <del>investigate some errors that happen when killing topology</del>

objmagic:
there is no error when killing topology. @kramasamy tried again and the error is gone. However, I'll improve the logging.

objmagic:
can be combined with #1159 now

objmagic:
closed via #1204

PR #1204, opened 2016-08-02
Improve logging on explorer, tracker, scheduler, etc...

objmagic:
This patch tries its best to unify logging infrastructure and properly propagate verbose level set from command line. Specifically:

1. `color.py` and `log.py` has been merged to `log.py`
2. `tar` with `-v` always prints verbose messages to stderr and `curl` without `-s` always print progress bar to stderr. `runSyncProcess` will always print out stderr from subprocess because we use `SEVERE` level. In this PR, we change to `FINE` level.
3. Bug fix. I didn't review some PRs carefully so that my previous PRs that fixed some logging problems were reverted.
4. Various tiny fixes including word usage, making log look better.

More commits may be pushed later. 

cc: @taishi8117 @maosongfu @kramasamy 

resolves: #1194 #1129


heron/cli/src/python/cli_helper.py
```diff
@@ -72,7 +72,7 @@ def run(command, parser, cl_args, unknown_args, action):
         "--command", command,
     ]
 
-    if opts.verbose():
+    if Log.getEffectiveLevel() == logging.DEBUG:
```
kramasamy: this does not seem to be correct. if the --verbose option is provided by the user then only we add the --verbose option to the java programs.

heron/cli/src/python/main.py
```diff
@@ -221,21 +223,14 @@ def main():
...
-
   # command to be execute
   command = command_line_args['subcommand']
 
   # file resources to be cleaned when exit
   files = []
 
-  if command != 'help' and command != 'version':
+  if command not in ('help', 'version'):
+    log.set_verbose(command_line_args)
```
kramasamy: this is not the correct logic - only if the --verbose is provided, then enable the debug level option.

objmagic: Please see [`set_verbose` implementation] (https://github.com/twitter/heron/pull/1204/files/47421526742c3418cdc1b62b7b6216ce808c7f57#diff-d7d30dde4ab9b43859ead4b233f151a7R85). I agree this name is confusing. I can change to `set_logging_level`.

heron/cli/src/python/cli_helper.py
kramasamy: this should be ok.

objmagic: Right. We compare to see if the logger object `Log` has the `DEBUG` level set. If it is set, we add ``--verbose`` flag to invoke Java program.

objmagic: this is basically about propagating logging level from Python executor down to Java scheduler.

heron/explorer/src/python/main.py
```diff
@@ -28,7 +28,7 @@
 import heron.explorer.src.python.topologies as topologies
 import heron.common.src.python.utils.config as config
 import heron.explorer.src.python.version as version
-from heron.common.src.python.color import Log
+import heron.common.src.python.utils.log as log
```
kramasamy: can you import as Log rather than log?

objmagic: Good suggestion but log module has the logger object called `Log`. There will be name clashing.

objmagic:
@maosongfu I have question. Do we need to [add `-v` to `tar`](https://github.com/twitter/heron/pull/1204/files#diff-49b803240c7fae81aca5e36ac0559735L233)? Is it necessary to list the files in tar file? Similarly, if `-s` is not provided to `curl`, `curl` will show progress bar. Is progress bar necessary?

heron/explorer/src/python/main.py
objmagic: `log` module has a logger object `Log` and a function `configure`. I first get the logger object out (see: https://github.com/twitter/heron/pull/1204/files#diff-762d509cb715b724772c61e8aae7f089R33) and then call `log.configure` under ``main`` function.

objmagic:
@maosongfu never mind. I now think those messages are necessary.

taishi8117:
👍

heron/common/src/python/handler/access/fetch.py
```diff
@@ -63,8 +65,8 @@ def fetch_url_as_json(fetch_url, default_value=None):
   end = time.time()
   duration = 1000 * (end - start)
 
-  logging.info("TIME: url fetch took %.2f ms server time %s", execution, fetch_url)
-  logging.info("TIME: url fetch took %.2f ms round trip  %s", duration, fetch_url)
+  Log.info("TIME: url fetch took %.2f ms server time %s", execution, fetch_url)
```
kramasamy: this should be Log.debug.

heron/common/src/python/handler/access/fetch.py
```diff
@@ -63,8 +65,8 @@ def fetch_url_as_json(fetch_url, default_value=None):
...
+  Log.info("TIME: url fetch took %.2f ms round trip  %s", duration, fetch_url)
```
kramasamy: same here as well.

heron/common/src/python/handler/access/fetch.py
objmagic: thanks. will fix.

heron/common/src/python/handler/access/fetch.py
objmagic: thanks. will fix.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
```diff
@@ -133,6 +133,7 @@ public HeronExecutorTask(final REEFFileNames fileNames,
         env,
         topologyJar,
         topologyDefFile,
+        false,
```
kramasamy: why give false here? should we pass the verbose value that was passed?

objmagic: because in this class, no verbose value has been set. I just put `false` here. cc @ashvina do you have any suggestion?

objmagic: if it's hard to propagate verbose flag to YARN scheduler, I'm not sure what to do. Maybe just put `false` here?

ashvina: Could you please elaborate what is the difficulty for YARN? A number of configurations are passed to the YARN scheduler. I think it is feasible to pass verbose flag too.

objmagic: There is no verbose option in [HeronExecutorTask's field](https://github.com/objmagic/heron/blob/6543f5f531d86521f341753f821dc89dc1fe69ff/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java#L85-L94).

objmagic: no verbose option in [HeronTaskConfiguration](https://github.com/twitter/heron/blob/e68943d422ea35f54c46fc8944c1c83ef6440d78/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronTaskConfiguration.java)

objmagic: I'll appreciate if you can add a verbose option in YARN scheduler.

objmagic:
@kramasamy btw, just as a reminder, please do not merge any other Python PR before this one.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
ashvina: Ok

ashvina: @objmagic Done #1208

objmagic: @ashvina Fantastic! We can merge #1208 first and make change on this PR accordingly.

kramasamy:
@objmagic - sure.

kramasamy:
👍

PR #1205, opened 2016-08-02
Parallel Pylint checking.

objmagic:


kramasamy:
@objmagic - while this is interesting, it might not be needed since bazel itself parallelizes the build. With two level parallelization it might be hard to identify where the bug should be.

objmagic:
I'm not sure how action listener in Bazel works in detail, but when you build project, bazel will say, for example, "46 files found by checkstyle" (when building target xxx). We can then use multiple processes to check style to accelerate. 

As for the bug reporting, Pylint usually dumps out information which clearly indicates which module has bad style. Parallelism will not create much hassle here.

billonahill:
For java we found the "N files found for checkstyle" output was misleading when inferring where build time was spent. It turned out that the checkstyle was in fact very fast compared to the compile, but since the checkstyle output log info it appeared (as a red herring) to be the time sink. Before adding parallelism I'd want to see some timing numbers showing how much savings we'd achieve. My suspicion is not much.

(Personally, I'd rather see the file that produces the style errors upon a failure, which is currently difficult to determine)

objmagic:
@billonahill @kramasamy I just used `time` to get the difference. It turns out you are right --- the difference is negligible. Let's close this PR.

PR #1207, opened 2016-08-03
Set PEX_ROOT for executor environment

### LINKED ISSUE #1206: Executor is failing on YARN containers trying to create ~/.pex directory
Stack trace in executor logs
```
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 45] Operation not supported: '/home/.pex'
```

This may be caused by this commit a9d4861813eb27a692dce7bfdeb8d3ad75fc0059

ashvina:
Fixes #1206 
* By default user home on NodeManagers may not be writable. Use container
directory as PEX_ROOT instead of user home.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
@@ -17,6 +17,7 @@ package com.twitter.heron.scheduler.yarn;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


objmagic:
I thought #1140 already fixed the problem? If not, could you please explain to me?

ashvina:
Hi @objmagic,

@kramasamy debugged this issue for me. Below is a explanation in his words. We had this discussion on slack

```
previous to the refactor
heron-executor depended only pyyam

https://github.com/twitter/heron/commit/a9d4861813eb27a692dce7bfdeb8d3ad75fc0059
heron executor refactor and unit tests (#1134) · twitter/heron@a9d4861 · GitHub
* switch order of monitor process * committing the unit tests * committing the unit tests * fix unit tests * updating comments * change most methods to be private * One last c... 
 
according to BUILD file
with the refactor it also includes - statemanager etc
that pulls in zookeeper changes
now when heron-executor itself is run - it needs a pex directory to explode itself
in this case it uses ~/.pex
when heron executor executes process like heron-shell etc
they use the container .pex directory
and hence you have two .pex directories
```

#1140 fixed the heron-shell issue. This one fixes the new error after the refactor commit.

objmagic:
@ashvina thanks a lot! This makes sense to me now.

kramasamy:
👍

PR #1214, opened 2016-08-04
Improve PackingPlanTest

billonahill:
Cleanup PackingPlanTest and add assertions for getComponentRamDistribution.

I'm going to add tests for PackingPlan serde to this one and wanted to commit other unrelated cleanup items first.

### DIFF
--- a/heron/spi/tests/java/com/twitter/heron/spi/packing/PackingPlanTest.java
+++ b/heron/spi/tests/java/com/twitter/heron/spi/packing/PackingPlanTest.java
@@ -40,7 +40,7 @@ public class PackingPlanTest {
       for (String instanceId : instanceList) {
         String componentName = instanceId.split(":")[1];
         Resource instanceResource = null;
-        if (componentName.equals("bolt")) {
+        if ("bolt".equals(componentName)) {
           instanceResource = new Resource(1.0, 2 * Constants.GB, 10 * Constants.GB);
         } else {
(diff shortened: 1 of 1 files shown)


heron/spi/tests/java/com/twitter/heron/spi/packing/PackingPlanTest.java
```diff
@@ -25,7 +25,7 @@
 import com.twitter.heron.spi.common.Constants;
 
 public class PackingPlanTest {
-  public static PackingPlan generatePacking(Map<String, List<String>> basePacking) {
+  private static PackingPlan generatePacking(Map<String, List<String>> basePacking) {
     Resource resource =
```
kramasamy: do we need this with the new change?

billonahill: Access isn't needed outside of this class and won't be when I add a serde test, since I was going to add it to this class as well.

kramasamy: I was not clear - I meant the variable "resource"

billonahill: Yes, it's still used in the ContainerPlan and the PackingPlan objects below. I just changed to a resource for each instance that was unique, so I could assert on that.

kramasamy:
👍

PR #1218, opened 2016-08-04
DO NOT MERGE - scaling topologies working branch

billonahill:
This PR is just to make it easier to view diffs and to have travis ci verify commits. This is a long-lived branch.

No need to review this branch. All reviews will happen on smaller PRs. You might want to unwatch this one to not get tons of commit emails that don't require action.

Working branch for #1292. This branch will be cherry-picked and merged into master piece by piece until it's empty.


heron/scheduler-core/src/java/com/twitter/heron/scheduler/ScalableScheduler.java
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.scheduler;
+
+public interface ScalableScheduler {
+  void addContainers(Integer count);
```
ashvina: This API would need additional information for heterogeneous schedulers. the YARN scheduler for eg, needs to know what resources to request for the new container. This information will come from the updated packing plan.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/ScalableScheduler.java
```diff
@@ -0,0 +1,19 @@
...
+  void removeContainers(Integer existingContainerCount, Integer count);
```
ashvina: It is assumed that the containers with the highest id will be removed?

billonahill: That's the current impl, but we might need to rethink this. For aurora if we remove a container that's not at the end, I'm not sure what the container ids would do and there are places where a continuous set of ids is expected so it feels safer to remove from the end. But it is presumptuous to say the last ones should go.

When doing the packing diff we could pull out the container ids that are now empty and pass those to removeContainers? We'd need to make sure there isn't a mess if we remove containers not at the end though.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/ScalableScheduler.java
billonahill: Yes, this seems reasonable. Aurora would need to ignore the Container info though which is unfortunate, since it can only clone.

ashvina: @billonahill I have created a patch with this change:  https://github.com/ashvina/heron/commit/4f38c29f2c7f58a38aa33a35696285af5df64e0d
Could you please review and merge this. Thanks !

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -0,0 +1,162 @@
...
+    // update parallelism in updatedTopology since TMaster checks that
+    // Sum(parallelism) == Sum(instances)
+    logFine("Deleted existing Topology: %s", stateManager.deleteTopology(topologyName));
+    logFine("Set new Topology: %s", stateManager.setTopology(updatedTopology, topologyName));
+
+    // update packing plan to trigger the scaling event
+    logFine("Deleted existing packing plan: %s", stateManager.deletePackingPlan(topologyName));
+    logFine("Set new PackingPlan: %s",
+        stateManager.setPackingPlan(proposedProtoPackingPlan, topologyName));
+
```
ashvina: There are 4 state manager interactions above. Failure in any one can result in state corruption and existing topology fail. In my opinion update operation must not negatively impact existing deployment. So I think state update (both packingPlan and topology) should be a transaction. Although curator (ZK) does support atomic batch operations, other state managers might not support it. I am thinking if we can implement a "current/active version marker" pattern. An update operation will look like this

1. Add new topology with a version number, say topology-v2
2. Add new packing plan, packing-v2
3. Atomically update topology's current/active version marker to v2
4. Asynchronously delete topology-v1 and packing-v1
5. consumers detect version marker change and read new packing and topology

What do you think?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -109,6 +118,28 @@ public boolean onRestart(Scheduler.RestartTopologyRequest request) {
     return controller.restartJob(containerId);
   }
 
+  @Override
+  public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
+    try {
+      updateTopologyManager.updateTopology(
```
ashvina: This is related to #1222 

Currently method `createAuroraProperties` provides instance distribution to containers. On topology update the instance distribution for existing containers, e.g. `C`, can change. However the aurora property map for container `C` will not change. As a result if container `C` restarts, it will use old instance plan. Will the executor ignore the old instance distribution?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/ScalableScheduler.java
billonahill: @ashvina could you make a PR against my branch?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: I agree that we should strive towards atomicity. The change described will we a considerable refactorting effecting much of the non-scaling functionality (i.e., distribution/update of physical plan, topology, etc during submit) so I'm inclined to ship an initial version of scaling that is weak transitionally but keeps submit as-is and then follow up with a refactoring to harden things from a transactional aspect.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
billonahill: Yes. See the changes to the ```heron_executor.py``` in this branch. On startup instead of doing:
```
self.update_instance_distribution(self.parse_instance_distribution(args[5]))
```
we do:
```
executor.register_packing_plan_watcher(executor)
```
which ignores the passed param for both submit/startup and update.

heron/cli/src/python/main.py
```diff
@@ -111,30 +113,21 @@ def run(command, parser, command_args, unknown_args):
...
+      'restart':restart,
+      'submit':submit,
+      'update':update,
+      'help':cli_help,
+      'version':version,
+  }
+
+  if command in runners.keys():
+    return runners[command].run(command, parser, command_args, unknown_args)
+  else:
```
objmagic: better to add a line `Log.error('Unknown subcommand: %s' % command)`

heron/executor/src/python/heron_executor.py
```diff
@@ -536,6 +546,79 @@ def prepare_launch(self):
         do_print("Failed to run command: %s. Exiting" % command)
```
objmagic: `do_print` is changed to `Log`. See: https://github.com/twitter/heron/pull/1252/files

Rebasing against master should fix this.

heron/cli/src/python/main.py
billonahill: Good call, will update. This code path will only happen if someone adds a new command to the parser, but doesn't register a runner, so it should get caught in testing, but with the logging it will help pinpoint where the change was missed.

An even better approach would be to somehow have these runners registered as actions with the parser so we don't need this map.

objmagic: maybe passing `runners` to `create_parser`? For example, do `kill.create_parser(subparsers, self.runners)` and inside `kill` module associates string `kill` with a lambda function like the following:

````Python
runners['kill'] = lambda *args: run(*args)
````

Back in `main`, refactor `run` to:
````
def run(command, *rest_args):
    try:
        return self.runners[command](command, *rest_args)
    except:
        Log.error('Unknown subcommand: %s' % command)
        return 1
````

billonahill: Yeah, that makes our invocation a bit more fancy, but we still need to maintain both the commands in the parser and the runners dict. I don't think that's a huge deal, but maybe a yak-shave for another day...

objmagic: agree...

billonahill:
Resolving, as all changes have been incorporated into master. Woot!

PR #1219, opened 2016-08-04
Adding PackingPlan protobuf object

billonahill:
Adding protobuf object to represent a `PackingPlan` object which will be used for scaling. Also including a serializer and deserializer tfrom `com.twitter.heron.spi.packing.PackingPlan` to `com.twitter.heron.proto.system.PackingPlans.PackingPlan`.


ashvina:
LGTM. CI failures need to be fixed.

ashvina:
@billonahill methods like `c.t.h.spi.packing.PackingPlan.getInstanceDistribution()` seem to be serializing packing plan into strings. Will this change make such custom methods redundant and removable?

billonahill:
@ashvina good observation. I'd love to remove then and that was my initial intent, but the executor needs that info and is written in python. Serializing it means we don't need to change the executor code, so I kept in there for now. Without it we'd need to reimplement that logic in python to reconstruct. I don't love having the logic implemented in two places, but I agree we might want to do that instead before we ship scaling, but in a later patch.

kramasamy:
Bill - it would be good to track this issue so that we don't drop it on the floor.

billonahill:
Created #1222

billonahill:
Ping.

billonahill:
I've removed the instance dist and component ram dist from the serialize proto. @ashvina  and @kramasamy please review.

heron/proto/packing_plan.proto
```diff
@@ -0,0 +1,31 @@
...
+message ContainerPlan {
+  required string id = 1;
+  repeated InstancePlan instance_plans = 2;
+  required Resource resource = 3;
+}
+
+message PackingPlan {
+  required string id = 1;
+  repeated ContainerPlan container_plans = 2;
+  required Resource resource = 3;
```
kramasamy: Is this the overall resource requirement for the entire plan?

heron/proto/packing_plan.proto
```diff
@@ -0,0 +1,31 @@
...
+message InstancePlan {
+  required string id = 1;
+  required string component_name = 2;
+  required Resource resource = 3;
+}
+
+message ContainerPlan {
+  required string id = 1;
+  repeated InstancePlan instance_plans = 2;
+  required Resource resource = 3;
```
kramasamy: I am assuming this is the resource requirement for the entire container?

heron/proto/scheduler.proto
```diff
@@ -28,6 +29,11 @@ message RestartTopologyRequest {
   required int32 container_index = 2;
 }
 
+message UpdateTopologyRequest {
+  required heron.proto.system.PackingPlan currentPackingPlan = 1;
+  required heron.proto.system.PackingPlan proposedPackingPlan = 2;
+}
+
```
kramasamy: Just wondering if this is a request send by Topology Master to Executor or something else?

kramasamy: If this is indeed a request - do we need a response? In general, for control messages, it will always good to send a response back.

heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlanProtoDeserializer.java
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.spi.packing;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.twitter.heron.proto.system.PackingPlans;
+
+/**
```
kramasamy: Instead of two classes - one for serialization and the other deserialization - can we have all of them in a single class? It will be easy to read and any modification will be localized - makes it easy to read and review code - in case one is changed but not the other?

heron/proto/packing_plan.proto
billonahill: Yes. This class is basically a protobuf version of [c.t.h.spi.packing.PackingPlan](https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java#L20).

heron/proto/packing_plan.proto
billonahill: Yes. See [c.t.h.spi.packing.ContainerPlan](https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java#L119).

heron/proto/scheduler.proto
billonahill: This is send from the [RuntimeManagerRunner](https://github.com/twitter/heron/pull/1218/files#diff-175aa5efeb9ca85f99dfb52dead01fa7R175) to the [Scheduler](https://github.com/twitter/heron/compare/master...billonahill:billg/scaling_proto#diff-4a29bbb233b3d74c8c379edb6cf37487R122) to invoke the scaling event. 

Look at the rest of scheduler.proto to get a sense for how request/responses are implemented. Each request has it's own type, but all responses use the common SchedulerResponse that includes a status code indicating success.

heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlanProtoDeserializer.java
billonahill: I've gone back and forth on this in the past. In other projects I've implemented both in one class before finding that most clients rely on one and not the other, and sometimes each has different deps. Hence I implemented them separately, which also keeps the class size small and easy to read.

heron/proto/scheduler.proto
kramasamy: thanks for the clarification.

heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlanProtoDeserializer.java
kramasamy: sounds good.

kramasamy:
👍  @ashvina - do you want to take a quick look before merging it?

ashvina:
:+1: LGTM

PR #1223, opened 2016-08-06
Prepare default handler for all use cases.

objmagic:
since some users complained they see error like "No handler found..."

@taishi8117 can you have a quick look?

### DIFF
--- a/heron/common/src/python/utils/log.py
+++ b/heron/common/src/python/utils/log.py
@@ -16,6 +16,9 @@ import logging
 from logging.handlers import RotatingFileHandler
 import colorlog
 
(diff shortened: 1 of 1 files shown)


kramasamy:
@objmagic - can you please comment why this is needed?

objmagic:
it is possible that user simply imports the `Log` object without calling function ``configure``. This may cause exception "No handler found". So we at top level add a `logging.basic_config()` to prepare a default handler so that `Log` will always be properly initialized.

kramasamy:
please add this comment in the code.

kramasamy:
👍

PR #1227, opened 2016-08-08
Allow overriding docker build environment variables

ajorgensen:
It is possible to build heron in an environment that does not include
the git repository (this can be the case in ci for instance). We should
allow people building override the derived variables to manually set the
version.


ajorgensen:
There is actually a heron version string that is passed in whose purpose was to be the version that the release was tagged with. What are people thoughts about either A removing that and allowing the override as I have implemented here or B removing the git parsing for the version string? (https://github.com/twitter/heron/pull/1227/files#diff-489b3a1dc33fb4234394f371a8e67fe1R104)

kramasamy:
Ship it!

PR #1233, opened 2016-08-09
Changed heron.options from system property to environment variable

### LINKED ISSUE #1230: Passing of heron.options to a python topology's main method when writing its defn file
In Java, when a topology's defn file is created by executing the topology's main method, its initial state and the defn file's destination path (tmp directory path) are [passed](https://github.com/twitter/heron/blob/62793b63413976cc4f9d2e7cceb0815d062523d4/heron/api/src/java/com/twitter/heron/api/HeronSubmitter.java#L53) by [Java's system property](https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/utils/Utils.java#L51) (giving -Dheron.options= flag to JVM at runtime). 

However for python topology, I'm stuck at doing the same thing, because python doesn't have an equivalent of Java's system property, and giving these options simply as command line arguments would restrict user's flexibility in writing a topology's main method. One possible solution is to set an environment variable (like HERON_RUNTIME_OPTION), but I don't know if that's a good idea.

Any suggestions? @maosongfu

taishi8117:
resolves #1230

### DIFF
--- a/heron/cli/src/python/execute.py
+++ b/heron/cli/src/python/execute.py
@@ -52,17 +52,22 @@ def heron_class(class_name, lib_jars, extra_jars=None, args=None, java_defines=N
   # Construct the command line for the sub process to run
   # Because of the way Python execute works,
   # the java opts must be passed as part of the list
-  all_args = [config.get_java_path(), "-client", "-Xmx1g", opts.get_heron_config()] + \
+  all_args = [config.get_java_path(), "-client", "-Xmx1g"] + \
              java_opts + \
              ["-cp", config.get_classpath(lib_jars + extra_jars)]
 
   all_args += [class_name] + list(args)
 
+  # set heron_config environment variable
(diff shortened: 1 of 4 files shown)


heron/cli/src/python/execute.py
```diff
@@ -106,6 +111,10 @@ def heron_pex(topology_pex, topology_class_name, tmp_dir):
   try:
     pex_loader.load_pex(topology_pex)
     topology_class = pex_loader.import_and_get_class(topology_pex, topology_class_name)
+    if opts.get_config("cmdline.topology.initial.state") == "PAUSED":
```
kramasamy: use Log.debug("Topology deployment status %s" % (opt.get_config(...))

so that it is displayed even in the case of active vs paused.

kramasamy:
👍

objmagic:
👍

PR #1237, opened 2016-08-09
Additions to heron/common for python topology

taishi8117:
- Added custom serializer support

### DIFF
--- a/heron/common/src/python/utils/misc/serializer.py
+++ b/heron/common/src/python/utils/misc/serializer.py
@@ -19,7 +19,7 @@ try:
 except:
   import pickle
 
-class HeronSerializer(object):
+class IHeronSerializer(object):
   """Serializer interface for Heron"""
   @abstractmethod
   def initialize(self, config):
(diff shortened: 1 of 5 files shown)


heron/common/src/python/utils/misc/serializer.py
```diff
@@ -19,7 +19,7 @@
 except:
   import pickle
 
-class HeronSerializer(object):
+class IHeronSerializer(object):
```
kramasamy: Why change to IHeronSerializer?

kramasamy:
👍

heron/common/src/python/utils/misc/serializer.py
objmagic: also wondering... we don't have to even follow the naming convention from Factory method of Java

taishi8117: I thought it might be better to indicate in the name that this is an interface.

objmagic:
👍

PR #1231, opened 2016-08-09
Added docs for Python topology support

taishi8117:
This is a working branch and is not ready to be merged.

resolves #1220

PR #1238, opened 2016-08-09
Python topology submission/execution support

taishi8117:
- Heron-cli can submit python topology as described in #1231


heron/cli/src/python/execute.py
```diff
@@ -106,13 +106,30 @@ def heron_tar(class_name, topology_tar, arguments, tmpdir_root, java_defines):
...
+      raise RuntimeError(err_str)
+  else:
+    try:
+      # loading topology from Topology's subclass (no main method)
+      os.environ["HERON_OPTIONS"] = opts.get_heron_config()
+      pex_loader.load_pex(topology_pex)
+      topology_class = pex_loader.import_and_get_class(topology_pex, topology_class_name)
+      topology_class.write()
+    except Exception:
+      traceback.print_exc()
```
kramasamy: do you want to use Log.debug(traceback.print_exc())

kramasamy:
It looks fine to me - rest of the stuff 👍  (once you fix the comment)

heron/cli/src/python/execute.py
```diff
@@ -106,13 +106,30 @@ def heron_tar(class_name, topology_tar, arguments, tmpdir_root, java_defines):
...
-    traceback.print_exc()
-    err_str = "Topology pex failed to be loaded. Bailing out..."
-    raise RuntimeError(err_str)
+  if topology_class_name == '-':
+    # loading topology by running its main method (if __name__ == "__main__")
+    heron_env = os.environ.copy()
+    heron_env['HERON_OPTIONS'] = opts.get_heron_config()
+
+    args = [topology_pex]
+    Log.debug('$> %s' % ' '.join(args))
```
objmagic: joining a list of one element?

objmagic: also, better rename it to `cmd`. `args` does not make sense to me.

heron/cli/src/python/execute.py
```diff
@@ -106,13 +106,30 @@ def heron_tar(class_name, topology_tar, arguments, tmpdir_root, java_defines):
...
+    Log.debug('Heron options: %s' % str(heron_env['HERON_OPTIONS']))
+
+    # invoke the command with subprocess and print error message, if any
+    status = subprocess.call(args, env=heron_env)
+    if status != 0:
+      err_str = "User main failed with status %d. Bailing out..." % status
```
objmagic: Not necessary because `main` execution failed. Not really being a troll, but a user can mistakenly submit a PEX even without `if __name__ = "__main__"`. It's also possible that this PEX has some module importing problem. In this case, user still will see error message about `main`...

Also, the `if topology_class_name == '-'` looks good to me for now, but I feel somehow this hack could be shaky under some circumstances.

heron/cli/src/python/execute.py
```diff
@@ -106,13 +106,30 @@ def heron_tar(class_name, topology_tar, arguments, tmpdir_root, java_defines):
...
+      raise RuntimeError(err_str)
+  else:
+    try:
+      # loading topology from Topology's subclass (no main method)
+      os.environ["HERON_OPTIONS"] = opts.get_heron_config()
+      pex_loader.load_pex(topology_pex)
+      topology_class = pex_loader.import_and_get_class(topology_pex, topology_class_name)
+      topology_class.write()
+    except Exception:
+      Log.debug(traceback.print_exc())
```
objmagic: should be `traceback.format_exc())`. Fix all the same errors if you think there are more.

heron/cli/src/python/submit.py
```diff
@@ -246,9 +246,9 @@ def submit_pex(cl_args, unknown_args, tmp_dir):
   topology_file = cl_args['topology-file-name']
   topology_class_name = cl_args['topology-class-name']
   try:
-    execute.heron_pex(topology_file, topology_class_name, tmp_dir)
+    execute.heron_pex(topology_file, topology_class_name)
   except Exception as ex:
-    Log.error("Unable to execute topology main class: " + ex.message)
+    Log.error("Error when loading a topology: " + ex.message)
```
objmagic: how about `Log.error("Error when loading a topology: %s", str(ex))`

PR #1245, opened 2016-08-11
add option to pass any additional classpath while launching a topology

kramasamy:
Needed for schedulers for YARN - might be useful later for others as well.


ashvina:
:+1: Thanks for this contribution
I tested the classpath on local yarn cluster by adding YARN jars and config xml files. It worked as expected. There is no need to copy jars anymore. I also tested with by adding `heron:submit:* --extra-launch-classpath <value>` to .heronrc

heron/cli/src/python/execute.py
```diff
@@ -54,7 +54,7 @@ def heron_class(class_name, lib_jars, extra_jars=None, args=None, java_defines=N
   # the java opts must be passed as part of the list
   all_args = [config.get_java_path(), "-client", "-Xmx1g"] + \
              java_opts + \
-             ["-cp", config.get_classpath(lib_jars + extra_jars)]
+             ["-cp", config.get_classpath(extra_jars + lib_jars)]
```
objmagic: em.. why switching order here?

objmagic:
👍

heron/tools/common/src/python/utils/classpath.py
```diff
@@ -0,0 +1,46 @@
...
+  # check if the class path directory exists
+  if path.endswith('*'):
+    Log.debug('Checking class path directory: %s', path[-1])
+    if os.path.isdir(path[-1]):
+      return True
+    return False
+
+  # check if the class path file exists
+  Log.debug('Checking class path file: %s', path)
+  if os.path.isfile(path):
```
ashvina: The classpath entry can be a directory also. For e.g. dir of hadoop config xml files. This check fails for a directory.

ashvina:
👍 LGTM
I tested on a local yarn cluster. Incorrect paths are flagged correctly.

PR #1246, opened 2016-08-11
Modified test_runner for python topology integration tests

taishi8117:


integration-test/src/python/test_runner/main.py
```diff
@@ -154,11 +154,26 @@ def runAllTests(conf, args):
...
+  total = len(test_topologies)
   current = 1
-  for topologyConf in conf["topologies"]:
+
+  for topologyConf in test_topologies:
     topologyName = ("%s_%s_%s") % (timestamp, topologyConf["topologyName"], str(uuid.uuid4()))
-    classPath = conf["topologyClasspathPrefix"] + topologyConf["classPath"]
+    if pkg_type == 'pex':
+      classPath = topologyConf["classPath"]
+    else:
```
kramasamy: it will be good if you can do else if and compare with jar - even though we filter it at the top. As people add more code - people will tend to forget and they might not add here. always good to add more checks.

integration-test/src/python/test_runner/main.py
```diff
@@ -154,11 +154,26 @@ def runAllTests(conf, args):
   successes = []
   failures = []
   timestamp = str(int(time.time()))
-  total = len(conf["topologies"])
+
+  if args.testsJarPath.endswith(".jar"):
+    test_topologies = conf["javaTopologies"]
+    pkg_type = "jar"
+  elif args.testsJarPath.endswith(".pex"):
```
objmagic: better to rename `testsJarPath` to something else

kramasamy:
👍

objmagic:
👍

PR #1252, opened 2016-08-12
Fix heron executor log problem

### LINKED ISSUE #1216: Heron Executor's use of logging module
Heron Executor currently uses its own print function [``do_print()``](https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py#L41), instead of heron's common logging module (``heron/common/src/python/utils/log.py``). In order to make the logging uniform, it's better for the executor to use this as well, but there are several potential issues:

1. Since stdout/stderr of the executor is already redirected to a file (e.g. heron-executor.stdout) by the scheduler, any logs from the executor should be output to stdout. That case, we can't use the existing [``StreamHandler``](https://github.com/twitter/heron/blob/master/heron/common/src/python/utils/log.py#L56), because  ugly escape characters from colorlog will also be in the log file. A possible solution is to create another StreamHandler that doesn't use colorlog.
2. Heron Executor does not use any other common modules (under ``heron/common/src/python``), meaning that importing the whole heron's python common library just for ``log.py`` is quite cumbersome (since it includes unnecessary dependencies, like tornado, as well). One possible solution is to create another BUILD rule for the python common, such as one just for logging. 

What would be the best solution?

Related PR: #1195

taishi8117:
Resolves #1216

### DIFF
--- a/heron/common/src/python/BUILD
+++ b/heron/common/src/python/BUILD
@@ -21,3 +21,11 @@ pex_library(
   deps = [],
 )
 
+pex_library(
+  name = 'common_log',
+  srcs = ["utils/log.py"],
+  reqs = [
+    "colorlog==2.6.1",
+  ]
+)
+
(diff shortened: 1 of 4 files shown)


kramasamy:
👍  @objmagic - can you check this as well?

objmagic:
👍

PR #1256, opened 2016-08-12
add address argument to heron-ui

lukess:
1. address argument to heron-ui
2. remove socket.gethostbyname(socket.gethostname()) because of multiple NICs

### DIFF
--- a/heron/ui/src/python/main.py
+++ b/heron/ui/src/python/main.py
@@ -17,7 +17,6 @@
 import logging
 import os
 import signal
-import socket
 
 import tornado.ioloop
 import tornado.options
@@ -142,7 +141,9 @@ def main():
   Log.info("Using tracker url: %s", command_line_args['tracker_url'])
 
   # pass the options to tornado and start the ui server
-  define_options(command_line_args['address'], command_line_args['port'], command_line_args['tracker_url'])
+  define_options(command_line_args['address'],
+                 command_line_args['port'],
+                 command_line_args['tracker_url'])
   http_server = tornado.httpserver.HTTPServer(Application())
   http_server.listen(command_line_args['port'], address=command_line_args['address'])
 
--- a/heron/ui/src/python/main.py
+++ b/heron/ui/src/python/main.py
@@ -136,12 +136,7 @@ def main():
   # log additional information
   command_line_args = vars(parsed_args)
 
-  address = '0.0.0.0'
-  if command_line_args['address']:
-    address = command_line_args['address']
-  else:
-    Log.warn("Bindings wildcard by default")
-  Log.info("Listening at http://%s:%d", address, command_line_args['port'])
+  Log.info("Listening at http://%s:%d", command_line_args['address'], command_line_args['port'])
   Log.info("Using tracker url: %s", command_line_args['tracker_url'])
 
   # pass the options to tornado and start the ui server
--- a/heron/ui/src/python/main.py
+++ b/heron/ui/src/python/main.py
@@ -136,7 +136,11 @@ def main():
   # log additional information
   command_line_args = vars(parsed_args)
 
-  address = '0.0.0.0' if command_line_args['address'] is None else command_line_args['address']
+  address = '0.0.0.0'
+  if command_line_args['address']:
+    address = command_line_args['address']
+  else:
+    Log.warn("Bindings wildcard by default")
   Log.info("Listening at http://%s:%d", address, command_line_args['port'])
   Log.info("Using tracker url: %s", command_line_args['tracker_url'])
 
--- a/heron/ui/src/python/main.py
+++ b/heron/ui/src/python/main.py
@@ -105,12 +105,14 @@ class Application(tornado.web.Application):
     tornado.web.Application.__init__(self, callbacks, **settings)
 
 
-def define_options(port, tracker_url):
+def define_options(address, port, tracker_url):
   '''
+  :param address:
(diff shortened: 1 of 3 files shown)


heron/ui/src/python/main.py
```diff
@@ -134,14 +136,15 @@ def main():
 
   # log additional information
   command_line_args = vars(parsed_args)
-  address = socket.gethostbyname(socket.gethostname())
+
+  address = '0.0.0.0' if command_line_args['address'] is None else command_line_args['address']
```
kramasamy: instead of this - can we do the following - 
```
address = socker.gethostbyname(socket.gethostname()) if not command_line_args['address'] else command_line_args['address']
```

kramasamy: or instead of the long line - 

```
address = socket.gethostbyname(socket.gethostname())
if command_line_args['address']:
   address = command_line_args['address']
```
easy to read as well, thoughts?

objmagic: :+1 this is easier to read.

kramasamy:
@lukess - as suspected the python check style does not seem to accept long lines.

objmagic:
Building and testing on your local machine can catch style error. Did you do it...?

lukess:
ah~ I followed the https://github.com/lukess/heron/blob/master/website/content/docs/operators/heron-ui.md

I think I should follow
http://twitter.github.io/heron/docs/developers/compiling/mac/
and
http://twitter.github.io/heron/docs/contributors/testing/

objmagic:
Yes. If you work on OS X, you can do ``bazel build --config=darwin heron/...`` and then ``bazel test --config=darwin heron/...`` before PR.

heron/ui/src/python/main.py
lukess: @kramasamy @objmagic 
In the multiple NICs env socket.gethostbyname(socket.gethostname()) would return virtual NIC address such as vpn, virualbox, docker...
Therefore, I added '0.0.0.0' as Tornado default bind for IPv4, and It's not a good solution too.

kramasamy: @lukess - if the user knows that there are multiple NICs, you address port combination will help.

kramasamy: Also, you might want to make the same changes to heron-tracker as well?

lukess: @kramasamy you are right 0.0.0.0 is not a good log. Make it clearer,
$ heron-ui, currently it prints Listening at http://{ip address from one of the interfaces}:8889, but it actually LISTEN 0.0.0.0/:::8889

kramasamy: @lukess - that is bad. It does this in the case of multiple NICs using socket.gethostbyname(), is that correct?

kramasamy: in my case it does the following
```
Listening at http://127.0.0.1:8889
```

lukess: @kramasamy honestly I don't know right now. I will test it later. My assumption is from https://github.com/tornadoweb/tornado/blob/7ec71d46a83eade88f6d588f8f568f9759676c89/tornado/netutil.py#L111.

kramasamy: @lukess - can you check how does your /etc/hosts file looks like?

lukess: @kramasamy sure,
127.0.0.1      	    localhost
255.255.255.255	broadcasthost
::1             localhost

May I know LISTEN when you run heron-ui too?
$ lsof -i -P | grep LISTEN

kramasamy: it is listening in IPv6 and IPv4 in port *.8889

lukess: ok~ it did listen to 0.0.0.0 and ::
I think the problem right now is if want to force address to socket.gethostbyname(socket.gethostname()) by default which always send Address into Tornado bind.

lukess: or add a warning log, and print 0.0.0.0 as well.

heron/ui/src/python/main.py
```diff
@@ -134,14 +135,21 @@ def main():
 
   # log additional information
   command_line_args = vars(parsed_args)
-  address = socket.gethostbyname(socket.gethostname())
+
+  address = '0.0.0.0'
```
kramasamy: you should use DEFAULT_ADDRESS

heron/ui/src/python/main.py
```diff
@@ -134,14 +135,21 @@ def main():
...
   Log.info("Using tracker url: %s", command_line_args['tracker_url'])
 
   # pass the options to tornado and start the ui server
-  define_options(command_line_args['port'], command_line_args['tracker_url'])
+  define_options(command_line_args['address'],
+                 command_line_args['port'],
+                 command_line_args['tracker_url'])
   http_server = tornado.httpserver.HTTPServer(Application())
-  http_server.listen(command_line_args['port'])
+  http_server.listen(command_line_args['port'], address=command_line_args['address'])
```
kramasamy: should it address=address?

lukess: my bad! I did not read create_parsers() carefully. The default arguments have been set up. Should I add tracker address in this pull request?

kramasamy: @lukess - please consider a seperate PR for the tracker address.

PR #1119, opened 2016-07-20
add support for heronrc file

prabhuinbarajan:
see readme

PR #1257, opened 2016-08-13
Keep argument value as list

objmagic:
A bug introduced by https://github.com/twitter/heron/pull/1119 takes only the first element of the argument value list and destructively updates the corresponding key argument in the dictionary. See: https://github.com/twitter/heron/pull/1119/files#r74668021

This problem is blocking PyHeron integration test so this is not good. This PR fixed it.

cc: @prabhuinbarajan @taishi8117

### DIFF
--- a/heron/cli/tests/python/heronparser_unittest.py
+++ b/heron/cli/tests/python/heronparser_unittest.py
@@ -134,8 +134,6 @@ class HeronParserTest(unittest.TestCase):
     args, _ = parser.parse_known_args(["activate", "devcluster/ads/PROD",
                                        "12313", "--config-property", "a=b"])
     namespace = vars(args)
-    with open('/Users/rli/a.txt', 'w') as f:
-      f.write(str(namespace))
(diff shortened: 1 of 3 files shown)


kramasamy:
👍

prabhuinbarajan:
sounds good to me... one call out on heronrc - when the same argument is provided at multiple levels, this was supposed to clobber the argument list and enforce the precedence. Now we should find another way to support that or have a convention in place where same argument is not supplied at multiple levels .

objmagic:
> when the same argument is provided at multiple levels, this was supposed to clobber the argument list and enforce the precedence

Do you mean for `.heronrc` file like:
````
........ --config-property a=b
...
...
........ --config-property c=d --config-property a=b
````
 
we should dedup list ``--config-property a=b --config-property c=d --config-property a=b`` to ``--config-property a=b --config-property c=d``? Is this what you meant by "clobber"?

Also, can you **elaborate** what you mean by "precedence"? and why we should preserve it?

PR #1260, opened 2016-08-15
heron cli cleanup

billonahill:
A few changes I've made while working on the scalding work that can merge now.

The `topology_name` initialization had to move out of the try block, since it's referenced in the rescue.

### DIFF
--- a/heron/cli/src/python/main.py
+++ b/heron/cli/src/python/main.py
@@ -85,6 +85,7 @@ def create_parser():
   kill.create_parser(subparsers)
   restart.create_parser(subparsers)
   submit.create_parser(subparsers)
+  update.create_parser(subparsers)
   version.create_parser(subparsers)
 
   return parser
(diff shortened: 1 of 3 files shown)


heron/cli/src/python/main.py
```diff
@@ -100,30 +100,21 @@ def run(command, parser, command_args, unknown_args):
...
+      'activate':activate,
+      'deactivate':deactivate,
+      'kill':kill,
+      'restart':restart,
+      'submit':submit,
+      'help':cli_help,
+      'version':version,
+  }
+
+  if command in runners.keys():
```
objmagic: `keys()` is not necessary here. Feel free to ignore since this is trivial...

objmagic:
👍

heron/cli/src/python/main.py
billonahill: I had no idea...

PR #1263, opened 2016-08-16
Fix flaky SpoutInstanceTest

windie:
SpoutInstanceTest should register the singleton objects before offering the message. Otherwise, TestBolt maybe has already processed several tuples before the singleton objects are registered.


kramasamy:
@maosongfu - can you please take a look at this?

maosongfu:
:+1: looks good to me

PR #1270, opened 2016-08-17
Remove duplicate code from state manager tests

billonahill:
Removing duplicate code from tests - no functional changes here. This will simplify adding new tests without copy and paste.


objmagic:
👍

PR #1276, opened 2016-08-18
move cli, explorer, tracker and ui and tools common files 

kramasamy:
into heron/tools


objmagic:
this breaks since the hardcoded level is changed:
https://github.com/twitter/heron/blob/master/heron/common/src/python/utils/config.py#L130

heron/tools/common/src/python/utils/config.py
```diff
@@ -151,7 +151,7 @@ def get_heron_dir_explorer():
   From heron-cli with modification since we need to reuse cli's conf
   :return: root location for heron-cli.
   """
```
taishi8117: btw, why do we need this? Isn't this the same with `get_heron_dir()`?

taishi8117:
heron-ui is still not working

kramasamy:
ui, tracker, cli and explorer are working. Let me know if this PR is good to go.

objmagic:
👍 LGTM

Issue #1264, opened 2016-08-16
Remaining things to complete for PyHeron


taishi8117:
The following is a list of remaining things to complete for the Python topology support.

- Streamparse API compatibility
  - Batching Bolt, Java Spout/Bolt, and Shell Spout/Bolt aren't supported yet.
  - Some methods are not fully implemented or supported.
- Compilation
  - We should include an automated script to compile a Python topology project to the deployable PEX file (maybe using Pants or Bazel).
- Testing with other schedulers
  - Python topologies are only tested locally as of now. We need to make sure they will work on other schedulers such as Aurora.
- More integration tests
  - Local integration tests (failure test) are not done yet.
  - Some functionalities, such as custom grouping, are not tested in integration tests (same for Java).
- Multi-thread Heron Instance
- Performance Improvement
  - We can try the C++ implementation for the Python Protocol Buffer.
  - Maybe upgrade to the Protocol Buffer v3 (refer to #1183).
- Default Metrics
  - Some default metrics are not implemented yet.
  - We should also create Python-process related metrics (something that is equivalent of JVM-related metrics).
- Updates to Heron Tracker and Heron UI
  - Some of their functionalities are currently exclusive to Java topologies (like JStack, Memory histrogram, etc.).
- More unit tests (#1251)
- Making system config singleton/global
  - Currently, system config is passed to many components via an argument. We can make it singleton or global variable.

PR #1278, opened 2016-08-18
Some fixes to PyHeron (#1264)

taishi8117:
Mainly to improve the compatibility with the Streamparse API.
- Add NotCompatibleError for the streamparse APIs that are invalid for PyHeron
- Add process_tick()


objmagic:
👍

PR #1281, opened 2016-08-19
Added Streamparse copyrights to PyHeron API.

### LINKED ISSUE #1259: Streamparse copyright necessary for under heron/pyheron

taishi8117:
Resolves #1259 

Note that [`component_spec.py`](https://github.com/taishi8117/heron/blob/pyheron/copyright/heron/pyheron/src/python/component/component_spec.py) is original.


billonahill:
Will this pass checkstyles?

taishi8117:
@billonahill it seems to have passed. I think we don't have a Python check style for copyright strings, which we should probably do?

billonahill:
👍 

(yes we should add a copywrite check for py)

objmagic:
👍 

@kramasamy any final comment? I think license related PR needs your final review

kramasamy:
👍

PR #1282, opened 2016-08-19
Adding packing plan methods to SchedulerStateManagerAdaptor

billonahill:
I forgot to include my changes to the adaptor in #1274.

PR #1284, opened 2016-08-19
Update Launcher to set PackingPlan in StateManager

billonahill:
Once #1282 ships, which adds `setPackingPlan` and `deletePackingPlan` to `SchedulerStateManagerAdaptor` this patch will pass CI and can then ship. This will cause PackingPlan to be persisted in `StateManager` upon submit.

Once this is shipped, we can update the heron executor to read PackingPlan on startup fron the state manager, then we can get rid of the dreaded instance distribution string.


objmagic:
restarted CI

billonahill:
Thanks @objmagic for the restart. This failed because I needed to merge in #1282, which I just did. The current run should pass.

objmagic:
👍

Issue #953, opened 2016-06-17
Annotation scheme for interfaces that are evolving, private, etc.


billonahill:
Hadoop has an annotation scheme to use to convey to users if public interfaces or methods are stable, unstable, evolving, public, or private. We should leverage something similar to make contracts clear.

https://hadoop.apache.org/docs/r2.7.2/api/org/apache/hadoop/classification/package-summary.html

PR #1285, opened 2016-08-19
Repacking Interface to be used for autoscaling

avflor:
Added a repacking interface to be used for scaling and made small edits to the packing interface to make it compatible with the repacking interface.


billonahill:
@avflor put this through a PR on `billg/scaling_proto` (see https://github.com/billonahill/heron/pull/4) and we did a round of reviews there. To provide a little more context, the key change proposed here (besides the new `IRepacking` interface) is to the `IPacking` interface:
```
-  void initialize(Config config, Config runtime);
+  void initialize(Config config, TopologyAPI.Topology topology);
```

The goal is to make the initialize method for `IPacking` the same as for `IRepacking` to simplify classes that implement both. By standardizing on `Config config, TopologyAPI.Topology topology` we can do that but it's of course a breaking change for `IPacking` impls in the wild. I think that's ok (assuming we're ok with this signature) since we're still evolving the APIs and the refactor to `IPacking` impls would be straight forward.

See discussion in https://github.com/billonahill/heron/pull/4#discussion_r74977716 for more context.

heron/packing/src/java/BUILD
```diff
@@ -7,6 +7,10 @@ packing_deps_files = [
     "//heron/spi/src/java:packing-spi-java",
 ]
 
+nullpacking_debs_files = \
```
billonahill: typo: debs

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -135,25 +145,46 @@ public void initialize(Config config, Config runtime) {
...
+
+  }
 
+  /**
+   * Estimate the per instance and topology resources for the packing plan created using
+   * FFD placement
+   *
+   * @return Resources required
+   */
+  private Resource estimateResources(Map<String, List<String>> ffdAllocation,
```
billonahill: There are more changes in this class than required for a new initialize signature. Is that because you're refactoring to make things easier when implementing `IRepacking` or are these functional changes?

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -99,10 +99,13 @@
...
   protected long instanceDiskDefault;
+
   protected long maxContainerRam;
   protected double maxContainerCpu;
   protected long maxContainerDisk;
 
+  protected int paddingPercentage;
+
   public static String getContainerId(int index) {
     return "" + index;
```
billonahill: `return Integer.toString(index);` is more idiomatic.

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
avflor: This is refactoring

billonahill:
Would anyone else like to review before merging? Let me know and I'll hold off.

/cc @kramasamy @maosongfu @ashvina @nlu90 @objmagic

billonahill:
Addresses part of #1292.

ashvina:
Now, the signature of `initialize` method in `IPacking` and IRepacking is not same as the one in `IScheduler`, `ILauncher`, etc. Is that ok?

billonahill:
I think the signature should match what the implementations typically (or contractually) need. In the case of `IPacking`, every implementation just fishes the Topology from the Runtime and then drops it:
```
this.topology = com.twitter.heron.spi.utils.Runtime.topology(runtime);
```
so I think it makes sense to be more explicit with the init signature and pass the Topology itself. Unless we foresee a reason why one might need runtime for something else.

IIRC during repacking we don't have the runtime available to us. @avflor correct me if I'm wrong. The `IScheduler` and `ILauncher` always has it at submission time though.

That said, I'm open for other suggestions.

avflor:
I also agree with Bill that the runtime is not needed. So far it has been used in packing just to extract the topology. In my first repacking algo implementation I have not used it at all. I think we should not include it in the signature unless there is evidence that a packing or repacking algorithm might need it.  I don't see this happening right away. Can we remove it for now and include it if there is need in the future or this is too messy?

billonahill:
+1 @avflor
 
I think it's ok for our interfaces to still be evolving at this stage. (I'd love to find time for #953 for this purpose.)

ashvina:
@avflor @billonahill I agree with your comments. Currently there is no need.

However I am worried about the cost of changing interface. A small change would require changing all implementations of IPacking, changing all test cases where methods are invoked directly, and potentially changing [power]mocks where methods could be invoked using reflection. The current pattern has enough headroom for future enhancements. It has some qualities of the Spring framework, where in the dependencies are injected into the implementation. For e.g. REEF is based on such a pattern.

My impression is that the confusion is created by incompleteness of the `runtime` config object. In this case I would prefer keeping the config object complete over changing the interface. IMO this change introduces a tight coupling.

billonahill:
If we're not concerned about breaking backward compatibility of external clients (which we should not be at this stage) than we should be able to evolve our interfaces with confidence to get them right. The reflection risk is the only one that we can't catch at compile/test time so we'll need to look for where that might happen, or better yet we should really never, ever be using reflection as it's a tool of the devil (frameworks like Guice achieve the same with compile-time checks).

I don't fully agree with the Spring DI comparison. That pattern promotes passing the objects depended on into the class that needs them, which in this case is the Topology. That pattern doesn't promote passing what's effectively a giant HashMap of un-typed Objects (i.e. runtime) deep into the code stack and having objects fish out what's important to them.

ashvina:
@billonahill Totally agree with you regarding untyped-giant-HashMap. I find it hard to debug and ugly. And DI is out of scope for this change. For now this interface change looks ok.

avflor:
@kramasamy Any feedback on this PR? Should we merge?

billonahill:
See #1304 for an approach for specifying interface audience and stability.

kramasamy:
Sorry to get back on this late. If I remember correctly, the reason why we did Config runtime is to future proof any other runtime config that we might need.  Now passing only Topology.TopologyAPI makes it restrictive. This means we need to change current implementations to use this interface and later if we need any additional stuff, we might have put back Config runtime - again leading to more changes - IMO.

billonahill:
@kramasamy I agree with the intent. That said, none of the existing IPacking impls need it and IRepacking doesn't have access to it so I'm ok with refactoring it out for now.

kramasamy:
+1 - let us go with the refactoring.

billonahill:
Thanks @avflor!

PR #1287, opened 2016-08-19
Fix code block rendering.

### LINKED ISSUE #1283: First Fit Decreasing Packing Algorithm docs broken
The example code snippet at the bottom of this page is not rendering.

http://twitter.github.io/heron/docs/developers/packing/ffdpacking/

objmagic:
fix #1283 

(Atom editor automatically removes trailing white spaces so one only needs to review the last part)

cc @billonahill


billonahill:
Why 4 backticks and not 3 as seen elsewhere in the md docs?

objmagic:
ah, I am just used to using four backticks... will change to three now.

billonahill:
So the issue was the wrong number of indent spaces?

👍

objmagic:
@billonahill no. It is because a new line was not inserted before the code block.

PR #1288, opened 2016-08-20
Clean up executor logging in LocalScheduler

billonahill:
Executor log files in LocalScheduler currently look like this:
```
heron-executor-d7e5809a-started.stderr
heron-executor-dc47dc90-started.stderr
heron-executor-edd18eeb-started.stderr
heron-executor-effb081c-started.stderr
heron-executor.stdout
```
Two issues:
1. Every run of the executor makes a new stderr file which causes file bloat of mostly 0-byte sized files. Each run shares the same stdout file though, which is inconsistent.
2. When running multiple executors, they all log to the same stdout file, which makes the file contents difficult to read.

This changes executor logging to look more like the logging of the other processes. If you have two executors you'd see something like this:
```
heron-executor-0.stderr
heron-executor-0.stdout
heron-executor-1.stderr
heron-executor-1.stdout
```

If one fails and is restarted, it continues to log to the same file as the previous instance.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -595,10 +595,10 @@ def signal_handler(signal_to_handle, frame):
   # Just catch the SIGTERM and then cleanup(), registered with atexit, would invoke
   sys.exit(signal_to_handle)
 
-def setup():
+def setup(shardid):
   # Redirect stdout and stderr to files in append mode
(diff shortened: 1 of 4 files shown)


billonahill:
Ping. Any comments on this one?

objmagic:
LGTM

👍

PR #1290, opened 2016-08-21
Add Heron list of ADOPTERS

caniszczyk:
Start a list of Heron adopters, it's convenient to see who is using Heron in production!

See Finagle as an example: https://github.com/twitter/finagle/blob/develop/ADOPTERS.md


objmagic:
looks good. we can also display this list on the website.

caniszczyk:
I'm sure @kramasamy can add more in this list, but it's just a start, it's more important we have a location to adopters to be listed :)

objmagic:
👍

PR #1293, opened 2016-08-22
Take profiling env var.

objmagic:
This PR makes `heron-executor` takes environmental variable `HERON_PROFILE` set before submitting topology so that we can profile Python topology at run-time. 

Not really sure if `HERON_PROFILE` is a good name candidate here.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -432,9 +432,16 @@ class HeronExecutor(object):
       Log.info("%s stderr: %s" %(name, process_stderr))
 
   def _run_process(self, name, cmd, env_to_exec=None):
+    if env_to_exec is None:
+      _env = os.environ.copy()
+    else:
+      _env = env_to_exec
+    _env["PEX_PROFILE"] = os.environ.get("HERON_PROFILE", "false")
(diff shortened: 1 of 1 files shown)


kramasamy:
How can you set the variable at run time?

objmagic:
one can set this env var in shell before submitting topology, and `heron-executor` takes this env var at run-time.

kramasamy:
It will work only in local mode - not in distributed mode, correct?

objmagic:
You're right... Maybe this feature is mostly targeting Heron developer?

taishi8117:
maybe we can add an option to heron-cli (or heronrc) so custom env var can be set on distributed mode too?

kramasamy:
@taishi8117 - that is a good suggestion. I am looking for something similar to Java, where you could enable it for a particular spout/bolt - during runtime. Is it possible?

objmagic:
@kramasamy I think that is possible by adding this profiling option into API. And we don't need to add this option in command-line tool options.

kramasamy:
@objmagic - my concern is to have some consistency across the languages.

objmagic:
temporarily close this since we need to reach consensus on design

PR #1296, opened 2016-08-23
Improving logging and retries of local int test runner

billonahill:
When a local integration test fails to get the test result it needs to also kill the topology. Otherwise retries will fail with "topology already exists" errors.

### DIFF
--- a/integration-test/src/python/local_test_runner/main.py
+++ b/integration-test/src/python/local_test_runner/main.py
@@ -68,6 +68,9 @@ def runTest(test, topologyName, params):
   while not processExists(processList, HERON_STMGR_CMD):
     processList = getProcesses()
 
+  _safe_delete_file(params['readFile'])
+  _safe_delete_file(params['outputFile'])
+
   # insert lines into temp file and then move to read file
   try:
     with open('temp.txt', 'w') as f:
@@ -132,12 +135,8 @@ def runTest(test, topologyName, params):
       return False
 
     # delete test files
-    try:
-      os.remove(params['readFile'])
-      os.remove(params['outputFile'])
-    except Exception as e:
-      logging.error("Failed to delete test files")
-      return False
+    _safe_delete_file(params['readFile'])
+    _safe_delete_file(params['outputFile'])
 
   # get actual and expected result
   # retry if results are not equal a predesignated amount of times
@@ -298,6 +297,14 @@ def processExists(processList, processCmd):
     if processCmd in process.cmd:
       return True
   return False
+
+def _safe_delete_file(file):
+  if os.path.isfile(file) and os.path.exists(file):
+    try:
+      os.remove(file)
+    except Exception as e:
+      logging.error("Failed to delete file: %s: %s", file, e)
+      return False
 
 def main():
   ''' main '''
--- a/integration-test/src/python/local_test_runner/main.py
+++ b/integration-test/src/python/local_test_runner/main.py
@@ -196,7 +196,6 @@ def submitTopology(heronCliPath, testCluster, testJarPath, topologyClassPath,
       '%d' % (len(TEST_INPUT))
   ]
   logging.info("Submitting topology: %s", splitcmd)
-  logging.info(splitcmd)
   p = subprocess.Popen(splitcmd)
   p.wait()
   logging.info("Submitted topology %s", topologyName)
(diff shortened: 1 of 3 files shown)


integration-test/src/python/local_test_runner/main.py
```diff
@@ -25,7 +25,7 @@
 ]
 # Retry variables in case the output is different from the input
 RETRY_COUNT = 5
```
nlu90: we can increase the RETRY_COUNT to allow the test has more chance while decreasing the RETRY_INTERVAL.

integration-test/src/python/local_test_runner/main.py
```diff
@@ -190,34 +195,35 @@ def submitTopology(heronCliPath, testCluster, testJarPath, topologyClassPath,
       '%s' % (outputFile),
       '%d' % (len(TEST_INPUT))
   ]
-  logging.info("Submitting topology: ")
+  logging.info("Submitting topology: %s", splitcmd)
   logging.info(splitcmd)
```
nlu90: the splitcmd is logged twice.

integration-test/src/python/local_test_runner/main.py
```diff
@@ -190,34 +195,35 @@ def submitTopology(heronCliPath, testCluster, testJarPath, topologyClassPath,
...
   splitcmd = [
       '%s' % (heronCliPath),
       'kill',
       '--verbose',
       '%s' % (testCluster),
       '%s' % (topologyName),
   ]
-  logging.info("Killing topology:")
+  logging.info("Killing topology: %s", splitcmd)
   logging.info(splitcmd)
```
nlu90: the splitcmd is logged twice.

integration-test/src/python/local_test_runner/main.py
```diff
@@ -235,11 +240,11 @@ def restartShard(heronCliPath, testCluster, topologyName, shardNum):
       '%s' % (topologyName),
       '%d' % shardNum
   ]
-  logging.info("Killing TMaster command:")
+  logging.info("Killing TMaster: %s", splitcmd)
   logging.info(splitcmd)
```
nlu90: the splitcmd is logged twice.

objmagic:
FYI, @maosongfu wanted me to improve the local integration test since he spotted some potential flaws. I have to admit that I spent some time on it but felt there is no fundamentally design/programming flaw. Maybe you would like to have a quick look?

integration-test/src/python/local_test_runner/main.py
billonahill: IMO if a test is failing more than 5 times it's either a bad test or our test framework is too brittle. The fact that we have to retry at all is suspect.

integration-test/src/python/local_test_runner/main.py
billonahill: Where do you see the other occurrence? I'm not seeing it. Below we log topologyName.

integration-test/src/python/local_test_runner/main.py
billonahill: where?

objmagic: `logging.info("Killing topology: %s", splitcmd)`

and again down below

`logging.info(splitcmd)`

billonahill:
Fixed split command dup logging.

@objmagic I see the following areas for improvement:
- There's too much sleeping and retrying. We should be able to determine when we can proceed by watching something or getting an event, but sleep/retry === brittle/slow.
- When a test fails there's little to no info re why it failed. For example look at [this failed CI](https://travis-ci.org/twitter/heron/builds/154576607) where the output file wasn't found. Why did it not appear? What process logs might show us? There's no way to know and if you can't reproduce you're stuck. You can investigate process logs like you would on a running system.
- The tests are too black box. They give input and check output but there's no white-box testing (e.g., assert state of processes or state manager along the way). For scaling int tests we'd need that to verify the state of the system with more granularity along the way.
- The test pattern should be more easily extensible. Currently it's submit/kill process/verify, but you could imagine more variations, like scaling.

objmagic:
@billonahill thx for the suggestions. I'll take care of them at some time.

kramasamy:
@objmagic - can you create issues for @billonahill 's suggestions so that we can track them.

PR #1299, opened 2016-08-23
kafka spout patching

nlu90:
Patching heron for it work with kafka-spout.

### DIFF
--- a/heron/storm/src/java/org/apache/storm/Config.java
+++ b/heron/storm/src/java/org/apache/storm/Config.java
@@ -272,7 +272,7 @@ public class Config extends com.twitter.heron.api.Config {
   /**
    * The root directory in ZooKeeper for metadata about TransactionalSpouts.
    */
-  public static final String TRANSACTIONAL_ZOOKEEPER_ROOT="transactional.zookeeper.root";
+  public static final String TRANSACTIONAL_ZOOKEEPER_ROOT = "transactional.zookeeper.root";
 
   /**
    * The session timeout for clients to ZooKeeper.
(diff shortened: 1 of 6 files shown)


mycFelix:
Great!!!

heron/storm/src/java/org/apache/storm/Config.java
```diff
@@ -260,6 +260,53 @@
...
+
+  /**
+   * The port Storm will use to connect to each of the ZooKeeper servers.
+   */
+  public static final String STORM_ZOOKEEPER_PORT = "storm.zookeeper.port";
+
+  /**
+   * The root directory in ZooKeeper for metadata about TransactionalSpouts.
+   */
+  public static final String TRANSACTIONAL_ZOOKEEPER_ROOT="transactional.zookeeper.root";
```
objmagic: umm, to keep uniformity, should we surround assignment operator using whitespace?

objmagic:
👍

kramasamy:
@nlu90 - This patch will suffice for running Storm kafka spout. Or we need this for heron-kafka spout as well?

heron/storm/src/java/org/apache/storm/Config.java
nlu90: good find

nlu90:
@kramasamy we need this for heron-kafka spout as well

kramasamy:
@nlu90 - instead of introducing another set of configuration for the spout, I was wondering, can we think about making it more simpler. Looking at the config - kafka spout - seems to adding a lot of config and hard coded as well. We want to move away from hard coding the config into the topology. We can introduce the notion of config per component - in a cleaner fashion.

nlu90:
@kramasamy If we want to introduce the notion of config per component, what are the changes needed?

kramasamy:
ship it 👍

PR #1301, opened 2016-08-24
YARN: Kill and request new container(s) on restart

### LINKED ISSUE #1244: YARN scheduler should allocate new container on restart
The current implementation of `restart` on a YARN cluster kills and restarts the executor. The worker container is retained. However this is different from the expected behavior. `restart` command is meant to move instances to a new container. This is primarily used when a node is faulty or container "hangs". Like other schedulers, the YARN scheduler should kill the container and allocate a new container.

ashvina:
Fixes #1244 
Instead of restarting executor process in the same container, restart kills
current container and requests for a new one.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
@@ -185,7 +185,7 @@ public class HeronMasterDriver {
 
     Optional<HeronWorker> worker = multiKeyWorkerMap.lookupByWorkerId(id);
     if (!worker.isPresent()) {
-      LOG.log(Level.WARNING, "Active container for {0} not found. Requesting a new one.", id);
+      LOG.log(Level.WARNING, "Requesting a new container for: {0}", id);
       ContainerPlan containerPlan = packing.containers.get(id);
       if (containerPlan == null) {
         throw new IllegalArgumentException(
@@ -285,7 +285,6 @@ public class HeronMasterDriver {
   void submitHeronExecutorTask(String workerId) {
     Optional<HeronWorker> worker = multiKeyWorkerMap.lookupByWorkerId(workerId);
     if (!worker.isPresent()) {
-      LOG.log(Level.SEVERE, "Container for id: {0} not found.", workerId);
       return;
     }
 
@@ -348,9 +347,9 @@ public class HeronMasterDriver {
       }
     }
 
-    HeronWorker lookupByEvaluatorId(String evaluatorId) {
+    Optional<HeronWorker> lookupByEvaluatorId(String evaluatorId) {
       synchronized (workerMap) {
-        return evaluatorWorkerMap.get(evaluatorId);
+        return Optional.fromNullable(evaluatorWorkerMap.get(evaluatorId));
       }
     }
 
@@ -453,21 +452,21 @@ public class HeronMasterDriver {
     @Override
     public void onNext(FailedEvaluator evaluator) {
       LOG.log(Level.WARNING, "Container:{0} failed", evaluator.getId());
-      HeronWorker worker = multiKeyWorkerMap.lookupByEvaluatorId(evaluator.getId());
-      if (worker == null) {
+      Optional<HeronWorker> worker = multiKeyWorkerMap.lookupByEvaluatorId(evaluator.getId());
+      if (!worker.isPresent()) {
         LOG.log(Level.WARNING,
             "Unknown executor id for failed container: {0}, skip renew action",
             evaluator.getId());
         return;
       }
       LOG.log(Level.INFO, "Trying to relaunch executor {0} running on failed container {1}",
-          new Object[]{worker.workerId, evaluator.getId()});
-      multiKeyWorkerMap.detachEvaluatorAndRemove(worker);
+          new Object[]{worker.get().workerId, evaluator.getId()});
+      multiKeyWorkerMap.detachEvaluatorAndRemove(worker.get());
 
       try {
-        launchContainerForExecutor(worker);
+        launchContainerForExecutor(worker.get());
       } catch (ContainerAllocationException e) {
-        LOG.log(Level.SEVERE, "Failed to relaunch failed container: " + worker.workerId, e);
+        LOG.log(Level.SEVERE, "Failed to relaunch failed container: " + worker.get().workerId, e);
       }
     }
   }
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
@@ -21,8 +21,10 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -86,6 +88,9 @@ public class HeronMasterDriver {
   private final String topologyJar;
   private final int httpPort;
   private final boolean verboseMode;
+
+  // This map contains all the active workers managed my this scheduler. The workers can be
+  // looked up by heron's executor id or REEF's container id.
   private MultiKeyWorkerMap multiKeyWorkerMap;
 
   // TODO: https://github.com/twitter/heron/issues/949: implement Driver HA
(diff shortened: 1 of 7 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -139,107 +142,129 @@ public HeronMasterDriver(EvaluatorRequestor requestor,
...
+        launchContainerForExecutor(new HeronWorker(TMASTER_CONTAINER_ID, 1, TM_MEM_SIZE_MB));
+    return container != null;
   }
 
   /**
    * Container allocation is asynchronous. Request containers serially to ensure allocated resources
    * match the required resources
    */
-  void scheduleHeronWorkers(PackingPlan topologyPacking) {
+  boolean scheduleHeronWorkers(PackingPlan topologyPacking) {
```
kramasamy: From a terminology perspective what does HeronWorker mean?

ashvina: A HeronWorker represents a YARN container for heron instances. Scheduler and TMaster containers are managed differently.

kramasamy:
looks good to me 👍 @billonahill and @maosongfu - do you have any comments?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -139,107 +142,129 @@ public HeronMasterDriver(EvaluatorRequestor requestor,
...
     });
+
+    try {
+      return result.get();
+    } catch (InterruptedException e) {
+      LOG.log(Level.SEVERE, "Interrupted while waiting for container", e);
+    } catch (ExecutionException e) {
+      LOG.log(Level.SEVERE, "Error while waiting for container", e);
+    }
+    return null;
```
billonahill: It appears like you're returning null here and above to mask an exceptional condition. Could you instead throw an exception?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -139,107 +142,129 @@ public HeronMasterDriver(EvaluatorRequestor requestor,
...
-    return evaluator;
+    try {
+      AllocatedEvaluator evaluator = allocatedContainerQ.take();
+      LOG.log(Level.INFO,
+          "Container {0} is allocated for executor, id: {1}",
+          new Object[]{evaluator.getId(), id});
+      return evaluator;
+    } catch (InterruptedException e) {
+      LOG.log(Level.WARNING, "Interrupted while waiting for container allocation", e);
+      return null;
```
billonahill: this seems to be an exceptional case. can we throw an exception instead of returning null?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -139,107 +142,129 @@ public HeronMasterDriver(EvaluatorRequestor requestor,
...
   }
 
-  /**
-   * REEF will retain all YARN containers. Currently running heron-executors/REEF-tasks
-   * will be killed and new REEF-tasks representing previously running heron-executor will be
-   * started.
-   */
-  public void restartTopology() {
-    restartContainer(null);
+  public boolean restartTopology() {
```
billonahill: Are you using the boolean as a substitute for an exception here? If so, we should throw an exception if things that are expected to execute fail to do so.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -318,10 +324,79 @@ void submitHeronExecutorTask(ActiveContext context) {
...
+      }
+    }
+
+    HeronWorker lookupByWorkerId(String workerId) {
+      HeronWorker worker;
+      synchronized (workerMap) {
+        worker = workerMap.get(workerId);
+      }
+      if (worker == null) {
+        LOG.log(Level.WARNING, "Container for executor id: {0} not found.", workerId);
```
billonahill: Is this an exceptional condition?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: IMO this is inline with the restart method definition: https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/scheduler/IScheduler.java#L85. If the restart operation fails, the core `runtime` logic expects a boolean: https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java#L118

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: Ok, will throw the exception.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: While I agree with your observation, I think returning null is the right thing to do in this case. This method allocates a new container. Either it works or there is an exception. The exceptions seems to be independent of the context and currently have a common reaction. IMO, not throwing this error up the chain would be better.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: This would not happen normally. This could happen if two events happen concurrently. For e.g. user invokes restartContainer and reef invokes containerFailure event. If it happens, lookup will fail and the calling method can decide the action. As of now I think it is worthwhile tracking if it happens.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
billonahill: Ugh. Ah yes, the code is riddled with these unfortunate booleans where each callers in the stack checks for false and logs an error and returns false. It's really a poor man's exception throw IMO.

My thought on this is to not propagate this pattern more than we need to. We should implement the boolean return type when the interface demands it, but otherwise we should write new code to expect success and throw an exception on failure. What do you think?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
billonahill: My dislike of nulls (one of them) is that they're so ambiguous in meaning. It's much more expressive to throw something like a a ContainerAllocationException, which could be extended with stats codes, reasons, response messages, call stacks, etc. Returning a null basically swallows all contextual information of what happened below, leaving the caller to just log an ambiguous message and move on.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
billonahill: If this is a condition that should not ever happen, that's more of a reason IMO to throw an verbose exception than to log and return null and make all other callers above have to check for null. The cleanest implementation in my opinion is to annotate methods with @NotNull and they either do what's expected or they throw. The calling code never needs to check for null and NPEs won't ever occur.

ashvina: Given the distributed and asynchronous nature of this system, this condition can happen. This `hashmap` would not know why a particular key is missing and if it is an error situation. So I think an exception should not be thrown in this case.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: I see your point. I will recheck this case.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
billonahill: If this value might not be found under normal circumstances, could we use Optional instead?
http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: Agree with improper use of `null` or `boolean`. I will fix update this instance as it may make the code more readable.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: Optional would be used once Heron's own implementation is available as we cannot depend on java-8 for now.

billonahill: We can use guava's, which I've already introduced here:
https://github.com/twitter/heron/pull/1218/files#diff-c7b1c2644302d7242028db94c6ea727cR45

ashvina: I modified the code to use Optional. To me use of Optional looked like a force fit. In absence of other java-8 features replacing null check with Optional.isPresent does not add any value. While I like the concept in general, I would like to skip its usage here. Thoughts?

billonahill: I tend to think that map like lookup functionality is an ideal place to return an Optional (guava's), regardless of whether we're using other Java 8 features or not.

ashvina: Updated to replace null with Optional.

billonahill: The caller of this code always seems to log when isPresent is false so let's not double log here. Instead this method could just return `Optional.fromNullable(workerMap.get(workerId))` inside the sync block.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -379,38 +449,25 @@ public void onNext(AllocatedEvaluator evaluator) {
...
-      String executorId = reefContainerToHeronExecutorMap.get(evaluator.getId());
-      LOG.log(Level.WARNING,
-          "Container:{0} executor:{1} failed",
-          new Object[]{evaluator.getId(), executorId});
-      if (executorId == null) {
-        LOG.log(Level.SEVERE,
-            "Unknown executorId for failed container: {0}, skip renew action",
+      LOG.log(Level.WARNING, "Container:{0} failed", evaluator.getId());
+      HeronWorker worker = multiKeyWorkerMap.lookupByEvaluatorId(evaluator.getId());
+      if (worker == null) {
```
billonahill: can we also use Optional in lookupByEvaluatorId(..)?

ashvina: Updated.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
ashvina: Fixed instances where missing container was logged twice.

billonahill:
👍 for merging once CI completes. Thanks @ashvina!

kramasamy:
@ashvina - can you resolve and update so that we can merge it?

ashvina:
@kramasamy The merge conflicts are resolved now. Thanks !

Issue #1292, opened 2016-08-22
Support dynamically scaling the number of bolts or spouts of a running topology


billonahill:
Provide a method for a topology owner to run a command at the command line to change the number of bolt/spouts of a running topology. This involves regenerating the topology packing plan and possibly starting/stoping containers and instances.

This work is a precursor for supporting auto-scaling functionality, which would allow automated algorithmic scaling based on observed load and behavior.

billonahill:
A prototype branch for this work can be seen in #1218.

PR #1302, opened 2016-08-25
Refactor HeronExecutor signal handling

billonahill:
Cherry picking refactorings that I made during scaling that can be committed first, since they don't relying on the scaling functionality (but will be needed in some cases). No new functionality here, just moving some things around to clean up before adding new functionality.

1. Refactored how exit signal handling is done. This will help when we introduce a state manager that needs closing.
1. Renamed some public methods (`prepare_launch()` -> `initialize()`, `monitor_process()` -> `start_process_monitor()`)
1. Added synchronization around `processes_to_monitor`. 

Relates to #1292

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -15,15 +15,16 @@
 #!/usr/bin/env python2.7
 ''' heron-executor '''
 import atexit
-import os
-import sys
-import signal
-import subprocess
-import time
-import json
 import base64
-import string
+import json
+import os
 import random
+import signal
+import string
+import subprocess
+import sys
+import threading
+import time
 import yaml
 
 from heron.common.src.python.utils import log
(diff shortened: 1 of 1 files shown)


kramasamy:
looks good to me 👍

heron/executor/src/python/heron_executor.py
```diff
@@ -539,40 +564,32 @@ def get_command_changes(self, current_commands, updated_commands):
...
-    commands = [create_folders, chmod_binaries]
+      # get the commands to kill, keep and start
+      commands_to_kill, commands_to_keep, commands_to_start = \
+          self.get_command_changes(current_commands, updated_commands)
+
+      Log.info("current commands: %s" % sorted(current_commands.keys()))
+      Log.info("new commands    : %s" % sorted(updated_commands.keys()))
+      Log.info("commands_to_kill: %s" % sorted(commands_to_kill.keys()))
+      Log.info("commands_to_keep: %s" % sorted(commands_to_keep.keys()))
+      Log.info("commands_to_start: %s" % sorted(commands_to_start.keys()))
```
objmagic: should these be `Log.debug`?

objmagic: oh I forgot `heron-executor` logs to a file. Then I think it's ok to do `Log.info` here.

billonahill: Yeah, I think this are ok as info. They happen rarely and we typically want to know what's going on with this process restarting part.

kramasamy:
furthermore, these statements are not seen by the user since heron executor is automatically started by the scheduler

PR #1304, opened 2016-08-25
Adding annotations to specify interface usage

### LINKED ISSUE #953: Annotation scheme for interfaces that are evolving, private, etc.

billonahill:
Fixes #953.

Let me know what folks think of this approach and the classifications. Borrowed heavily from how Hadoop does things.

### DIFF
--- a/heron/api/src/BUILD
+++ b/heron/api/src/BUILD
@@ -1,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "processor-service-conf",
+    srcs = ["META-INF/services/javax.annotation.processing.Processor"]
+)
--- a/heron/api/src/META-INF/services/javax.annotation.processing.Processor
+++ b/heron/api/src/META-INF/services/javax.annotation.processing.Processor
@@ -1,0 +1,1 @@
+com.twitter.heron.classification.HeronAnnotationProcessor
--- a/heron/api/src/java/BUILD
+++ b/heron/api/src/java/BUILD
@@ -12,13 +12,13 @@ api_deps_files =  \
 
 java_library(
     name = "api-java",
-    srcs = glob(["**/*.java"]),
+    srcs = glob(["com/twitter/heron/api/**/*.java"]),
     deps = api_deps_files,
 )
 
 java_binary(
     name = "api-unshaded",
-    srcs = glob(["**/*.java"]),
+    srcs = glob(["com/twitter/heron/api/**/*.java"]),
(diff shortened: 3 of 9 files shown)


kramasamy:
@billonahill - this is great. I was wondering if there is a way to discourage people using an API - like warnings etc. I see sometimes people are using directly the Heron API - which should be completely discouraged. If we can find a way to enforce, it will be great!

billonahill:
@kramasamy yes we should be able to do things like check at compile time if a class not in `com.twitter.heron` is implementing a Private or Unstable interface and emit a warning.

I don't think we detect at runtime (or even compile time) if a class is calling a class that they shouldn't.

billonahill:
Update: @kramasamy it seems annotations can't be inherited which means that if a class _extends a superclass_ that has a given annotation we can detect that at compile time and warn, but if a class _implements an interface_ that has a given annotation we can not detect that.

kramasamy:
@billonahill - ok. I think it is better than the current state which is nothing.

kramasamy:
👍

billonahill:
@kramasamy I just added an annotations processor which will emit warnings at compile time if a non-heron class extends a class marked as Private, LimitedPrivate or Unsable. 
```
$ javac -cp "bazel-bin/heron/spi/src/java/libheron-spi.jar:bazel-bin/heron/api/src/java/libclassification.jar:bazel-bin/heron/proto/libproto_topology_java.jar" billg/*.java com/twitter/heron/*.java
billg/BillPacking.java:17: warning: billg.BillPacking extends from a class annotated with com.twitter.heron.classification.InterfaceStability.Unstable
public class BillPacking extends com.twitter.heron.NullPacking {
       ^
billg/BillPacking.java:17: warning: billg.BillPacking extends from a class annotated with com.twitter.heron.classification.InterfaceAudience.LimitedPrivate
public class BillPacking extends com.twitter.heron.NullPacking {
       ^
2 warnings
```

kramasamy:
@billonahill - this is great!.

kramasamy:
@billonahill - should we merge this?

billonahill:
@kramasamy this is merged.

kramasamy:
sorry - I did not see the change in status at the top :(

PR #1307, opened 2016-08-26
Refactor state manager locations to prevent #1303

billonahill:
When adding state manager locations, we shouldn't have to remember to add code in various places like `initTree`. This change makes that so by using an Enum to hold all state manager locations. The Enum allows us to do logic on all registered locations.


billonahill:
@ashvina @maosongfu @kramasamy this change will shield us from bugs like #1303 in the future.

kramasamy:
nice cleanup. Look good to me 👍

ashvina:
Nice :+1:

PR #1313, opened 2016-08-30
Change executor to load packing plan from StateManager

billonahill:
Change to no longer use the instance distribution passed but instead watch the state manager for packing plan updates. Will remove the instance distribution entirely in a follow-up change.


objmagic:
LGTM if Python style is fixed.

PR #1317, opened 2016-08-30
add kafka spout for heron

nlu90:
Updated the storm kafka spout code to make it work with Heron. Updates are:

1. remove Kafka bolt and trident related code.

2. change Scheme related interface to accept the byte[] instead of ByteBuffer

3. change the pom.xml file for compiling

Contrib code is put in Heron repo temporarily,  it should be moved to a separate heron-contrib repo finally. Setting up such a repo takes time so currently the code is put here to unblock others.


kramasamy:
@nlu90 - can we write a build file so that we can build using bazel under the config, if needed.

kramasamy:
@nlu90 - the reason why I suggested BUILD file, we can occasionally use it for release purposes and also ensure that the code build manually.

kramasamy:
I was wondering do they have unit tests as well

kramasamy:
should we change src/jvm to src/java directory to be consistent with others?

nlu90:
@kramasamy 

1. For release purpose, you can use the pom.xml file and maven to do a binary release. And also since the code will finally be moved into a separate repo, we should keep it decoupled with other part of the heron repo. Adding a build file also requires changes in the WORKSPACE file to include the extra binary dependencies, changes in other build scripts to integrate with the release scripts. Those changes need to be reverted once the code is moved. I think it's not a good practice.

2. This PR also includes the unit tests

3. I'm fine with either. I can change it to src/java

contrib/kafka-spout/README.md
```diff
@@ -0,0 +1,375 @@
+Storm Kafka
+====================
+
+Provides core Storm and Trident spout implementations for consuming data from Apache Kafka 0.8.x.
```
mycFelix: i was wondering do we need to modify README.md to mention users that current release only support `core api`,`Trident api` is not included.

kramasamy:
@nlu90 - can you make changes as @mycFelix mentioned?

nlu90:
@mycFelix, thanks for mentioning it, I've updated it.

@kramasamy the BUILD file for bazel is also added.

kramasamy:
Ship it!

PR #1318, opened 2016-08-30
Delete back-pressure message to avoid memory leak

maosongfu:


kramasamy:
👍

objmagic:
Did this actually cause memory leak before?

maosongfu:
Yes it did cause memory leak; but this message does not occur a lot and every time the size leaked is small. So the leak would build up slowly.

PR #1319, opened 2016-08-30
Global metrics for PyHeron

objmagic:
This PR adds Global Metrics module for Python topology, which is equivalent to [GlobalMetrics.java](https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/metric/GlobalMetrics.java). See test case in `global_metrics_unittest.py` below.

Related to #1264 

cc @taishi8117


taishi8117:
Looks good to me!

Issue #1323, opened 2016-08-31
Refactor packing plan object model classes for immutability and to use Sets


billonahill:
There are two improvements we should make to the packing plan model objects (i.e. `ContainerPlan`, `PackingPlan`, `InstancePlan`):

1. [Favor immutability](http://www.ibm.com/developerworks/java/library/j-jtp02183/index.html#N10064). These objects have public members that can be modified after object creation, which is an anti-pattern. Instead we should use getter methods.

2. The objects require `Map`s where `Set`s could be used, which would simplify working with these classes. The objects follow the following unnecessary pattern:

Given a class Foo, with an id and an equals/hashcode method:
```
class Foo {
  String fooId;
  boolean equals(Foo other) {..}
  String hashcode() {..}
}
```
To create a `Bar` object with a set of `Foo`s, instead of passing in a set of `Foo`s:
```
public Bar(Set<Foo> foos) {..}
```
first, you have to make a `Map<String, Foo>` where the string key is `Foo.fooId`. Requiring a `Map` instead of a `Set` would make sense if we were mapping `Foo`s to some other object, but in this case we're mapping them to themselves.

PR #1324, opened 2016-08-31
Change PackingPlan model objects to be immutable

billonahill:
Addresses issues 1 in #1323: favor immutability.

Changes model objects in `heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java` to be immutable.


objmagic:
👍 if CI passes

PR #1325, opened 2016-08-31
Change PackingPlan models to use Sets instead of Maps

billonahill:
Using Maps in this API adds unnecessary friction without benefit. This clean-up is the second half of #1323.

See `heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java` for the key changes. The rest of the classes are just updated to reflect the `PackingPlan` API change.

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
@@ -16,16 +16,24 @@ package com.twitter.heron.spi.packing;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
+
+import com.google.common.base.Optional;
 
 public class PackingPlan {
   private final String id;
-  private final Map<String, ContainerPlan> containers;
+  private final Map<String, ContainerPlan> containersMap;
+  private final Set<ContainerPlan> containers;
   private final Resource resource;
 
-  public PackingPlan(String id, Map<String, ContainerPlan> containers, Resource resource) {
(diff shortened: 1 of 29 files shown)


heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
```diff
@@ -16,16 +16,24 @@
...
+  private final Map<String, ContainerPlan> containersMap;
+  private final Set<ContainerPlan> containers;
   private final Resource resource;
 
-  public PackingPlan(String id, Map<String, ContainerPlan> containers, Resource resource) {
+  public PackingPlan(String id, Set<ContainerPlan> containers, Resource resource) {
     this.id = id;
     this.containers = containers;
     this.resource = resource;
+    containersMap = new HashMap<>();
```
ashvina: As mentioned in #1324, this map could be a immutable map?

billonahill: That map is initialized in the constructor and never exposed outside the class so making it immutable isn't really necessary. If there was a getter to return it, I would agree.

ashvina: Agree. Perhaps I was looking at this method `getContainers()` which returns the set of containers. The `Set` returned by this method should be immutable?

billonahill: Yeah, I was thinking about that. Updated to copy into an ImmutableSet during initialization.

ashvina:
LGTM 👍

maosongfu:
LGTM :+1:

PR #1326, opened 2016-09-02
Change PackingPlan Resource class to be immutable

### LINKED ISSUE #1323: Refactor packing plan object model classes for immutability and to use Sets

billonahill:
See `heron/spi/src/java/com/twitter/heron/spi/packing/Resource.java` for changes. All other classes are modified to reflect the change. The two scheduler changes were to remove a mutation that was having no effect.

Fixed #1323.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
@@ -348,10 +348,11 @@ public class RoundRobinPacking implements IPacking {
     for (PackingPlan.ContainerPlan containerPlan : plan.getContainers()) {
       for (PackingPlan.InstancePlan instancePlan : containerPlan.getInstances()) {
         // Safe check
-        if (instancePlan.getResource().ram < MIN_RAM_PER_INSTANCE) {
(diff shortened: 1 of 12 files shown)


kramasamy:
Ship it 👍

PR #1332, opened 2016-09-02
Remove PackingPlan.resource, fixes #1330

### LINKED ISSUE #1330: Remove PackingPlan.resources
Instances and Containers have resources which are well defined concepts used to start instances and containers, respectively. In contrast the PackingPlan resource is not well defined, it's calculated (at times incorrectly or inconsistently) and set but never used and it's really not useful. This issues is to remove it.

billonahill:
Changes PackingPlan and packing_plan.proto to remove resources. Also changes all the other code that set this, as well as the tests that read it.

fixes #1330


objmagic:
👍

PR #1333, opened 2016-09-03
Adding LocalScheduler scaling impl with UpdateTopologyManager

billonahill:
Now that `IScalable` is committed, committing the `LocalScheduler` implementation, along with `UpdateTopologyManager`, which is to be shared among all scheduler implementations. The scheduler changes are minimal with most of the logic contained in `UpdateTopologyManager`.

Required for #1292.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -137,7 +137,7 @@ public class UpdateTopologyManager {
    * added and those to be removed. Whether to add or remove a container is determined by the id of
    * the container. Proposed containers with an id not in the existing set are to be added, while
    * current container ids not in the proposed set are to be removed.
-   * </P>
+   *
    * It is important to note that the container comparison is done by id only, and does not include
    * the InstancePlans in the container, which for a given container might change in the proposed
    * plan.
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -81,9 +81,6 @@ public class UpdateTopologyManager {
 
     SchedulerStateManagerAdaptor stateManager = Runtime.schedulerStateManagerAdaptor(runtime);
 
-    // assert found PackingPlan is same as existing PackingPlan.
-    validateCurrentPackingPlan(existingProtoPackingPlan, topologyName, stateManager);
-
     // fetch the topology, which will need to be updated
     TopologyAPI.Topology updatedTopology =
         getUpdatedTopology(topologyName, proposedPackingPlan, stateManager);
@@ -119,17 +116,6 @@ public class UpdateTopologyManager {
     TopologyAPI.Topology updatedTopology = stateManager.getTopology(topologyName);
     Map<String, Integer> proposedComponentCounts = proposedPackingPlan.getComponentCounts();
     return mergeTopology(updatedTopology, proposedComponentCounts);
-  }
-
-  @VisibleForTesting
-  void validateCurrentPackingPlan(PackingPlans.PackingPlan existingProtoPackingPlan,
-                                  String topologyName,
-                                  SchedulerStateManagerAdaptor stateManager) {
-    PackingPlans.PackingPlan foundPackingPlan = stateManager.getPackingPlan(topologyName);
-    assertTrue(foundPackingPlan.equals(existingProtoPackingPlan),
-        "Existing packing plan received does not equal the packing plan found in the state "
-            + "manager. Not updating updatedTopology. Received: %s, Found: %s",
-        existingProtoPackingPlan, foundPackingPlan);
   }
 
   /**
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -1,0 +1,249 @@
+// Copyright 2016 Twitter. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.scheduler;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
+import com.google.protobuf.Descriptors;
+
+import com.twitter.heron.api.generated.TopologyAPI;
+import com.twitter.heron.proto.system.PackingPlans;
+import com.twitter.heron.spi.common.Config;
+import com.twitter.heron.spi.packing.PackingPlan;
+import com.twitter.heron.spi.packing.PackingPlanProtoDeserializer;
+import com.twitter.heron.spi.scheduler.IScalable;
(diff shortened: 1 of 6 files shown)


kramasamy:
Looks good to me! @ashvina and @avflor - can you review this please?

ashvina:
@kramasamy I can review it partially. I have authored a part of this commit (https://github.com/billonahill/heron/pull/2) which @billonahill had reviewed earlier.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -0,0 +1,249 @@
...
+    logFine("Deleted existing Topology: %s", stateManager.deleteTopology(topologyName));
+    logFine("Set new Topology: %s", stateManager.setTopology(updatedTopology, topologyName));
+
+    // update packing plan to trigger the scaling event
+    logFine("Deleted existing packing plan: %s", stateManager.deletePackingPlan(topologyName));
+    logFine("Set new PackingPlan: %s",
+        stateManager.setPackingPlan(proposedProtoPackingPlan, topologyName));
+
+    // delete the physical plan so TMaster doesn't try to re-establish it on start-up.
+    logFine("Deleted Physical Plan: %s", stateManager.deletePhysicalPlan(topologyName));
```
ashvina: In the past we discussed about issues when multiple components were updating a `PackingPlan` (`Launcher` while submitting a topology and `Scheduler` while updating the topology). In this case, `PhysicalPlan` is also getting updated my multiple components; `RuntimeManager` during kill, TMaster during submit and UpdateManager during update. Is it a concern?

billonahill: Typically the launcher (or the `TMaster` in the case of physical plan) initially writes object and `RuntimeManager` deletes them on kill. That pattern is somewhat standard.

Where we're deviating is with the `Scheduler` updating `PackingPlan` and `Topology` and deleting `PhysicalPlan` (`TMaster` will re-create it). I think this is ok for this iteration, but we might want to consider messaging the `TMaster` to do these operations as a future enhancement.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -0,0 +1,249 @@
...
+    Map<String, Integer> proposedComponentCounts = proposedPackingPlan.getComponentCounts();
+    return mergeTopology(updatedTopology, proposedComponentCounts);
+  }
+
+  @VisibleForTesting
+  void validateCurrentPackingPlan(PackingPlans.PackingPlan existingProtoPackingPlan,
+                                  String topologyName,
+                                  SchedulerStateManagerAdaptor stateManager) {
+    PackingPlans.PackingPlan foundPackingPlan = stateManager.getPackingPlan(topologyName);
+    assertTrue(foundPackingPlan.equals(existingProtoPackingPlan),
```
ashvina: Just curious, in what cases will the two packing plans differ? And if that happens what is the user expected to do?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -0,0 +1,249 @@
...
+  @VisibleForTesting
+  static TopologyAPI.Topology mergeTopology(TopologyAPI.Topology topology,
+                                            Map<String, Integer> proposedComponentCounts) {
+    TopologyAPI.Topology.Builder builder = TopologyAPI.Topology.newBuilder().mergeFrom(topology);
+    for (String componentName : proposedComponentCounts.keySet()) {
+      Integer parallelism = proposedComponentCounts.get(componentName);
+
+      boolean updated = false;
+      for (TopologyAPI.Bolt.Builder boltBuilder : builder.getBoltsBuilderList()) {
+        if (updateComponent(boltBuilder.getCompBuilder(), componentName, parallelism)) {
```
ashvina: `updateComponent` can return false. Should merge operation fail if that happens? `false` case should not happen so I think the case could be treated as an exception?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: They will differ if they somehow get our of sync, like during a race when multiple agents are trying to update packing at the same time. This is a poor version of checking though. I'd like to keep this in place for now, but follow up with real transnational support.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: updateComponent will return false under normal circumstances (see the updateComponent javadocs). Because in the protobuf world Bolt and Spout are both Components, but without inheritance, we have to try to update each to see if we have a match. We use true/false so if we get a match on bolt we don't try again on spout.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -0,0 +1,249 @@
...
+      }
+      this.containersToAdd = Collections.unmodifiableSet(toAdd);
+
+      Set<PackingPlan.ContainerPlan> toRemove = new HashSet<>();
+      for (PackingPlan.ContainerPlan curentContainerPlan : currentContainers) {
+        if (!proposedContainers.contains(curentContainerPlan)) {
+          toRemove.add(curentContainerPlan);
+        }
+      }
+      this.containersToRemove = Collections.unmodifiableSet(toRemove);
```
avflor: I'm a little confused here. Do we remove a container if its instances do not exactly match the previous ones? What happens if the current container contains instances 1,2,3 and the proposed container contains 1,2,4? Do we remove this container and re-allocate it?

ashvina: I agree with @avflor. 
My impression is that scheduler will allocate new containers for executorIds not present in the current plan and remove container's for executorIds not present in new packing plan. Moreover this logic can result in adding an existing container and then removing it in the `updateTopology` method. In `ContainerDelta()` method, comparison should be based on container ID only and not based on container instances.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
ashvina: I see the point. This looks fine now. Just wondering if there is a way to simplify the logic here? For e.g. by deserializing the topology protobuf object and then updating it?

billonahill: Updating the protobug Topology object is indeed clunky, due to the protobuf api, but I thing it's safer and easier (and more efficient) than going through a deserialization/serialization process, and the risk of ending up with a different Topology object than we started with. It's fairly large with nested config hanging from all components.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: Great catch. Updating now to use container ids.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
ashvina: By multiple agents, do you mean concurrent update requests? 
In my opinion this check should be removed. Jut in case scheduler plan is not in sync with persisted plan, scheduler would be able to persist the new desired plan. This check enforces topology restart.

billonahill: Yes. My thought was that if they're not in sync, something is happening that we're unable to reason so unexpected results could occur if we continue. I could be convinced to remove it though, just sharing my thinking when implementing.

billonahill: I've removed the check and filed #1353.

billonahill:
@ashvina and @avflor all comments have been addressed. Let me know how things look.

avflor:
@billonahill Looks good to me. Maybe we should add a comment at ContainerDelta to clarify that the code assumes that the size of an existing container cannot change from currentPlan to proposedPlan. 
If the scheduler provides this functionality we would need to change this code since it might return wrong results.

PR #1334, opened 2016-09-03
RuntimeManager changes required for scaling

billonahill:
- Changes `RuntimeManagerMain` to receive scaling requests.
- Changes `RuntimeManagerRunner` to handle scaling requests. Basically, receive the params, invoke the repacking algorithm and pass the new `PackingPlan` to the scheduler.

This does not make the feature available yet, since the new params need to also be added to the heron-cli.

Required for #1292.


heron/config/src/yaml/conf/local/packing.yaml
```diff
@@ -1,2 +1,3 @@
 # packing algorithm for packing instances into containers
 heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.RoundRobinPacking
+heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking
```
kramasamy: Why a new configuration variable? Will there be any issues when mixing packing algorithms?

billonahill: There are separate interfaces for each (`IPacking` and `IRepacking`) so implementors can implement one or the other. An `IRepacking` impl should not assume anything about what `IPacking` impl was used at submit.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -159,23 +207,25 @@ protected boolean cleanState(
...
+    // set. Just log but don't consider it a failure and don't return false
     result = statemgr.deleteTMasterLocation(topologyName);
     if (result == null || !result) {
-      // We would not return false since it is possible that TMaster didn't write physical plan
       LOG.warning("Failed to clear TMaster location. Check whether TMaster set it correctly.");
     }
 
+    result = statemgr.deletePackingPlan(topologyName);
+    if (result == null || !result) {
+      LOG.warning("Failed to clear packing plan. Check whether TMaster set it correctly.");
```
ashvina: I think the packing plan is set by `Launcher` or `UpdateManager`. If so the log message needs to be fixed.

Coincidentally, I saw this issue today (#1347) and created a PR #1348. I am ok if it is marked a duplicate of this and closed.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -196,4 +246,71 @@ protected boolean cleanState(
...
+      }
+      Integer newValue = changeRequests.get(component);
+      Integer delta = newValue - componentCounts.get(component);
+      if (delta == 0) {
+        changeRequests.remove(component);
+      } else {
+        changeRequests.put(component, delta);
+      }
+    }
+    return changeRequests;
```
ashvina: The input request Map is modified by this method. I think it would be better to create a new structure with the delta to avoid any confusion.

avflor: I think we need to add a check that delta should not be negative. For now we do not support scaling down. We can remove it once I implement this case in repack.

billonahill: Both good catches. Fixed.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
billonahill: Fixed. Let's go with this change and cancel #1348 to keep the relevant updates together.

kramasamy:
👍

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
ashvina: The log message is still incorrect?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -159,23 +207,25 @@ protected boolean cleanState(
...
+    result = statemgr.deletePackingPlan(topologyName);
+    if (result == null || !result) {
+      LOG.warning("Failed to clear packing plan. Check whether TMaster set it correctly.");
+    }
+
     result = statemgr.deletePhysicalPlan(topologyName);
     if (result == null || !result) {
-      // We would not return false since it is possible that TMaster didn't write physical plan
-      LOG.warning("Failed to clear physical plan. Check whether TMaster set it correctly.");
+      LOG.warning("Failed to clear physical plan. Check whether Launcher set it correctly.");
```
ashvina: The log message is incorrect?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
billonahill: my bad, fixed.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
billonahill: fixed

billonahill:
@ashvina and @avflor all comments have been addressed. Let me know if you'd like more time to review.

PR #1329, opened 2016-09-02
DO NOT MERGE Convert container id type to int from String

ashvina:
Fixes #1295 
Container Id is declared as of type String. The value of id is always an int.
The users always assume id to be an int and the code is sprinkled with
int-to-string-to-int conversion. This update fixes it.

PR #1337, opened 2016-09-04
Revert "DO NOT MERGE Convert container id type to int from String"

objmagic:
Reverts twitter/heron#1329


ashvina:
Hi @objmagic. I think there is no need to revert this patch. I tested #1329 with the local and yarn schedulers. While it works on local, it has some issues on the YARN scheduler. I will have a fix ready for YARN today. So the build will be stable then. Thoughts?

kramasamy:
@ashvina - that is fine.

objmagic:
I think this is fine. Sorry again for the hassle.

PR #1341, opened 2016-09-05
modify `heron-api` and `heron-storm` artifactId in kafka-spout pom.xml

mycFelix:
There seems a little mistake in kafka-spout `pom.xml`.

We used `local-heron-api` and `local-heron-storm` as test artifactId.

Sorry for carelessness ......


mycFelix:
@nlu90 - Should we remove the `local` prefix in pom.xml?

nlu90:
@mycFelix yeah 👍

Issue #1295, opened 2016-08-23
Should container's id be of type Integer?


ashvina:
Currently a container's `id` is of `String` type in `PackingPlan` https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java#L201. However it is assumed to be of type Integer in most places in the code. For e.g.

1. https://github.com/twitter/heron/blob/master/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java#L363
1. https://github.com/twitter/heron/blob/master/heron/schedulers/src/java/com/twitter/heron/scheduler/local/LocalScheduler.java#L43
1. https://github.com/twitter/heron/blob/master/heron/proto/scheduler.proto#L29

As a result there is a lot of Integer-String conversion going around in the code base. IMO this clutter and unseen errors could be minimized if we accept container's Id to be an Integer. Thoughts?

objmagic:
integer being type of container's id looks reasonable to me. any reason why `String` type shows up occasionally? maybe this happened during some refactor before?

billonahill:
+1 this has bothered me as well.

Semi-related grievance: Another area for potential cleanup is in the instance plan id, which is colon delimited string containing things like `global_task_id` and `component_index`. It gets parsed out on the executor to create the instance id. It's fine to use these indexes as part of a descriptive string id, but they should never be expected and parsed out. They should instead become first class entities of the component plan.

kramasamy:
I will be great if we could do this cleanup before it is too late - IMO.

PR #1342, opened 2016-09-05
Use executor id as int type in the YARN scheduler

ashvina:
These fixes the YARN scheduler and is related to container id type conversion, see #1295 and #1329.


objmagic:
👍

billonahill:
👍

PR #1363, opened 2016-09-08
Update heron cli to support update command for scaling

billonahill:
Last change required to make scaling available by adding the `update` command to the heron client.

Required for #1292.

### DIFF
--- a/heron/tools/cli/src/python/update.py
+++ b/heron/tools/cli/src/python/update.py
@@ -21,6 +21,7 @@ import heron.tools.common.src.python.utils.config as config
 import argparse
 import logging
 import re
+import traceback
 
 def create_parser(subparsers):
   """ Create the parse for the update command """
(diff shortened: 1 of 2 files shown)


heron/tools/cli/src/python/update.py
```diff
@@ -0,0 +1,97 @@
...
+    # invoke the runtime manager to kill the topology
+    execute.heron_class(
+        'com.twitter.heron.scheduler.RuntimeManagerMain',
+        lib_jars,
+        extra_jars=[],
+        args=new_args
+    )
+
+  except Exception as ex:
+    print 'Error: %s' % str(ex)
```
objmagic: maybe use `Log.debug(traceback.format_exc(ex))`?

objmagic:
👍

heron/tools/cli/src/python/update.py
billonahill: good catch, fixed.

kramasamy:
👍

PR #1360, opened 2016-09-08
AuroraSchedler implements IScalable

billonahill:
By implementing `IScalable`, we can scale jobs in aurora.

Required for #1292.

PR #1364, opened 2016-09-08
Repacking Implementation for the First Fit Decreasing algorithm

avflor:
This is the first repacking implementation that supports scale up only using the FFD algorithm.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
@@ -128,7 +128,7 @@ public class FirstFitDecreasingPacking implements IPacking, IRepacking {
 
     this.maxContainerResources = new Resource(
         TopologyUtils.getConfigWithDefault(topologyConfig, TOPOLOGY_CONTAINER_MAX_CPU_HINT,
-                this.defaultInstanceResources.getCpu() * DEFAULT_NUMBER_INSTANCES_PER_CONTAINER),
+            this.defaultInstanceResources.getCpu() * DEFAULT_NUMBER_INSTANCES_PER_CONTAINER),
         TopologyUtils.getConfigWithDefault(topologyConfig, TOPOLOGY_CONTAINER_MAX_RAM_HINT,
             this.defaultInstanceResources.getRam() * DEFAULT_NUMBER_INSTANCES_PER_CONTAINER),
         TopologyUtils.getConfigWithDefault(topologyConfig, TOPOLOGY_CONTAINER_MAX_DISK_HINT,
@@ -156,7 +156,9 @@ public class FirstFitDecreasingPacking implements IPacking, IRepacking {
   }
 
   /**
-   * Get a new packing plan given an existing packing plan and component-level changes
+   * Get a new packing plan given an existing packing plan and component-level changes.
+   * The current implementation assign the new instances into newly created containers and
+   * does not affect the contents of existing containers.
    *
    * @return new packing plan
    */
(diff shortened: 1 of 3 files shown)


heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -204,6 +238,17 @@ public void close() {
...
 
   /**
+   * Get the instances' allocation based on the First Fit Decreasing algorithm
+   *
+   * @return Map &lt; containerId, list of InstanceId belonging to this container &gt;
+   */
+  private Map<Integer, List<String>> getFFDAllocation(PackingPlan currentPackingPlan,
+                                                      Map<String, Integer> componentChanges) {
+    int numContainers = currentPackingPlan.getContainers().size();
+    return placeInstances(componentChanges, numContainers);
```
billonahill: If I understand correctly, `placeInstances` adds new containers beyond `numContainers`. Is that correct? If so can we make the signature and javadocs for `placeInstances` reflect that, because currently the method description and `numContainers` field imply that the method places instances within `numContainers` not beyond.

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -151,6 +155,36 @@ public PackingPlan pack() {
...
+    Map<Integer, List<String>> ffdAllocation =
+        getFFDAllocation(currentPackingPlan, componentChanges);
+
+    // Construct the PackingPlan
+    Map<String, Long> ramMap = TopologyUtils.getComponentRamMapConfig(topology);
+
+    Set<PackingPlan.ContainerPlan> containerPlans = PackingUtils.buildContainerPlans(
+        ffdAllocation, ramMap, defaultInstanceResources, paddingPercentage);
+
+    //merge the two plans
```
billonahill: We should add a comment to clarify that this merge approach only works because the `ffdAllocation` generated above is done by appending new instances into new containers so there is no container intersection in the merge.

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -151,6 +155,36 @@ public PackingPlan pack() {
...
+    Set<PackingPlan.ContainerPlan> totalContainerPlans = new HashSet<>();
+    totalContainerPlans.addAll(containerPlans);
+    totalContainerPlans.addAll(currentPackingPlan.getContainers());
+
+    LOG.info(String.format("Created a packing plan with %d containers",  containerPlans.size()));
+    for (PackingPlan.ContainerPlan containerPlan : containerPlans) {
+      LOG.info(String.format(
+          "Container %s consists of %s", containerPlan.getId(), containerPlan.getInstances()));
```
billonahill: For a large topology this could be a lot of output on the user console. We could keep it for now, but could you sort the container plans by container id? In #1360 I've added a comparator for unit tests, which we might want to pull into `src`. (Or we implement Comparable for ContainerPlan and all of it's child members)

heron/packing/tests/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPackingTest.java
```diff
@@ -306,4 +323,68 @@ public void testInvalidRamInstance() throws Exception {
...
+        BOLT_NAME, SPOUT_NAME, boltRam, instanceDefaultResources.getRam(), null);
+
+    int numScalingInstances = 3;
+    Map<String, Integer> componentChanges = new HashMap<>();
+    componentChanges.put(BOLT_NAME, numScalingInstances);
+    PackingPlan newPackingPlan = getFirstFitDecreasingPackingPlanRepack(topologyExplicitRamMap,
+        packingPlanExplicitRamMap,
+        componentChanges);
+
+    Assert.assertEquals(4, newPackingPlan.getContainers().size());
```
billonahill: should we also be asserting an expected container size (ram), here and above?

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
avflor: Thanks for catching this. I agree it shouldn't be here. I used this code only for debugging purposes and forgot to delete it afterwards.

avflor:
@billonahill I have addressed all the comments. Let me know if anything else is needed.

billonahill:
👍

PR #1366, opened 2016-09-09
refactor local test runner

billonahill:
Refactors local_test_runner integration test to prepare for adding a scaling test

- Change to object-oriented from proceedural
- Use the template pattern to allow for test subclassing (see `TestTemplate.run_test()`)
- Change method and variable names to be snake_case instead of camelCase (idiomatic python)

Required for #1292.

### DIFF
--- a/integration-test/src/python/local_test_runner/test_template.py
+++ b/integration-test/src/python/local_test_runner/test_template.py
@@ -1,0 +1,300 @@
+# Copyright 2016 Twitter. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


kramasamy:
👍 - nice cleanup!

integration-test/src/python/local_test_runner/test_template.py
```diff
@@ -0,0 +1,288 @@
...
+      logging.info("Expected result ---------- \n" + expected_result)
+      return True
+    else:
+      logging.error("Actual result did not match expected result for test %s", self.testname)
+      logging.info("Actual result ---------- \n" + actual_result)
+      logging.info("Expected result ---------- \n" + expected_result)
+      return False
+
+  # pylint: disable=no-self-use
+  def get_pid(self, processName, heronWorkingDirectory):
```
objmagic: still camel case

billonahill: good catch, fixed.

Issue #1372, opened 2016-09-12
Failed to submit topo using extra_launch_classpath arg


mycFelix:
hi~ 

It seems something wrong submitting topo using extra_launch_classpath arg if directory ends with '*' 

```
Felix@FelixWork heron-orig$ heron submit yarn ~/.heron/examples/heron-examples.jar com.twitter.heron.examples.ExclamationTopology ExclamationTopology --deploy-deactivated --extra-launch-classpath /usr/local/Cellar/hadoop/2.7.1/libexec/share/hadoop/yarn/*:/usr/local/Cellar/hadoop/2.7.1/libexec/share/hadoop/common/lib/* --verbose
INFO: Using config file under /Users/Felix/.heron/conf/yarn
DEBUG: {'config_property': [], 'topology-file-name': '/Users/Felix/.heron/examples/heron-examples.jar', 'override_config_file': '/var/folders/wf/t23yrky14rl6lktg6bstd2b00000gn/T/tmppzRJYx/override.yaml', 'config_path': '/Users/Felix/.heron/conf/yarn', 'subcommand': 'submit', 'deploy_deactivated': 'True', 'cluster': 'yarn', 'topology_main_jvm_property': [], 'extra_launch_classpath': '/usr/local/Cellar/hadoop/2.7.1/libexec/share/hadoop/yarn/*:/usr/local/Cellar/hadoop/2.7.1/libexec/share/hadoop/common/lib/*', 'role': 'Felix', 'environ': 'default', 'topology-class-name': 'com.twitter.heron.examples.ExclamationTopology', 'verbose': 'True'}
DEBUG: Checking classpath entry suffix as directory: *
ERROR: One of jar or directory in extra launch classpath does not exist
INFO: Elapsed time: 0.000s.
Felix@FelixWork heron-orig$
```

mycFelix:
closed #1373

PR #1373, opened 2016-09-12
Failed to submit topo using extra_launch_classpath arg  #1372

mycFelix:
It seems something wrong submitting topo using `extra_launch_classpath` arg if directory ends with '*' .

this PR try to fix it


kramasamy:
ship it. 👍

Issue #1376, opened 2016-09-12
InstanceId string format should never be parsed or assumed


billonahill:
A Heron Instance id (i.e., `instanceId`) is [generated](https://github.com/twitter/heron/blob/master/heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java#L145) using the following format:
```
String.format("%d:%s:%d:%d", containerIdx, componentName, instanceIdx, componentIdx)
```
The id provides uniqueness and contains helpful identifying information, which is fine, but it's contents and format are assumed by parts of the system, which is bad. For example:

* If the `instanceIdx` part of the id is not globally unique, tuples can be mis-routed when using shuffle grouping. This is because the `heron-executor` [parses](https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py#L379) `taskId` from `instanceId` and passes that to the started instance. This value becomes the globally unique `taskId` on the `PhysicalPlan` which is used extensively. 
* The packing package passes maps of `instanceIds` around and [parses out](https://github.com/twitter/heron/blob/master/heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java#L150) the `componentName`. Instead it should work with `InstancePlan` objects and call `instnacePlan.getComponentName()`.

There should be no hidden meaning or expectations of that id beyond it being human readable. That id format should be safe to change without breaking underlying systems. If a concept like `instnaceIdx` is required by the system, it should be added to `InstancePlan` as a typed concept that can be accessed without parsing tokens from the id.

PR #1377, opened 2016-09-12
Repacking to use globally unique task ids

billonahill:
Repacking was reusing global task ids (i.e., restarting from 1) for new instnaces added. This caused strange tuple routing behavior.

Opened #1376 to not rely on `instanceIndex` parsed from `instanceId`.


billonahill:
@avflor let me know what you think.

avflor:
@billonahill I reviewed the code a few hours ago. It looks good. Thanks for catching the bug with the globaltaskIndex!

PR #1381, opened 2016-09-13
Speed up local test runner and enhancements for scaling

billonahill:
Improved the local int tests:

- No longer just sleeping and trying, now polling for topology being up before continuing. This speeds up the tests.
- When checking for stream manager running, take into account possibility of multiple stream managers
- When checking for pplan, make sure expected instance count exists
- Fixed bug during scaling by restarting stream manager during scaling
- The test template framework now gets the pplan and injects it into the test case

Required for #1292.


kramasamy:
👍

PR #1382, opened 2016-09-13
Replacing colon-delimited Instance.id with typed InstanceId object

### LINKED ISSUE #1376: InstanceId string format should never be parsed or assumed

billonahill:
See #1376 for a more detailed description, but this removes the colon-delimited `InstancePlan.id` and replaces that with an `InstanceId` class. We no longer parse the colon-delimited id to pull things out of it. Key changes are seen in the following classes:

- `PackingPlan.InstancePlan`
- `InstanceId`
- `heron_executor.py`
- proto `PackingPlans.InstancePlan`
- `PackingUtils`
Fixes #1376.

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/InstanceId.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/InstanceId.java
@@ -1,0 +1,66 @@
+// Copyright 2016 Twitter. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.spi.packing;
+
+/**
+ * Encapsulates the unique components of an Instance identifier.
+ */
+public final class InstanceId {
+
+  private final String componentName;
+  private final int taskId;
+  private final int componentIndex;
+
+  public InstanceId(String componentName, int taskId, int componentIndex) {
+    this.componentName = componentName;
+    this.taskId = taskId;
+    this.componentIndex = componentIndex;
+  }
+
+  public String getComponentName() {
+    return componentName;
+  }
+
+  int getTaskId() {
+    return taskId;
+  }
+
+  int getComponentIndex() {
+    return componentIndex;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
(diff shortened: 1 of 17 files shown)


heron/spi/src/java/com/twitter/heron/spi/packing/InstanceId.java
```diff
@@ -0,0 +1,66 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.spi.packing;
+
+/**
+ * Encapsulates the unique components of an Instance identifier.
+ */
+public final class InstanceId {
```
ashvina: the instance Id used to container container index also `String.format("%d:%s:%d:%d", containerIdx, componentName, instanceIdx, componentIdx)`. Is it not needed any more?

ashvina:
Very useful change, improves quality of the code 👍

heron/spi/src/java/com/twitter/heron/spi/packing/InstanceId.java
billonahill: No, it's actually no longer needed. It was only used to be parsed by the packing classes and the executor. Now that it's contents are part of InstanceId, it's no longer needed. There is still a convenient task identifier generated and used by the executor so simplify things like log and pid files and to avoid naming collisions, but that's it. From [the executor](https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py#L382):
```
instance_id = "container_%s_%s_%s" % (str(self.shard), component_name, str(global_task_id))
```

kramasamy:
Nice! 👍

PR #1392, opened 2016-09-15
Improve Python program metrics collector and add test

objmagic:
* use `MultiAssignableMetrics` to record threads' metric
* bug fix on GC metrics collector naming
* add test using Mock


kramasamy:
👍

Issue #1362, opened 2016-09-08
Support parallelism scale-down in FirstFitDecreasingPacking


billonahill:
`FirstFitDecreasingPacking` is the only current impl of `IRepacking`. It supports scaling up but not down.

Related to #1292.

Issue #1311, opened 2016-08-29
Favor exceptions and Optionals over boolean and nulls


billonahill:
There's an anti-pattern in the code of returning nulls when an exceptional case occurs. This causes any context from the source of the exceptional case to be lost and it results in lots of null checks and logging at all higher stacks of the code. **Ideally methods never return null.**

### Favor Exceptions over nulls ###
If a method should always return a value, except during an exception, than mark it as `@NotNull` and throw an exception if an exceptional case occurs. The thrown exception could be a custom exception, defined for the situation. This provides the following benefits:

- Lowered risk of NPEs
- More concise code written to the happy path, without null checks.
- Cleaner logging, since everyone who handles the null doesn't need to blab about it.
- More context about the root of the issue (e.g., rich exception messages, stack traces, etc)
- A more informed caller, since the calling code can handle different exception types or exception response codes differently.

### Favor Exceptions over booleans ###

If an action should always succeed under normal circumstances, favor a `void` return type and indicate failure with an exception. Using a boolean true/false to indicate success/failure has similar downsides to using null to indicate failure. NPE's are no longer an issue with this approach, but returning false will be equally void of meaningful context about the failure. Should the calling code retry? Should it give up? Custom exceptions or response codes can help the caller decide.

### Favor Optional over null ###

If a method may or may not return a value as part of normal behavior, favor returning an [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) of the given return type. [This article](http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html) explains why.

Caveat: Optional was introduced in Java 8, so it should not be used in code that would be invoked within a topology to keep use Java 7 compatible.

ashvina:
Is a "Best Practices" doc page a expected deliverable for this issue?

kramasamy:
@ashvina - that is a start. But we need to continue to be mindful of it during the code review. Then we need to clean up our existing code.

billonahill:
Favor immutability (#1323) is another best practice.

PR #1394, opened 2016-09-15
Support for scale down

avflor:
This is related to #1362

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/Container.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/Container.java
@@ -17,6 +17,8 @@ package com.twitter.heron.packing;
 
 import java.util.ArrayList;
 
+import com.google.common.base.Optional;
+
 import com.twitter.heron.spi.packing.InstanceId;
 import com.twitter.heron.spi.packing.Resource;
 
@@ -82,36 +84,31 @@ public class Container {
    * corresponding resources.
    *
    * @return instanceId if the instance is removed the container.
-   * Return null if the container is empty or
-   * does not contain an instance of the given component.
+   * Return void if an instance is not found
    */
-  public InstanceId remove(Resource resource, String component) {
-    if (!instanceIds.isEmpty()) {
-      InstanceId instanceId = instanceOfComponent(component);
-      if (instanceId != null) {
-        usedRam -= resource.getRam();
-        usedCpuCores -= resource.getCpu();
-        usedDisk -= resource.getDisk();
-        instanceIds.remove(instanceId);
-        return instanceId;
-      } else {
-        return null;
-      }
+  public Optional<InstanceId> removeAnyInstanceOfComponent(Resource resource, String component) {
+    Optional<InstanceId> instanceId = getAnyInstanceOfComponent(component);
+    if (instanceId.isPresent()) {
+      usedRam -= resource.getRam();
+      usedCpuCores -= resource.getCpu();
+      usedDisk -= resource.getDisk();
+      instanceIds.remove(instanceId.get());
+      return Optional.of(instanceId.get());
     }
-    return null;
+    return Optional.absent();
   }
 
   /**
-   * Find whether an instance of a particular component is assigned to the container
+   * Find whether any instance of a particular component is assigned to the container
    *
-   * @return the instanceId if an instance is found, null otherwise
+   * @return the instanceId if an instance is found, void otherwise
    */
-  public InstanceId instanceOfComponent(String component) {
+  public Optional<InstanceId> getAnyInstanceOfComponent(String component) {
     for (int i = 0; i < instanceIds.size(); i++) {
       if (instanceIds.get(i).getComponentName().equals(component)) {
-        return instanceIds.get(i);
+        return Optional.of(instanceIds.get(i));
       }
     }
-    return null;
+    return Optional.absent();
   }
 }
--- a/heron/packing/src/java/com/twitter/heron/packing/Container.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/Container.java
@@ -15,11 +15,11 @@
 
 package com.twitter.heron.packing;
 
-import java.util.ArrayList;
+import java.util.HashSet;
 
 import com.google.common.base.Optional;
 
-import com.twitter.heron.spi.packing.InstanceId;
+import com.twitter.heron.spi.packing.PackingPlan;
 import com.twitter.heron.spi.packing.Resource;
 
 /**
@@ -28,12 +28,7 @@ import com.twitter.heron.spi.packing.Resource;
  */
 public class Container {
 
-  private ArrayList<InstanceId> instanceIds;
-
-  //Resources currently used by the container.
-  private long usedRam;
-  private double usedCpuCores;
-  private long usedDisk;
+  private HashSet<PackingPlan.InstancePlan> instances;
 
   //Maximum resources that can be assigned to the container.
   private long maxRam;
@@ -41,13 +36,10 @@ public class Container {
   private long maxDisk;
 
   public Container(Resource resource) {
-    this.usedRam = 0;
-    this.usedCpuCores = 0;
-    this.usedDisk = 0;
     this.maxRam = resource.getRam();
     this.maxCpuCores = resource.getCpu();
     this.maxDisk = resource.getDisk();
-    instanceIds = new ArrayList<InstanceId>();
+    instances = new HashSet<PackingPlan.InstancePlan>();
   }
 
   /**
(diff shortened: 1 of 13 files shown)


heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -231,36 +226,135 @@ public void close() {
...
+
+  /**
+   * Assigns instances to containers
+   * @param containers helper data structure that describes the containers' status
+   * @param allocation existing packing plan
+   * @param parallelismMap component parallelism
+   * @param firstTaskIndex first taskId to use for the new instances
+   */
+  private void assignInstancesToContainers(
+      ArrayList<Container> containers, Map<Integer, List<InstanceId>> allocation,
```
billonahill: I feel like we're stretching our use of Lists and Maps in the signature here, by passing a mutable list of containers and a map of containers to instance lists. Instead can we use an object that encapsulated all of that? Like a PackingPlan object or a variant of it? Some container object that is able to effectively mutate a plan?

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -313,4 +457,22 @@ private int allocateNewContainer(ArrayList<Container> containers) {
     containers.add(new Container(maxContainerResources));
     return containers.size();
   }
+
+  private class IdPair {
```
billonahill: Can we use Pair instead? https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/basics/Pair.java#L24

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -279,31 +373,81 @@ public void close() {
...
+    int containerId = 0;
+    for (int i = 0; i < containers.size() && !removed; i++) {
+      InstanceId instanceId = containers.get(i).remove(instanceResource, component);
+      if (instanceId != null) {
+        removed = true;
+        containerId = i + 1;
+        return new IdPair(containerId, instanceId);
+      }
+    }
+    return null;
```
billonahill: Please use Optional instead of null. Or make this method return void if the caller doesn't care either way. See #1311

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -231,36 +226,135 @@ public void close() {
...
+          maxInstanceIndex + 1);
+    }
+    cleanAllocation(allocation);
+    return allocation;
+  }
+
+  /**
+   * Removes containers from tha allocation that do not contain any instances
+   */
+  private void cleanAllocation(Map<Integer, List<InstanceId>> allocation) {
```
billonahill: `removeEmpyContainers(..)`?

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -231,36 +226,135 @@ public void close() {
...
+    Iterator<Map.Entry<Integer, List<InstanceId>>> entries = allocation.entrySet().iterator();
```
billonahill: Instead of iterating over entries it's more idiomatic (and reads easier IMO since it's always obvious what the key and value represents by their naming) to iterate over keys:
```
Iterator<Integer> containerIds = allocation.keySet().iterator();
while (containerIds.hasNext() {
  Integer containerId = containerIds.next();
  if (allocation.get(containerId) == null)
   allocation.remove(containerId);
...
```
Or even better using `for (Integer containerId : allocation.keySet())` if that doesn't cause issues during deletion.

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -59,14 +65,53 @@ private boolean hasSpace(long ram, double cpuCores, long disk) {
...
+    }
+    return null;
+  }
+
+  /**
+   * Find whether an instance of a particular component is assigned to the container
+   *
+   * @return the instanceId if an instance is found, null otherwise
+   */
+  public InstanceId instanceOfComponent(String component) {
```
avflor: Yes, but this method returns only one of them. I don't care about returning all of them. ANy one would work. How about public InstanceId getAnInstanceOfComponent(String component)?

avflor:
@billonahill I addressed all the comments except for the packing builder object. I need to think more about it.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -136,4 +138,26 @@ public static boolean isValidInstance(Resource instanceResources,
 
     return containerPlans;
   }
+
+  /**
+   * Sort the container plans based on the container Ids
+   * @param containers
+   * @return sorted array of container plans
+   */
+  public static PackingPlan.ContainerPlan[] sortOnContainerId(
```
billonahill: Could you add back the `PackingPlanTest.testContainerSorting` from the earlier review as `PackingUtilsTest.testContainerSorting`?

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -59,14 +67,48 @@ private boolean hasSpace(long ram, double cpuCores, long disk) {
...
+   * Remove an instance of a particular component from a container and update its
+   * corresponding resources.
+   *
+   * @return instanceId if the instance is removed the container.
+   * Return void if an instance is not found
+   */
+  public Optional<InstanceId> removeAnyInstanceOfComponent(Resource resource, String component) {
+    Optional<InstanceId> instanceId = getAnyInstanceOfComponent(component);
+    if (instanceId.isPresent()) {
+      usedRam -= resource.getRam();
```
billonahill: Feel free to disregard this comment, as we don't follow this standard in many places.

avflor:
@billonahill I addressed all the comments. We have some ideas about the packing builder. Will submit that later.

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -23,50 +28,94 @@
...
+   *
+   * @return the corresponding instance plan if the instance is removed the container.
+   * Return void if an instance is not found
+   */
+  public Optional<PackingPlan.InstancePlan> removeAnyInstanceOfComponent(String component) {
+    Optional<PackingPlan.InstancePlan> instancePlan = getAnyInstanceOfComponent(component);
+    if (instancePlan.isPresent()) {
+      PackingPlan.InstancePlan plan = instancePlan.get();
+      this.instances.remove(plan);
+      return Optional.of(plan);
```
billonahill: Instead of creating a new Optional you could return the same one removed: `instancePlan`.

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -23,50 +28,94 @@
...
-  private double usedCpuCores;
-  private long usedDisk;
+  private HashSet<PackingPlan.InstancePlan> instances;
 
   //Maximum resources that can be assigned to the container.
   private long maxRam;
   private double maxCpuCores;
   private long maxDisk;
 
   public Container(Resource resource) {
```
billonahill: to make it clean what this resource is, can we call it something like `capacity`?

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -23,50 +28,94 @@
...
 public class Container {
 
-  //Resources currently used by the container.
-  private long usedRam;
-  private double usedCpuCores;
-  private long usedDisk;
+  private HashSet<PackingPlan.InstancePlan> instances;
 
   //Maximum resources that can be assigned to the container.
   private long maxRam;
```
billonahill: Instead of storing each part of the resource passed, can we just store the resource itself?

heron/packing/tests/java/com/twitter/heron/packing/PackingUtilsTest.java
```diff
@@ -0,0 +1,97 @@
...
+        .build();
+  }
+
+  /**
+   * Tests the sorting of containers based on the container Id.
+   */
+  @Test
+  public void testContainerSortOnId() {
+
+    PackingPlan packingPlan = getFirstFitDecreasingPackingPlan(topology);
```
billonahill: This is an indirect way to test container sorting. Could we instead write this test without a dep on FirstFitDecreasingPacking and just pass a Set of containers to sortOnContainerId and assert the returned ordering? That way we can directly test the utility method without going through the FirstFitDecreasingPacking code, which should be tested in it's own test suite/class.

heron/packing/tests/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPackingTest.java
```diff
@@ -81,6 +81,35 @@ private static PackingPlan getFirstFitDecreasingPackingPlanRepack(
     return packing.repack(currentPackingPlan, componentChanges);
   }
 
+  PackingPlan doScalingTest(TopologyAPI.Topology t,
```
billonahill: `TopologyAPI.Topology topology`

avflor: @billonahill This gives me a checkstyle error that that the topology hides a field because the class contains a topology member. This is the reason I used t.

billonahill: You could make the method static and pass `totalInstances` in as `totalExpectedInstances`

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -30,15 +30,14 @@
...
 
-  public Container(Resource resource) {
-    this.maxRam = resource.getRam();
-    this.maxCpuCores = resource.getCpu();
-    this.maxDisk = resource.getDisk();
+  /**
+   * Creates a container with a specific capacity
+   * @param maxResource the capacity of the container in terms of cpu, ram and disk
+   */
+  public Container(Resource maxResource) {
```
billonahill: Instead of using both capacity and maxResource, how about just capacity?

billonahill:
One last naming nit and then lgtm! Thanks!

avflor:
@billonahill Thanks for the comments. I think we can merge now.

PR #1395, opened 2016-09-15
Use container id provided by packing plan

### LINKED ISSUE #1393: LocalScheduler should use container ID provided in packing plan to properly support scale down
If a topology needs N containers, `LocalScheduler` launches processes with ids ranging from 1 to N. During scale up, the id assigned to new container is equal to number of active containers + 1. This assumes may not hold once scale down is implemented, as scale down can result in removing of a container mid range. As a result number of active containers can be less than N. This will result in duplication of id and error. `LocalScheduler` should assign the same ids as provided by the packing plan to avoid such errors.

ashvina:
This is required to properly support scale up/down events. Using id provided in
the container plan avoids conflics resulting after a continer is removed.

Fixes #1393

### DIFF
--- a/heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
+++ b/heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
@@ -72,8 +72,8 @@ public class LocalSchedulerTest {
   public void testOnSchedule() throws Exception {
     Mockito.doNothing().
         when(scheduler).startExecutorMonitor(Mockito.anyInt(), Mockito.any(Process.class));
-    Process[] mockProcesses = new Process[3];
-    for (int i = 0; i < 3; i++) {
+    Process[] mockProcesses = new Process[4];
+    for (int i = 0; i < 4; i++) {
       mockProcesses[i] = Mockito.mock(Process.class);
       Mockito.doReturn(mockProcesses[i]).when(scheduler).startExecutorProcess(i);
     }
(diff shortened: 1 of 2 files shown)


heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
```diff
@@ -74,18 +72,25 @@ public void testClose() throws Exception {
...
+    Process[] mockProcesses = new Process[3];
+    for (int i = 0; i < 3; i++) {
+      mockProcesses[i] = Mockito.mock(Process.class);
+      Mockito.doReturn(mockProcesses[i]).when(scheduler).startExecutorProcess(i);
+    }
 
     PackingPlan packingPlan = Mockito.mock(PackingPlan.class);
+    Set<PackingPlan.ContainerPlan> containers = new HashSet<>();
+    containers.add(PackingTestUtils.testContainerPlan(1));
+    containers.add(PackingTestUtils.testContainerPlan(2));
```
billonahill: We could use container id 3 here instead of 2 to assert that it's ok to not be consecutive.

heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
```diff
@@ -159,7 +169,7 @@ public void testRemoveContainer() throws Exception {
     containersToRemove.add(containerToRemove);
     scheduler.removeContainers(containersToRemove);
     existingContainers.remove(containerToRemove);
-    Assert.assertEquals(existingContainers.size(), scheduler.getProcessToContainer().size());
+    Assert.assertEquals(1 + existingContainers.size(), scheduler.getProcessToContainer().size());
```
billonahill: If we're not already doing so, we should assert proper container ids as well as total container count.

heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
ashvina: Good point. Non-consecutive ids is getting verified in `testAddContainer#L123`.

heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
ashvina: Ok. The containerIds are indirectly verified in L185.

heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
ashvina: Fixed

heron/schedulers/tests/java/com/twitter/heron/scheduler/local/LocalSchedulerTest.java
ashvina: Fixed

kramasamy:
@ashvina and @billonahill - is this ready to go?

ashvina:
@kramasamy I do not have any pending changes related to this PR. In addition to the CI, I have also tested this change locally.

PR #1398, opened 2016-09-17
Persist scheduled container size in packing plan

### LINKED ISSUE #1396: Preserve container resources requested by scheduler in state manager
A scheduler may request more resources than what is suggested by the packing algorithm. For e.g.  the `AuroraScheduler` requests homogeneous containers. Currently this information is not preserved anywhere. It would be useful to retain this information. For e.g. implementations of re-packing algorithm can efficiently pack new instances if the available space is available. 

I am planning to add `scheduledResource` to `PackingPlan.ContainerPlan`. A scheduler will update  the packing plan in `onSchedule` method and replace the one in `.../packingplans/` in state manager.

Required for #1292

ashvina:
* Persist actual container size in packing plan as containerPlan.scheduledResource
* Rename containerPlan.resource to containerPlan.requiredResource: this is the minimum resource requirement as suggested by packing plan 

Fixes #1396

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
@@ -282,7 +282,7 @@ public class PackingPlan {
     /**
      * Returns a {@link ContainerPlan} with updated scheduledResource
      */
-    public ContainerPlan cloneWithScheduledResource(Resource resource) {
+    private ContainerPlan cloneWithScheduledResource(Resource resource) {
       return new ContainerPlan(getId(), getInstances(), getRequiredResource(), resource);
     }
 
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
@@ -192,20 +192,20 @@ public class PackingPlan {
   public static class ContainerPlan {
     private final int id;
     private final Set<InstancePlan> instances;
-    private final Resource resource;
+    private final Resource requiredResource;
     private final Optional<Resource> scheduledResource;
 
-    public ContainerPlan(int id, Set<InstancePlan> instances, Resource resource) {
-      this(id, instances, resource, null);
+    public ContainerPlan(int id, Set<InstancePlan> instances, Resource requiredResource) {
+      this(id, instances, requiredResource, null);
     }
 
     public ContainerPlan(int id,
                          Set<InstancePlan> instances,
-                         Resource resource,
+                         Resource requiredResource,
                          Resource scheduledResource) {
       this.id = id;
       this.instances = ImmutableSet.copyOf(instances);
-      this.resource = resource;
+      this.requiredResource = requiredResource;
       this.scheduledResource = Optional.fromNullable(scheduledResource);
     }
 
@@ -217,8 +217,8 @@ public class PackingPlan {
       return instances;
     }
 
-    public Resource getResource() {
-      return resource;
+    public Resource getRequiredResource() {
+      return requiredResource;
     }
 
     public Optional<Resource> getScheduledResource() {
@@ -238,7 +238,7 @@ public class PackingPlan {
 
       return id == that.id
           && getInstances().equals(that.getInstances())
-          && getResource().equals(that.getResource())
+          && getRequiredResource().equals(that.getRequiredResource())
           && getScheduledResource().equals(that.getScheduledResource());
     }
 
@@ -246,7 +246,7 @@ public class PackingPlan {
     public int hashCode() {
       int result = id;
       result = 31 * result + getInstances().hashCode();
-      result = 31 * result + getResource().hashCode();
+      result = 31 * result + getRequiredResource().hashCode();
       if (scheduledResource.isPresent()) {
         result = 31 * result + getScheduledResource().get().hashCode();
       }
@@ -255,8 +255,8 @@ public class PackingPlan {
 
     @Override
     public String toString() {
-      String str = String.format("{container-id: %s, instances-list: %s, estimated-resource: %s",
-          id, getInstances().toString(), getResource());
+      String str = String.format("{container-id: %s, instances-list: %s, required-resource: %s",
+          id, getInstances().toString(), getRequiredResource());
       if (scheduledResource.isPresent()) {
         str = String.format("%s, scheduled-resource: %s", str, getScheduledResource().get());
       }
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
@@ -15,6 +15,7 @@
 package com.twitter.heron.spi.packing;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
@@ -33,6 +34,31 @@ public class PackingPlan {
     for (ContainerPlan containerPlan : containers) {
       containersMap.put(containerPlan.getId(), containerPlan);
     }
+  }
+
+  /**
+   * Creates a clone of {@link PackingPlan}. It also computes the maximum of all the resources
+   * required by containers in the packing plan and updates the containers of the clone with the
+   * max resource information
+   */
+  public PackingPlan cloneWithHomogeneousScheduledResource() {
+    double maxCpu = 0;
+    long maxRam = 0;
+    long maxDisk = 0;
+    for (ContainerPlan containerPlan : getContainers()) {
+      maxCpu = Math.max(maxCpu, containerPlan.getRequiredResource().getCpu());
+      maxRam = Math.max(maxRam, containerPlan.getRequiredResource().getRam());
+      maxDisk = Math.max(maxDisk, containerPlan.getRequiredResource().getDisk());
+    }
+
+    Resource maxResource = new Resource(maxCpu, maxRam, maxDisk);
+    Set<ContainerPlan> updatedContainers = new HashSet<>();
+    for (ContainerPlan container : getContainers()) {
+      updatedContainers.add(container.cloneWithScheduledResource(maxResource));
+    }
+
+    PackingPlan updatedPackingPlan = new PackingPlan(getId(), updatedContainers);
+    return updatedPackingPlan;
   }
 
   public String getId() {
@@ -251,6 +277,13 @@ public class PackingPlan {
         result = 31 * result + getScheduledResource().get().hashCode();
       }
       return result;
+    }
+
+    /**
+     * Returns a {@link ContainerPlan} with updated scheduledResource
+     */
+    public ContainerPlan cloneWithScheduledResource(Resource resource) {
+      return new ContainerPlan(getId(), getInstances(), getRequiredResource(), resource);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -431,16 +434,45 @@ public static boolean curlAndExtractPackage(
...
+    PackingPlan updatedPackingPlan = new PackingPlan(packingPlan.getId(), updatedContainers);
+    return updatedPackingPlan;
+  }
+
+  /**
+   * Replaces persisted packing plan in state manager.
+   */
+  public static void persistUpdatedPackingPlan(String topologyName,
+                                               PackingPlan updatedPackingPlan,
+                                               Config runtime) {
```
billonahill: Could you pass a stateManager instead of a runtime here? Better to not bury the stateManager initialization code.

Actually, if PackingPlan is updateable in ZK why is this method needed? Couldn't we just need to do thisL:
```
stateManager.setPackingPlan(serializer.toProto(updatedPackingPlan), topologyName)
```

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -431,16 +434,45 @@ public static boolean curlAndExtractPackage(
   /**
    * Get a resource that requires the maximum amount of ram, cpu and disk
    */
-  public static Resource getMaxRequiredResource(PackingPlan packingPlan) {
+  public static PackingPlan getHomogenizedContainerPlan(PackingPlan packingPlan) {
```
billonahill: Please add a unit test for this method.

billonahill:
Also what about the resource usage in `com.twitter.heron.packing.Container`? Are those required or scheduled or should those resources remain generic?

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
ashvina: I have already created a unit test for this method, see SchedulerUtilsTest.updatesPackingPlanWithUniformSizedContainers#L180

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
ashvina: I will remove the runtime from `persistUpdatedPackingPlan` method.

Regarding the `setPackingPlan`, I may have borrowed it from `UpdateTopologyManager.updateTopology #L-96`. I am not sure if you would know why it was implemented like that in the UpdateTopologyManager?

ashvina:
`com.twitter.heron.packing.Container` is a helper class used by packing algorithms. This change does not impact or is impacted by that class.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
```diff
@@ -168,4 +175,63 @@ public void testSchedulerCommandArgs() throws Exception {
...
+    PackingPlan packing = createPackingPlan();
+    SchedulerUtils.persistUpdatedPackingPlan("topology", packing, runtime);
+    InOrder inOrder = Mockito.inOrder(adaptor);
+    inOrder.verify(adaptor).deletePackingPlan("topology");
+    inOrder.verify(adaptor).setPackingPlan(Mockito.any(PackingPlans.PackingPlan.class),
+        Mockito.eq("topology"));
+  }
+
+  @Test
+  public void updatesPackingPlanInStateManager() {
```
billonahill: Can we name this after the method it's testing, like testHomogenizedContainerPlan?

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
```diff
@@ -168,4 +175,63 @@ public void testSchedulerCommandArgs() throws Exception {
...
+    PackingPlan plan = createPackingPlan();
+    PackingPlan newPlan = SchedulerUtils.getHomogenizedContainerPlan(plan);
+
+    PackingPlan.ContainerPlan[] containerArray
+        = newPlan.getContainers().toArray(new PackingPlan.ContainerPlan[2]);
+    Assert.assertEquals(2, newPlan.getContainers().size());
+
+    PackingPlan.ContainerPlan largeContainer = containerArray[0];
+    PackingPlan.ContainerPlan smallContainer = containerArray[1];
```
billonahill: I'm confused, shouldn't each of these be the same size (i.e. homogeneous)?

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
billonahill: I see it now. I was searching for the method string and missed it.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
billonahill: Good eye. I suspect it was implemented like that in UTM initially because setPackingPlan was setting overwrite to false initially in LocalFileSystemStateManager. I don't think it should be doing a delete there as well.

billonahill:
Re `com.twitter.heron.packing.Container` yes, that class in completely independent. I was just wondering if the naming change from resource to required and scheduled resource was something that that class should also adopt. I'm really not sure if it should or not.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
ashvina: The tests submits two different sized containers. Here `largeContainer` and `smallContainer` refers to the original size of the container before they are homogenized, see `SchedulerUtilsTests L#220`.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
ashvina: Ok. A public method can have more than one test method to test different control paths. So I plan to follow the following rules:

1. Include some test information in the name. That way if the test fails, a hint will be available to the developer. 
1. To avoid making the test name too long avoid redundant word `test`. The method belongs to a test class and is `Test` annotated. When the test fails, the word test does not add any value, just makes the method name long.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
ashvina: This is what I have understood, the implementations of `setPackingPlan` method will take care of updating the packing plan if it exists, or will create a node/file if it is not present. Is this correct?

billonahill: Correct, we shouldn't need to delete first.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
billonahill: +1

Having some notion of the test method in the name (when possible) helps to clarify the scope of what the test if intended to verify. I find this useful for tests of limited scope, which I guess most should be ideally.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
billonahill: Right, but large and smallContainer here are taken from containerArray which is taken form newPlan, which is supposed to be homogeneous right? So I'd expect newPlan to have same size containers, not large and small. Am I misunderstanding the expected behavior?

ashvina: I see the source of confusion. The resources estimated by packing plan are in contained in `requiredResource`. This object will not be modified by `getHomogenizedContainerPlan`. The max resource value will be updated in `scheduledResources` object. This is verified in L#208-L#217. Required resource of large container will be greater than that of small container. However `scheduledResources` will be equal.

billonahill: I see now, thanks for walking me though it. What would you think about renaming `getHomogenizedContainerPlan` to something like `withHomogenousScheduledResources`?

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -431,16 +434,45 @@ public static boolean curlAndExtractPackage(
...
+      maxCpu = Math.max(maxCpu, containerPlan.getRequiredResource().getCpu());
+      maxRam = Math.max(maxRam, containerPlan.getRequiredResource().getRam());
+      maxDisk = Math.max(maxDisk, containerPlan.getRequiredResource().getDisk());
     }
 
-    return new Resource(maxCpu, maxRam, maxDisk);
+    Resource maxResource = new Resource(maxCpu, maxRam, maxDisk);
+    Set<PackingPlan.ContainerPlan> updatedContainers = new HashSet<>();
+    for (PackingPlan.ContainerPlan container : packingPlan.getContainers()) {
+      PackingPlan.ContainerPlan updatedContainer = new PackingPlan.ContainerPlan(container.getId(),
```
billonahill: This seems like a useful method to add directly to `ContainerPlan`:
```
public PackingPlan.ContainerPlan cloneWithScheduledResource(Resource resource);
```
We did something similar with `Resource.cloneWithRam`. Then the new constructor on ContainerPlan could be private and we use the clone method instead.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
ashvina: I like your suggestion. Will update the method name.

heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
ashvina: Good suggestion. Will add it.

heron/spi/tests/java/com/twitter/heron/spi/utils/SchedulerUtilsTest.java
ashvina: On second thoughts, I am planning to add `PackingPlan.cloneWithHomogeneousScheduledResources`. Thoughts?

heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
```diff
@@ -253,6 +279,13 @@ public int hashCode() {
       return result;
     }
 
+    /**
+     * Returns a {@link ContainerPlan} with updated scheduledResource
+     */
+    public ContainerPlan cloneWithScheduledResource(Resource resource) {
```
billonahill: This could be private

billonahill:
One last public -> private comment, then lgtm! Thanks!

heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
ashvina: `clone` methods are public. Could you please help me understand why this should be private?

billonahill: Just because nobody else currently needs it besides the class itself. I typically keep things as restrictive as necessary and increase visibility when necessary, or when ready to provide a public, stable API.

ashvina: Done

billonahill:
👍 once CI completes

PR #1406, opened 2016-09-21
Adding exception handling to scheduler responses

billonahill:
If a handler throws an exception, the scheduler doesn't catch it. This causes the exception to disappear without being logged while a response never gets sent to the client. This fixes that.

It would be great to take this one step further and allow the server to pass an error message back to the client (in addition to the `NOTOK` status code) but that's a much larger refactor for another day.


billonahill:
Whoops, I merged this one too soon. @kramasamy et al please let me know if you have comments and I'll put up a patch for them.

PR #1407, opened 2016-09-21
Ability to run just one test in local_test_runner

billonahill:
Allow passing just one testname to local_test_runner.


kramasamy:
👍  - once CI passes.

PR #1408, opened 2016-09-21
local state manager to check for path before trying to read file

billonahill:
Trying to read from a node that doesn't exist throws a `FilePathNotFoundException` instead of returning an empty byte array.


kramasamy:
👍

PR #1409, opened 2016-09-21
Use an enum for the TMaster commands instead of a string

billonahill:
Plus random minor cleanup to `TopologyUtils.java` and `ResourceCompliantRRPacking.java`.


kramasamy:
👍 - once CI passes.

PR #1412, opened 2016-09-21
Deactivate topology before scaling

### LINKED ISSUE #1390: Support topology deactivation period before scaling
This is a simple way to ensure stateful data correctness for many topologies, via @pankajgupta.

The approach used for scaling at Twitter currently is the following:
1. Deactivate the topology for X seconds. This halts new tuples from being emitted by spouts and causes time-based caches to be drained, where X > the cache interval.
2. Redeploy the topology as activated. New tuples are emitted and processed by the correct instances when using fields grouping. Caches are initially empty (i.e. drained) and correct.

A similar feature should be built into the scaling event. The topology update command would take an optional parameter to indicate that before the scaling event occurs, deactivate the topology for X seconds. The default deactivation interval could also be provided in the topology config, which would allow frameworks built on top of Heron to inject this setting based on their configured cache settings. This would eliminate the need for the user to set this properly.

Related to #1292.

billonahill:
Deactivates a topology before scaling and then reactivates after scaling is complete. The reactivation happens asynchronously since it needs to wait for the scaled topology to come back up.

Adds support for 2 new topology configs:
- `heron.topology.update.deactivate.seconds`: time to sleep after deactivating before scaling
- `heron.topology.update.reactivate.seconds`: time to wait for topology to come up after scaling before reactivating.

Related to #1292. Fixes #1390.


heron/spi/src/java/com/twitter/heron/spi/common/keys.yaml
```diff
@@ -62,6 +62,12 @@ SCHEDULER_IS_SERVICE:                heron.scheduler.is.service
 SCHEDULER_JAR:                       heron.jars.scheduler
 
 ################################################################################
+# keys for Topology Update command
+################################################################################
+TOPOLOGY_UPDATE_DEACTIVATE_SECONDS:  heron.topology.update.deactivate.seconds
+TOPOLOGY_UPDATE_REACTIVATE_SECONDS:  heron.topology.update.reactivate.seconds
```
billonahill: @kramasamy is this change required topology configs? I see many entries in this file that are not referenced in the codebase so wondering what it's primary function is.

kramasamy: Is this config used in scheduler.yaml? If that is the case it is needed in keys.yaml and you need an accessor function in one of the Context/Config to access this config. If there are some default values to be supplied for this argument, you need to add this in defaults.yaml. It is a messy design - we need to cleanup some point in time.

billonahill: No, `scheduler.yaml` doesn't use this. The current intent is that it can be set in one of two ways:

1. in the topology main code by calling `setUpdateDeactivateWaitDuration(Map<String, Object> conf, int seconds)` on `Config`.
2. By passing it as a command line arg vai `--config-property` when calling `heron update`.

If that's the case, it seems I don't need the entires in `keys.yaml`, `defaults.yaml` and `Context`.

billonahill:
Ping. Any other comments on this PR?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -104,12 +126,124 @@ public void updateTopology(final PackingPlans.PackingPlan existingProtoPackingPl
...
+    long deactivateSleepSeconds = TopologyUtils.getConfigWithDefault(
+        topologyConfig, TOPOLOGY_UPDATE_DEACTIVATE_WAIT_SECS, 0L);
+
+    logInfo("Deactivating topology %s before handling update request", topology.getName());
+    assertTrue(TMasterUtils.transitionTopologyState(
+            topology.getName(), TMasterUtils.TMasterCommand.DEACTIVATE, stateManager,
+            TopologyAPI.TopologyState.RUNNING, TopologyAPI.TopologyState.PAUSED),
+        "Failed to deactivate topology %s. Aborting update request", topology.getName());
+
+    if (deactivateSleepSeconds > 0) {
```
ashvina: How are users expected to use this configuration? Is it based on size of the topology?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -104,12 +126,124 @@ public void updateTopology(final PackingPlans.PackingPlan existingProtoPackingPl
...
     }
   }
 
   @VisibleForTesting
+  void deactivateTopology(SchedulerStateManagerAdaptor stateManager,
+                          final TopologyAPI.Topology topology)
+      throws InterruptedException {
+
+    List<TopologyAPI.Config.KeyValue> topologyConfig = topology.getTopologyConfig().getKvsList();
+    long deactivateSleepSeconds = TopologyUtils.getConfigWithDefault(
```
ashvina: If wait time is a topology configuration, user can set it only once. If it were update operation specific configuration, then auto-scaling could use a different value based on need. Should this config be adaptive?

heron/scheduler-core/tests/java/com/twitter/heron/scheduler/UpdateTopologyManagerTest.java
```diff
@@ -89,9 +90,17 @@ public void requestsToAddAndRemoveContainers() throws Exception {
 
     Mockito.doReturn(null).when(spyUpdateManager).
         getUpdatedTopology(null, proposedPacking, mockStateMgr);
+    Mockito.doNothing().when(spyUpdateManager)
+        .deactivateTopology(eq(mockStateMgr), any(TopologyAPI.Topology.class));
```
ashvina: These operations should be Mockito.verified?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: That is the intent yes, which is that the setting can be configured at topology submission time in the main method, or overridden at update time from the command line.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: Some typologies flush internal state to external systems every N seconds. In that case they would set this value to something lager than N. This assures that the topology deactivated (pausing new tuple consumption from spouts) and state is flushed before scaling occurs.

heron/spi/src/java/com/twitter/heron/spi/common/keys.yaml
kramasamy: @billonahill - should we just keep only --config-property? IMO - it might not be a good idea to provide API since our goal long term is to remove them from being hard coded.
-

heron/scheduler-core/tests/java/com/twitter/heron/scheduler/UpdateTopologyManagerTest.java
billonahill: Yes, good catch. Updated.

heron/spi/src/java/com/twitter/heron/spi/common/keys.yaml
billonahill: @kramasamy Tsar has a use case to set this when topologies are created based on the cache settings so I think it's ok to keep the API. The values wouldn't be hard-coded, they'd be set during topology creation based on the setting from the higher level framework.

ashvina: @billonahill The activate and deactivate wait durations are used to ensure runtime states are flushed/recovered as you explained below. It may not be always feasible to predict the time in all use cases. Moreover it would be mess up the api if we add one for each use case. Would it be better if the scheduler could `callback` user implementations of some hooks to control this behavior?

billonahill: @ashvina implementing callbacks would be great but that's a much more involved feature that we'd need to think through and ideally have a solid use case for before building it. Currently we don't support any network callbacks to other systems so to do so we'd need to design the protocol, the register/deregister callback mechanisms, etc. I suggest we look into that once we have a solid use case for it's use.

ashvina: @billonahill I am not thinking of a network callback. Say a new interface `UpdateCallback` is added. A new `scheduler` config could be added using which the callback handler could be configured. The default implementation could be no op handler. User could configure a provided time based callback handler or a custom handler. This would not need a network call or protocol implementation.

billonahill: @ashvina and I spoke offline and agreed to move forward with this approach and consider an event listener approach once we have a concrete use case for it.

billonahill:
I had to make a lot of changes for tunneling support for scheduler, which I broke out into it's own PR in #1456. Once we ship that I'll merge master into this and this review will get much smaller again.

billonahill:
I've merged in the tunneling changes and this PR is small again. Let me know if there are any add'l comments.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -26,28 +33,46 @@
...
     this.runtime = runtime;
     this.scalableScheduler = scalableScheduler;
     this.deserializer = new PackingPlanProtoDeserializer();
+    this.reactivateExecutorService = new ScheduledThreadPoolExecutor(1);
+    this.reactivateExecutorService.setMaximumPoolSize(1);
+  }
+
+  @Override
+  public void close() {
+    this.reactivateExecutorService.shutdown();
```
ashvina: Should this be shutdownNow as any pending reactivate commands should be terminated?

ashvina:
LGTM 👍

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
billonahill: Cood catch, fixed. I changed to `shutdown()` when I was trying to get the CLI to not shutdown and kill the scheduler work before the reactivation finished. Instead I just waited on the future with `future.get(timeout)`.

PR #1415, opened 2016-09-22
Use C++ auto keyword when possible

congwang:
We already use the C++11 auto keyword in our current code base, we should use it in other places too.


kramasamy:
👍  ship it.

PR #1419, opened 2016-09-23
Docker Uploader for Heron Topologies

chris-pardy:
This adds support for uploading Topology definitions to a Docker repository by building them into a Docker Image. A base image can be specified in the config file, as well as a repository prefix for the image.

The DockerUploader creates a Docker image from a given base image (eg. ubuntu:trusty) and adds the topology definition and binary to the docker image, it then optionally uploads that image to a docker repository.

This allows schedulers downstream to deploy the generated docker image, or a new docker image based on the generated docker image. The purpose is to enable running Heron topologies within a Docker container. In practice this allows a correctly configured Aurora job to use the Docker container to setup the job sandbox. It could also serve as the basis for other schedulers that would like to schedule heron instances inside docker containers.

All docker functionality is provided by the Docker command line interface, so if Docker is not installed or configured uploading via the DockerUploader will fail.

In practice how we're using this:
- We've pre-built a docker image from a minimal trusty + corporate stuff + heron core
- When we "upload" topologies we create a new docker image from the heron core base that includes our topology definition, his is pushed into our private docker repository.
- We use a custom Aurora job file that knows how to use leverage these docker containers to start the job.


kramasamy:
@chris-pardy - Can you provide an elaborate description about what this PR does and how is intend to be used so that we can understand.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
```diff
@@ -0,0 +1,187 @@
...
+ * IUploader Implementation that creates a DockerUploader Image for the Topology
+ * <p>
+ * The created DockerUploader Image will be based on the given base image and include the topology definition
+ * and binary in a fixed location. The uploader optionally pushes the docker image to a remote repository.
+ * Downstream schedulers must take responsibility for leveraging this image, by either building an additional
+ * DockerUploader image based on it with the correct START command
+ * </p>
+ * <p>
+ * The topology definition will be written to "/home/[role]/[topology-name]"
+ * if no role is given "heron" is used
```
kramasamy: if the user does not provide a role in the command line - we supply the role as "default" and the environment as "default"

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
```diff
@@ -0,0 +1,187 @@
...
+ * <p>
+ * The Dockerfile generated will be as follows:
+ * <code>
+ * FROM [base-image]
+ * ADD [topology.tar]   /home/[role | "heron"]/[topology-name]
+ * </code>
+ * Assuming your topology was named "MyTopology" and based on the ubuntu:trusty image with the role
+ * "MyRole" and packed into a tarfile called "MyTopology-3424.tar" the result would be.
+ * <code>
+ * FROM ubuntu:trusty
```
kramasamy: the base image should be a heron core image - that already contains the core image of heron?

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
chris-pardy: Is there a way to check if the user didn't supply the role? I don't like the idea of checking for default.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
chris-pardy: That's how we're using it but it's not necessary, a scheduler could do that work, or configure the scheduled task to download the core. I can change the documentation to call out that way of leveraging the base image if you think it's worthwhile.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
kramasamy: It is at the CLI level - by the time it invokes the submitter which in turn calls the uploader, it might be too late.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/DockerUploader.java
kramasamy: you mean scheduler will download the docker image for heron core and the docker image for topology? or the heron core is a tar image (like we have it today) and the docker image for the topology will know where to download the core when docker image is starting? Is that correct?

kramasamy: got it from the detailed explanation.

kramasamy:
one more meta question - is a single docker container runs the entire topology or fragments of the topologies run in multiple containers constitute a single topology?

chris-pardy:
@kramasamy - much like the other uploaders this leaves all the questions of "how" to run the topology to the schedulers. Assuming you have a base image with the Heron Core libs and then you use this uploader the result will be a docker image with all the code needed to run a topology container. In the case of something like Aurora we're starting multiple jobs that use this container each of which run a part of the topology.

kramasamy:
@chris-pardy - one more question, you might know this already. In the case of Aurora, you need to supply *.aurora file. In the aurora file, you need to have explicit instructions to download the docker image from the registry before you can start it, correct? If the topology image is different for different topologies, how can you specify the different image URLs for those - since the aurora file is part of the configuration and is static. 

BTW, If you are interested in the heron committers slack channel? If so, I would need your email address.

chris-pardy:
@kramasamy the Scheduler is passed the URL that the uploader outputs in the Runtime Config, since we're sending the Docker tag as the URI here (technically it's not a URI but...) the Scheduler can export it. In the case of the Aurora scheduler you have:
`...
String heronCoreReleasePkgURI = Context.corePackageUri(config);
String topologyPkgURI = Runtime.topologyPackageUri(runtime).toString();

auroraProperties.put("CORE_PACKAGE_URI", heronCoreReleasePkgURI);
auroraProperties.put("TOPOLOGY_PACKAGE_URI", topologyPkgURI);
...`

So the heron.aurora file could download the image by doing something like `docker pull {{TOPOLOGY_PACKAGE_URI}}`. Aurora will handle the templating.

And re: Slack chris.pardy@gmail.com

kramasamy:
Got it. The heron.aurora file need to be changed to use docker pull if we are using docker uploader for that cluster configuration. Furthermore, launching of the executor needs to be changed as well  - the docker container should be informed on how to start the heron executor.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/sample.yaml
```diff
@@ -0,0 +1,11 @@
+# uploader class for transferring the topology jar/tar files to storage
```
kramasamy: Similar to this file, can you add a sample for heron.aurora - if somebody wants to use docker uploader. You could add this under the directory heron/config/src/yaml/aurora - with the file name heron.aurora.docker.

heron/uploaders/src/java/BUILD
```diff
@@ -107,3 +107,22 @@ genrule(
     outs = ["heron-scp-uploader.jar"],
```
kramasamy: can you write a page documentation similar to other uploaders so that we have information about how to use this?

chris-pardy:
@kramasamy I've added a documentation page and a sample heron.docker.aurora file. Please let me know if there's anything else needed.

heron/config/src/yaml/conf/aurora/heron.docker.aurora
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+heron_core_release_uri = '{{CORE_PACKAGE_URI}}'
+heron_topology_image_uri = '{{TOPOLOGY_PACKAGE_URI}}'
+core_release_file = "heron-core.tar.gz"
+
+# --- processes ---
+fetch_heron_system = Process(
+  name = 'fetch_heron_system',
+  cmdline = 'curl %s -o %s && tar zxf %s' % (heron_core_release_uri, core_release_file, core_release_file)
```
kramasamy: @chris-pardy - how does this work? If the docker image is supposed to contain both the core image as well as topology image, we need to just use docker pull to get the image, is it correct?

chris-pardy: This assumes that the docker image only contains the topology, so we fetch core from a remote url first. The container = ... bit tells Aurora to use the Docker image that the uploader built for the service code. If you wanted to also include the core files you would need to first build a docker image with those libraries. That's going to be an exercise left up to the reader.

ajorgensen:
I'm curious how you are getting the log files out of the docker sandbox and into a place where they can be accessed by heron/aurora. We played around with this approach at first but found that getting at the log files in the docker container was more complex than was worth the effort of running docker. If your answer is you're not then this implementation will break multiple components the of the UI that allow for streaming the logs from the aurora container

chris-pardy:
@ajorgensen the docker image is just used to distribute the topology binaries and yaml config. The log files are written into the working directory which is the Aurora Sandbox.

ajorgensen:
Sorry I must have misunderstood the purpose then, this is just using docker as a transportation mechanism for a tarball? My experience with docker and aurora is limited so sorry if this is a naive question but how does the` /home/<role>/<topology>` folder (which correct me if i'm wrong is in the docker container) in such a way that it can be access in the aurora sandbox? I see your aurora command is referencing by its absolute path `/home/<role>/<topology>`

chris-pardy:
@ajorgensen no, the Aurora Scheduler with a correctly configured heron.aurora file (per the example given) will work. I'm not sure what the details of this are but effectively Aurora will take care of getting the docker image from the registry and inflating it, it will then configure the sandbox such that it's reachable from the executable that are in the docker image. Basically using this serves 2 purposes.
1) It makes it so that fetching the topology tar doesn't rely on something like 'curl' or 'hdfs' being installed on the mesos slaves. Generally speaking making the slaves stupid enough to just use docker is the advice I've gotten for Aurora.
2) It allows multiple instances of the job on the same machine to leverage the same docker image download preventing it from repeating downloads.

As part of 2 if you were to build a docker image with all the docker core libraries and set it as the base image in the uploader config Mesos/Aurora is smart enough to cache that base image on the executors which means that if you're running a number of different topologies they'll each only download the bits that make them different. I haven't explicitly configured things for this use case as I'm expecting that each user will end up wanting <their preferred linux distro> + <their required software> + <heron core libs>.

kramasamy:
@chris-pardy - one more question, can we do docker all the way where we run the images directly in the container? similar to what is used at http://aurora.apache.org/documentation/latest/reference/configuration/#docker-object

The docker image when it starts needs to start heron-executor and a clean way to view logs in the running container.  This will be really useful since Kubernetes/ECS can leverage this e2e docker for deploying containers. thoughts?

chris-pardy:
@kramasamy I'm not sure what exactly is required by the new Mesos/Aurora Docker support. However I'm guessing it would be the addition of a "CMD" directive at the end of the Dockerfile basically saying what to run. Because the heron core binaries have to be compiled to the target system (Ubuntu, CentOs, etc.) you can't install those as part of creating this image without tying yourself to a specific distribution so you would need to rely on base image that was specified having them installed already. Additionally at the point when the uploader is invoked you don't know most of the information you need to actually run the topology (ie. Ram Map) so in practice what would be required is a CMD directive that looks a lot like the command_to_start_executor in the heron.aurora file, which is to say a lot of variables that can be specified. The scheduler would then need to specify things like the path to the Core binaries to the docker command so that they can be turned into environment variables.

I like the idea of getting this Docker container to be executable but I think it's beyond the scope of this PR. Assuming that the base image on which this container is built includes the core binaries it's possible to take the output of this uploader and run it by specifying the executor binary to the docker run command.

chris-pardy:
Is this good to be merged or does it need additional changes?

ajorgensen:
Sorry for the delay i wasn't able to verify this change locally this weekend. I'll do that today.

ajorgensen:
I am getting the following when trying to compile on ubuntu 14.04 with docker:

```
ERROR: /scratch/heron/uploaders/src/java/BUILD:120:12: in deps attribute of java_binary rule //heron/uploaders/src/java:docker-uploader-unshaded: deps not allowed without srcs; move to runtime_deps?
ERROR: Analysis of target '//scripts/packages:tarpkgs' failed; build aborted.
```

Here's the command:
`./docker/build-artifacts.sh ubuntu14.04 docker ~/Downloads/heron-artifacts/heron-docker`

Do you mind just verifying that it works for you?

ajorgensen:
I see whats going on, this .tarignore file is a bit too aggressive: https://github.com/twitter/heron/blob/master/docker/.tarignore#L2

heron/config/src/yaml/conf/aurora/heron.docker.aurora
```diff
@@ -0,0 +1,50 @@
...
+jobs = [
+  Job(
+    name = '{{TOPOLOGY_NAME}}',
+    cluster = '{{CLUSTER}}',
+    role = '{{ROLE}}',
+    environment = '{{ENVIRON}}',
+    service = True,
+    task = heron_task,
+    instances = '{{NUM_CONTAINERS}}',
+    announce = Announcer(primary_port = 'http')
```
ajorgensen: You are missing a `,` at the end of this line.

heron/config/src/yaml/conf/aurora/heron.docker.aurora
```diff
@@ -0,0 +1,50 @@
...
+    container = Container(docker = Docker(image = heron_topology_image_uri))
```
ajorgensen: I believe this is supposed to just be `container = Docker(image = heron_topology_image_uri)` according to the docs here: http://aurora.apache.org/documentation/latest/features/containers/

ajorgensen: Sorry  it looks like the docs differ in a few places on this one. I tried this: http://livewyer.com/blog/2015/04/13/deploying-docker-containers-using-apache-aurora and it seemed to parse correctly.

I think you do need to interpolate the `{{heron_topology_image_uri}}` though instead of just having it be a string.

chris-pardy: The containder = Docker is only supported in newer versions of Aurora, but you're right I need to wrap that in {{ }}. We're using a patched version of Aurora with different config files so it's harder for me to test this change. Thanks.

ajorgensen: Ok. Coincidentally we dont have docker turned on in our aurora cluster and the aurora vagrant setup seems to not be working for me so im trying to figure out a way i can test this change out. With the missing comma I got up to the point where aurora is telling me that docker is not enabled. Going to try the local vagrant route again and see if i can verify this all works.

I see what you mean about the interpolation. I believe you're correct given that its already interpolated it should be able to just be used as a normal variable.

ajorgensen:
Was able to get a little bit farther. The container is now existing with a error code when attempting to be run:
image

I was able to run the container directly with docker and see the topology resources were in it. I'll see if i can keep playing with it to get the container to actually run in aurora and not exist with an error.

ajorgensen:
Yeah so that appears to be it. If i create an aurora file that looks like:

```
hello_loop = Process(
  name = 'hello',
  cmdline = """
    while true; do
      echo hello world
      sleep 10
    done
  """)

task = Task(
  processes = [hello_loop],
  resources = Resources(cpu=1, ram=1*MB, disk=8*MB)
)

jobs = [
  Service(
    cluster = 'devcluster',
    environment = 'devel',
    role = 'vagrant',
    name = 'hello_docker',
    task = task,
    container = Docker(image = 'test-container')
  )
]
```

and a docker container specification like:
```
FROM ubuntu:14.04

ADD heron-examples.jar /mnt/heron-examples.jar
```

I get the same exit with 127 that I do with your change here. I think this indicates that you cannot really use docker in the way you are intending to here (as a transportation mechanism only) because aurora will try and run it but have another to run. Also i am fairly positive that by default nothing from inside the container is linked into the aurora sandbox without you explicitly creating the mount points to do so. 

Sorry for the long conclusion but I do not believe that this PR will work as is and will need to either be re-worked to support running the heron topology fully inside the docker container (which shouldn't be too much of a stretch from this I dont think).

heron/uploaders/src/java/com/twitter/heron/uploader/docker/sample.yaml
```diff
@@ -0,0 +1,11 @@
...
+heron.class.uploader:         com.twitter.heron.uploader.docker.DockerUploader
+
+#base image
+heron.uploader.docker.base:                      ubuntu:latest
```
ajorgensen: The base here actually has to be a container with python installed in it. Aurora does some crazy stuff to get a aurora executor running inside the docker container so it can monitor its process and do health checking and things. The default ubuntu container does not have python installs so you'll need to set `python:2.7` or something similar as the base container.

ajorgensen: The container attempted to run yes but overall it did not work.

chris-pardy: I'm personally using a Docker base image with the binaries, but writing them into the sandbox should be fine. I'm going to verify this tonight.

ajorgensen: Do you mean the sandbox in aurora? AFAIK you dont have a mount point inside the docker container into the aurora sandbox and any command you setup in aurora will be run into the docker container itself which means that you need to be able to get to the binaries inside the docker container.

kramasamy: If you need the heron binaries in a container, we can build it as a part of the release process and ship it. Let me know if this will simplify it.

ajorgensen: yeah I think building out the full docker support wouldn't be too much more work from where this PR is now and would actually simplify how all this works.

kramasamy: great! if you and @chris-pardy can let us what will be needed, we can get this done.

chris-pardy:
It doesn't nessesarily make sense to build a base image since people are
likely going to want to build off task or organization specific base images.

I believe the problem you're seeing is caused by permissions in the docker
container, can you run docker and do an ls -l on the directory where the
topology files are written and see if they're readable by everyone?

kramasamy:
@chris-pardy - one thought we had was - if we build a standard base image then we can do docker all the way including the launch of container and within the container starting various process in the appropriate sequence etc.  Of course, enterprise can construct a new image based on this image - to address their needs - especially with respect to their needs.

heron/uploaders/src/java/com/twitter/heron/uploader/docker/sample.yaml
chris-pardy: @ajorgensen when you do go down that route you'll find that the heron executor tries to run chmod +x on a number of the base files. That won't work if they're not owned by the user who's running the executor. That's definitely not going to be the case for a Docker Image deployed to Aurora. It's a simple change to the executor to treat those as non-terminal failures.
I can put together a Docker file for some base images easily enough, however that's not going to change this Uploader at all. I'm working on figuring out why it's not working for you but the problem I'm currently having is that I can't get the docker build to complete locally, it just hangs with 177 tasks remaining.

chris-pardy:
@ajorgensen Doing a build of Heron via Docker seems to hang with 177 tasks left. So I'm not able to do a complete verification yet.
The Base image needs to be something with both python and java installed, I'm not sure if there are standard images out on github you can use but this may be a good use case for building a base image.

kramasamy:
@chris-pardy - I build heron using docker for all our releases and it seems to doing fine. Typically, it takes 30-35 minutes to make a release. If you are doing it for the first time, docker image building takes some time since it installs various software. But after the first time, it should go faster.

kramasamy:
@ajorgensen - any update on this PR?

chris-pardy:
@kramasamy I think the next step at this point is to build a base docker image that can be uploaded to docker hub for people to use. I don't have time to do that right now but I may be able to get to it in the next week.

chris-pardy:
@kramasamy I obviously don't have time to finish this now, but please feel free to appropriate what you will in the future!

kramasamy:
@chris-pardy - no worries. We are moving into a direction where the docker images are prebuilt for each heron release and is available for you (based on your's and John's). The only remaining aspect is how to upload the job jar and pull it in the docker image that is running. We are currently using an URL based approach. Your PR gave a lot of valuable insights. Currently we are running seamlessly with this approach in DC/OS - Marathon.

PR #1420, opened 2016-09-23
More configurable Aurora

chris-pardy:
We'd like to keep using the builtin Aurora scheduler but we'd like to.
1) Move the default file to a different location, and not have it be named heron.aurora (we're using json for other aurora jobs)
2) Sometimes specify a different file for different topologies.

This PR is intended to enable this workflow by allowing configs to override the location and/or name of the default aurora template. Additionally it looks for a System Property which if set will take precedence. Because this method falls back to the old behavior if the properties aren't set this should be a non-breaking change however there is a very small possibility that someone is running the Heron CLI with the environment property HERON_SCHEDULER_JOB_TEMPLATE set to some garbage. I think the risk is minimal enough that it's not required to require a config be turned on to check the environment property.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
@@ -14,16 +14,24 @@
 
 package com.twitter.heron.scheduler.aurora;
 
+import java.io.File;
+
 import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.Context;
 
 public final class AuroraContext extends Context {
   public static final String JOB_LINK_TEMPLATE = "heron.scheduler.job.link.template";
+  public static final String JOB_TEMPLATE = "heron.scheduler.job.template";
 
   private AuroraContext() {
   }
 
   public static String getJobLinkTemplate(Config config) {
     return config.getStringValue(JOB_LINK_TEMPLATE);
+  }
+
+  public static String getHeronAuroraPath(Config config) {
+    return config.getStringValue(JOB_TEMPLATE,
+        new File(Context.heronConf(config), "heron.aurora").getPath());
   }
 }
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
@@ -14,7 +14,6 @@
 
 package com.twitter.heron.scheduler.aurora;
 
-import java.io.File;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -98,7 +97,7 @@ public class AuroraScheduler implements IScheduler, IScalable {
 
     Map<String, String> auroraProperties = createAuroraProperties(updatedPackingPlan);
 
-    return controller.createJob(getHeronAuroraPath(), auroraProperties);
+    return controller.createJob(AuroraContext.getHeronAuroraPath(config), auroraProperties);
   }
 
   @Override
@@ -158,10 +157,6 @@ public class AuroraScheduler implements IScheduler, IScalable {
         javaOpts.getBytes(Charset.forName("UTF-8")));
 
     return String.format("\"%s\"", javaOptsBase64.replace("=", "&equals;"));
-  }
-
-  protected String getHeronAuroraPath() {
-    return new File(Context.heronConf(config), "heron.aurora").getPath();
   }
 
   protected Map<String, String> createAuroraProperties(PackingPlan packing) {
--- a/heron/schedulers/tests/java/BUILD
+++ b/heron/schedulers/tests/java/BUILD
@@ -70,6 +70,7 @@ java_tests(
     "com.twitter.heron.scheduler.aurora.AuroraSchedulerTest",
     "com.twitter.heron.scheduler.aurora.AuroraLauncherTest",
     "com.twitter.heron.scheduler.aurora.AuroraControllerTest",
+    "com.twitter.heron.scheduler.aurora.AuroraContextTest",
   ],
   runtime_deps = [ ":aurora-tests" ],
   size = "small",
--- a/heron/schedulers/tests/java/com/twitter/heron/scheduler/aurora/AuroraContextTest.java
+++ b/heron/schedulers/tests/java/com/twitter/heron/scheduler/aurora/AuroraContextTest.java
@@ -23,13 +23,13 @@ public class AuroraContextTest {
 
   @Test
   public void testUsesConfigString() {
-    final String aurora_template = "/dir/test.aurora";
+    final String auroraTemplate = "/dir/test.aurora";
     Config config = Config.newBuilder()
-        .put(AuroraContext.JOB_TEMPLATE, aurora_template)
+        .put(AuroraContext.JOB_TEMPLATE, auroraTemplate)
         .put(ConfigKeys.get("HERON_CONF"), "/test")
         .build();
     Assert.assertEquals("Expected to use value from JOB_TEMPLATE config",
-        aurora_template, AuroraContext.getHeronAuroraPath(config));
+        auroraTemplate, AuroraContext.getHeronAuroraPath(config));
   }
 
   @Test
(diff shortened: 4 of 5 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
```diff
@@ -14,16 +14,40 @@
...
+  public static final String JOB_TEMPLATE = "heron.scheduler.job.template";
+  public static final String JOB_TEMPLATE_DIR = "heron.scheduler.job.template.dir";
 
   private AuroraContext() {
   }
 
   public static String getJobLinkTemplate(Config config) {
     return config.getStringValue(JOB_LINK_TEMPLATE);
   }
+
```
kramasamy: Instead of using environment variable, a better approach would be to check if these new configs are present, if these are present, use those - otherwise, default to what it is used to be. Alternatively, we can make this config mandatory - with the default pointing to heron.aurora file in the cluster directory.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
```diff
@@ -14,16 +14,40 @@
...
+   * Get the path to the Aurora Job template to be used by heron
+   * If the environment parameter HERON_SCHEDULER_JON_TEMPLATE is defined then it is used
+   * Otherwise checks the configuration properties heron.scheduler.job.template and
+   * heron.scheduler.job.template.dir for the location of the template.
+   * If heron.scheduler.job.template is not defined then "heron.aurora" is used
+   * If heron.scheduler.job.template.dir is not defined then the cluster specific conf directory
+   * is used.
+   */
+  public static String getHeronAuroraPath(Config config) {
+    String template = System.getenv(JOB_TEMPLATE_ENV);
```
billonahill: Instead of mixing both Heron configs and environment variables can we just use configs? Splitting the logic between configs and env variables makes for an inconsistent experience with hidden features. If we use a config param you can still pass it at the command line to override the default:
```
$ heron help submit | grep config-property
  --config-property (key=value; a config key and its value; default: [])
```

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
```diff
@@ -14,16 +14,40 @@
...
+    if (template != null){
+      return new File(template).getPath();
```
billonahill: We should always join the configured path with the file name for consistency.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
chris-pardy: @billonahill I didn't know the --config-property was a thing, thanks. I'll drop the environment config, I actually much prefer the commandline option.

kramasamy: +1 please use --config-property, that is a better approach.

heron/config/src/yaml/conf/aurora/scheduler.yaml
```diff
@@ -8,4 +8,10 @@ heron.class.launcher:                        com.twitter.heron.scheduler.aurora.
...
 # Invoke the IScheduler as a library directly
-heron.scheduler.is.service:                  False
\ No newline at end of file
+heron.scheduler.is.service:                  False
+
+# Override the default Heron Aurora template
+#heron.scheduler.job.template:                heron.aurora
+
+# Override the default directory for the Aurora template
+#heron.scheduler.job.template.dir:            /etc/heron
```
kramasamy: By default, this is not installed in /etc/heron. It is installed in where the cluster config resides. Location of where the config resides is provided as configPath in SubmitterMain - you need to use this configPath. Probably, a better approach would be if the config is not present, fall back to the default approach?

chris-pardy: I wanted to put something in this file to indicate that it could be set to a different, fixed directory. The Aurora Context file already does the fall back if the tag is not defined. I thought that having it here but commented out made sense since it would easy to simply uncomment the line and change the directory to make use of it.

chris-pardy:
@kramasamy @billonahill I believe I've addressed all the PR comments. The CI check seems to be failing because I'm missing JavaDoc on a Unit Test that I modified. I'm guessing this is the result of a mis-configured checkstyle rule. Would you prefer if I edited the rule to ignore tests or added JavaDoc to the test?

heron/config/src/yaml/conf/aurora/scheduler.yaml
kramasamy: sounds good. perhaps, we can make the comment little more verbose 

- in the sense that these two parameters are optional
- Any situations it is used as well

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
```diff
@@ -14,16 +14,26 @@
...
 
   public static String getJobLinkTemplate(Config config) {
     return config.getStringValue(JOB_LINK_TEMPLATE);
   }
+
+  public static String getHeronAuroraPath(Config config) {
+    return new File(config.getStringValue(JOB_TEMPLATE_DIR, Context.heronConf(config)),
+        config.getStringValue(JOB_TEMPLATE,"heron.aurora")).getPath();
+  }
+
```
kramasamy: @chris-pardy - in the case of both the config being absent, how does this work? Is the JOB_TEMPLATE_DIR set to default somewhere?

kramasamy:
@chris-pardy - meta comment. Do we need two configs for the path? A single config will be sufficient, isn't?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
chris-pardy: The second parameter to the config.getStringValue method is the default value, I'm using the same values here that used to be in AuroraScheduler.

chris-pardy:
@kramasamy I think you're right that the 2 configs is overkill, I was mirroring the structure of the code (conf dir + file) but I think it makes more sense just to have one config. I've updated the code + added some documentation, I've also changed the conf files to call out that the line is commented out and what it will do more clearly.

billonahill:
+1 for not requiring javadocs on unit test if you're able to add that inclusion. Best to do so in a separate PR please.

kramasamy:
@chris-pardy - can you please fill in the CLA?  https://engineering.twitter.com/opensource/cla

chris-pardy:
@kramasamy I just filled out the CLA. Is there a way I can get this PR to  build again?

kramasamy:
@chris-pardy - no idea why the ci did not kick in. Try another commit and update the pull request?

kramasamy:
@chris-pardy - some Java style issue, "unused import"

kramasamy:
This seems to be simple fix - 

```
git diff
diff --git a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
index 64ecffb..82db52c 100644
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
@@ -14,7 +14,6 @@
 
 package com.twitter.heron.scheduler.aurora;
 
-import java.io.File;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -34,7 +33,6 @@ import com.twitter.heron.common.basics.FileUtils;
 import com.twitter.heron.proto.scheduler.Scheduler;
 import com.twitter.heron.scheduler.UpdateTopologyManager;
 import com.twitter.heron.spi.common.Config;
-import com.twitter.heron.spi.common.ConfigKeys;
 import com.twitter.heron.spi.common.Context;
 import com.twitter.heron.spi.common.Misc;
 import com.twitter.heron.spi.packing.PackingPlan;
```

But after fixing the checkstyle, it seems to be failing in the unit test

chris-pardy:
@kramasamy not sure what happened to the CI build but it doesn't look related to my PR?

kramasamy:
@chris-pardy - looks good. there were some flaky integration tests. we are taking a look at them.

kramasamy:
@chris-pardy - is this ready to go?

chris-pardy:
Yep

kramasamy:
👍

PR #1432, opened 2016-09-26
Must delete packing plan before updating it

billonahill:
The scheduler updates the packing plan with the scheduled resources. When doing so it must delete the first packing plan, otherwise zookeeper will fail.

cc/ @ashvina 

Relates to #1292

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
@@ -436,14 +436,10 @@ public final class SchedulerUtils {
                                                SchedulerStateManagerAdaptor stateManager) {
     LOG.log(Level.INFO, "Updating scheduled-resource in packing plan: {0}", topologyName);
     PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
-    if (stateManager.getPackingPlan(topologyName) != null) {
-      stateManager.deletePackingPlan(topologyName);
-    }
 
-    boolean result =
-        stateManager.setPackingPlan(serializer.toProto(updatedPackingPlan), topologyName);
-    if (!result) {
-      throw new RuntimeException(String.format("Failed to save %s's packing plan", topologyName));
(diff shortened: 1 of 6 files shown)


heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java
```diff
@@ -436,6 +436,10 @@ public static void persistUpdatedPackingPlan(String topologyName,
                                                SchedulerStateManagerAdaptor stateManager) {
     LOG.log(Level.INFO, "Updating scheduled-resource in packing plan: {0}", topologyName);
     PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
+    if (stateManager.getPackingPlan(topologyName) != null) {
+      stateManager.deletePackingPlan(topologyName);
```
ashvina: There is a test for this util method, `persistUpdatedPackingPlanWillUpdatesStateManager`. Is verification of the delete operation needed?

ashvina: What do you think about reusing this method in `UpdateTopologyManager.updateTopology Line#96`? Or even adding a new `updatePacking` method to `SchedulerStateManagerAdaptor` to avoid confusion?

billonahill: This operation works fine in the local filesystem state manager, which is used in the tests, but not the ZK state manager unfortunately.

billonahill: Yeah, I also considered that. Let me give that a shot.

billonahill: I've added update commands to the state manager adaptor. Also changes `AuroraScheduler` to not update state manager as a side effect of calling `createAuroraProperties`.

ashvina:
LGTM 👍

PR #1433, opened 2016-09-26
Permit the ability to scale down

billonahill:
Before we supported scale down we had a check in place to prevent it. Removing that check.

cc/ @avflor 

Relates to #1292.


ashvina:
👍

PR #1434, opened 2016-09-26
Turn off Javadoc Checks for tests/.*

chris-pardy:
Turning off the Javadoc requirements for Tests (probably since I can't seem to reproduce checkstyle failures locally)

### DIFF
--- a/tools/java/src/com/twitter/bazel/checkstyle/suppressions.xml
+++ b/tools/java/src/com/twitter/bazel/checkstyle/suppressions.xml
@@ -14,5 +14,6 @@
   -->
 
   <suppress files="Empty.java" checks=".*"/>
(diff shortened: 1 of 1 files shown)


tools/java/src/com/twitter/bazel/checkstyle/suppressions.xml
```diff
@@ -14,5 +14,6 @@
   -->
 
   <suppress files="Empty.java" checks=".*"/>
+  <suppress files="tests/.*" checks="JavadocMethod|JavadocType|JavadocStyle" />
```
billonahill: Can we keep `JavadocStyle` in place so if there are javadocs we verify they are valid.

kramasamy:
👍  ship it

PR #1436, opened 2016-09-27
ZkStateManager configuration

chris-pardy:
Changing ZkStateManager and StateManager to accept a set of host, port pairs rather than a single host port. This allows us to specify all the members of a ZK Quorum rather than needing to choose 1 member which is potentially more fragile.

### DIFF
--- a/heron/statemgrs/src/python/statemanager.py
+++ b/heron/statemgrs/src/python/statemanager.py
@@ -90,7 +90,7 @@ class StateManager:
         socket.create_connection(hostport, 2)
         return True
       except:
-        LOG.info("StateManager %s Unable to connect to host: %s port %i" % (self.name,hostport[0], hostport[1]))
+        LOG.info("StateManager %s Unable to connect to host: %s port %i" % (self.name, hostport[0], hostport[1]))
         continue
     return False
 
--- a/heron/statemgrs/src/python/statemanager.py
+++ b/heron/statemgrs/src/python/statemanager.py
@@ -109,6 +109,7 @@ class StateManager:
       self.tunnel.append(subprocess.Popen(
           ('ssh', self.tunnelhost, '-NL127.0.0.1:%d:%s:%d' % (localport, host, port))))
       localportlist.append(('127.0.0.1', localport))
+    print (localportlist)
     return localportlist
 
   def terminate_ssh_tunnel(self):
--- a/heron/statemgrs/src/python/statemanager.py
+++ b/heron/statemgrs/src/python/statemanager.py
@@ -71,14 +71,8 @@ class StateManager:
     """ Setter for the tunnelhost to create the tunnel if host is not accessible """
     self.__tunnelhost = newTunnelHost
 
-  @abc.abstractmethod
-  def __init__(self, hostportlist, rootpath, tunnelhost):
-    """
-    @param hostportlist - Hosts and ports where the states are stored
-    @param rootpath - Path where the heron states are stored
-    @param tunnelhost - Host to which to tunnel through if state host is not directly accessible
-    """
-    pass
+  def __init__(self):
+    self.tunnel = []
 
   def is_host_port_reachable(self):
     """
@@ -90,7 +84,8 @@ class StateManager:
         socket.create_connection(hostport, 2)
         return True
       except:
-        LOG.info("StateManager %s Unable to connect to host: %s port %i" % (self.name, hostport[0], hostport[1]))
+        LOG.info("StateManager %s Unable to connect to host: %s port %i"
+                 % (self.name, hostport[0], hostport[1]))
         continue
     return False
 
@@ -117,9 +112,8 @@ class StateManager:
     return localportlist
 
   def terminate_ssh_tunnel(self):
-    if hasattr(self, 'tunnel') and self.tunnel:
-      for tunnel in self.tunnel:
-        tunnel.terminate()
+    for tunnel in self.tunnel:
+      tunnel.terminate()
 
   @abc.abstractmethod
   def start(self):
--- a/heron/statemgrs/src/python/statemanager.py
+++ b/heron/statemgrs/src/python/statemanager.py
@@ -16,6 +16,8 @@ import abc
 import socket
 import subprocess
 
+from heron.statemgrs.src.python.log import Log as LOG
+
 HERON_EXECUTION_STATE_PREFIX = "{0}/executionstate/"
 HERON_PACKING_PLANS_PREFIX = "{0}/packingplans/"
 HERON_PPLANS_PREFIX = "{0}/pplans/"
@@ -88,6 +90,7 @@ class StateManager:
         socket.create_connection(hostport, 2)
         return True
       except:
+        LOG.info("StateManager %s Unable to connect to host: %s port %i" % (self.name,hostport[0], hostport[1]))
         continue
     return False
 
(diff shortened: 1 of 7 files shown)


heron/statemgrs/src/python/statemanager.py
```diff
@@ -96,11 +83,13 @@ def is_host_port_reachable(self):
...
     """
-    try:
-      socket.create_connection((self.host, self.port), 2)
-      return True
-    except:
-      return False
+    for (host,port) in self.hostportlist:
+      try:
+        socket.create_connection((host, port), 2)
+        return True
```
kramasamy: If you are unable to connect, can you please print a log message for the host/port that you are unable to connect. I know it is not there previously - but it will be nice to have.

heron/statemgrs/src/python/statemanager.py
```diff
@@ -42,24 +42,12 @@ def name(self, newName):
     self.__name = newName
```
kramasamy: you might want to change the example configuration files or given a commented example in

 heron/tools/config/src/yaml/tracker/heron_tracker.yaml

heron/statemgrs/src/python/statemanager.py
```diff
@@ -113,17 +105,21 @@ def pick_unused_port(self):
...
+      localport = self.pick_unused_port()
+      self.tunnel.append(subprocess.Popen(
+          ('ssh', self.tunnelhost, '-NL127.0.0.1:%d:%s:%d' % (localport, host, port))))
+      localportlist.append(('127.0.0.1', localport))
+    return localportlist
 
   def terminate_ssh_tunnel(self):
-    if hasattr(self, 'tunnel') and self.tunnel:
-      self.tunnel.terminate()
+    if hasattr(self, 'tunnel'):
```
billonahill: Does this assure that `self.tunnel != None`?

heron/statemgrs/tests/python/statemanagerfactory_unittest.py
```diff
@@ -0,0 +1,35 @@
...
+  """Unittest for statemanagerfactory"""
+
+  def test_all_zk_supports_comma_separated_hostports(self):
+    """Verify that a comma separated list of host ports is ok"""
+    conf = Config()
+    conf.set_state_locations([{'type':'zookeeper', 'name':'zk', 'hostport':'localhost:2181,localhost:2281',
+                              'rootpath':'/heron', 'tunnelhost':'localhost'}])
+    statemanagers = statemanagerfactory.get_all_zk_state_managers(conf)
+    # 1 state_location should result in 1 state manager
+    self.assertEquals(1,len(statemanagers))
```
billonahill: space after comma

heron/statemgrs/tests/python/zkstatemanager_unittest.py
```diff
@@ -0,0 +1,91 @@
...
+'''ZkStateManager unittest'''
+import unittest2 as unittest
+
+from heron.statemgrs.src.python.zkstatemanager import ZkStateManager
+
+
+class ZkStateManagerTest(unittest.TestCase):
+  """Unittest for ZkStateManager"""
+
+  class NoseFlute:
```
billonahill: This took me a little while to figure out the NoseFlute naming, I'm a little slow. :) Can we call this MockKazoo or something more boring, but more informative?

billonahill:
Does https://github.com/twitter/heron/blob/master/heron/statemgrs/src/python/configloader.py#L22 still work without changes? It's used to provide a state manager to the heron cli.

Also, please test that Heron UI and Heron Tracker works with this change, and provide config upgrade instructions in the summary. We're going to need to update our internal configs which live outside this repo.

chris-pardy:
@billonahill this should be completely backwards compatible unless someone was specifically calling the host or port properties. There was never anything in the config loader that mandated host:port formatting, that was a requirement in the statemanagerfactory. Additionally there should be no config file changes required statemanagerfactory now just correctly handles the case of a string of comma separated values. The only case where behavior would change from before this that I can come up with is if you had a zookeeper statemanager and you were relying on it failing to get configured by passing a set of comma seperated host:ports that's a incredibly weird behavior so I don't think it needs to be called out.

billonahill:
👍 once CI is fixed

kramasamy:
@chris-pardy - checkstyle did not pass in python

chris-pardy:
@kramasamy @billonahill Looking at the CI it looks like my unittests are failing. Running locally I'm seeing the following failure:
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
============================= test session starts ==============================
platform darwin -- Python 2.7.11 -- py-1.4.27 -- pytest-2.6.4
ERROR: file not found: heron/statemgrs/tests/python/zkstatemanager_unittest.py

===============================  in 0.00 seconds ===============================

The cause seems to be the script that is generated to run the unit tests as part of the Pex Test target which is:
PYTHONDONTWRITEBYTECODE=1 /private/var/tmp/_bazel_cpardy/9b99ba7514b33ef96b48f910b2c02c50/execroot/heron/bazel-out/local-fastbuild/bin/heron/statemgrs/tests/python/statemanager_unittest.pex /private/var/tmp/_bazel_cpardy/9b99ba7514b33ef96b48f910b2c02c50/execroot/heron/heron/statemgrs/tests/python/configloader_unittest.py heron/statemgrs/tests/python/zkstatemanager_unittest.py heron/statemgrs/tests/python/statemanagerfactory_unittest.py

you can see that the existing tests (configloader_unittest.py) has generated a path in the bazel build directory while my added unit tests have not. I'm not sure if I can get the logs on the CI server but I'm guessing there's a similar problem. 
If I copy and paste the code above into my terminal and run it the result is passing tests, I also made the tests fail and verified that they do actually run. I dug through the pex rule and it doesn't seem to be doing anything that would treat my new tests differently. I'm guessing at this point that the problem is something with bazel itself. Any thoughts?

kramasamy:
@chris-pardy - let me give it a try with your code in my machine.

kramasamy:
@chris-pardy - I was able to reproduced the error. When I looked at the BUILD file, I realized that you need to use one pex_test macro per test file. Hence I modified the BUILD file into 

```
load("/tools/rules/pex_rules", "pex_test")

pex_test(
    name = "configloader_unittest",
    srcs = [
        "configloader_unittest.py",
    ],
    deps = [
        "//heron/statemgrs/src/python:statemgr-py",
    ],
    reqs = [
        "py==1.4.27",
        "pytest==2.6.4",
        "unittest2==0.5.1",
    ],
    size = "medium",
)

pex_test(
    name = "zkstatemanager_unittest",
    srcs = [
        "zkstatemanager_unittest.py",
    ],
    deps = [
        "//heron/statemgrs/src/python:statemgr-py",
    ],
    reqs = [
        "py==1.4.27",
        "pytest==2.6.4",
        "unittest2==0.5.1",
    ],
    size = "medium",
)

pex_test(
    name = "statemanagerfactory_unittest",
    srcs = [
        "statemanagerfactory_unittest.py",
    ],
    deps = [
        "//heron/statemgrs/src/python:statemgr-py",
    ],
    reqs = [
        "py==1.4.27",
        "pytest==2.6.4",
        "unittest2==0.5.1",
    ],
    size = "medium",
)
```

And all the unit tests pass.

PR #1440, opened 2016-09-27
create serializer with default constructor

nlu90:
As reported in [heron-users](https://groups.google.com/forum/#!topic/heron-users/LLmiQtxXMRM) google group, heron fails to create the serializer with default constructor. This PR fixed the issue.

### DIFF
--- a/heron/storm/src/java/backtype/storm/serialization/SerializationFactory.java
+++ b/heron/storm/src/java/backtype/storm/serialization/SerializationFactory.java
@@ -163,6 +163,12 @@ public final class SerializationFactory {
       // do nothing
     }
 
+    try {
+      return serializerClass.newInstance();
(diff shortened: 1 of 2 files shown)


billonahill:
👍

kramasamy:
nice - thanks for fixing this @nlu90

mycFelix:
@nlu90 - Shall we do the same change for `org.apache.storm.serialization.SerializationFactory` in Heron at [line]  (https://github.com/twitter/heron/blob/master/heron/storm/src/java/org/apache/storm/serialization/SerializationFactory.java#L159) since the namespace was changed after Storm 1.0.x released.

nlu90:
@mycFelix nice point

mycFelix:
👍

kramasamy:
@nlu90 - is this good to go?

PR #1456, opened 2016-09-29
Add support for SSH sock proxying to scheduler

billonahill:
In #1412 we ran into a need for socks proxying from the scheduler to TMaster. This patch fixes that by creating two tunneling approaches: port forwarding (existing behavior required for ZK since ZK client doesn't support proxies) and sock proxy (preferred approach for tunneling HTTP).

The previous tunneling approach was written specifically for ZK. This pulls the tunneling configs out of ZK and into it's own config loader that can be used by state manager, scheduler, etc. See `NetworkUtils` and `ShellUtils` for the bulk of the changes.

Also removed a few occurances of `verbose` flags, which were not used and should be replaces by the logging level controls that now handle that.

Related to #1292.


heron/config/src/yaml/conf/aurora/scheduler.yaml
```diff
@@ -8,4 +8,25 @@ heron.class.launcher:                        com.twitter.heron.scheduler.aurora.
...
+heron.scheduler.tunnel.connection.retry.count:             2
+
+# The interval in ms between two retries when testing direct access to remote host
+heron.scheduler.tunnel.retry.interval.ms:                  1000
+
+# The number of retries when verifying connectivity via tunnel
+heron.scheduler.tunnel.verify.count:                       10
+
+# SSH tunnel host
+heron.scheduler.tunnel.host:                               "my.tunnel.host"
```
maosongfu: Add a new line at the end?

heron/config/src/yaml/conf/aurora/statemgr.yaml
```diff
@@ -10,3 +10,23 @@ heron.statemgr.root.path:                       /vagrant/.herondata/repository/s
...
+heron.statemgr.tunnel.connection.retry.count:             2
+
+# The interval in ms between two retries when testing direct access to remote host
+heron.statemgr.tunnel.retry.interval.ms:                  1000
+
+# The number of retries when verifying connectivity via tunnel
+heron.statemgr.tunnel.verify.count:                       10
+
+# SSH tunnel host
+heron.statemgr.tunnel.host:                               "my.tunnel.host"
```
maosongfu: Add a new line at the end?

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
```diff
@@ -31,23 +33,111 @@
...
+
+    public String getShortName() {
+      return shortName;
+    }
+  }
+
   private NetworkUtils() {
   }
 
+  public static class TunnelConfig {
```
maosongfu: Should we make it separate class or a nested class?

billonahill: I could go either way on this. I made it an inner class since it's tightly couple to, and only intended for, `NetworkUtils` but if we think it should stand on it's own I'm ok with it.

maosongfu:
looks good to me. 👍

PR #1459, opened 2016-09-30
Prevent concurrent scaling events from occuring

### LINKED ISSUE #1353: Prevent concurrent topology parallelism changes
We need a way to assure that multiple scaling events don't collide.

Related to #1292.

billonahill:
Does two things to prevent concurrent scaling events:
1. Takes out a local lock
2. checks the submitted packing plan is same as that in state manager

Fixes #1353 
Relates to #1292


billonahill:
I realized that this approach on it's own is not sufficient. The scheduler can be run as a library, so every agent invoking an update command will have it's own instance. Hence there's no shared locking in that scenario. Please hold on reviewing this.

kramasamy:
@billonahill - can we do simpler one? Wondering if a solution that creates an ephemeral node in ZK with the topology name by the client or whoever is responsible and hold on to it until it is done?

billonahill:
@kramasamy yes I'm working on a ZK impl now. The `UpdateTopologyManager` code will look mostly the same but the `Lock` will be backed by ZK and fetched via `StateManager`.

billonahill:
This patch has been updates with the inclusion of distributed locks in #1463 and is now ready for final review.

ashvina:
LGTM 👍

billonahill:
Failing test will be fixed once #1472 merges in.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -104,7 +139,6 @@ public void updateTopology(final PackingPlans.PackingPlan existingProtoPackingPl
     assertTrue(newContainerCount + removableContainerCount == 0 || scalableScheduler.isPresent(),
```
mycFelix: Hi, @billonahill . I'm reading this part of code. I was wondering if we could use `Preconditions.checkArgument` or `Preconditions.checkState` provided by Google guava instead of assertTrue method?

PR #1463, opened 2016-10-03
Adding distributed locking support

billonahill:
We need a distributed locking mechanism to assure that multiple Heron commands aren't be executed at the same time by different actors. This is required for the update command for scaling, but other commands could also leverage this approach.

Implemented as a new method on `StateManager` to get a `Lock` impl backed by the state manager impl. For ZK the lock is backed by curator, and for local file system the lock is backed by the file system.

Required for #1353 and #1459

### DIFF
--- a/heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
+++ b/heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
@@ -42,8 +42,7 @@ public class LocalFileSystemStateManager extends FileSystemStateManager {
 
   /**
    * Local filesystem implementation of a lock that mimics the file system behavior of the
-   * distributed lock. Only intended for local mode testing, not production. Not guaranteed to be
-   * truly atomic due to limitations in the underlying file system read/write semantics.
+   * distributed lock.
    */
   private final class FileSystemLock implements Lock {
     private String path;
--- a/heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
+++ b/heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
@@ -14,7 +14,9 @@
 
 package com.twitter.heron.statemgr.localfs;
 
+import java.nio.charset.Charset;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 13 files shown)


heron/statemgrs/tests/java/com/twitter/heron/statemgr/zookeeper/curator/CuratorStateManagerTest.java
```diff
@@ -108,7 +108,7 @@ public void testInitialize() throws Exception {
     // Verify curator client is invoked
     Mockito.verify(mockClient).start();
     Mockito.verify(mockClient).blockUntilConnected(Mockito.anyInt(), Mockito.any(TimeUnit.class));
-    Mockito.verify(mockClient, Mockito.times(6)).createContainers(Mockito.anyString());
+    Mockito.verify(mockClient, Mockito.times(7)).createContainers(Mockito.anyString());
```
objmagic: why change 6 to 7 here?

billonahill: This tests the state managers `initTree` method, which creates all `StateLocation`s and we've added a new one with `/locks`.

https://github.com/twitter/heron/blob/master/heron/statemgrs/src/java/com/twitter/heron/statemgr/FileSystemStateManager.java#L42

objmagic:
Is there any test for ZK curator lock?

billonahill:
Yes:
https://github.com/apache/curator/blob/master/curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessSemaphoreMutex.java

Our tests verify via mocks that we call the proper methods on the lock. Curator tests verify that the lock impl works.

heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
```diff
@@ -31,12 +33,50 @@
...
+   * truly atomic due to limitations in the underlying file system read/write semantics.
+   */
+  private final class FileSystemLock implements Lock {
+    private String path;
+
+    private FileSystemLock(String path) {
+      this.path = path;
+    }
+
+    @Override
```
kramasamy: for local file system, there is a file system call that you can use open system call with OCREAT - that automatically creates if a local file does not exist or fails if the file exists. It might be better solution, I think

kramasamy:
Gave one feedback - other than that it looks good 👍

heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java
billonahill: The main use case for locking is the distributed one, where multiple users are trying to make concurrent changes. Since the local file system is only used for compatibility during development by an individual user, do you think it's worth it to invest in a robust local implementation?

billonahill: I've updated the file creation to be atomic. The `Files` API supports this:
http://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#CREATE_NEW

Issue #1410, opened 2016-09-21
Failed to display topology config informations on heron-ui


mycFelix:
Hi,guys

I'm using 0.14.3 version. I found heron-ui failed to display topology config informations on `http://${ip}:${port}/topologies/${cluster}/${environ}/${topology}/config` page. I'm running example topology on my local environment. The submission command is following

`heron submit local ~/.heron/examples/heron-examples.jar com.twitter.heron.examples.AckingTopology AckingTopology`

The Chrome's console log is following

````
url: /topologies/local/default/ExclamationTopology/physicalplan.json  
Uncaught SyntaxError: Unexpected token A in JSON at position 0
   render                     @       config.js:100
   (anonymous function)       @       react.js:6029
````

I thought there might something wrong at [line](https://github.com/twitter/heron/blob/master/heron/tools/ui/resources/static/js/config.js#L100). I printed logs in my local `config.js`. It seems `configValue.value` is `Java object`, which couldn't be parsed by JSON. But it works in 0.14.2 version.

Here is a screenshot
![image](https://cloud.githubusercontent.com/assets/4719414/18773354/f4621a5e-8180-11e6-8e41-3184b00d1c53.png)

objmagic:
Thx for reporting this issue. I will look into it.

maosongfu:
Solved at #1464

PR #1464, opened 2016-10-04
Add back the "import StringIO" to javaobj.py

maosongfu:
Earlier, the import of StringIO was removed mistakenly, causing heron-tracker
failing to parse bytes into java object. Then the heron-ui would not be able to
fetch the correct topology_config value, and thus the config could not be displayed
normally.

This patch fixes the issue #1410.


kramasamy:
👍

objmagic:
👍

PR #1469, opened 2016-10-04
Remove usage of NullFoo objects

billonahill:
For all `Foo` interfaces that use reflection, we implement a `NullFoo` object with an empty implementation so we can test classes that rely on reflection to load `Foo`s. Refactoring to use a more idiomatic approach using mock objects instead. This has the following benefits:

1. We don't need to maintain dummy classes. This is tedious when adding methods and refactoring and it clutters the code base
2. We don't need to maintain build targets and dependency chains for the dummy classes.
3. We can customize the expected method call behavior of the mock objects based on the context of their use, instead of relying on hard-coded implementations.

This pr only removes usage of the `NullFoo` objects. Removing the actual classes can happen in a follow-up.

### DIFF
--- a/heron/config/src/yaml/conf/yarn/uploader.yaml
+++ b/heron/config/src/yaml/conf/yarn/uploader.yaml
@@ -1,5 +1,5 @@
 # uploader class for transferring the topology jar/tar files to storage
-heron.class.uploader:                            com.twitter.heron.uploader.NullUploader
+heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
 
 # name of the directory to upload topologies for local file system uploader
 #heron.uploader.localfs.file.system.directory:    ${HOME}/.herondata/repository/topologies/${CLUSTER}/${ROLE}/${TOPOLOGY} 
--- a/heron/scheduler-core/tests/java/com/twitter/heron/scheduler/RuntimeManagerMainTest.java
+++ b/heron/scheduler-core/tests/java/com/twitter/heron/scheduler/RuntimeManagerMainTest.java
@@ -16,16 +16,22 @@ package com.twitter.heron.scheduler;
 
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
 
 import com.twitter.heron.proto.system.ExecutionEnvironment;
 import com.twitter.heron.scheduler.client.ISchedulerClient;
 import com.twitter.heron.spi.common.Command;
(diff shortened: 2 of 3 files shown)


heron/config/src/yaml/conf/yarn/uploader.yaml
```diff
@@ -1,5 +1,5 @@
 # uploader class for transferring the topology jar/tar files to storage
-heron.class.uploader:                            com.twitter.heron.uploader.NullUploader
+heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
```
ashvina: The YARN scheduler does not need to upload heron core, A use case for a `no-op` uploader. Replacing it with this uploader will result in unnecessary copies of the binaries. Will it be ok to let the `NullUploader` exist?

billonahill: I changed this because all other uploaders use the local fs uploader, which just moves it to another directory, so I thought that made sense for yarn as well. Is the yarn workflow different than all the others with respect to it's uploader needs?
```
$ grep -r heron.class.uploader heron/config/.
heron/config/./src/yaml/conf/aurora/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/local/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/localzk/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/marathon/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/mesos/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/slurm/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.localfs.LocalFileSystemUploader
heron/config/./src/yaml/conf/yarn/uploader.yaml:heron.class.uploader:                            com.twitter.heron.uploader.NullUploader
```

kramasamy: which means it will copy to another directory - which incurs overhead and also a configuration of directory where the file needs to be moved as well. Probably, it might be a good idea to have NullUploader.

ashvina: @billonahill Yes, the YARN scheduler skips the uploader step. The heron binaries are added to **classpath** by the launcher in this case.

billonahill: I've changed back to `NullUploader` for now, but the proper fix is really to change the yarn impl (or the shared classes it relies on) to not invoke an uploader at all. This is a cleaner approach than calling a class written to do nothing.

PR #1470, opened 2016-10-04
Using static imports for Mockito and Assert in scheduler-core tests

billonahill:
This is just a clean up PR, no logic changes.

Using static imports for `Mockito` and `Assert` reduces the verbosity of our unit tests and makes them much more readable.


ashvina:
LGTM 👍

PR #1471, opened 2016-10-05
Repack implementation for the Resource Compliant RR algorithm

### LINKED ISSUE #1439: Implement repack in Round Robin packing
It will be used during scale up/down to generate a packing plan using the RoundRobin algorithm.

avflor:
Fixes #1439


avflor:
@billonahill Can you please take a look at this PR? It contains the repacking implementation of RR.

heron/packing/src/java/com/twitter/heron/packing/Container.java
```diff
@@ -89,6 +101,7 @@ private Resource getTotalUsedResources() {
    * @return true if the instance can be added to the container, false otherwise
    */
   public boolean add(PackingPlan.InstancePlan instancePlan) {
+   // System.out.println(this.capacity.toString());
```
billonahill: pls remove.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+    }
+    return new Resource(cpu, ram, disk);
+  }
+
+  /**
+   * Computes the additional resources needed to accommodate a scale up and down operation
+   *
+   * @return Additional resources needed
+   */
+  public static Resource getAdditionalResources(Resource scaleupResource,
```
billonahill: let's camelCase consistently with elsewhere (e.g., `getScaleUpResource`) by using `scaleUpResource` and `scaleUpResource` here and elsewhere in this patch.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+                                                Resource scaledownResource) {
+    double cpuDifference =  scaleupResource.getCpu() - scaledownResource.getCpu();
+    double cpu = cpuDifference < 0 ? 0 : cpuDifference;
```
billonahill: better to use `Math.max(0, cpuDifference)` here and elsewhere

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+      } else {
+        ram += parallelismChange * defaultInstanceResources.getRam();
+      }
+    }
+    return new Resource(cpu, ram, disk);
+  }
+
+  /**
+   * Computes the additional resources needed to accommodate a scale up and down operation
+   *
```
billonahill: pls clarify what is expected by a scaleupResource and a scaledownResource? For example should one be larger than the other? What do they represent?

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+    return new Resource(cpu, ram, disk);
+  }
+
+
+  /**
+   * Computes number of containers needed to accommodate a specific resource request given the container size
+   *
+   * @return Number of containers needed
+   */
+  public static double getRequiredNumContainers(Resource resourceRequest, Resource capacity) {
```
billonahill: This could also be a method on `Resource`:
```
Resource divideBy(Resource other);
```

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+    return Math.max(numContainersCpu, Math.max(numContainersRam, numContainersDisk));
+  }
+
+  /**
+   * Removes containers from tha allocation that do not contain any instances
+   */
+  public static void removeEmptyContainers(Map<Integer, List<InstanceId>> allocation) {
+    Iterator<Integer> containerIds = allocation.keySet().iterator();
+    while (containerIds.hasNext()) {
+      Integer containerId = containerIds.next();
```
billonahill: Typically we want to iterate directly on the keySet here, which is an `Iterable<Integer>`. There first 3 lines can be replaced with:
```
for (Integer containerId : allocation.keySet()) {
```

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -153,7 +163,45 @@ public PackingPlan pack() {
...
+  /**
+   * Get a new packing plan given an existing packing plan and component-level changes.
+   *
+   * @return new packing plan
+   */
+  public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> componentChanges) {
+    int adjustments = 0;
+    this.numAdjustments = 0;
+    this.numContainers = currentPackingPlan.getContainers().size();
+    this.containerId = 1;
```
billonahill: What does `this.containerId` represent and why is it 1 when starting a repack?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -153,7 +163,45 @@ public PackingPlan pack() {
...
+    int additionalContainers = computeNumAdditionalContainers(componentChanges, currentPackingPlan);
+    increaseNumContainers(additionalContainers);
+    LOG.info(String.format("Allocated "
+            + "%s additional containers for repack bring the number of containers to %s.",
+        additionalContainers, this.numContainers));
+    // Get the instances using Resource Compliant Round Robin allocation
+    Optional<Map<Integer, List<InstanceId>>> resourceCompliantRRAllocation =
+        getResourceCompliantRRAllocation(currentPackingPlan, componentChanges);
+
+    while (!resourceCompliantRRAllocation.isPresent()) {
```
billonahill: can you explain in comments what's happening here? it seems like we just keep calling `getResourceCompliantRRAllocation` until it returns something, but nothing changes in the values we pass to it so it's unclear why we should expect a different result each time.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+    }
+    return new Resource(cpu, ram, disk);
+  }
+
+  /**
+   * Identifies the resources needed for the components that will be scaled up
+   *
+   * @return Total resources needed
+   */
+  public static Resource getScaleUpResource(TopologyAPI.Topology topology, Map<String, Integer>
```
billonahill: There's a lot of duplicate logic in `getScale[Up|Down]Resource` and `getComponentsToScale[Up|Down]` that could be replaced by refactoring into generic methods that take an enum that represents whether we're scaling up or down:
```
  enum ScalingDirection {
    UP,
    DOWN;
    
    boolean includes(int parallelismChange) {
      switch (this) {
        case UP:
          return parallelismChange > 0;
        case DOWN:
          return parallelismChange < 0;
        default:
          throw new IllegalArgumentException("...");
      }
    }
  }
```
Also, I think we could name this method more clearly. Maybe something like this maybe?
```
Resource computeTotalResourceChange(
  TopologyAPI.Topology topology,
  Map<String, Integer> componentChanges,
  Resource defaultInstanceResources,
  ScalingDirection scalingDirection);
```

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -198,65 +245,264 @@ public void close() {
...
-                instanceResource))) {
-          allocation.get(containerId).add(new InstanceId(component, globalTaskIndex, i));
-        } else {
-          //Automatically adjust the number of containers
-          increaseNumContainers();
-          LOG.info(String.format("Increasing the number of containers to "
-              + "%s and attempting packing again.", this.numContainers));
-          return null;
+        boolean sufficientNumContainers = true;
+        if ("strict".equals(policyType)) {
```
billonahill: better to represent PolicyType as an enum with a method that encapsulated this logic.

heron/packing/tests/java/com/twitter/heron/packing/PackingUtilsTest.java
```diff
@@ -37,6 +54,36 @@
...
+
+    for (int containerId : basePacking.keySet()) {
+      List<InstanceId> instanceList = basePacking.get(containerId);
+
+      Set<PackingPlan.InstancePlan> instancePlans = new HashSet<>();
+
+      for (InstanceId instanceId : instanceList) {
+        String componentName = instanceId.getComponentName();
+        Resource instanceResource;
+        if ("bolt".equals(componentName)) {
```
billonahill: can we `switch` on componentName and have an explicit case for "bolt" and "spout" and fail on default?

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
```diff
@@ -219,4 +222,142 @@ public static int allocateNewContainer(ArrayList<Container> containers, Resource
...
+        ram += parallelismChange * ramMap.get(component);
+      } else {
+        ram += parallelismChange * defaultInstanceResources.getRam();
+      }
+    }
+    return new Resource(cpu, ram, disk);
+  }
+
+  /**
+   * Computes the additional resources needed to accommodate a scale up and down operation
```
avflor: @billonahill This is not a proper subtract. It nevers returns a negative value. It is optimized for the packing algo context and should not be used in other settings

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
avflor: @billonahill same as before

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
billonahill: We could change the name to imply the operation (e.g. `subtractAbsolute`?) or just document that subtract will never go negative. Since it's a mathematical operation between two `Resource` items it seems fitting to have this as a method on `Resource` IMO. It's more OO than static helpers.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
billonahill: @avflor see my other comment re mathematical operations on the object that does them and OO vs helpers.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -153,7 +163,45 @@ public PackingPlan pack() {
...
+   * @return new packing plan
+   */
+  public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> componentChanges) {
+    int adjustments = 0;
+    this.numAdjustments = 0;
+    this.numContainers = currentPackingPlan.getContainers().size();
+    this.containerId = 1;
+
+    int additionalContainers = computeNumAdditionalContainers(componentChanges, currentPackingPlan);
+    increaseNumContainers(additionalContainers);
```
billonahill: `increaseNumContainers` takes a delta to increase by not an absolute, so I was proposing two methods:
```
increaseNumContainers(delta)
resetNumContainers(absolute)
```
it's just about being consistent with our use of encapsulation. When we want to add containers we call a method to increase, which handles the operation. When we want to reset to a value though, we instead manipulate the values directly in multiple places. we'd have better encapsulation and consistency if we did both operations through methods that handles modifying of the underlying values.

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
billonahill:
```
for (Integer containerId : allocation.keySet()) {
  if (allocation.get(containerId).isEmpty()) {
    allocation.remove(containerId);
  }
}
```

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
avflor: @billonahill In the algo I only need the increaseNumContainers method. There is no place in the code that does a reset. Even if I had to use reset, I would still compute the additionalContainers and then add them to the current containers size and then invoke reset. Am I missing sth here?

heron/packing/src/java/com/twitter/heron/packing/PackingUtils.java
billonahill: Ugh, that's right, we need to use the iterator.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
billonahill: Sorry, I missed that `increaseNumContainers` doesn't change `numAdjustments` and `containerId`, only `numContainers` so I was confusing things by lumping that method in with the reset I was suggesting. I'm suggesting a method that resets the following, which happen in multiple places:
```
this.numAdjustments = 0;
this.containerId = 1;
```

avflor:
@billonahill I think I addressed all your comments. Can you please take a look?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -198,65 +255,297 @@ public void close() {
...
+        }
+      }
+      throw new RuntimeException("Cannot remove instance."
+          + " No more instances of component " + component + " exist"
+          + " in the containers.");
+    }
+  }
+
+  private enum PolicyType {
+    strictPolicy, flexiblePolicy;
```
billonahill: For consistency please use capital snakcase for enum values: `STRICT, FLEXIBLE`

billonahill:
👍 once tests pass, thanks!

avflor:
@billonahill @kramasamy Should we merge this one?

PR #1472, opened 2016-10-05
Change location of local lock to be consistent depth with other state objects

billonahill:
This is required to fix tests in #1459.

Instead of storing local FS locks at:
```
/state/local/locks/topologyName/lockName
```
we will store them at:
```
/state/local/locks/topologyName__lockName
```

### DIFF
--- a/heron/statemgrs/src/java/com/twitter/heron/statemgr/FileSystemStateManager.java
+++ b/heron/statemgrs/src/java/com/twitter/heron/statemgr/FileSystemStateManager.java
@@ -69,7 +69,7 @@ public abstract class FileSystemStateManager implements IStateManager {
     }
 
     public String getNodePath(String root, String topology, String extraToken) {
-      return concatPath(getNodePath(root, topology), extraToken);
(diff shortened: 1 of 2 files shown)


kramasamy:
@billonahill - what is the advantage of the new location vs old location?

billonahill:
The advantage is consistency of depth. The locks have an extra node for `<lockName>` in the path so without this change all state paths except lock follow this pattern:
```
/state/local/pplan/<topologyName>
```
where `<topologyName>` is the data file, but locks were initially implemented like this:

```
/state/local/locks/<topologyName>/<lockName>
```
where `<lockName>` was the file and `<topologyName>` a dir. This became a special case due to the depth, where `/state/local/locks/<topologyName>` doesn't exist, causing an exception.

By doing this instead, all files are at the same depth in the tree, which makes for more consistent code paths:
```
/state/local/locks/<topologyName>__<lockName>
```

kramasamy:
@billonahill - understood.

PR #1473, opened 2016-10-05
Make metrics mgr fail fast when unexpected errors happen.

maosongfu:
Make stream mgr fail fast when unexpected errors happen to avoid dangling process.

This is a bug reported by Twitter internally.

### DIFF
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
@@ -97,6 +97,8 @@ public class MetricsManager {
   private final String topologyName;
   private final String metricsmgrId;
 
+  private final long mainThreadId;
+
   /**
    * Metrics manager constructor
    */
@@ -122,6 +124,8 @@ public class MetricsManager {
             this.metricsManagerServerLoop);
     this.metricsCollector = new MetricsCollector(metricsManagerServerLoop, metricsQueue);
     this.heronMetricsExportIntervalSec = systemConfig.getHeronMetricsExportIntervalSec();
+
+    this.mainThreadId = Thread.currentThread().getId();
 
     // Set up the internal Metrics Export routine
     setupInternalMetricsExport();
@@ -192,8 +196,8 @@ public class MetricsManager {
     if (args.length != 6) {
       throw new RuntimeException(
           "Invalid arguments; Usage: java com.twitter.heron.metricsmgr.MetricsManager "
-          + "<id> <port> <topname> <topid> <heron_internals_config_filename> "
-          + "<metrics_sinks_config_filename>");
+              + "<id> <port> <topname> <topid> <heron_internals_config_filename> "
+              + "<metrics_sinks_config_filename>");
     }
 
     String metricsmgrId = args[0];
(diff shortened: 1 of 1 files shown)


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
```diff
@@ -328,24 +332,32 @@ public void start() {
...
-      String threadName = thread.getName();
       LOG.log(Level.SEVERE,
-          "Exception caught in thread: " + threadName + " with thread id: " + thread.getId(),
+          "Exception caught in thread: " + thread.getName() + " with thread id: " + thread.getId(),
           exception);
 
-      String sinkId = threadName;
-      Integer thisSinkRetryAttempts = sinksRetryAttempts.remove(sinkId);
+      if (exception instanceof Error || thread.getId() == mainThreadId) {
+        LOG.severe("Would not recover from error. Metrics Manager halts immediately");
```
billonahill: Instead of logging the exception separately above, can we have an if/then with a different log message in each. That way it's clear what's causing what.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
```diff
@@ -328,24 +332,32 @@ public void start() {
...
+        Runtime.getRuntime().halt(1);
+      }
+
+      String sinkId = thread.getName();
```
billonahill: sinceThreadName? it's not the thread id, it's the name.

kramasamy:
@maosongfu - the title says that it is the stream manager instead of metrics manager? Can you please correct?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
maosongfu: In current implementation, we enforce the thread-name the same as sink-id. Do you think we need to add an extra map to record it?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
maosongfu: Sorry but I don't understand. This implementation logs exception at the start once; and then followings are handling logic with a different if/then ... Those are not causes but handling logics?

billonahill: My suggestion was instead of doing this:
```
- log all the details of the exception
if it's the main thread
  - log that we have to die (due to the previously logged exception)
```

to the log reader it will be more clear if we consolidate those two:
```
if it's the main thread
  - log all the details of the exception and that we have to die
else
  - log all the details of the exception
```

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
billonahill: I wasn't suggesting an extra map, just consistency in variable naming. Above we assign `getId()` to a variable named `mainThreadId`:
```
mainThreadId = Thread.currentThread().getId();
```
bit here we assign `getName()` to a variable named `sinkId` (instead of `sinkName`):
```
sinkId = thread.getName();
```

this is a minor nit though and can be disregarded if you think `sinkId` is more representative bases on it's usage.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
maosongfu: Sounds good. Will do that.

kramasamy:
@maosongfu - is this completed?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
maosongfu: Added comments in the source code to avoid confusion.

maosongfu:
@kramasamy Updated.

billonahill:
👍

PR #1475, opened 2016-10-05
Allow server to discard large packets

maosongfu:
- Allow server to discard large packets. Integrated tested locally.

Sometimes client can send large packet to server, and destroy the server. For instance, a client can send a very large MetricsMessage (due to it contains huge amount of exception stack trace), and the MetricsManager will throw OutOfMemoryError and crash when trying to deserialize the byte[].

This pull request allows server to discard large packets.

- It is a bug reported by Twitter internally.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
+++ b/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
@@ -311,6 +311,13 @@ public class SystemConfig {
   public static final String METRICSMGR_NETWORK_OPTIONS_SOCKET_SEND_BUFFER_SIZE_BYTES
       = "heron.metricsmgr.network.options.socket.send.buffer.size.bytes";
 
+  /**
+   * The maximum packet size to handle by metrics mgr in bytes
+   */
+  public static final String METRICSMGR_SERVER_MAX_PACKET_SIZE_BYTES
+      = "heron.metricsmgr.server.max.packet.size.bytes";
+
+
   private Map<String, Object> config = new HashMap<>();
 
   public SystemConfig() {
@@ -538,6 +545,12 @@ public class SystemConfig {
     return TypeUtils.
         getInteger(this.config.get(
             SystemConfig.METRICSMGR_NETWORK_OPTIONS_SOCKET_SEND_BUFFER_SIZE_BYTES));
+  }
+
+  public int getMetricsmgrServerMaxPacketSizeBytes() {
+    Object value = this.config.get(
+        SystemConfig.METRICSMGR_SERVER_MAX_PACKET_SIZE_BYTES);
+    return value == null ? Integer.MAX_VALUE : TypeUtils.getInteger(value);
   }
 
   public Object put(String key, Object value) {
--- a/heron/common/src/java/com/twitter/heron/common/network/HeronServer.java
+++ b/heron/common/src/java/com/twitter/heron/common/network/HeronServer.java
@@ -46,6 +46,7 @@ public abstract class HeronServer implements ISelectHandler {
   // Define the address where we need to listen on
   private InetSocketAddress endpoint;
   private HeronSocketOptions socketOptions;
+  private int maxPacketSize;
   // Our own looper
   private NIOLooper nioLooper;
   // All the clients that we have connected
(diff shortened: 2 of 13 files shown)


kramasamy:
@maosongfu - what issue does that resolve?

maosongfu:
@kramasamy 
Added in the description.

kramasamy:
@maosongfu - instead of rejecting the large packet, can we gracefully consume it? In the PR, you are reading the packet size and ignoring it. But it might not work in a TCP connection since you have to consume the entire packet before discarding it, correct?

maosongfu:
@kramasamy In fact, here by saying "reject" I mean "discard". It works in TCP connnection (in fact we tested in TCP connection). It still consumes the entire packet in byte array but not to parse (deserialize) them. On the other hand, the client later would also limit the size of message to send.

Besides that, do you have any recommandation how to gracefully consume it? In production, the MetricsManager server can throw exceptoin (OOM Error) when trying to parse it. Or should we try to discard the data byte by byte when reading it from socket?

kramasamy:
@maosongfu - if you look at the problem more deeper, it is initiated because user chose to define a metrics that is pretty large. When you collect the metrics and it is beyond a certain size, we should reject right there. This will force the user rethink the metrics and how it needs to be structured into smaller chunks. Thoughts?

maosongfu:
@kramasamy 
That is a good point; in fact discussed with Bill about this yesterday. 
The key is: server could not rely on/expect on client side's behaviors too much; that's not safe. Similarly in practice, the security check is always required on server side, no matter whether client side has the security check or not. For heron, we could not prevent people writing a client sending large packets (well, in fact we can prevent this by careful code reviews and asking people to take their own risks if to use clients not reviewed by us....).

So the plan is 2-step:
1. Allow server to dicard packets
2. Allow client to discard packets in future pull requests

Thoughts?

kramasamy:
@maosongfu - it is ok to safeguard the server in general -  when some large packets are thrown in. Do all servers need protection like this from various Heron components perspective? If we introduce this, then you need to look at the batch size and number of tuples send from stmgr to Heron instance and vice versa and the config becomes complex.

If this is a purely a metrics manager issue, my suggestion would be to restrict the size of the metriics at the point of receiving it - probably in the api call itself?

congwang:
I think you need to handle the OOM exception from unpackMessage() instead of checking the bytes, it needs to be handled anyway regardless to size, in theory OOM could happen even for allocating 1 byte.

Also, why 131072? The max IPv4 packet size is 65536 (needs GRO).

maosongfu:
@kramasamy 

> Do all servers need protection like this from various Heron components perspective? Do all servers need protection like this from various Heron components perspective? If we introduce this, then you need to look at the batch size and number of tuples send from stmgr to Heron instance and vice versa and the config becomes complex.

In this pull request's implementation, yes all java servers will have this protection. But by default, the value is Integer.MAX_VALUE; people can use the default value by not passing this config.


> If this is a purely a metrics manager issue, my suggestion would be to restrict the size of the metriics at the point of receiving it - probably in the api call itself?

This issue sourced from Metrics Manager, since for data tuple, we have already done size limitation, for instance, # of tuples in a batch, max size in bytes for a batch, and so on. So if we do the control on client's side solely, then we need to update all clients; or all metrics mgr clients, (java, python, ...).
Thoughts?

maosongfu:
> I think you need to handle the OOM exception from unpackMessage()

On the one hand, OOM is an error (not a checked exception) should be bubbled up rather than be caught and handled. On the other hand, this pull request is only trying to resolve the crashes caused by large packets. In the case you mentioned " it needs to be handled anyway regardless to size, in theory OOM could happen even for allocating 1 byte.", the appropriate solution I believe should be to increase the memory for JVM.

> Also, why 131072? The max IPv4 packet size is 65536 (needs GRO).
Just consider 128k is a decent value. Sorry, for packet I mean, the IncomingPacket object defined by us, rather than the IPv4 packet. I could modify the name if needed.

kramasamy:
@maosongfu - if all the java servers have this protection, we might be restricting everybody just for one component's behavior, is that correct?

kramasamy:
For simplicity, let us restrict the size of the metrics received - then we can tackle the problem of server protection differently or in different PR, thoughts?

maosongfu:
sounds good.

kramasamy:
@maosongfu - do we need this PR or we can close it?

maosongfu:
Let's close it.

PR #1477, opened 2016-10-06
Do not let external processes block on IO buffers in ShellUtils.runSyncProcess

windie:
In ShellUtils.runSyncProcess, "output from the external process can exhaust the available buffer for the output or error stream because neither stream is read while waiting for the process to complete. If either buffer becomes full, it can block the external process as well, preventing all progress for both the external process and the Java program." [1]

This PR just launches threads to read outputs asynchronously to fix this issue. The added unit tests can reproduce this bug.

[1] https://www.securecoding.cert.org/confluence/display/java/FIO07-J.+Do+not+let+external+processes+block+on+IO+buffers

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
@@ -94,7 +94,7 @@ public final class ShellUtils {
           try {
             input.close();
           } catch (IOException e) {
-            LOG.log(Level.SEVERE, "Failed to close the input stream", e);
+            LOG.log(Level.WARNING, "Failed to close the input stream", e);
           }
         }
       }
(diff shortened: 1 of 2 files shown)


kramasamy:
thanks @windie for the PR - @objmagic / @maosongfu, can you please check this?

objmagic:
LGTM 👍

windie:
Updated.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -82,13 +82,36 @@ public static int runSyncProcess(
...
+    Thread thread = new Thread() {
+      @Override
+      public void run() {
+        try {
+          out.append(inputstreamToString(input));
+        } finally {
+          try {
+            input.close();
+          } catch (IOException e) {
+            LOG.log(Level.SEVERE, "Failed to close the input stream", e);
```
billonahill: since we're logging this and moving on without throwing, we should change this to Level.WARN actually.

windie: > since we're logging this and moving on without throwing, we should change this to Level.WARN actually.

Fixed.

billonahill:
👍

PR #1478, opened 2016-10-06
Only process update requests if there is a parallelism change

### LINKED ISSUE #1455: Update topology should validate that requested parallelism is not what exists
Currently if parallelism = X and the user requests to change parallelism to X, we go through a scalding event. We should instead return messaging that the parallelism is already set to X and halt.

billonahill:
Fixes #1455 by verifying that the update request represents a change in parallelism.


kramasamy:
👍

PR #1480, opened 2016-10-07
Avoid duplicated exception wrapper during reportError()

maosongfu:
Earlier during reportError(Throwable error), heron would wrap the error into another Exception.

This is bad, since:
1. It introduces extra stack trace which is useless and makes debug harder.
2. It hides the actual exception point into the detailed stack trace, and
replace the exception message (exception.toString()) with the wrapped one,
which makes debug harder.
3. It also introduces extra cost to report and transfer this error.


billonahill:
👍

PR #1481, opened 2016-10-07
Limit the count of exceptions inside a MetricPublisherPublishMessage

maosongfu:
#1475

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
+++ b/heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
@@ -59,6 +59,12 @@ public class SystemConfig {
       = "heron.metrics.export.interval.sec";
 
   /**
+   * The maximum number of exceptions in one MetricPublisherPublishMessage protobuf
+   */
+  public static final String HERON_METRICS_MAX_EXCEPTIONS_COUNT
+      = "heron.metrics.max.exceptions.count";
+
+  /**
    * The queue capacity (num of items) in bolt for buffer packets to read from stream manager
    */
   public static final String INSTANCE_INTERNAL_BOLT_READ_QUEUE_CAPACITY
@@ -538,6 +544,12 @@ public class SystemConfig {
     return TypeUtils.
         getInteger(this.config.get(
             SystemConfig.METRICSMGR_NETWORK_OPTIONS_SOCKET_SEND_BUFFER_SIZE_BYTES));
+  }
+
+  public int getHeronMetricsMaxExceptionsCount() {
+    Object value = this.config.get(
+        SystemConfig.HERON_METRICS_MAX_EXCEPTIONS_COUNT);
+    return value == null ? Integer.MAX_VALUE : TypeUtils.getInteger(value);
   }
 
   public Object put(String key, Object value) {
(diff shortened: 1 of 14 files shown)


heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
```diff
@@ -540,6 +546,12 @@ public int getMetricsMgrNetworkOptionsSocketSendBufferSizeBytes() {
             SystemConfig.METRICSMGR_NETWORK_OPTIONS_SOCKET_SEND_BUFFER_SIZE_BYTES));
   }
 
+  public int getHeronMetricsMaxExceptionsCount() {
+    Object value = this.config.get(
+        SystemConfig.HERON_METRICS_MAX_EXCEPTIONS_COUNT);
+    return value == null ? Integer.MAX_VALUE : TypeUtils.getInteger(value);
```
billonahill: This pattern is repeated in this class. Could we create a methods like this in this class to simplify this:
```
public static int getValue(Map<String, Object> config, String key, Integer default);
```
might be better to do in another pr.

heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
```diff
@@ -59,6 +59,12 @@
       = "heron.metrics.export.interval.sec";
 
   /**
+   * The maximum number of exceptions in one MetricPublisherPublishMessage protobuf
+   */
+  public static final String HERON_METRICS_MAX_EXCEPTIONS_COUNT
+      = "heron.metrics.max.exceptions.count";
```
billonahill: This name on it's own doesn't clearly convey it's meaning. Can we change it to convey that it's the max exceptions in a protobuf? Something like `heron.metrics.max.exceptions.per.message.count` or something else more clear.

heron/common/src/java/com/twitter/heron/common/utils/logging/ErrorReportLoggingHandler.java
```diff
@@ -75,13 +85,20 @@ public void publish(LogRecord record) {
     // Convert Log
     Throwable throwable = record.getThrown();
     if (throwable != null) {
-      StringWriter sink = new StringWriter();
-      throwable.printStackTrace(new PrintWriter(sink, true));
-      String trace = sink.toString();
       synchronized (ExceptionRepositoryAsMetrics.INSTANCE) {
+        // We would not include the message if already exceeded the exceptions limit
+        if (ExceptionRepositoryAsMetrics.INSTANCE.getExceptionsCount() >= exceptionsLimit) {
```
billonahill: What's the scope of this counter? Does this mean that no more exceptions will ever be published, or just for this payload? Same for `droppedExceptionsCount`. Is that count per message payload, count since the jvm started, or what?

maosongfu: Both this counter and `droppedExceptionsCount ` are per message payload. 

Every time we collect the metrics, we would invoke IMetric.getAndReset(), send the message out and these two counter will reset to 0.

heron/common/src/java/com/twitter/heron/common/config/SystemConfig.java
maosongfu: will do it in another pr

billonahill:
👍

PR #1482, opened 2016-10-07
Add try and catch block when handling exceptions

maosongfu:
It prevent new exceptions stop the handling thread, which could potentially lead to
dangling process. For instance, it can lead to the death of MetricsManagerServer thread,
while the MetricsManager process keeps alive and doing nothing.

### DIFF
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
@@ -332,87 +332,96 @@ public class MetricsManager {
      * Handler for uncaughtException
      */
     public void uncaughtException(Thread thread, Throwable exception) {
-      // We would fail fast when errors occur
-      if (exception instanceof Error) {
-        LOG.log(Level.SEVERE,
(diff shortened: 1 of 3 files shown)


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/MetricsManager.java
```diff
@@ -332,87 +332,96 @@ public void start() {
...
-      // We would fail fast when errors occur
-      if (exception instanceof Error) {
-        LOG.log(Level.SEVERE,
-            "Error caught in thread: " + thread.getName()
-                + " with thread id: " + thread.getId() + ". Process halting...",
-            exception);
-        Runtime.getRuntime().halt(1);
-      }
+      // Add try and catch block to prevent new exceptions stop the handling thread
+      try {
```
billonahill: Generally try blocks should be small and this one is pretty large. Could you move the logic inside the block into a private method and just delegate to that inside the try?

maosongfu: sounds good. Will do it.

billonahill:
👍

PR #1483, opened 2016-10-08
Return valid json when converting topology config in tracker's 

maosongfu:
When getting config of a topology, tracker should always return valid json object, which is expected by the ui page. Otherwise, the ui page for topology config would crash.


kramasamy:
Ship it

PR #1486, opened 2016-10-11
count bolt should ack

objmagic:
According to https://github.com/twitter/heron/commit/c2c3725f4093b95e4fe560be848f9891d8653c66#diff-01591666ae24a19c3ebc913806717cc9R36, count bolt should ack tuple


kramasamy:
👍  - can you check if this is similar to Java?

PR #1487, opened 2016-10-11
Simplify how to run aurora int tests locally

billonahill:
Adds some features to the aurora integration tests:

- ability to specify a different cli config path
- ability to only run a subset of tests


objmagic:
👍

PR #1488, opened 2016-10-12
Refactor integration test to idiomatic python

billonahill:
Refactoring to use snake_case for method and variable names which is idiomatic python.


kramasamy:
👍

Issue #1489, opened 2016-10-13
GC logs are never rolled over or cleaned up 


ajorgensen:
The GC log files are never rolled over and will continue to grow indefinitely until the topology is re-deployed. We can pass flags into the JVM to automatically roll the gc log files similarly to how the main topology log files are rolled over.

`-XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=<number of files> -XX:GCLogFileSize=<size>`

For a data point we have a topology thats been running for 5 days and has 2.8 GB of gc logs for a single component. I'm not sure what the most sensible defaults are here but it might be good to keep it consistent with the other log files which I believe is 5 files with a max size of 100MB per log.

PR #1490, opened 2016-10-13
Add jvm arguments for gc log rotation

ajorgensen:
Currently the gc logs are not rotated and will continue to grow
indefinitely. These jvm options allow for the gc logs to be
automatically managed and rotated. These settings are consistent with
the number and frequency that the component logs are managed.

This closes https://github.com/twitter/heron/issues/1489


billonahill:
👍

PR #1492, opened 2016-10-14
Add the ability to store state as json in http_server for testing

billonahill:
To test scaling remotely the local client and the topology spouts need a way to exchange state. Adding a new http endpoint to permit storing and retrieving json on `http_server`.


billonahill:
I've got a few PRs piled up behind this one. Any comments?

objmagic:
👍

PR #1493, opened 2016-10-14
Use argparse + make chmod not fail in all cases

dmarchand:
This PR has two components to it:

- Refactor executor script a bit to use argparse. This shouldn't impact any runtime behavior. This just makes it so adding and removing parameters no longer requires shifting that giant array of arguments around. Shifting of parameters here caused some tricky to debug issues with our usage of Heron, and this should make figuring things out a bit easier.

- Change the way permission handling works a bit. The script originally just blindly attempted to chmod the log dirs and the various binaries. In a docker-based environment this ended up causing some issues for us. The executor had permission to execute these binaries already, so it didn't need to chmod them. However, it did not have permission to chmod them. As a result, this script would fail and abort even though it didn't actually need to change permissions. This solution seems a bit hacky, so I'm definitely open to something cleaner if anyone has any ideas.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -238,11 +238,18 @@ class HeronExecutor(object):
     parser.add_argument("scheduler_port")
     parser.add_argument("python_instance_binary")
 
-    return parser.parse_args(args[1:])
+    parsed_args, unknown_args = parser.parse_known_args(args[1:])
+
+    if unknown_args:
+      Log.error('Unknown argument: %s' % unknown_args[0])
+      parser.print_help()
+      sys.exit(1)
+
+    return parsed_args
 
   def run_command_or_exit(self, command):
     if self._run_blocking_process(command, True, self.shell_env) != 0:
-      Log.info("Failed to run command: %s. Exiting" % command)
+      Log.error("Failed to run command: %s. Exiting" % command)
       sys.exit(1)
 
   def initialize(self):
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -15,6 +15,7 @@
 #!/usr/bin/env python2.7
 """ The Heron executor is a process that runs on a container and is responsible for starting and
 monitoring the processes of the topology and it's support services."""
+import argparse
 import atexit
 import base64
 import json
@@ -24,11 +25,14 @@ import signal
 import string
 import subprocess
 import sys
+import stat
 import threading
 import time
 import yaml
 
 from functools import partial
+
+
 
 from heron.common.src.python.utils import log
 # pylint: disable=unused-import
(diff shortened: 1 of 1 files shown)


billonahill:
This makes me so happy, I was just thinking about this broken window! :)

I'll review soon. Would you please accept the CLA:
http://twitter.github.io/heron/docs/contributors/community/

heron/executor/src/python/heron_executor.py
```diff
@@ -200,8 +248,19 @@ def initialize(self):
...
     create_folders = 'mkdir -p %s' % self.log_dir
-    chmod_binaries = 'chmod a+rx . && chmod a+x %s && chmod +x %s %s %s' \
-        % (self.log_dir, self.tmaster_binary, self.stmgr_binary, self.heron_shell_binary)
+    chmod_x_binaries = [self.tmaster_binary, self.stmgr_binary, self.heron_shell_binary]
+    binaries_to_chmod = []
+
+    for binary in chmod_x_binaries:
+      stat_result = os.stat(binary)[stat.ST_MODE]
+      if not stat_result & stat.S_IXOTH:
+        binaries_to_chmod.append(binary)
```
billonahill: stat seems like a good approach, provided it works on all distros.

There's no need to do all chmods in one concatenated command, so better to break it up to make the code more readable. No need to build up arrays and concated strings, we can just do this:

```
- chmod log dir

- for each binary
  - if chmod needed
    - do chmod 
```

dmarchand: Works for me, I'll make that change. Stat appears to work on everything major, but I'll do some additional digging and testing on some VMs to be sure.

heron/executor/src/python/heron_executor.py
```diff
@@ -192,6 +201,45 @@ def __init__(self, args, shell_env):
 
     self.state_managers = []
 
+  @staticmethod
+  def parse_args(args):
+    """Uses python argparse to collect positional args"""
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("shard")
```
billonahill: elsewhere inthe code we use `--shard` instead. What's the difference in usage when using `shard`?

heron/executor/src/python/heron_executor.py
```diff
@@ -125,27 +129,31 @@ class HeronExecutor(object):
...
+                         rstrip('"').replace('&equals;', '='))
+    self.classpath = parsed_args.classpath
+    self.master_port = parsed_args.master_port
+    self.tmaster_controller_port = parsed_args.tmaster_controller_port
+    self.tmaster_stats_port = parsed_args.tmaster_stats_port
+    self.heron_internals_config_file = parsed_args.heron_internals_config_file
     self.component_rammap =\
-        map(lambda x: {x.split(':')[0]: int(x.split(':')[1])}, args[16].split(','))
+        map(lambda x: {x.split(':')[0]:
+                           int(x.split(':')[1])}, parsed_args.component_rammap.split(','))
```
billonahill: can args.component_rammap do this and just return a map?

kramasamy:
thanks @dmarchand - this is a great PR, we have been wanting to do this. Great to see you picked this up.

heron/executor/src/python/heron_executor.py
```diff
@@ -192,6 +201,50 @@ def __init__(self, args, shell_env):
...
+    parser.add_argument("cluster")
+    parser.add_argument("role")
+    parser.add_argument("environ")
+    parser.add_argument("instance_classpath")
+    parser.add_argument("metrics_sinks_config_file")
+    parser.add_argument("scheduler_classpath")
+    parser.add_argument("scheduler_port")
+    parser.add_argument("python_instance_binary")
+
+    return parser.parse_args(args[1:])
```
objmagic: Since you removed [this part](https://github.com/twitter/heron/pull/1493/files#diff-2a669e084143b86b5766a49245f8e357L650), can you use `parse_known_args` as in [heron/tools/explorer/src/python/main.py#L172](https://github.com/twitter/heron/blob/12425366a9ade8501675c1e29c2ea34cb36044cb/heron/tools/explorer/src/python/main.py#L172), and check if there is any unknown args?

heron/executor/src/python/heron_executor.py
```diff
@@ -192,6 +201,50 @@ def __init__(self, args, shell_env):
...
+
+  def run_command_or_exit(self, command):
+    if self._run_blocking_process(command, True, self.shell_env) != 0:
+      Log.info("Failed to run command: %s. Exiting" % command)
```
billonahill: Log.error

heron/executor/src/python/heron_executor.py
objmagic: I don't understand this comment. Can you explain?

billonahill: I was wondering if this logic could be encapsulated into the argparse logic via custom actions. I'm not certain it could, and maybe it's not worth the effort, so feel free to disregard.

heron/executor/src/python/heron_executor.py
```diff
@@ -192,6 +201,57 @@ def __init__(self, args, shell_env):
...
+    parser.add_argument("metrics_sinks_config_file")
+    parser.add_argument("scheduler_classpath")
+    parser.add_argument("scheduler_port")
+    parser.add_argument("python_instance_binary")
+
+    parsed_args, unknown_args = parser.parse_known_args(args[1:])
+
+    if unknown_args:
+      Log.error('Unknown argument: %s' % unknown_args[0])
+      parser.print_help()
```
objmagic: since we didn't write any help message for any argument, it doesn't really matter if we'll print help message here...

dmarchand: In that case argparse is supposed to dump the list of required arguments: https://docs.python.org/3/library/argparse.html#usage

It's probably not the best possible help message and we should likely take some time at some point to document the individual arguments, but this will at least clue the developer in as to which argument they added doesn't belong.

objmagic: Thanks for pointing out. Agreed.

billonahill:
👍

objmagic:
👍

PR #1494, opened 2016-10-15
Refactor integration tests to share more code

billonahill:
This refactor shares duplicate code, while allowing the ability to add custom logic to subclasses. It also changes to use proper arg parsing of topology arguments. This will simplify writing test with custom logic or args, like scaling tests.

Two main classes to look at, `AbstractTestTopology.java` and `main.py`. The rest are all basically the same refactor to all tests that now subclass `AbstractTestTopology.java`.


kramasamy:
ship it - once CI passes.

PR #1495, opened 2016-10-15
add warning log of tunnelHost before throw exception

Yaliang:


heron/statemgrs/src/java/com/twitter/heron/statemgr/zookeeper/curator/CuratorStateManager.java
```diff
@@ -77,6 +77,7 @@ public void initialize(Config newConfig) {
 
       String newConnectionString = tunneledResults.first;
       if (newConnectionString.isEmpty()) {
+        LOG.warning("Cannot connect to tunnelHost: " + tunnelConfig.getTunnelHost());
         throw new IllegalArgumentException("Bad connectionString: " + connectionString);
```
Yaliang: Fixed in PR [#1498](https://github.com/twitter/heron/pull/1498)

PR #1496, opened 2016-10-17
Set role to None if execution_state is None.

objmagic:
because strangely, execution state of some topology is None. This only happens to one devel topology inside of Twitter environment. We can discuss more about this issue internally...

cc @nlu90

### DIFF
--- a/heron/tools/tracker/src/python/handlers/topologieshandler.py
+++ b/heron/tools/tracker/src/python/handlers/topologieshandler.py
@@ -64,7 +64,12 @@ class TopologiesHandler(BaseHandler):
     for topology in topologies:
       cluster = topology.cluster
       environ = topology.environ
-      topo_role = topology.execution_state.role
+
+      if topology.execution_state:
+        topo_role = topology.execution_state.role
+      else:
+        topo_role = None
+
       if not cluster or not topo_role or not environ:
         continue
 


maosongfu:
We can investigate why the execution-state is missing; it should not happen.

objmagic:
@maosongfu it is a internal topology called "nlu-ex". Besides execution_state, its cluster and environ variables are also None.

kramasamy:
@objmagic - let us investigate why this happened

objmagic:
sure. considering it happens only internally, we'd better continue this discussion on internal Slack channel.

billonahill:
My suspicion is that that topology is one that @nlu90 was using to test with that has gotten into a bad state.

objmagic:
update: we could not find this topology on Aurora. However, for some unknown reasons, when talking to Zookeeper, tracker gets this topology's info (name only, no any other info)

kramasamy:
@objmagic - can you launch other topologies and see how they work?

kramasamy:
@objmagic - Is this PR relevant?

objmagic:
@kramasamy not related to the viz code. viz are fine now. we still don't understand why some ghost topology could show up in zookeeper...

objmagic:
close this PR since it's because of some ghost topologies in our internal ZK. @billonahill bombed them out and we think this PR is not necessary anymore.

PR #1500, opened 2016-10-18
Change integration test harness to support scaling tests

billonahill:
Made a number of changes to support scaling tests:

- If json test configs in `updateArgs`, the test is considered a scaling (i.e., update) test
- Added `MultiPhaseTestSpout` which outputs tuples, polls for `topology_updated` state changes on the `http_server` and outputs more. This class relies on `HttpGetCondition` that polls for expected state.
- Changed the python test runner to handle update tests by:
  - poll for `topology_started` state.
  - update the topology
  - proceed with the result verification

Apologies that this PR got large, but it was a hard one for me to break up.

There is still more work to be done to handle tests the scale down. Currently the aggregator bolt waits for teminal tuples from all upstream instances so removing one leaves it hanging. Will tackle that in a separate review.


integration-test/src/java/com/twitter/heron/integration_test/core/IntegrationTestBolt.java
```diff
@@ -86,6 +86,9 @@ public void execute(Tuple tuple) {
         collector.emit(Constants.INTEGRATION_TEST_CONTROL_STREAM_ID,
             tuple,
             new Values(Constants.INTEGRATION_TEST_TERMINAL));
+      } else {
+        LOG.info(String.format(
+            "Received a terminals, waiting to receive %s more", terminalsToReceive));
```
kramasamy: it should be Received a terminal or Received terminals

integration-test/src/python/test_runner/main.py
```diff
@@ -14,18 +14,18 @@
 # The location of default configure file
 DEFAULT_TEST_CONF_FILE = "integration-test/src/python/test_runner/resources/test.json"
 
-RETRY_ATTEMPTS = 15
+RETRY_ATTEMPTS = 25
```
kramasamy: why the change from 15 to 25?

billonahill: When there are legit failures retrying 25 times is excessive and it takes over 4 minutes of results checking per test. Even 15 times seems high. If we have to retry anything this many times for simple, small test topologies, something is wrong.

That said, I can keep it if our tests are known to be too flaky when set below 25.

kramasamy: Most of the times the integration tests not flaky, I believe.

billonahill: That's what I thought. Let's reduce it and see if it causes problems.

billonahill:
All comments addressed and CI is passing. Let me know if anyone has additional comments.

objmagic:
LGTM 👍

PR #1503, opened 2016-10-19
Send back_pressure related metrics to TMaster

maosongfu:
Add back_pressure related metrics into the whitelist of metrics
sending to TMaster, in the TMaster Sink's config in metrics_sinks.yaml

Tested in local scheduler integration.

### DIFF
--- a/heron/config/src/yaml/conf/aurora/metrics_sinks.yaml
+++ b/heron/config/src/yaml/conf/aurora/metrics_sinks.yaml
@@ -53,6 +53,8 @@ tmaster-sink:
     "__jvm-memory-used-mb": LAST
     "__jvm-memory-mb-total": LAST
     "__jvm-gc-collection-time-ms": LAST
+    "__server/__time_spent_back_pressure_initiated": SUM
+    "__time_spent_back_pressure_by_compid/": SUM
 
 ### Config for scribe-sink
 # scribe-sink:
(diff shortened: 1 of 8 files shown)


objmagic:
just curious, `SUM` means sum of the metrics in a window, right?

maosongfu:
@objmagic 
That means, in a given window:

For "SUM", when we received multiple values for a metric_name, we would sum them up and consider the aggregated value as the actual value.
For "AVG", we would consider the mean value as the actual value.
For "LAST", we would consider the last value received as the actual value.

objmagic:
thanks for the explanation. 

LGTM 👍

objmagic:
👍

kramasamy:
@maosongfu - can we get both? Can we know the container that is in back pressure and also the instance that triggered the back pressure? Sometimes, the back pressure is initiated by stmgr only due to network issues. Wondering if we can capture this as well.

maosongfu:
@kramasamy 
yes both of them are added in the whitelist.

kramasamy:
@maosongfu - how to distinguish between the two?

kramasamy:
Also how can we keep track of the back pressure for the last 3 min, 10 min, 1 hrs, 3 hrs, etc? so that we can bring this information up in the UI.

maosongfu:
"__server/__time_spent_back_pressure_initiated": SUM   -- is the one initiated by stmgr
 "__time_spent_back_pressure_by_compid": SUM -- is the one triggered by instance

Notice these are the prefix of the metrics_name. The actual full metric name would contain the instance id

kramasamy:
👍

Issue #1497, opened 2016-10-17
storm-hdfs (+ other storm-externals) not compatible


otter-in-a-suit:
When trying to use storm-hdfs (1.0.2) with Heron 0.14.3, the storm-internal classes from storm-core (e.g. org/apache/storm/utils/TupleUtils) are not found, resulting in a ClassNotFoundException:

`Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/storm/utils/TupleUtils
	at org.apache.storm.hdfs.bolt.AbstractHdfsBolt.getComponentConfiguration(AbstractHdfsBolt.java:219)
	at org.apache.storm.topology.IRichBoltDelegate.getComponentConfiguration(IRichBoltDelegate.java:70)
	at com.twitter.heron.api.topology.BaseComponentDeclarer.<init>(BaseComponentDeclarer.java:35)
	at com.twitter.heron.api.topology.BoltDeclarer.<init>(BoltDeclarer.java:34)
	at com.twitter.heron.api.topology.TopologyBuilder.setBolt(TopologyBuilder.java:118)
	at org.apache.storm.topology.TopologyBuilder.setBolt(TopologyBuilder.java:44)
	at com.otterinasuit.twitter.Main.main(Main.java:89)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
Caused by: java.lang.ClassNotFoundException: org.apache.storm.utils.TupleUtils
	at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 12 more`

storm-hdfs is invoked as such:

    <dependency>
[...]

PR #1505, opened 2016-10-20
add TupleUtils class

nlu90:
Trying to solve: #1497 

Only newer storm 1.0.x has this TupleUtils class available, thus we only add it into the corresponding path.


billonahill:
👍 once ci passes

PR #1507, opened 2016-10-21
Add version in tracker and UI

objmagic:
This PR

1. Add `version` subcommand for `heron-tracker` and `heron-ui`
2. `API_VERSION` in `heron/tools/tracker/src/python/constants.py` picks up version number from `release.yaml`.
3. Add `VERSION` constant in `heron/tools/ui/src/python/consts.py`, which also picks up version number from `release.yaml`

The key to solve the long-existing problem that `heron-ui` built with `zip_safe = False` flag is agnostic about the location of `release.yaml` is to zip `release.yaml` into `heron-ui` PEX file at compile time. See [here1](https://github.com/twitter/heron/compare/tracker-ui-version?expand=1#diff-20422e673bb9505a470407f3d2397ed2R34) and [here2](https://github.com/twitter/heron/compare/tracker-ui-version?expand=1#diff-2030bd7a10d92365e9bb96545104d00fR32).

Can verify `heron`, `heron-explorer`, `heron-tracker`, and especially `heron-ui` work fine.

cc @billonahill @kramasamy @maosongfu

### DIFF
--- a/heron/tools/common/src/python/utils/config.py
+++ b/heron/tools/common/src/python/utils/config.py
@@ -37,6 +37,7 @@ ETC_DIR = "etc"
 LIB_DIR = "lib"
 CLI_DIR = ".heron"
 RELEASE_YAML = "release.yaml"
+ZIPPED_RELEASE_YAML = "scripts/packages/release.yaml"
 OVERRIDE_YAML = "override.yaml"
 
 # directories for heron sandbox
@@ -138,9 +139,22 @@ def get_heron_dir():
 
   :return: root location for heron-cli.
   """
-  path = "/".join(os.path.realpath(__file__).split('/')[:-9])
+  go_above_dirs = 9
+  path = "/".join(os.path.realpath(__file__).split('/')[:-go_above_dirs])
   return normalized_class_path(path)
 
+def get_zipped_heron_dir():
+  """
+  This will extract heron directory from .pex file,
+  with `zip_safe = False' Bazel flag added when building this .pex file
+
+  Specifically designed for heron-ui
+
+  :return: root location for heron-ui.
+  """
+  go_above_dirs = 7
+  path = "/".join(os.path.realpath(__file__).split('/')[:-go_above_dirs])
+  return normalized_class_path(path)
 
 ################################################################################
 # Get the root of heron dir and various sub directories depending on platform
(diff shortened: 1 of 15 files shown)


heron/tools/common/src/python/utils/config.py
```diff
@@ -138,9 +139,22 @@ def get_heron_dir():
...
+  go_above_dirs = 9
+  path = "/".join(os.path.realpath(__file__).split('/')[:-go_above_dirs])
   return normalized_class_path(path)
 
+def get_zipped_heron_dir():
+  """
+  This will extract heron directory from .pex file,
+  with `zip_safe = False' Bazel flag added when building this .pex file
+
+  Specifically designed for heron-ui
```
objmagic: complicated to explain but in short --- because it is built with [`zip_safe=False`](https://github.com/twitter/heron/blob/master/heron/tools/ui/src/python/BUILD#L32)

objmagic: I am thinking any tool in the future that uses `zip_safe=False` can reuse this method, and naming this method in this way then makes sense.

heron/tools/common/src/python/utils/config.py
```diff
@@ -151,12 +165,12 @@ def get_heron_dir_explorer():
   From heron-cli with modification since we need to reuse cli's conf
   :return: root location for heron-cli.
   """
-  path_list = os.path.realpath(__file__).split('/')[:-10]
+  go_above_dirs = 10
```
objmagic: actually, i realized `heron-explorer` was moved to the same directory as `heron`. This means this method is no longer needed.

heron/tools/common/src/python/utils/config.py
objmagic: right. will remove it

heron/tools/common/src/python/utils/config.py
```diff
@@ -136,27 +137,38 @@ def get_heron_dir():
...
+  This means the variable `go_above_dirs` below is 7.
+
+  Specifically designed for heron-ui
+
+  :return: root location for heron-ui.
   """
-  path_list = os.path.realpath(__file__).split('/')[:-10]
-  path_list.append(CLI_DIR)
-  path = "/".join(path_list)
+  go_above_dirs = 7
```
billonahill: if we still need go_above_dirs in this method and/or in get_heron_dir, we should document the value.

objmagic: I briefly mentioned it [here](https://github.com/twitter/heron/pull/1507/files/e72fe571568deba0aed16582d8367e10a48b4e2e#diff-c8ec498d0b564ed129baec8a64dceb09R159) and I am not sure how to have a better wording here.

heron/tools/common/src/python/utils/config.py
objmagic: removed via https://github.com/twitter/heron/pull/1507/commits/e72fe571568deba0aed16582d8367e10a48b4e2e

billonahill:
👍

PR #1510, opened 2016-10-21
Adding IUpdatable for instances that need to be aware of runtime changes to plan

### LINKED ISSUE #1509: During scaling new physical plan info needs to be passed to running instances
Tracks how we will update the internal topology state that running instances hold during a scaling event. Required for #1292.

billonahill:
The PR is just being posted to share for discussion as a possible approach to fixed #1509. It is intended for comments about the approach.

- New method `IInstance.update(PhysicalPlanHelper)`
- `SpoutInstance` and `BoltInstance` implement `IInstance.update(..)`
- Added `IUpdatable.update(TopologyContext)` which user bolts and spouts can implement.


ashvina:
@billonahill Are `BoltInstance` and `SpoutInstance` internal/core classes? If yes, is interface UpdateableInstance` needed? Can the method `updateTopologyContext` be added to `IInstance`?

billonahill:
@ashvina that's a good point, those are internal classes. We could add `updateTopologyContext` into `IInstance` and just use `UpdateableInstance` for user implementations of `Bolt` and `Spout`.

billonahill:
@ashvina following your suggestion I added `update(PhysicalPlanHelper)` to `IInstance` which will simplify the implementation considerably. Internal classes will use `PhysicalPlanHelper` similar to how they do today, but user bolts and spouts only need the `TopologyContext` object.

The TODO in the PR will be satisfied after #1516 ships and the update method is made available.

heron/api/src/java/BUILD
```diff
@@ -6,6 +6,7 @@ load("/tools/rules/heron_deps", "heron_java_proto_files")
 
 api_deps_files =  \
     heron_java_proto_files() + [
+        ":classification",
```
kramasamy: @billonahill - this might not work since there is no target defined in this file?

heron/api/src/java/com/twitter/heron/api/topology/UpdatableComponent.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Bolt or spout instances should implement this method if they wish to be informed of changed to
+ * the topology context during the lifecycle of the instance. The TopologyContext is backed by
+ * elements of the physical plan, which can change during a scaling event. Instances that need to
+ * adapt to such changes should implement this interface to receive a callback when a change occurs.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public interface UpdatableComponent {
```
kramasamy: Instead of calling it UpdatableComponent should we call it IUpdatable to keep interface names consistent.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -96,6 +97,14 @@ public BoltInstance(PhysicalPlanHelper helper,
   }
 
   @Override
+  public void update(PhysicalPlanHelper physicalPlanHelper) {
+    if (bolt instanceof UpdatableComponent) {
+      ((UpdatableComponent) bolt).update(physicalPlanHelper.getTopologyContext());
+    }
+    // TODO: update collector with new physicalPlanHelper
```
kramasamy: Wondering if there is more code to be added here?

heron/api/src/java/BUILD
billonahill: This target is defined in this file.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
kramasamy: ok

heron/api/src/java/BUILD
kramasamy: my bad - i did not realize it is at end

heron/api/src/java/com/twitter/heron/api/topology/UpdatableComponent.java
billonahill: Renamed.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -96,6 +97,14 @@ public BoltInstance(PhysicalPlanHelper helper,
   }
 
   @Override
+  public void update(PhysicalPlanHelper physicalPlanHelper) {
+    if (bolt instanceof IUpdatable) {
+      ((IUpdatable) bolt).update(physicalPlanHelper.getTopologyContext());
```
ashvina: We invoke user code here which could fail/block. What is the expectation in such cases? I think the `executor` will log the error and launch a new instance if an error occurs?

kramasamy: Good point - this needs to be addressed. If the function execution exceeds beyond a certain time - some action needs to be taken.

ashvina:
The approach looks good to me. My understanding is that the UpdateManager will delete the physical plan and upload a new packing plan. TMaster will use the new packing plan to create a new physical plan and inform all the instances using control message. This control message is received by `Slave` to update the HeronInstance. Is this correct?

kramasamy:
@ashvina - yes that is the approach. It maximizes the use of existing messages and code path.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
billonahill: This is something we should look into separately. The same is true for all other calls to `bolt` in this class (e.g.,  `bolt.prepare(..)`,  `bolt.cleanup(..)`,  `bolt.execute(..)`). None have exception handling or timeouts.

PR #1516, opened 2016-10-24
Refactor usage of PhysicalPlanHelper, dedup simulator code

billonahill:
No new functionality here, refactoring how `PhysicalPlanHelper` is used by `SpoutInstance` and `BoltInstnance` to make it easier to update the `TopologyContext`, which is required for #1509. I encountered a lot of duplicate code between the instance classes and the simulator classes so I removed much of that as well.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/utils/misc/CustomStreamGroupingHelper.java
+++ b/heron/common/src/java/com/twitter/heron/common/utils/misc/CustomStreamGroupingHelper.java
@@ -16,43 +16,39 @@ package com.twitter.heron.common.utils.misc;
 
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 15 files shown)


kramasamy:
Looks good to me! @maosongfu - can you take a look at the instance changes? @nlu90 - can you take a look at the simulator code?

objmagic:
LGTM 👍

billonahill:
Next I've got some more structural changes to make around the interaction between and  PhysicalPlanHelper/TopologyContextImpl but I'll break those into another review.

billonahill:
Sorry for all the commits on this one but I realized I can simplify the approach. I've reverted some edits and am setting us up to allow `IInstance` to take an `update(PhysicalPlanHelper)` method since this is a core (i.e., internal only) class. The user bolts/spouts would then implement `UpdatableInstnace` which takes an `update(TopologyContext)` method.

kramasamy:
New changes look good to me. 👍

PR #1517, opened 2016-10-25
Fix redirectErrorStream for runASyncProcess

windie:
`Process. redirectErrorStream` is no-op. Should use `Process.redirectErrorStream(Boolean)` instead.

This PR also fixes the wrong unit test `testRunAsyncProcess`.

### DIFF
--- a/heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
+++ b/heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
@@ -88,15 +88,25 @@ public class ShellUtilsTest {
 
   @Test
   public void testRunAsyncProcess() throws IOException {
-    String testString = "testString";
-    StringBuilder stdout = new StringBuilder();
-    StringBuilder stderr = new StringBuilder();
-    // Sleep 1 second and echo some text.
-    Process p = ShellUtils.runASyncProcess(
-        String.format("sleep 1 && echo %s", testString));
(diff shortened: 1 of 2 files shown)


heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
```diff
@@ -88,15 +88,25 @@ public void testLargeOutput() throws IOException, InterruptedException {
...
   public void testRunAsyncProcess() throws IOException {
-    String testString = "testString";
-    StringBuilder stdout = new StringBuilder();
-    StringBuilder stderr = new StringBuilder();
-    // Sleep 1 second and echo some text.
-    Process p = ShellUtils.runASyncProcess(
-        String.format("sleep 1 && echo %s", testString));
-    // Test process is running and input stream is empty
-    wait(10, TimeUnit.MILLISECONDS);
-    Assert.assertEquals(0, p.getInputStream().available());
```
windie: This old test actually didn't test the behavior. `sleep 1 && echo ...` will be split to `sleep`, `1`, `&&`, `echo`, `...`, and the code will launch the process `sleep` with the other strings as its argument. Then `sleep` will fail at once. Since `redirectErrorStream` didn't work before, the error info will go into stderr. So `p.getInputStream().available()` will always be 0.

kramasamy:
Looks good 👍

billonahill:
👍

PR #1520, opened 2016-10-26
Remove unused packing plan code

billonahill:


objmagic:
do we need to change [FirstFitDecreasingPackingTest.java](https://github.com/twitter/heron/blob/691f19ddba080e84ee824acc30de893907ffa3f9/heron/packing/tests/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPackingTest.java#L109) accordingly?

billonahill:
@objmagic no, there are no consumers of these methods. They're leftover from a previous refactor that used another approach.

objmagic:
You are right I forgot to check if arguments match.

objmagic:
👍

PR #1521, opened 2016-10-27
Add PackingPlanBuilder

### LINKED ISSUE #946: Make PackingPlan builder pattern
Make PackingPlan builder pattern, allowing people to construct PackingPlan easier and safer.

billonahill:
Adds `PackingPlanBuilder` to fix #946. This reduces almost all container state management code (i.e., map/list manipulation) from the packing algorithms and leaves just the algorithm logic itself.

There are a few more edits I'd need to make (possibly in others PR) but wanted to share this to get feedback on the approach:
- Add checking that all required setters were called before adding/removing instances (DONE)
- `Packing.pack` and `Packing.repack` should throw `PackingException` instead of returning null. I'd do this in another PR.
- Some of the `PackingUtils` code is now only used by `PackingPlanBuilder`, so it could move there as internal utils.
- Some unit tests for just `PackingPlanBuilder`.

cc/ @avflor @ashvina @maosongfu

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
@@ -13,7 +13,6 @@
 // limitations under the License.
 package com.twitter.heron.packing;
 
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -40,8 +39,7 @@ public class PackingPlanBuilder {
   private int requestedContainerPadding;
   private int numContainers;
 
-  private Map<Integer, List<InstanceId>> containerInstances;
-  private List<Container> containers;
+  private Map<Integer, Container> containers;
 
   public PackingPlanBuilder(String topologyId) {
     this(topologyId, null);
@@ -50,7 +48,6 @@ public class PackingPlanBuilder {
   public PackingPlanBuilder(String topologyId, PackingPlan existingPacking) {
     this.topologyId = topologyId;
     this.existingPacking = existingPacking;
-    this.containerInstances = new HashMap<Integer, List<InstanceId>>();
     this.numContainers = 0;
   }
 
@@ -84,22 +81,18 @@ public class PackingPlanBuilder {
   public PackingPlanBuilder addInstance(Integer containerId,
                                         InstanceId instanceId) throws ResourceExceededException {
     initContainers();
-    if (containerInstances.get(containerId) == null) {
-      containerInstances.put(containerId, new ArrayList<InstanceId>());
-    }
 
     Resource instanceResource = PackingUtils.getResourceRequirement(
         instanceId.getComponentName(), this.componentRamMap, this.defaultInstanceResource,
         this.maxContainerResource, this.requestedContainerPadding);
 
-    if (!containers.get(containerId - 1).add(
+    if (!containers.get(containerId).add(
         new PackingPlan.InstancePlan(instanceId, instanceResource))) {
       throw new ResourceExceededException(String.format(
           "Insufficient container resources to add instance %s with resources %s to container %d.",
           instanceId, instanceResource, containerId));
     }
 
-    containerInstances.get(containerId).add(instanceId);
     LOG.fine(String.format("Added to container %d instance %s", containerId, instanceId));
     return this;
   }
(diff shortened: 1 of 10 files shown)


heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+public class PackingPlanBuilder {
+  private static final Logger LOG = Logger.getLogger(PackingPlanBuilder.class.getName());
+
+  private final String topologyId;
+  private final PackingPlan existingPacking;
+  private Resource defaultInstanceResource;
+  private Resource maxContainerResource;
+  private Map<String, Long> componentRamMap;
+  private int requestedContainerPadding;
+  private int numContainers;
```
avflor: It's a bit confusing that both the builder and the individual algorithms track the number of containers. Would it be easier to use it in one place only?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -183,78 +183,43 @@ public PackingPlan pack() {
...
 
-    while (!resourceCompliantRRAllocation.isPresent()) {
-      //The number of containers has been updated
-      if (this.numAdjustments > adjustments) {
+    int adjustments = 0;
+    while (adjustments <= this.numAdjustments) {
+      try {
+        PackingPlanBuilder planBuilder = newPackingPlanBuilder(currentPackingPlan);
+        planBuilder.updateNumContainers(numContainers);
+        planBuilder = getResourceCompliantRRAllocation(
```
avflor: If get resource compliant allocation fails with an exception, then the packingbuilder will still have the updaed number of containers? Shouldn't we bring it back to the old value in the catch clause? It might not make a difference eventually since we update again at the next iteration but it guarantees consistency. Now it is a little tricky to me to check whether the numContainers in the builder is always the correct one.

heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+    this.componentRamMap = ramMap;
+    return this;
+  }
+
+  public PackingPlanBuilder setRequestedContainerPadding(int percent) {
+    this.requestedContainerPadding = percent;
+    return this;
+  }
+
+  public PackingPlanBuilder updateNumContainers(int count) {
```
avflor: I think it looks better if we allocate/deallocate  containers when the user invokes the updateNumContainers command and not leave it to addInstance which invokes init.

billonahill: I haven't yet documented this, but the design intent (which we can enforce) is that the builder is used as follows:

1. Call the setters to set required settings (e.g., resource settings)
2. Set the container size
3. Add/remove instances
  3a. If more containers are required go to step 2.
4. Build

As such, there are a few reasons why allocation/deallocation happens in add/remove instances.
- setting container size is required before adding/removing because the system is bounded by container size. (Actually we should support either bounding by container size or containerMaxResources. Currently we require bounding by both.)
- instances can be added and removed, followed by build without the need to call set container size between the two.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
billonahill: I think you're describing a scenario where a container is added and we then catch a `ResourceExceedException` and nothing has been added so we have an empty container. [We dont't have a check for that case and we should but] What currently happens is that we add a container and a number of instnaces are added to it until it's full and _then_ we get a `ResourceExceedException`, so we're consistent and the new container is full and it's time to add another.

heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
avflor: I'm a little confused here. The command to add a container is the following: PackingUtils.allocateNewContainer(containers, containers.get(0).getCapacity(), this.requestedContainerPadding) when someone adds a new container. Why can't we do this in the updatecommand? It doesn't have any connection to an instance.

Overall, what makes it hard for me to verify the correctness of the code is that the numContainers in the builder can change without allocating the necessary containers right away. Wouldn't it be easier to enforce that by requiring that when the numContainers is updated in the builder, the List<Containers> is also updated in a "transaction".

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
avflor: Ok makes sense.

heron/packing/src/java/com/twitter/heron/packing/PackingPlanBuilder.java
billonahill: We can't allocate containers in the `updateNumContainers` because `setRequestedContainerPadding` might not have been called yet and we need the container padding. Hence we call 'initContainers' before each add/remove and we get our internal data structures consistent.

My plan was to document and enforce that steps 1 and 2 in my sequence above are done (in any order) to set the relevant params, before adding/removing instances. If make the decision to change the internal behavior to init containers upon update, we'll need to enforce more restrictions on the user ordering of actions that are not necessary and wouldn't be caught until runtime.

billonahill:
I've addressed all the TODOs in the diff except for throwing PackingException unstead of returning null, which I'll do separately. I've also simplified methods that are now redundant and the PackingPlanBuilder implementation. Let me know if there are additional comments. This is ready to commit IMO.

kramasamy:
@avflor @ashvina - can you let us know if you have any comments?

ashvina:
LGTM 👍

PR #1522, opened 2016-10-28
Optimize container allocation 950

### LINKED ISSUE #950: Support fast homogeneous container allocation in YARN scheduler
YARN scheduler supports packing based heterogeneous container allocation. This implementation prevents over utilization of cluster resources. However it is serial and hence could be slow. YARN scheduler could support a Aurora like container allocation mode. 

This enhancement may not be needed if YARN-4879 is fixed.

ashvina:
Fixes #950 

The containers allocated by YARN could be bigger than the requested size. This is a consequence of a YARN scheduler optimization, which allocate resources in fixed memory and cpu increments. For e.g. if the memory increment is 100 MB, YARN may allocated a container of 200 MB when 150 MB is requested. As a result, the actual resource allocation may be little more than what was requested. Moreover YARN does not provide any mechanism to match the container request to allocated container. So the AM has no way to know which ContainerPlan to launch in the allocated container.

To avoid confusion, in the first implementation the AM would requests one container at a time. Only when the container is requested and configured, a new container request is made. This serial request-allocation cycle could be slow. In this new implementation, the serial container request mechanism is replaced with concurrent requests. The AM now provides a method to perform find a largest fitting heron ContainerPlan in the allocated container to address the issue described above.

Concurrent implementation seems to have improve topology start time by 3x to 5x. In one of the test involving 200 Containers, the new implementation took 45 seconds compared to 260 seconds earlier.


kramasamy:
@ashvina - can we provide a detailed explanation?

mycFelix:
I think it will much more efficient than the previous implementation.

ashvina:
@kramasamy - Updated the PR description and javadoc for the YARN AM.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -464,17 +503,41 @@ private void launchScheduler() {
...
-        // Fail deployment of topology if there is a error starting any worker
-        LOG.log(Level.SEVERE, "Container allocation for id:{0} failed. Trying to close currently"
-            + " allocated resources for {1}", new Object[]{evaluator.getId(), topologyName});
-        killTopology();
-        throw new RuntimeException("Failed to allocate container for workers", e);
+      EvaluatorDescriptor descriptor = evaluator.getEvaluatorDescriptor();
+      LOG.log(Level.INFO, String.format("New container received, id: %s, mem: %d, cores: %d",
+          evaluator.getId(), descriptor.getMemory(), descriptor.getNumberOfCores()));
+
+      Optional<HeronWorker> result;
```
ashvina: Yes, it could be inside. Just to be sure, did you observe any functional issues with it other than code readability?

kramasamy:
@ashvina - thanks. Looks like some syntax in compilation.

ashvina:
@kramasamy The compilation errors are in kafka spout, unrelated to yarn.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
```diff
@@ -163,30 +165,86 @@ void scheduleHeronWorkers(PackingPlan topologyPacking) throws ContainerAllocatio
...
+  }
+
+  /**
+   * YARN allocates resources in fixed increments of memory and cpu. As a result, the actual
+   * resource allocation may be little more than what was requested. This method finds the biggest
+   * heron containerPlan that will fit the allocated YARN container. In some cases the YARN CPU
+   * scheduling may be disabled, resulting in default core allocation to each container. This
+   * method can ignore core fitting in such a case.
+   */
+  @VisibleForTesting
```
ashvina: No, the YARN scheduler continues to work with heterogeneous container sizes. YARN does not have a notion of request id in the allocated container. This method ensures that the allocated container is utilized properly by finding the biggest heron ContainerPlan that can fit in it.

PR #1528, opened 2016-11-03
Reduce logging during packing plan change

billonahill:
The PackingPlan can be really large for big topologies so changing logging for that to debug from info. Also fixing some typos where we referenced packing instead of physical plan.


objmagic:
👍 if style is fixed

PR #1529, opened 2016-11-03
Move methods from PackingUtils into PackingPlanBuilder

billonahill:
Now that logic is centralized in the builder there are methods in utils that are only used by the builder which we can make private in the builder.

Various other cleanup of unused code.

No logic changes here, just moving things around.


kramasamy:
👍  @ashvina @avflor - any comments?

avflor:
Will review that tomorrow

avflor:
@billonahill The changes look good to me.

PR #1530, opened 2016-11-03
IPacking and IRepacking now throws PackingException

billonahill:
Moved `PackingException` into `c.t.heron.spi.packing` and `pack()` and `repack()` now throw it.


kramasamy:
👍

PR #1531, opened 2016-11-03
Remove numAdjustments from ResourceCompliantRRPacking

billonahill:
`numAdjustments` is no longer needed so removing it and simplifying some placement code.


kramasamy:
👍

PR #1532, opened 2016-11-04
Change std::list to std::deque for tuple cache

congwang:
An STL implemented deque usually allocates memory in pools, which reduces calls into memory allocator. This also makes deque much more cache friendly than list.

In production, we observe 2.5x performance improvement with this change. Locally, I saw ~5% improvement.


objmagic:
👍

kramasamy:
👍

PR #1540, opened 2016-11-08
Catch error when killing proceses if process doesn't exist

### LINKED ISSUE #1536: Executor left in bad state during scaling if expected process doesn't exist.
If an executor needs to kill and start a new stream manager, for instance, and the expected process doesn't exist, an exception is thrown during the kill that prevents the start from happening.
```
11/07/2016 20:06:36 +0000:INFO: Killing stmgr-39 process with pid 46555: ...
11/07/2016 20:06:36 +0000:ERROR: [Errno 3] No such process
Traceback (most recent call last):
  File ".../.pex/install/kazoo-1.3.1-py2.7.egg.9e0018919913c23864670dfb4abcbd86b8867ea6/kazoo-1.3.1-py2.7.egg/kazoo/recipe/watchers.py", line 135, in _log_func_exception
    result = self._func(data, stat)
  File "heron/statemgrs/src/python/zkstatemanager.py", line 259, in watch_packing_plan 
    callback(packing_plan) 
  File "heron/executor/src/python/heron_executor.py", line 692, in on_packing_plan_watch File "heron/executor/src/python/heron_executor.py", line 664, in launch File "heron/executor/src/python/heron_executor.py", line 552, in _kill_processes 
  File "/usr/lib64/python2.7/subprocess.py", line 1564, in kill
    self.send_signal(signal.SIGKILL)
  File "/usr/lib64/python2.7/subprocess.py", line 1554, in send_signal
    os.kill(self.pid, sig)
OSError: [Errno 3] No such process
```

Required for #1292

billonahill:
Fixed #1536. Also improves some logging.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -547,9 +547,16 @@ class HeronExecutor(object):
         for process_info in self.processes_to_monitor.values():
           if process_info.name == command_name:
             del self.processes_to_monitor[process_info.pid]
-            Log.info("Killing %s process with pid %s: %s" %
+            Log.info("Killing %s process with pid %d: %s" %
                      (process_info.name, process_info.pid, ' '.join(command)))
-            process_info.process.kill()
+            try:
+              process_info.process.kill()
+            except OSError, e:
+              if e.errno == 3: # No such process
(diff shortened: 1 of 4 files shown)


kramasamy:
👍

PR #1541, opened 2016-11-08
Use pipe to properly block clients when server hasn't entered event loop.

objmagic:
The previous way of using `usleep` cannot guarantee the clients are blocked when server hasn't entered event loop, which makes``heron/common/tests/cpp/network:http_unittest`` flaky.

This PR switches to using pipe. Server will send a byte to pipe prior to entering event loop. Clients will wait(block) until the byte is read on the other side of the pipe. This makes sure that clients are properly blocked.

Test done via script
```Bash
#!/bin/sh

for i in `seq 1 1000`;
do
    bazel test --config=darwin --no_cache_test_results heron/common/tests/cpp/network:http_unittest
    if [[ $? != 0 ]]; then
      exit 1
    fi
done
```

Thanks @congwang for suggestions.

### DIFF
--- a/heron/common/tests/cpp/network/http_unittest.cpp
+++ b/heron/common/tests/cpp/network/http_unittest.cpp
@@ -36,7 +36,7 @@ static const sp_uint32 SERVER_PORT = 60000;
 
 extern void start_http_client(sp_uint32 _port, sp_uint64 _requests, sp_uint32 _nkeys);
 
-extern void start_http_server(sp_uint32 _port, sp_uint32 _nkeys, int fd, int sent);
+extern void start_http_server(sp_uint32 _port, sp_uint32 _nkeys, int fd);
 
 void TerminateRequestDone(HTTPClient* client, IncomingHTTPResponse* response) {
   delete response;
@@ -68,13 +68,13 @@ void StartTest(sp_uint32 nclients, sp_uint64 requests, sp_uint32 nkeys) {
   int fds[2];
 
   // choose an arbitrary number for ``sent'' as token
-  int recv, sent = 19583;
+  int recv;
   if (::pipe(fds) < 0) {
     std::cerr << "Unable to open pipe" << std::endl;
     GTEST_FAIL();
   }
 
-  std::thread sthread(start_http_server, SERVER_PORT, nkeys, fds[1], sent);
+  std::thread sthread(start_http_server, SERVER_PORT, nkeys, fds[1]);
 
   // block clients from reading from server using pipe
   read(fds[0], &recv, sizeof(int));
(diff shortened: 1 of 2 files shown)


heron/common/tests/cpp/network/http_unittest.cpp
```diff
@@ -64,9 +65,24 @@ void TerminateServer(sp_uint32 port) {
 
 void StartTest(sp_uint32 nclients, sp_uint64 requests, sp_uint32 nkeys) {
   // start the server thread
-  std::thread sthread(start_http_server, SERVER_PORT, nkeys);
+  int fds[2];
 
-  ::usleep(1000);
+  // choose an arbitrary number for ``sent'' as token
+  int recv, sent = 19583;
+  if (pipe(fds) < 0) {
```
congwang: Use ::pipe().

heron/common/tests/cpp/network/http_unittest.cpp
```diff
@@ -64,9 +65,24 @@ void TerminateServer(sp_uint32 port) {
...
+    std::cerr << "Unable to open pipe" << std::endl;
+    GTEST_FAIL();
+  }
+
+  std::thread sthread(start_http_server, SERVER_PORT, nkeys, fds[1], sent);
+
+  // block clients from reading from server using pipe
+  read(fds[0], &recv, sizeof(int));
+  if (recv != sent) {
```
congwang: No need to check the value, we only have one reader and one writer.

heron/common/tests/cpp/network/http_unittest.cpp
objmagic: resolved via https://github.com/twitter/heron/pull/1541/commits/0238865309396146ecb897aac5eec9260b7015d8

heron/common/tests/cpp/network/http_unittest.cpp
objmagic: resolved via https://github.com/twitter/heron/pull/1541/commits/593f35ad033a133938057f381844285804be373a

kramasamy:
Good one! Thanks for fixing this.

heron/common/tests/cpp/network/http_unittest.cpp
```diff
@@ -35,7 +36,7 @@ static const sp_uint32 SERVER_PORT = 60000;
 
 extern void start_http_client(sp_uint32 _port, sp_uint64 _requests, sp_uint32 _nkeys);
 
-extern void start_http_server(sp_uint32 _port, sp_uint32 _nkeys);
+extern void start_http_server(sp_uint32 _port, sp_uint32 _nkeys, int fd, int sent);
```
congwang: 'sent' is no longer needed either.

objmagic: resolved via https://github.com/twitter/heron/pull/1541/commits/dc185672ac8d8b9bd773ecf9246c4a32ef6a3bc0

congwang:
Looks good to me.

billonahill:
Should fix #1544.

Issue #949, opened 2016-06-17
Make YARN scheduler's master highly available 


ashvina:
YARN scheduler is implemented using Apache REEF. Currently REEF does not support YARN AM HA (REEF-345). In absence of this feature, YARN will kill all containers if AM container crashes. Currently Heron TMaster runs in a separate container. Which mean HA for Heron TMaster and workers is supported.

PR #1542, opened 2016-11-09
Yarn colocate tmaster and scheduler

### LINKED ISSUE #951: YARN scheduler must start TMaster and Heron Scheduler in the same container
Currently Heron scheduler is co-located with topology AM and TMaster is started in its own container. It is better to keep Scheduler and TMaster in the same container. Which means a heron topology on YARN will have n+2 containers. 1 for AM, 1 for TMaster+Scheduler and N worker nodes.

ashvina:
Fixes #951 

In an attempt to keep YARN AM lightly loaded, TMaster was launched in a different container. As a result TMaster and scheduler were split in two different containers. This change resolves this issue. More resources are allocated to AM and TMaster is launched in the same container as the YARN scheduler.


kramasamy:
@ashvina  - does it mean that the the master containers - running TM, scheduler, etc will require more resources than the other data containers - in the case of homogenous containers?

ashvina:
@kramasamy - Not necessary. A container with instances is typically 4GB/4cores in size. The master out of the box will be 2GB/1 core. The memory allocated to master is configurable.

kramasamy:
thanks @ashvina

kramasamy:
@mycFelix - please test this and let us know, how it works.

nlu90:
I'm new to yarn scheduler, so my question might be noob. 

Is the HeronMasterDriver luanched in separate containers for each topology? Will there be a case that multiple tmasters running in the same container after this change?

ashvina:
@nlu90 - All questions are welcomed :-)

Yes, one HeronMasterDriver is launched for each topology. In case of YARN, each topology is a YARN application and each app needs an app-master (called Driver in our implementation).

kramasamy:
@nlu90 = any other questions?

mycFelix:
@kramasamy - Sorry for the delay.  It runs beautiful. 

BTW @ashvina , shall we update the doc since #949 and #951 have been solved?

PR #1543, opened 2016-11-09
Clean up ZK lock resources upon kill

### LINKED ISSUE #1537: Topologies need to clean up their ZK storm/heron/states/locks/<name> node
Upon scaling a lock is written to ZK under `storm/heron/states/locks/<name>`. The locks node needs to be cleaned up by the topology upon exit.

billonahill:
Releasing a ZK lock does not remove the node, since others might be contending for the same one. We need to clean all lock resources upon exit.

Fixes #1537. Relates to #1292.


ashvina:
LGTM 👍

PR #1546, opened 2016-11-09
Change local integration tests to use test status objects instead of True/False

billonahill:
The integration test frameworks return True/False, or sometimes "success"/"fail" to denote a successful or failed test. This changes that so tests can either return a `Success` object or raise a `TestFailure` exception.

This simplifies the code because all methods in the chain no longer need to check for True/False and log before proceeding. Instead tests proceed assuming success and handle the failure exceptions if thrown. We also don't need topology cleanup calls scattered about, since we can just catch the failure and clean up.

This changes the local integration test, but I plan to change the remote tests similarly in another PR.

### DIFF
--- a/integration-test/src/python/local_test_runner/test_template.py
+++ b/integration-test/src/python/local_test_runner/test_template.py
@@ -82,15 +82,13 @@ class TestTemplate(object):
       return result
 
     except status.TestFailure as e:
-      if topology_submitted:
-        logging.error("Test failed, attempting to clean up")
-        self.cleanup_test()
       raise e
     except Exception as e:
+      raise status.TestFailure("Exception thrown during test", e)
+    finally:
       if topology_submitted:
         logging.error("Test failed, attempting to clean up")
         self.cleanup_test()
-      return status.TestFailure("Exception thrown during test", e)
 
   def submit_topology(self):
     _submit_topology(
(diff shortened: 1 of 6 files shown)


integration-test/src/python/local_test_runner/test_template.py
```diff
@@ -52,56 +54,54 @@ def __init__(self, testname, params):
...
+      return result
 
-    # finally verify the expected results
-    return self._check_results()
+    except status.TestFailure as e:
+      if topology_submitted:
+        logging.error("Test failed, attempting to clean up")
+        self.cleanup_test()
+      raise e
+    except Exception as e:
```
objmagic: duplicate ``except ... as e``?

integration-test/src/python/local_test_runner/test_template.py
```diff
@@ -52,56 +54,54 @@ def __init__(self, testname, params):
...
+      if topology_submitted:
+        logging.error("Test failed, attempting to clean up")
+        self.cleanup_test()
+      return status.TestFailure("Exception thrown during test", e)
```
objmagic: how about using ``try .. except .. finally`` clause? https://docs.python.org/2.7/tutorial/errors.html#defining-clean-up-actions

objmagic:
besides comments and style issues, 👍 👍 👍 👍 👍

integration-test/src/python/local_test_runner/test_template.py
billonahill: That's ok since `e` is scoped within each except clause so only 1 will be defined without collision.

PR #1548, opened 2016-11-10
Remove failure logging for cleanup and duplicate cleanup method call

billonahill:
Patch for #1546 to not call cleanup twice and to not log failure in finally block.


kramasamy:
👍

objmagic:
👍

PR #1421, opened 2016-09-23
Stream manager performance improvement

congwang:
Major changes include:
1. Add memory pool for dedicated protobuf messages to reuse object, reducing the expensive cost to new/delete a protobuf message.
2. Reduce duplicated protobuf message, which was design to be immutable and be easy for reasoning. But it is super inefficient.
3. Optimize *.proto to reduce duplicated serialization/deserialization cost. For instance, when one stmgr receives a message from another stmgr, it does not need to parse the TupleSet message; it just need to send its serialized byte array to corresponding instance.
4. Reduce duplicated computation. For instance, protobuf Message.ByteSize(), which is expensive and re-calculate the result (not reading from cache value computed) every time invoked.
5. Use a C++11 feature to optimize protobuf performance.

6X performance improvement for clang (Tested on mac); 4x performance improvement for gcc on Linux (Tested on Centos7 for large scale).

Joint work with Maosong Fu.

PR #1549, opened 2016-11-10
Add tuple-serialization-time-ns back

congwang:
This code was removed by accident, we should add it back.


kramasamy:
👍

objmagic:
Pleases do the same for [SpoutOutputCollectorImpl.java](https://github.com/twitter/heron/pull/1421/files#diff-58eb0480b3f9560b193ed37bc373ddc9L233)

objmagic:
👍

PR #1550, opened 2016-11-10
Using common python logging for int tests

billonahill:
Standardizing int test framework to use the common python logging format. This makes logs look much better because the int test logs and the heron cli logs are consistent. It also adds color-coding to the log level which helps when reviewing logs in CI on the console (thanks for that formatter @objmagic!).

This means the `local_test_runner`, can no longer be invoked directly due to the pex dep. Now we need to build and then run the pex, similar to what we currently do with `test_runner`.

### DIFF
--- a/integration-test/src/python/local_test_runner/main.py
+++ b/integration-test/src/python/local_test_runner/main.py
@@ -26,6 +26,7 @@ import sys
 from collections import namedtuple
 
 import status
+from heron.common.src.python.utils import log
 
 # import test_kill_bolt
 import test_kill_metricsmgr
@@ -46,7 +47,7 @@ TEST_CLASSES = [
 ]
 
 # The location of default configure file
-DEFAULT_TEST_CONF_FILE = "resources/test.conf"
+DEFAULT_TEST_CONF_FILE = "integration-test/src/python/local_test_runner/resources/test.conf"
 
 ProcessTuple = namedtuple('ProcessTuple', 'pid cmd')
 
@@ -100,8 +101,7 @@ def _random_port():
 
 def main():
   """ main """
-  root = logging.getLogger()
-  root.setLevel(logging.DEBUG)
+  log.configure(level=logging.DEBUG)
 
   # Read the configuration file from package
   conf_file = DEFAULT_TEST_CONF_FILE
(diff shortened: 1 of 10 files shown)


kramasamy:
👍

integration-test/src/python/local_test_runner/main.py
```diff
@@ -132,16 +129,26 @@ def main():
...
+    class_tokens = first_arg.split(".")
+    if first_arg == "-h" or len(class_tokens) < 2:
       usage()
+
     import importlib
-    test_classes = [getattr(importlib.import_module(class_name[0]), class_name[1])]
+    package_tokens = class_tokens[:-1]
+    test_class = class_tokens[-1]
+    if len(package_tokens) == 1: # prepend base packages for convenience
+      test_module = "integration-test.src.python.local_test_runner." + package_tokens[0]
```
objmagic: "for convenience" of what? i don't quite follow here...

billonahill: So you can run a single test like this:
```
./bazel-bin/integration-test/src/python/local_test_runner/local-test-runner \
  test_template.TestTemplate
```
instead of like this:
```
./bazel-bin/integration-test/src/python/local_test_runner/local-test-runner \
  integration-test.src.python.local_test_runner.test_template.TestTemplate
```

objmagic:
I see. 👍


On Nov 10, 2016, at 11:58, Bill Graham <notifications@github.com> wrote:

*@billonahill* commented on this pull request.
------------------------------

In integration-test/src/python/local_test_runner/main.py
<https://github.com/twitter/heron/pull/1550>:

>      import importlib
-    test_classes = [getattr(importlib.import_module(class_name[0]),
class_name[1])]
+    package_tokens = class_tokens[:-1]
+    test_class = class_tokens[-1]
+    if len(package_tokens) == 1: # prepend base packages for convenience
+      test_module = "integration-test.src.python.local_test_runner."
+ package_tokens[0]

So you can run a single test like this:

./bazel-bin/integration-test/src/python/local_test_runner/local-test-runner \
  test_template.TestTemplate

instead of like this:

./bazel-bin/integration-test/src/python/local_test_runner/local-test-runner \
  integration-test.src.python.local_test_runner.test_template.TestTemplate

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/twitter/heron/pull/1550>, or mute the thread

PR #1551, opened 2016-11-10
Use std::vector for fields grouping indices

congwang:
We always push back, so std::vector is enough and faster than std::list.


objmagic:
👍

PR #1553, opened 2016-11-10
Refactor test_runner to use status classes instead of success/fail

billonahill:
Instread of returning `"success"`/`"fail"` strings, `test_runner` should return `TestSuccess` or raise `TestFailure`.

Moved the `status` module classes to the `common` module to share between test frameworks.


objmagic:
👍

PR #1554, opened 2016-11-11
Fix string formatting bug from migration to TestFailure

billonahill:
`TestFailure(message, error=None)` is the signature we should now be using, so the error is no longer part of the string.


objmagic:
👍

PR #1563, opened 2016-11-15
Clean up some logging found while testing

billonahill:
While testing I found a few places where we our logging could be improved. Fixing those and deleting `status.py` which should have been removed when it got moved to `common`.


objmagic:
👍

PR #1565, opened 2016-11-15
Broke out packing builder and utils into separate packages

billonahill:
Moved some classes to their own packages to provide better isolation ok classes like Container.


ashvina:
LGTM 👍

Issue #1560, opened 2016-11-14
Scaling down should remove instances from the container with the most imbalance


billonahill:
The repacking algos currently remove instances in a round-robin fashion from container 1, which can result in the imbalanced packing described below. Instead instances should be removed from the containers that have the most imbalance.

For example:

1. Deploy the `ExclamationTopology`:
`~/bin/heron submit local ~/.heron/examples/heron-examples.jar com.twitter.heron.examples.ExclamationTopology ExclamationTopology --deploy-deactivated --verbose`
<img width="130" alt="heron_admin-0" src="https://cloud.githubusercontent.com/assets/23758/20283689/dc5b7eca-aa6e-11e6-90a8-37a6729684bf.png">

2. Scale up from 4 to 6:
`~/bin/heron update local ExclamationTopology --component-parallelism=exclaim1:6`
<img width="130" alt="heron_admin-1" src="https://cloud.githubusercontent.com/assets/23758/20283695/e0604e1a-aa6e-11e6-8ac4-b75c2fdb8325.png">

3. Scale back down to 4:
`~/bin/heron update local ExclamationTopology --component-parallelism=exclaim1:4`
<img width="149" alt="heron_admin-2" src="https://cloud.githubusercontent.com/assets/23758/20283703/e52b06a6-aa6e-11e6-807a-98ef248bcf0c.png">

Instead the packing in 3 should look like 1. The proposed algorithm is as follows:

First, compute the idealized allocation (representation?) factor for component A on a given container as the ratio of component A's parallelism to the sum of all component parallelisms. For example, after Step 2 above the idealized allocation factor for `exclaim1` would be 6/8 or 0.75, since the topology has 8 total instances of which 6 are As.

Second, compute the actual allocation factor on a given container for component A as the ratio A's  instance count to the count of all instances on the container. Following the example above, after Step 2 the allocation factors for A on each of the nodes is [2/3, 2/3, 2/2] or [0.67, 0.67, 1.0]. 

Finally, the offset from the ideal could be computed as [-0.08, -0.08, +0.25], which shows that container 3 is the most over allocated and should be targeted for instance A removal.

avflor:
@billonahill This is an interesting idea. I think the algorithm can be simplified to work as follows:
[...]

PR #1572, opened 2016-11-18
Add a scoring-based approach to modifying packing plan

billonahill:
Instances are often added and removed to containers based on a ranking of all containers. This adds methods to the PackingPlanBuilder API to make it easy to do so without requiring the caller to rank, sort and iterate through containers. Instead, packing algorithms can now add/remove instances by passing a class that scores Components based on some heuristic. Multiple scorers can be passed for primary, secondary, tertiary, etc ranking.

This patch includes an common implementation that is used for round robin placement, but other custom implementations can be passed to score based on balance, resource utilization, etc.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
@@ -24,7 +24,7 @@ import java.util.logging.Logger;
 import com.twitter.heron.api.generated.TopologyAPI;
 import com.twitter.heron.packing.RamRequirement;
 import com.twitter.heron.packing.ResourceExceededException;
-import com.twitter.heron.packing.builder.IdBasedContainerScorer;
+import com.twitter.heron.packing.builder.ContainerIdScorer;
 import com.twitter.heron.packing.builder.PackingPlanBuilder;
 import com.twitter.heron.packing.utils.PackingUtils;
 import com.twitter.heron.spi.common.Config;
@@ -285,7 +285,7 @@ public class FirstFitDecreasingPacking implements IPacking, IRepacking {
     ArrayList<RamRequirement> ramRequirements =
         getSortedRAMInstances(componentsToScaleDown.keySet());
 
-    IdBasedContainerScorer scorer = new IdBasedContainerScorer(1, this.numContainers);
+    ContainerIdScorer scorer = new ContainerIdScorer();
 
     for (RamRequirement ramRequirement : ramRequirements) {
       String component = ramRequirement.getComponentName();
@@ -306,10 +306,8 @@ public class FirstFitDecreasingPacking implements IPacking, IRepacking {
       planBuilder.updateNumContainers(++numContainers);
     }
 
-    IdBasedContainerScorer scorer = new IdBasedContainerScorer(1, this.numContainers);
-
     try {
-      planBuilder.addInstance(scorer, instanceId);
+      planBuilder.addInstance(new ContainerIdScorer(), instanceId);
     } catch (ResourceExceededException e) {
       planBuilder.updateNumContainers(++numContainers);
       planBuilder.addInstance(numContainers, instanceId);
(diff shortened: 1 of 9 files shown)


billonahill:
Any comments on this one? I think this change will make it much easier for us to compose rankings for how instance placement and removal should occur.

avflor:
@billonahill Will take a look later today

heron/packing/src/java/com/twitter/heron/packing/builder/ContainerIdScorer.java
```diff
@@ -0,0 +1,48 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.twitter.heron.packing.builder;
+
+/**
+ * Sorts containers ascending by container id. Id firstId and maxId as passed, starte with
```
wangli1426: a typo: starte -> start

heron/packing/src/java/com/twitter/heron/packing/binpacking/FirstFitDecreasingPacking.java
```diff
@@ -284,11 +285,13 @@ private void removeInstancesFromContainers(PackingPlanBuilder packingPlanBuilder
     ArrayList<RamRequirement> ramRequirements =
         getSortedRAMInstances(componentsToScaleDown.keySet());
 
+    ContainerIdScorer scorer = new ContainerIdScorer();
```
wangli1426: Is a reverse scorer for instance removal better? IMO, inserting and removing instances among the containers in reverse direct will result in better resource utilization.

wangli1426:
@billonahill I have reviewed this PR. It is smart to use a scorer here. Great work! I have two comments:

1)  Since instance insertion is done by searching the containers from the smallest container id to the largest one until a container with enough resource is found, you may want to remove the instances of the scaling-down component from the containers in a reverse order. In this way, the container with the largest id is more likely to be empty eventually, making it possible to release this empty container in order to improve the resource utilization.

2) If a container become empty after component scaling-down, will the container be released?

Thanks.

billonahill:
@wangli1426 thanks for the comments.

1. Yes, we should improve the scale down logic. In this PR though I'd keep all functionality the same and just changing the implementation to introduce the ranked approach. In a follow up we'll change the scoring logic for scale down, most likely when addressing #1560.
2. Yes, in the `PackingPlanBuilder.pack()` method, empty containers get removed from the built `PackingPlan`. This happens in the `buildContainerPlans()` method, where you'll see that we `continue` in the loop if the number of container instances is zero.

heron/packing/src/java/com/twitter/heron/packing/builder/PackingPlanBuilder.java
```diff
@@ -107,13 +112,80 @@ public PackingPlanBuilder addInstance(Integer containerId,
...
+  }
+
+  @SuppressWarnings("JavadocMethod")
+  /**
+   * Remove an instance from the first container possible ranked by score. If a scoring tie exists,
+   * uses the next scorer in the scorers list to break the tie.
+   * @return containerId of the container the instance was removed from
+   * @throws com.twitter.heron.spi.packing.PackingException if the instance could not be removed
+   */
+  public int removeInstance(List<Scorer<Container>> scorers, String componentName) {
```
avflor: In the RCRR implementation, during scale down we do not always start from the first container. For example if we need to scale down 2 components by 1 instance from containers [1...5] and the first component is removed from instance 2 then the removal for the next component will start from container 3. This will help with load balancing. This remove instance method does not capture that. However, it is fine if a good scoring method is used when we change the scale down approach as discussed in the other thread.

billonahill:
I believe we've captured that used case. See how we use the two different
constructors to the ContainerIDScorer.

heron/packing/src/java/com/twitter/heron/packing/builder/PackingPlanBuilder.java
avflor: @billonahill I'm not sure if that is the same. In the previous example if we pass as input to the constructor container id 3 and containrId 5 then the search will stop at 5, right? Wehat we ideally want is to start from 3 and use this sequence: 4,5,1,2. Is that behavior captured?

billonahill:
Yes, that's why the constructor requires the maxId, so it can loop back to
the lowest and continue.

billonahill:
@avflor @wangli1426 any other comments?

avflor:
@billionahill LGTM

wangli1426:
LGTM

Issue #1567, opened 2016-11-16
Increased stream manager memory usage


ajorgensen:
We've observed what appears to be the stream manager not releasing memory properly or holding onto memory indefinitely on heron `0.14.4`. The following image is the rss of the heron stream manager, the red vertical line represents the same topology deployed on 0.14.4 on the left of the red line and 0.14.1 on the right of the red line. We can see that on 0.14.4 the rss for the heron stream manager grows until it pushed the container above its allocated memory and then forcibly gets killed while the same topology running on 0.14.1 shows a fairly consistent memory usage.

![image](https://cloud.githubusercontent.com/assets/509779/20330834/795ffb50-ab6f-11e6-98fb-ad464a211664.png)

I do not currently have a test case setup that demonstrates this behavior but will work on getting one setup. It appears to only happen on topologies that utilize acking and we have not seen this behavior on those that do not.

congwang:
@ajorgensen How much does the following one-line fix help?
```

diff --git a/heron/common/src/cpp/network/mempool.h b/heron/common/src/cpp/network/mempool.h
index a8467df..875e0ac 100644
--- a/heron/common/src/cpp/network/mempool.h
+++ b/heron/common/src/cpp/network/mempool.h
@@ -75,6 +75,7 @@ class MemPool {
     }
     B* t = pool.back();
     pool.pop_back();
+    pool.shrink_to_fit();
     return static_cast<M*>(t);
   }
 

```

I still can't reproduce it locally with AckingTopology.

ajorgensen:
I can try tomorrow. Ill also try to work out a reproducible test case for you if I can.

congwang:
Just in case, make sure you didn't change the default config for tuple cache:

```
# The frequency in ms to drain the tuple cache in stream manager
heron.streammgr.cache.drain.frequency.ms: 10

# The sized based threshold in MB for draining the tuple cache
heron.streammgr.cache.drain.size.mb: 100
```

ajorgensen:
No we have both of those still at the default. @congwang what would happen if the spout could produce tuples faster than the bolt could consume them and the `maxSpoutPending` value was high? Would the spout know to stop calling `nextTuple` or would the buffer in the stream manager continue to increase until it ran out of memory?

congwang:
The stream manager will send back pressure to the spolt in this case.

ajorgensen:
@congwang were you able to reproduce the issue with the topology I sent you?

congwang:
[...]

PR #1573, opened 2016-11-22
Add a configurable memory cap for mempool

congwang:
Currently mempool is unbounded and memory in mempool grows and settles on steady state. This causes stream manager could use a lot of memory in production. We need a mechanism to add a cap there, the heap usage of stmgr is reduced down to 10% and meanwhile performance is almost not affected at all.

This should resolve https://github.com/twitter/heron/issues/1567


objmagic:
LGTM 👍

kramasamy:
@congwang - how is the performance after moving to size in MB?

congwang:
I don't have a chance to test it in data center yet.

PR #1578, opened 2016-11-23
Follow symbolic links when searching status files.

objmagic:
because in some OS X environment, `find` does not follow symlinks (`bazel-out/`), which means that Bazel will generate empty `release.yaml`, which means that `heron version` will output nothing.


billonahill:
👍

Issue #1579, opened 2016-11-23
Packing algorithms don't check for componentId collisions


billonahill:
The repacking algorithms (i.e., `FirstFitDecreasingPacking` and `ResourceCompliantRRPacking`) assert that the global task ids are unique, but they don't check that the `componentIndex` is unique to all components of a given type. As a result we currently have collisions when scaling.

The fix is to make it so that the packing algorithms don't even need to manage this. This logic can be entirely managed in the `PackingPlanBuilder` and the algos can just add/remove instances by `componentName` instead of by `InstanceId`.

PR #1580, opened 2016-11-23
Add more tests for PackingPlanBuilder

billonahill:
Added a bunch of tests and in doing so found some edge cases and bugs.

- Improved error messaging in a number of places
- change `return boolean` methods to instead return void and throw methods
- better handling of edge cases and defensive programming (adding/removing to/from container that doesn't exist, adding instances with taskId, componentIndex collisions)
- Identified bug #1579 to be fixed in a follow up patch to keep this one from ballooning.
- added a much easier way to test changes to packing (see how new tests in `PackingPlanBuilderTest` use `PackingTestHelper` and `AssertPacking`)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+website/content/docs/contributors/community.md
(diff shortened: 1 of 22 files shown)


kramasamy:
👍

PR #1581, opened 2016-11-28
Fix fd leak in http_unittest

congwang:
We need to close pipe fd's when the test finishes. Also update the comments.


kramasamy:
👍

objmagic:
👍

PR #1582, opened 2016-11-28
Move taskId/componentIndex management into PackingPlanBuilder

### LINKED ISSUE #1579: Packing algorithms don't check for componentId collisions

billonahill:
Fixes #1579.

The current repacking algos can cause collisions with `componentIndex`. This patch fixes that by moving managment of `taskId` and `compoenentIndex` into `PackingPlanBuilder` since the packing algos shouldn't need to handle these.

This has the added benefit of simplifying the `PackingPlanBuilder` API, since add/remove calls can work just with `componentNames`, instead of `InstanceIds`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+website/content/docs/contributors/community.md
--- a/heron/examples/src/java/com/twitter/heron/examples/WordCountTopology.java
+++ b/heron/examples/src/java/com/twitter/heron/examples/WordCountTopology.java
@@ -149,7 +149,6 @@ public final class WordCountTopology {
         Integer val = countMap.get(key);
         countMap.put(key, ++val);
       }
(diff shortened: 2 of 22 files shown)


billonahill:
Any comments on this?

billonahill:
@avflor @ashvina please review when you can.

avflor:
@billonahill This LGTM.

billonahill:
thanks @avflor for reviewing.

PR #1584, opened 2016-11-28
Change scale down to try to free containers

### LINKED ISSUE #1560: Scaling down should remove instances from the container with the most imbalance

billonahill:
Fixes #1560.

Updates `ResourceCompliantRRPacking` to scale down instances from the containers that are most imbalanced first. Also adds a bunch of unit tests for Scorers. The ranking algorithm prioritizes instances for removal as follows:

- Instances from containers with the most homogeneity (e.g., AAAA before AABB).
- When a tie occurs, instances from the container with the fewest instances
- When a tie occurs, instances from the container with the largest it.

Only making the change now to `ResourceCompliantRRPacking` to limit the scope of the PR. Once the approach is agreed upon, I can make similar changes to `FirstFitDecreasingPackingTest` which should also benefit from the same approach.

Best to review the 2 unit tests first to see what's happening.


heron/packing/src/java/com/twitter/heron/packing/builder/HomogeneityScorer.java
```diff
@@ -0,0 +1,51 @@
...
+    if (totalInstances == 0) {
+      return 0;
+    }
+
+    for (PackingPlan.InstancePlan instancePlan : container.getInstances()) {
+      if (componentName.equals(instancePlan.getComponentName())) {
+        totalComponentInstances++;
+      }
+    }
+    return (double) totalComponentInstances / totalInstances;
```
avflor: I think you lose opportunities for removing containers with this scoring function. Let's take a container BBCD and one ABC and you want to remove 1 B. It is better to remove it from ABC than BBCD because you maximize the probability to remove this container by doing that. The current scoring function will remove from BBCD. I think you need to categorize containers and use total number of instances as the scoring function within each category. I have another example in the discussion thread. In general optimizing for imbalance will not let you remove containers easily. You can cover all cases by using a two-step approach as discussed in #1560

billonahill: As implemented we're trying to achieve balance while also removing containers where we can (despite how I named this PR). So I would argue that we would want to go from [BBCD, ABC] to [BCD, ABC] which is a more balanced state. To expand on this example, if we instead had [BBBBBBCD, ABC], we should try to removing as many Bs as possible from the first container to reduce our imbalance before trying to remove the seconds container.

avflor: I think there is a confusion here. The imbalance metric you have will differentiate correctly between homogeneous and heterogeneous containers (e.g., [BBCD, BBB]). But it will not work well in a set of mixed containers. The reason is that usually the imbalance metric will try many times to remove from the most loaded container which mean that you reduce significantly the chances to remove containers overall. In your example with BBBBBBCD, ABC you have more chances getting rid of a container if you remove always from the least loaded one. Your solution currently does not remove multiple components at once. You remove one component at a time. In this setting, removing from the least loaded container will typically result in fewer containers overall. 

If you want to optimize scale down for imbalance then this approach is ok. But I personally believe that users would want to see resources been released when scaling down otherwise why would they even do it?

ashvina: @billonahill - If a user requests to remove an instance of A from [AB, AABCD], this score will return [.5, .4], and it will result in [B, AABCD].

billonahill:
@avflor I've updated the algorithm I believe as you've proposed. Please review the unit tests to see if I've captured all the examples we've discussed.

kramasamy:
@avflor - can you please check and let me know?

billonahill:
@avflor or @ashvina any comments on this?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/ResourceCompliantRRPacking.java
```diff
@@ -334,9 +339,14 @@ private void removeInstancesFromContainers(PackingPlanBuilder packingPlanBuilder
...
    */
   private void removeRRInstance(PackingPlanBuilder packingPlanBuilder,
-                                String component) throws RuntimeException {
-    ContainerIdScorer scorer = new ContainerIdScorer(this.containerId, this.numContainers);
-    this.containerId = nextContainerId(packingPlanBuilder.removeInstance(scorer, component));
+                                String componentName) throws RuntimeException {
+    List<Scorer<Container>> scorers = new ArrayList<>();
+    scorers.add(new HomogeneityScorer(componentName, true));  // all-same-component containers first
+    scorers.add(new InstanceCountScorer());                   // then fewest instances
+    scorers.add(new HomogeneityScorer(componentName, false)); // then least homogeneous
```
billonahill: Good catch. That's a bug in my comment, we are doing most homogeneous first. Updated.

avflor:
@billonahill Thanks Bill. The code LGTM.

PR #1585, opened 2016-11-28
Use BaseMemPool for tuple cache

congwang:
Use BaseMemPool directly instead of implementing it again. No functional change.


kramasamy:
👍

PR #1586, opened 2016-11-28
Adding new BytesAmount class for use by Resource

### LINKED ISSUE #1576: Model bytes/megabytes as a typed object in the packing classes
Ram and disk space are represented as longs of bytes in the `Resource` class and in all of it's many consumers. This requires manipulation of those bytes by multiplying and dividing `Constants.MB` to the size values in many place in the code, which is tedious and error prone. I forgot to multiply by `Constants.MB` once which without a proper unit test to catch the my mistake, could have caused a really bad bug where the computed number of containers required was (O)10k.

The work involved is to replace all longs that refer to bytes/megabytes with a typed object that clearly and consistently handles units. If done properly `Constants.MB` could be remove entirely.

billonahill:
Fixes #1576

The key classes to review are:
```
heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
heron/common/tests/java/com/twitter/heron/common/basics/ByteAmountTest.java
heron/spi/src/java/com/twitter/heron/spi/packing/Resource.java
```
The rest are the downstream refactors to replace `longs` with `ByteAmounts`. Also I removed the `Constants` class since it's no longer needed, since `MB` and `GB` are no longer used.

Please review and provide feedback on the design, naming, etc of the proposed `BytesAmount` class. 

Passing a long for bytes results in a lot of unit conversion math and is subject to hard to track bugs. If instead we create a class to represent bytes and their units, heron code will be more robust and byte math can be simplified into methods on the object.


heron/spi/src/java/com/twitter/heron/spi/common/ByteAmount.java
```diff
@@ -0,0 +1,111 @@
...
+  public long asBytes() {
+    return bytes;
+  }
+
+  public long asMegabytes() {
+    return bytes / Constants.MB;
+  }
+
+  public long asGigabytes() {
+    return bytes / Constants.GB;
```
objmagic: Divide two `long`s could yield 0, which I guess is not what we want to see...?

billonahill: Yes, that needs to be fixed. It should divide and round. Will update that.

billonahill:
Any other comments on the propose new class?

heron/spi/src/java/com/twitter/heron/spi/common/ByteAmount.java
```diff
@@ -56,6 +51,22 @@ public ByteAmount minus(ByteAmount other) {
...
+  public ByteAmount plus(ByteAmount other) {
+    return ByteAmount.fromBytes(this.asBytes() + other.asBytes());
+  }
+
+  public ByteAmount multiply(int factor) {
+    return ByteAmount.fromBytes(this.asBytes() * factor);
+  }
+
+  public ByteAmount increaseBy(int percentage) {
+    return ByteAmount.fromBytes(asBytes() * (1 + (percentage / 100)));
```
objmagic: what's the usage of this function since the one below looks perfect?

billonahill: Great point, removed.

heron/spi/src/java/com/twitter/heron/spi/common/ByteAmount.java
```diff
@@ -0,0 +1,127 @@
...
+    return ByteAmount.fromBytes(this.asBytes() * factor);
+  }
+
+  public ByteAmount divide(int factor) {
+    return ByteAmount.fromBytes(this.asBytes() / factor);
+  }
+
+  public ByteAmount increaseBy(double percentage) {
+    return ByteAmount.fromBytes(Math.round(asBytes() * (1 + (percentage / 100))));
+  }
```
objmagic: should be the last comment so bear with me...

Do we also need a `decreaseBy`, maybe it's useful when we scale down?

billonahill: I haven't yet found that computation done in the code, but if I do, I'll add it. Otherwise I say we leave it out until it's needed.

objmagic: ok 👍

objmagic:
fixed the CI and then 👍

billonahill:
Thanks @objmagic for the reviews! CI is broken because I now need to refactor a ton of downstream code that is broken due to the proposed changes. I'll make that refactor and resubmit.

billonahill:
Just pushed the large refactor of all the downstream classes. I would recommend reviewing the following 3 classes primarily:
```
heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
heron/common/tests/java/com/twitter/heron/common/basics/ByteAmountTest.java
heron/spi/src/java/com/twitter/heron/spi/packing/Resource.java
```

@ashvina you might want to check on the yarn classes to check on my work there:
```
heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronMasterDriver.java
heron/schedulers/tests/java/com/twitter/heron/scheduler/yarn/HeronMasterDriverTest.java
```

kramasamy:
@ashvina - can you please check?

heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
```diff
@@ -0,0 +1,247 @@
...
+   * @param other ByteValue to subtract
+   * @return this ByteValue minus other ByteValue
+   * @throws IllegalArgumentException if subtraction would overshoot Long.MIN_VALUE
+   */
+  public ByteAmount minus(ByteAmount other) {
+    if (Long.MIN_VALUE + other.asBytes() > asBytes()) {
+      throw new IllegalArgumentException(String.format(
+          "Subtracting %s from %s would overshoot Long.MIN_LONG", other, this));
+    }
+    return ByteAmount.fromBytes(asBytes() - other.asBytes());
```
ashvina: The method is returning a new object. However, this object should be reduced as per the javadoc?

billonahill: I think ByteAmounts should be immutable, so returning a new one makes sense. I'll update the javadoc to reflect that

heron/schedulers/tests/java/com/twitter/heron/scheduler/yarn/HeronMasterDriverTest.java
```diff
@@ -315,9 +316,9 @@ private void verifyFittingContainer(Set<HeronMasterDriver.HeronWorker> container
   @Test
   public void fitBiggestContainerIgnoresCoresIfMissing() {
     Set<HeronMasterDriver.HeronWorker> workers = new HashSet<>();
-    workers.add(new HeronMasterDriver.HeronWorker(1, 3, 3 * 1024));
+    workers.add(new HeronMasterDriver.HeronWorker(1, 3, ByteAmount.fromMegabytes(3)));
```
ashvina: ByteAmount.fromGigabytes(3)

heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  /**
+   * Converts the number of bytes to megabytes, rounding if there is a remainder. Because of loss
+   * of precision due to rounding, it's strongly advised to only use this method when it is certain
+   * that megabytes were using during the construction of the object via fromMegabytes. Only In that
+   * scenario is there a guarantee that precision will not be lost when calling this method.
+   * @return returns the ByteValue in MBs or 0 if the value is < (1024 * 1024) / 2
+   */
+  public long asMegabytes() {
```
ashvina: what do you think about using abbreviations in this method, asMB() ?

ashvina:
LGTM 👍

heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Converts the number of bytes to gigabytes, rounding if there is a remainder. Because of loss
+   * of precision due to rounding, it's strongly advised to only use this method when it is certain
+   * that gigabytes were using during the construction of the object via fromGigabytes. Only In that
+   * scenario is there a guarantee that precision will not be lost when calling this method.
+   * @return returns the ByteValue in GBs or 0 if the value is < (1024 * 1024 * 1024) / 2
+   */
+  public long asGigabytes() {
+    return Math.round((double) bytes / GB);
```
ashvina: Agree. 
We could return float or double from these methods, and let the caller perform context specific transformation?

heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
billonahill: I debated this one as well. I like the conciseness but not the resulting inconsistency between asBytes, asMB and asGB.

heron/schedulers/tests/java/com/twitter/heron/scheduler/yarn/HeronMasterDriverTest.java
billonahill: great catch, thanks.

heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java
billonahill: Another thought is we could add an enum to the signature where the caller specifies how the rounding should occur: ceil, floor, round, or strict where we fail if there is a remainder. That way the caller has to acknowledge the potential for rounding and decide how rounding should be handled.

ashvina: Works for me.

billonahill: Cool. Since this patch is already large and at risk of drifting out of sync with master, I'd like to commit as-is and address this in a follow-up.

PR #1589, opened 2016-11-30
Extending S3 updater

fabianmenges:
Changes:
Upgrading AWS library
Upgrading and extending the S3 updater
Fixing proxy code
Adding commons-codec dependency for AWS

All changes should be compatible with existing configurations.

The current S3 updater requires hard coded credentials. These changes allow to the S3 updater to just use the AWS [Default Credential Provider Chain](http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html). In addition it allows the configuration of an AWS Profile to signal which credentials to use from an AWS credentials file.

The Proxy code of the current version only supported reverse proxies, now reverse proxies and authenticated http proxies are supported.

`commons-codec` is required by the AWS library for authenticated proxies.
I upgraded to the latest AWS library version, which simplified usage of `CredentialsProvider`s.

### DIFF
--- a/heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
+++ b/heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
@@ -17,20 +17,23 @@ package com.twitter.heron.uploader.s3;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import com.amazonaws.AmazonClientException;
+
 import com.amazonaws.ClientConfiguration;
-import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.Protocol;
+import com.amazonaws.SdkClientException;
+
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.BasicAWSCredentials;
-import com.amazonaws.regions.Region;
-import com.amazonaws.regions.Regions;
-import com.amazonaws.services.s3.AmazonS3Client;
-import com.amazonaws.services.s3.S3ClientOptions;
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 
 import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.Context;
@@ -39,8 +42,9 @@ import com.twitter.heron.spi.uploader.IUploader;
 /**
  * Provides a basic uploader class for uploading topology packages to s3.
  * <p>
- * By default this uploader will write topology packages to s3://&lt;bucket&gt;/&lt;topologyName&gt;/topology.tar.gz. This is a known
- * location where you can then download the topology package to where it needs to go in order to run the topology.
+ * By default this uploader will write topology packages to s3://&lt;bucket&gt;/&lt;topologyName&gt;/topology.tar.gz
+ * trying to obtain credentials using the default credential provider chain. The package destination serves as known
+ * location which can be used to download the topology package in order to run the topology.
  * <p>
  * This class also handles the undo action by copying any existing topology.tar.gz package found in the folder to
  * previous_topology.tar.gz. In the event that the deploy fails and the undo action is triggered the previous_topology.tar.gz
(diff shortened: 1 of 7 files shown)


heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -79,52 +84,68 @@ public void initialize(Config config) {
...
       throw new RuntimeException("Missing heron.uploader.s3.bucket config value");
     }
 
-    if (accessKey == null || accessKey.isEmpty()) {
-      throw new RuntimeException("Missing heron.uploader.s3.access_key config value");
-    }
+    // If an accessKey is specified, use it. Otherwise check if an aws profile
+    // is specified. If neither was set just use the DefaultAWSCredentialsProviderChain
+    // by not specifying a CredentialsProvider.
+    if ((accessKey != null && !accessKey.isEmpty())
```
billonahill: If we add a helper method in this class to do `foo != null && !foo.isEmpty()` it would improve the readability of this code:

```
private static boolean isEmpty(String... value) { ... }
```

We could then simplify these conditionals to do:
```
if (!isEmpty(accessKey, accessSecret) {
  // set credentials
} else if (!isEmpty(awsProfile) {
  // set creentials
} else ...
```

heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -79,52 +84,68 @@ public void initialize(Config config) {
...
-    if (uri != null && !uri.isEmpty()) {
-      URI uriObj;
+    if (proxy != null && !proxy.isEmpty()) {
+      URI proxyUri;
 
       try  {
-        uriObj = new URI(uri);
+        proxyUri = new URI(proxy);
       } catch (URISyntaxException e) {
         throw new RuntimeException("Invalid heron.uploader.s3.uri config value");
```
billonahill: please include proxy in the exception message here

heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -169,7 +190,7 @@ public URI uploadPackage() {
     }
 
     try {
-      return new URI(resourceUrl);
+      return resourceUrl.toURI();
     } catch (URISyntaxException e) {
       LOG.log(Level.SEVERE, e.getMessage());
```
billonahill: please include a helpful error message here, including the bad field value

kramasamy:
@fabianmenges - the unit tests failed. Can you please take a look?

kramasamy:
@ajorgensen - Can you please take a look at this PR since your team uses it heavily?

fabianmenges:
@billonahill Thanks for your feedback. I'm a Twitter employee, I probably don't need to accept the contributors CLA.

billonahill:
Whoops, sorry bout that @fabianmenges!

kramasamy:
@billonahill - wondering if your feedbacks have been taken care off, before I merge it?

heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -79,52 +86,67 @@ public void initialize(Config config) {
...
-      URI uriObj;
+    if (!Strings.isNullOrEmpty(proxy)) {
+      URI proxyUri;
 
       try  {
-        uriObj = new URI(uri);
+        proxyUri = new URI(proxy);
       } catch (URISyntaxException e) {
-        throw new RuntimeException("Invalid heron.uploader.s3.uri config value");
+        throw new RuntimeException("Invalid heron.uploader.s3.proxy_uri config value");
```
billonahill: please include the invalid value in the exception

heron/uploaders/src/java/com/twitter/heron/uploader/s3/sample.yaml
```diff
@@ -9,13 +9,24 @@ heron.uploader.s3.bucket:     heron-topologies-company-com
...
-# Credentials that have write access to the bucket specified above
-heron.uploader.s3.access_key: access_key
-heron.uploader.s3.secret_key: secret_access_key
+# AWS Credentials
+# By default the S3 Uploader will use the Default Credential Provider Chain for accessing the S3 bucket -
+# see http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#using-the-default-credential-provider-chain
+# However you have the option to directly specify aws credentials:
+# heron.uploader.s3.access_key: access_key
+# heron.uploader.s3.secret_key: secret_access_key
+# If you have multiple AWS profiles in your credentials file (~/.aws/credentials)
```
billonahill: We should clarify that if both access_key, secret_key and aws_profile are all included the latter will be ignored. Or probably better to throw an exception if all 3 are specified, just to make sure things are clear. That's consistent with how we throw an exception if just aws_profile and access_key are set for example.

kramasamy:
@billonahill - wondering if your review recommendations have been taken care?

heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -100,6 +100,12 @@ public void initialize(Config config) {
     // is specified. If neither was set just use the DefaultAWSCredentialsProviderChain
     // by not specifying a CredentialsProvider.
     if (!Strings.isNullOrEmpty(accessKey) || !Strings.isNullOrEmpty(accessSecret)) {
+
+      if (!Strings.isNullOrEmpty(awsProfile)) {
+        throw new RuntimeException("Supply either provide heron.uploader.s3.access_key, "
```
billonahill: I think what we want to convey her is that the user must specify either a profile or a key/secret, but not both. Also the wording is off a bit. We should either use supply or provide but not both.

heron/uploaders/src/java/com/twitter/heron/uploader/s3/S3Uploader.java
```diff
@@ -120,7 +126,7 @@ public void initialize(Config config) {
       try  {
         proxyUri = new URI(proxy);
       } catch (URISyntaxException e) {
-        throw new RuntimeException("Invalid heron.uploader.s3.proxy_uri config value");
+        throw new RuntimeException("Invalid heron.uploader.s3.proxy_uri config value", e);
```
billonahill: Please include `proxy` in the message.

kramasamy:
@fabianmenges - can you please review the changes @billonahill proposed?

fabianmenges:
@billonahill fixed, thx.

PR #1590, opened 2016-11-30
Add a configurable memory cap for mempool

### LINKED ISSUE #1567: Increased stream manager memory usage

congwang:
Currently mempool is unbounded and memory in mempool grows and settles on steady state. This causes stream manager could use a lot of memory in production. We need a mechanism to add a cap there, the memory usage of stmgr is reduced down to 10% and meanwhile performance is almost not affected at all.

Note, I use a global limit for all instances of BaseMemPool (including MemPool), this relies on the C++ type system.

This should resolve #1567

### DIFF
--- a/heron/common/src/cpp/config/heron-internals-config-reader.cpp
+++ b/heron/common/src/cpp/config/heron-internals-config-reader.cpp
@@ -200,8 +200,8 @@ sp_int32 HeronInternalsConfigReader::GetHeronStreammgrCacheDrainSizeMb() {
   return config_[HeronInternalsConfigVars::HERON_STREAMMGR_CACHE_DRAIN_SIZE_MB].as<int>();
 }
 
-sp_int32 HeronInternalsConfigReader::GetHeronStreammgrMempoolSize() {
-  return config_[HeronInternalsConfigVars::HERON_STREAMMGR_MEMPOOL_SIZE].as<int>();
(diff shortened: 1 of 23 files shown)


ajorgensen:
I deployed one of our topologies with this change and I think we're still seeing the stream manager memory usage spike up.

image

Anything I can do to dig into where its coming from?

srkukarni:
Is this pr still valid?

ajorgensen:
I think it was closed because it turned out to not solve the original issue, im not sure if the commits here can still be used but likely that can.

congwang:
Right, the original back-pressure-related memory issue is addressed by https://github.com/twitter/heron/pull/1785.

About this PR itself, the mempool does not contribute to memory pressure, we will hit the hard limit anyway no matter we have mempool or not, the problem with mempool is that it may hit the soft limit eventually since this memory never shrinks.

PR #1217, opened 2016-08-04
move to protobuf version 3

kramasamy:

PR #1591, opened 2016-12-02
Update protobuf to 3.1.0

### LINKED ISSUE #1183: Upgrade Protobuf version
Protobuf team just [released version 3.0](https://github.com/google/protobuf/releases/tag/v3.0.0). Improvement of Python implementation looks promising.

congwang:
Update protobuf from 2.5.0 to 3.1.0 to get potentially better performance. Meanwhile, still keep the proto2 syntax.

resolves #1183


third_party/protobuf/BUILD
```diff
@@ -15,6 +15,20 @@ genrule(
     outs = [
         "bin/protoc",
 
+        "include/google/protobuf/arena.h",
```
billonahill: How were you able to determine the long list of required includes here? Could you provide a link in a comment?

billonahill:
Serialized protobuf objects are stored in the Heron state manager for the duration of the topology livecycle. Is this change backwards compatible so the new code could continue to read protobuf objects stored with the previous code?

third_party/protobuf/protobuf-3.1.0.patch
```diff
@@ -0,0 +1,29 @@
+diff -Nrup protobuf-3.1.0.orig/src/google/protobuf/io/zero_copy_stream_impl_lite.h protobuf-3.1.0/src/google/protobuf/io/zero_copy_stream_impl_lite.h
```
congwang: This patch already exists for 2.5.0, I just regenerate it for 3.1.0. The reasons why we need it are:

1) The C++11 version is faster, it saves an if check on super hot path.
2) We compile our stream manager with C++11, so it is safe to just remove the macro
3) I don't see any configure parameter to enable it, so we have to patch the source directly.

Not sure if it is really worth a README, since we don't change the patch often. It is generated manually actually.

congwang:
@billonahill I think the API and ABI compatibility is a minimum requirement for protobuf as a library. I didn't touch the new incompatible API under proto3 (unlike PR #1217 ).

billonahill:
@congwang can you verify the compatibility by deploying heron-tracker with the new version and trying to read data for all of the currently persisted fields? A quick search doesn't reveal anything in the proto3 docs that clarifies ABI compatibility so we should test on our own.

https://developers.google.com/protocol-buffers/docs/overview#introducing-proto3

congwang:
@billonahill I am confused, we don't use proto3 even with this PR, so why should I verify proto3 compatibility? I explicitly specify proto2 in the code actually.

third_party/protobuf/protobuf-3.1.0.patch
congwang: Will do it in a separate PR.

billonahill:
Maybe I'm confused, is protobuf 3.1.0 not proto3?

Either way, do we have any protobuf documentation that 3.x is guaranteed to be ABI compat with 2.x? Often times major semver changes are specific not. Without that we should test this change.

congwang:
@billonahill protobuf-3.x has proto2 which is compatible with protobuf-2.x, and also proto3 which is an incompatible API. We only use proto2, so I don't think there is any incompatible issue.

objmagic:
so the proto2 of protobuf-3.x also has improvement? 

And we should run internal integration test to make sure everything works fine as before. Also, it would be great to give some concrete numbers...

kramasamy:
@billonahill @objmagic - According to what I have read proto2 and proto3 are not compatible (from binary perspective). But @congwang is using proto2 binary format using proto3 version - which should be binary compatible. @congwang - I would suggest to run a test whether they are compatible or not. Also, it will be great to quantify the performance improvement.

kramasamy:
@congwang - any update?

rohanag12:
Any updates on this?  
Protobuf 3 support would be nice to have. We have a few protobufs that are written in the proto 3 syntax which we would like to use with Heron.

srkukarni:
@rohanag12 This is slated for the 0.15.3 release.

srkukarni:
BTW any users of Heron can use proto2 or proto3. This ticket is just for the internal implementation of certain Heron components and does not have any ramifications from user perspective.

nicknezis:
Latest Heron uses Protobuf 3.8.0. Closing this PR. Let me know if anything from this PR is still needed. Thanks.

Issue #1597, opened 2016-12-08
Duplicated code is found in SubmitterMain.java and SchedulerConfig.java


mycFelix:
Hi

There is a duplicated code in [SubmitterMain.java](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java#L66) and [SchedulerConfig.java](https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerConfig.java#L48).

Is it necessary to add an new method named `FileUtils.getPkgType(String topologyBinaryFile)` in `FileUtils`?   If so, I'll send PR for this.

kramasamy:
Sure - go for it.

objmagic:
resolved via #1599 and #1609

PR #1599, opened 2016-12-08
Remove duplicated code in SubmitterMain and SchedulerConfig

mycFelix:
As #1597 mentioned, the duplicated code is found in SubmitterMain and SchedulerConfig. 

This PR is trying to fix it and the changes are following:

* add an new method ` String getPkgType(String topologyBinaryFile)` in FileUtils
* invoke `getPkgType` method in SubmitterMain
* invoke `getPkgType` method in SchedulerConfig

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
+++ b/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
@@ -123,4 +123,17 @@ public final class FileUtils {
   public static String getBaseName(String file) {
     return new File(file).getName();
   }
+
+  public static String getPkgType(String topologyBinaryFile) {
+    String pkgType;
+    String basename = FileUtils.getBaseName(topologyBinaryFile);
+    if (FileUtils.isOriginalPackagePex(basename)) {
+      pkgType = "pex";
+    } else if (FileUtils.isOriginalPackageJar(basename)) {
+      pkgType = "jar";
+    } else {
+      pkgType = "tar";
+    }
+    return pkgType;
+  }
 }
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
@@ -60,16 +60,7 @@ public class SubmitterMain {
   protected static Config topologyConfigs(
       String topologyPackage, String topologyBinaryFile, String topologyDefnFile,
       TopologyAPI.Topology topology) {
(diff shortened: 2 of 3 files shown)


objmagic:
👍

heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
```diff
@@ -123,4 +123,17 @@ public static boolean isOriginalPackagePex(String packageFilename) {
   public static String getBaseName(String file) {
     return new File(file).getName();
   }
+
+  public static String getPkgType(String topologyBinaryFile) {
```
billonahill: Thanks for the removal of dup code!

Could you also please change to not abbreviate pkg (e.g., packageType and getPackageType).

Also instead of returning strings, better to use an enum, maybe as an inner enum in FileUtils:
```
enum PackageType {
  PEX,
  JAR,
  TAR
}
```

objmagic: sorry for merging this too fast. I'll do this.

objmagic: actually, returning enum might not be what we want, since we need to put value as string into config. See: https://github.com/twitter/heron/pull/1599/files#diff-1043de1ec8f8e4c0194d4e6035e654cfL80

objmagic: where do you think is appropriate to put this? I am really bad at finding an appropriate class to add code😔

mycFelix: @objmagic - hi, Is there anything I could do? The method name which @billonahill mentioned is better . However, I don't think adding an new enum class is a good idea. Perhaps we could add new static constants in FileUtils instead of handling string variable explicitly, for example

```
private static final String PEX = "pex";
```

objmagic: any reason why using enum is not a good idea?

mycFelix: @objmagic - Cause `String pkgType` is used for Config's key. Just in my opion, this scene is applicable to the principle of Occam's razor -- `If not necessary, do not increase the entity`

objmagic: @mycFelix imo Bill has better arguments. I will submit a PR to use enum.

PR #1601, opened 2016-12-08
Separate config params for SCP/SSH options and target host

chgl:
Currently, the `heron.uploader.scp.command.options` config value is incorrectly used to build and execute a SCP command, which causes the topology package upload to fail when using the ScpUploader. 

A correct scp invocation looks like this: `scp <options> <src> <user@host>:<dest>`
The current implementation of ScpController formats the command line like this: `scp <src> <heron.uploader.scp.command.options>:<dest>`

Because scp expects options to be set before specifying the source file, the upload fails if `heron.uploader.scp.command.options` contains both options and a user@host pair. Note that [the documentation](http://twitter.github.io/heron/docs/operators/deployment/uploaders/scp/) gives an incorrect example.

This PR aims to fix this issue by separating the SCP options and user@host pair into two distinct configuration parameters which are then used by the ScpController to format a correct command line.

I'm still unsure whether or not the name `connection` for the user@host pair is appropriate...

### DIFF
--- a/heron/uploaders/src/java/com/twitter/heron/uploader/scp/ScpContext.java
+++ b/heron/uploaders/src/java/com/twitter/heron/uploader/scp/ScpContext.java
@@ -18,18 +18,30 @@ import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.Context;
 
 public class ScpContext extends Context {
-  public static final String HERON_UPLOADER_SCP_COMMAND = "heron.uploader.scp.command.options";
-  public static final String HERON_UPLOADER_SSH_COMMAND = "heron.uploader.ssh.command.options";
+  public static final String HERON_UPLOADER_SCP_OPTIONS = "heron.uploader.scp.command.options";
+  public static final String HERON_UPLOADER_SCP_CONNECTION =
+      "heron.uploader.scp.command.connection";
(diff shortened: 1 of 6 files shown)


supunkamburugamuve:
@chgl Thanks for the pull request. Could you update the sample.yaml with the new configurations as well?

objmagic:
1. how about using `destination` instead of `connection?
2. please update [`sample.yaml`](https://github.com/twitter/heron/blob/master/heron/uploaders/src/java/com/twitter/heron/uploader/scp/sample.yaml)
3. please sign CLA at https://engineering.twitter.com/opensource/cla

chgl:
Oops, didn't see the sample.yaml. Updated now. @objmagic I signed the CLA before with my GitHub username, just to clarify: do I need to resign for every PR (with the URL)?

I'm not sure how I feel about `destination`, seems confusing in the context of copying files/scp.

kramasamy:
@chgl - if you have signed a CLA before there is no need to sign again.

kramasamy:
Looks like you are committing to this code base, first time.

kramasamy:
@supunkamburugamuve - everything looks ok?

supunkamburugamuve:
@kramasamy changes are looking great. thanks @chgl

kramasamy:
👍

PR #1606, opened 2016-12-08
Uses consistent scale-down algorithm for both IRepacking classes

billonahill:
Update `FirstFitDecreasingPacking` to use the same scale down approach as we've added to RCRRP.

In doing this I wanted to share scale-down unit tests across the two so I created CommonPackingTests. This had the added benefit of sharing a lot of other duplicate test code that already existed in both sets of tests.


billonahill:
@avflor, @ashvina please review when you have a chance

avflor:
@billonahill LGTM

Issue #1598, opened 2016-12-08
Replace `Charset.forName("UTF-8")` to `StandardCharsets.UTF_8`


mycFelix:
Hi，

Is it necessary to replace `Charset.forName("UTF-8")` or some codes like `new InputStreamReader(is, "UTF-8")` to the constant variable `StandardCharsets.UTF_8` which JDK 7 has provided

```java
public final class StandardCharsets {
 ……………………

/**
     * Eight-bit UCS Transformation Format
     */
    public static final Charset UTF_8 = Charset.forName("UTF-8");
………………
}
```

kramasamy:
@maosongfu - can you respond?

billonahill:
+1, better to use these constants than the string version.

mycFelix:
@kramasamy @maosongfu @billonahill - I'll send PR for this. And `UnsupportedEncodingException` doesn't need to be captured if we use `StandardCharsets.UTF_8`

mycFelix:
merged #1607 and closed

dwy189:
Why shall we Replace `Charset.forName("UTF-8")` to `StandardCharsets.UTF_8`?

billonahill:
It's more strongly typed and has compile time checks built in.

PR #1607, opened 2016-12-09
using constant instead of utf8 string in heron package

mycFelix:
As #1598 mentioned, this PR is trying to use constant instead of `UTF-8` string on Java side in heron package. I've just modify the code in heron package, the contrib package is not included. If so, I can modify them as well. I'm not sure if I've found all the `UTF-8` strings in heron package. Please let me know if I have any omissions.

* remove `import java.nio.charset.Charset` and add `java.nio.charset.StandardCharsets`
* replace `Charset.forName("UTF-8")` to `StandardCharsets.UTF_8`

There are four classes included:

* heron/spi/utils/SchedulerUtils.java
* heron/scheduler/aurora/AuroraScheduler.java
* heron/spi/utils/ShellUtils.java
* heron/integration_test/common/HdfsHelper.java

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
@@ -14,7 +14,7 @@
 
 package com.twitter.heron.scheduler.aurora;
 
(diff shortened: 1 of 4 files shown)


billonahill:
👍

PR #1608, opened 2016-12-09
use startsWith to filter prefix of metric name

tobecontinued:
MetricsFilter Specifies the metric-prefix or metric-name we need to keep, so startsWith is more suitable than contains to check  prefix of metricName.

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/metricsmgr/metrics/MetricsFilter.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/metricsmgr/metrics/MetricsFilter.java
@@ -40,7 +40,7 @@ public class MetricsFilter {
 
   public boolean contains(String metricName) {
     for (String prefix : prefixToType.keySet()) {
-      if (metricName.contains(prefix)) {
+      if (metricName.startsWith(prefix)) {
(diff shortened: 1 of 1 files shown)


billonahill:
This class seems to interchange prefix with contains quit a bit, in both the comments, the method naming and the impl. I don't think this proposed change is safe since contains is also used to match on named, which I suspect are not prefixes. @maosongfu thoughts?

We need to have unit test for this class.

tobecontinued:
@billonahill 
Using prefix to filter the metric is better than using contains to. Suppose a case,  there is a topology used to monitor some jvm servers. it has a stream named 'jvm-memory-used'  connect to a blot colletcting these servers' memory-used metrics. Any metric generated by the blot(eg .__emit_count/jvm-memory-use )will flood to sink which it may just want ot collect jvms' metrics (like __jvm-memory-use， jvm-uptime-sec and etc)  of instances of topology.
The MetricsFilter::contains means whether the metric's prefix is in prefix set we shoudl pass, its meaning is very different from meaning  of String::contains, So it doesn't conflit with the purpose that specifing the metric-prefix or metric-name we need to keep.

PR #1609, opened 2016-12-09
Use enum for package type

objmagic:
This PR refactors so that enum is used for topology package type. It addresses discussions at https://github.com/twitter/heron/pull/1599#discussion_r91573936

Follow-up of #1599 

cc @billonahill @mycFelix

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
+++ b/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
@@ -120,20 +120,25 @@ public final class FileUtils {
     return packageFilename.endsWith(".pex");
   }
 
+  public static boolean isOriginalPackageTar(String packageFilename) {
+    return packageFilename.endsWith(".tar");
+  }
+
   public static String getBaseName(String file) {
     return new File(file).getName();
   }
 
-  public static String getPkgType(String topologyBinaryFile) {
-    String pkgType;
+  public static PackageType getPackageType(String topologyBinaryFile) {
     String basename = FileUtils.getBaseName(topologyBinaryFile);
     if (FileUtils.isOriginalPackagePex(basename)) {
-      pkgType = "pex";
+      return PackageType.PEX;
     } else if (FileUtils.isOriginalPackageJar(basename)) {
-      pkgType = "jar";
+      return PackageType.JAR;
+    } else if (FileUtils.isOriginalPackageTar(basename)) {
+      return PackageType.TAR;
     } else {
-      pkgType = "tar";
+      throw new RuntimeException(String.format("Unknown package type of file: %s",
+          topologyBinaryFile));
     }
-    return pkgType;
   }
 }
--- a/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
+++ b/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
@@ -112,33 +112,8 @@ public final class FileUtils {
     return isDirectoryExists(file) && new File(file).list().length > 0;
   }
 
-  public static boolean isOriginalPackageJar(String packageFilename) {
-    return packageFilename.endsWith(".jar");
-  }
-
-  public static boolean isOriginalPackagePex(String packageFilename) {
-    return packageFilename.endsWith(".pex");
-  }
-
-  public static boolean isOriginalPackageTar(String packageFilename) {
-    return packageFilename.endsWith(".tar");
-  }
-
   public static String getBaseName(String file) {
     return new File(file).getName();
   }
 
-  public static PackageType getPackageType(String topologyBinaryFile) {
-    String basename = FileUtils.getBaseName(topologyBinaryFile);
-    if (FileUtils.isOriginalPackagePex(basename)) {
-      return PackageType.PEX;
-    } else if (FileUtils.isOriginalPackageJar(basename)) {
-      return PackageType.JAR;
-    } else if (FileUtils.isOriginalPackageTar(basename)) {
-      return PackageType.TAR;
-    } else {
-      throw new RuntimeException(String.format("Unknown package type of file: %s",
-          topologyBinaryFile));
-    }
-  }
 }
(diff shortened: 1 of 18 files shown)


heron/common/src/java/com/twitter/heron/common/basics/PackageType.java
```diff
@@ -0,0 +1,36 @@
...
+/***
+ * This enum defines commands topology package type
+ */
+public enum PackageType {
+  PEX,
+  JAR,
+  TAR;
+
+  @Override
+  public String toString() {
```
billonahill: Why is a custom toString necessary? IIRC the enum toString prints nicely.

heron/common/src/java/com/twitter/heron/common/basics/PackageType.java
```diff
@@ -0,0 +1,36 @@
...
+    switch (this) {
```
billonahill: `this.name()` does this

heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
```diff
@@ -120,20 +120,25 @@ public static boolean isOriginalPackagePex(String packageFilename) {
...
+
   public static String getBaseName(String file) {
     return new File(file).getName();
   }
 
-  public static String getPkgType(String topologyBinaryFile) {
-    String pkgType;
+  public static PackageType getPackageType(String topologyBinaryFile) {
     String basename = FileUtils.getBaseName(topologyBinaryFile);
     if (FileUtils.isOriginalPackagePex(basename)) {
```
billonahill: This if statement can be replaced with a static factory method on the enum

```
public enum PackageType {
  PEX,
  JAR,
  TAR
  
  public static PackageType getType(String filename) {
    ...
  }
}
```

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -186,7 +186,7 @@ protected String formatJavaOpts(String javaOpts) {
     auroraProperties.put("COMPONENT_RAMMAP", Runtime.componentRamMap(runtime));
     auroraProperties.put("COMPONENT_JVM_OPTS_IN_BASE64",
         formatJavaOpts(TopologyUtils.getComponentJvmOptions(topology)));
-    auroraProperties.put("TOPOLOGY_PACKAGE_TYPE", Context.topologyPackageType(config));
+    auroraProperties.put("TOPOLOGY_PACKAGE_TYPE", Context.topologyPackageType(config).toString());
```
billonahill: We don't want to rely on `toString()` here since that should only ever really be used for logging. If we need a string better to call `.name()` on the enum, or create a `value` field in the enum that is defined as part of the enum (i.e., `TAR("tar")`).

heron/common/src/java/com/twitter/heron/common/basics/PackageType.java
objmagic: I am not sure if it will return a lowercase one or an uppercase one.

heron/common/src/java/com/twitter/heron/common/basics/PackageType.java
```diff
@@ -0,0 +1,51 @@
...
+  private static boolean isOriginalPackagePex(String packageFilename) {
+    return packageFilename.endsWith(".pex");
+  }
+
+  private static boolean isOriginalPackageTar(String packageFilename) {
+    return packageFilename.endsWith(".tar");
+  }
+
+  public static PackageType getPackageType(String topologyBinaryFile) {
+    String basename = FileUtils.getBaseName(topologyBinaryFile);
```
billonahill: We can remove ~20 lines of code and just use guava Files class here:
```
String extension = Files.getFileExtension(topologyBinaryFile)
return PackageType.valueOf(extension.toUpperCase();
```
```

File: heron/spi/src/java/com/twitter/heron/spi/utils/SchedulerUtils.java [hunk 6]
```diff
@@ -212,7 +212,7 @@ private SchedulerUtils() {
     commands.add(Context.systemConfigSandboxFile(config));
     commands.add(Runtime.componentRamMap(runtime));
     commands.add(SchedulerUtils.encodeJavaOpts(TopologyUtils.getComponentJvmOptions(topology)));
-    commands.add(Context.topologyPackageType(config));
+    commands.add(Context.topologyPackageType(config).toString());
```
billonahill: better to use `.name()` than toString. Also check on the cases, think both return caps.

heron/common/tests/java/com/twitter/heron/common/basics/PackageTypeTest.java
```diff
@@ -0,0 +1,44 @@
...
+    Assert.assertEquals(PackageType.getPackageType(jarFile), PackageType.JAR);
+
+    String tarFile = "b.tar";
+    Assert.assertEquals(PackageType.getPackageType(tarFile), PackageType.TAR);
+
+    String pexFile = "c.pex";
+    Assert.assertEquals(PackageType.getPackageType(pexFile), PackageType.PEX);
+  }
+
+  @Test(expected = RuntimeException.class)
```
billonahill: Is this RuntimeException or a subclass? If a subclass we should expect that instead.

heron/common/src/java/com/twitter/heron/common/basics/PackageType.java
```diff
@@ -0,0 +1,32 @@
...
+
+/***
+ * This enum defines commands topology package type
+ */
+public enum PackageType {
+  PEX,
+  JAR,
+  TAR;
+
+  public static PackageType getPackageType(String topologyBinaryFile) {
```
billonahill: We should rename topologyBinaryFile to something more generic like filename.

objmagic: how about `packageFile`, since this class is `PackageType`? `filename` is too general

billonahill: sure

billonahill:
👍 once CI passes

mycFelix:
👍

Issue #1612, opened 2016-12-12
Refactor UpdateTopologyManager assert method


mycFelix:
Hi~

I was reading [this part of code](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java#L128). I was wondering if we could use guava `Preconditions.checkArgument` or `Preconditions.checkState` instead of `assertTrue` method. I'll send a PR for this.

mycFelix:
merged #1613 and closed

mycFelix:
Try to do the same in `ByteAmount`

PR #1613, opened 2016-12-12
using guava Preconditions in UpdateTopologyManager

mycFelix:
As #1612 mentioned, this PR is trying to use guava Preconditions in UpdateTopologyManager class for checking conditions and replace the `assertTrue` method. 

* add `com.google.common.base.Preconditions`
* remove `assertTrue` method
* use `Preconditions.checkArgument` or `Preconditions.checkState`

cc: @billonahill


billonahill:
Changes look good, but can we not prefix "ERROR:" to every message string? These messages are going to be used as the message to a RuntimeException, which should be ultimately logged as SEVERE, so no need to also put ERROR in the message.

mycFelix:
@billonahill Thanks for your suggestion and the PR was updated. BTW, any comment if we do the same change on ByteAmount class, is it necessary?

billonahill:
👍 thanks!

Doing the same in `ByteAmount` would be great thanks.

mycFelix:
@billonahill - I'll send another PR for `ByteAmount`

objmagic:
👍

PR #1614, opened 2016-12-12
command-line parser support for dry run.

objmagic:
Add command line parser support for dry run mode. User can pass `--dry-run` flag to `submit` and `update` subcommands.


billonahill:
This change seems reasonable (we should really create a way to share common options across main classes) but let's not commit this until there is some functionality backing it.

PR #1615, opened 2016-12-12
Adding metrics sink: WebSink

fabianmenges:
Adding a new metrics sink, WebSink, that publishes metrics on a http endpoint. 

Background:
In our environment processes publish metrics on an configureable http endpoint. We are running a metrics collector that hits these endpoints to collect metrics in specific intervals.

### DIFF
--- a/heron/metricsmgr/src/java/BUILD
+++ b/heron/metricsmgr/src/java/BUILD
@@ -16,11 +16,12 @@ java_library(
         "//heron/proto:proto_metrics_java",
         "//heron/proto:proto_tmaster_java",
         "//heron/metricsmgr/src/thrift:thrift_scribe_java",
+        "//third_party/java:guava",
         "//third_party/java:jackson",
+        "//third_party/java:logging", # slf4j is only used for thrift in scribe-sink
         "@org_yaml_snakeyaml//jar",
         "@com_google_protobuf_protobuf_java//jar",
         "@org_apache_thrift_libthrift//jar",
-        "//third_party/java:logging", # slf4j is only used for thrift in scribe-sink
     ],
 )
 
--- a/heron/metricsmgr/src/java/BUILD
+++ b/heron/metricsmgr/src/java/BUILD
@@ -19,9 +19,9 @@ java_library(
         "//third_party/java:guava", # only used in WebSink
         "//third_party/java:jackson",
         "//third_party/java:logging", # slf4j is only used for thrift in scribe-sink
-        "@org_yaml_snakeyaml//jar",
         "@com_google_protobuf_protobuf_java//jar",
         "@org_apache_thrift_libthrift//jar",
+        "@org_yaml_snakeyaml//jar",
     ],
 )
 
--- a/heron/metricsmgr/src/java/BUILD
+++ b/heron/metricsmgr/src/java/BUILD
@@ -16,7 +16,7 @@ java_library(
         "//heron/proto:proto_metrics_java",
         "//heron/proto:proto_tmaster_java",
         "//heron/metricsmgr/src/thrift:thrift_scribe_java",
-        "//third_party/java:guava",
+        "//third_party/java:guava", # only used in WebSink
         "//third_party/java:jackson",
         "//third_party/java:logging", # slf4j is only used for thrift in scribe-sink
         "@org_yaml_snakeyaml//jar",
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
@@ -149,7 +149,6 @@ public class WebSink implements IMetricsSink {
       });
       httpServer.start();
     } catch (IOException e) {
-      LOG.log(Level.SEVERE, "Could not create HttpServer on port " + port, e);
       throw new RuntimeException("Failed to create Http server on port " + port, e);
     }
   }
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
@@ -23,11 +23,15 @@ import java.nio.file.Paths;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+
+import com.google.common.base.Strings;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 
 import com.sun.net.httpserver.HttpExchange;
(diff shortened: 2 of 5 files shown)


kramasamy:
@maosongfu - can you review this code?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * A metrics sink that publishes metrics on a http endpoint
+ */
+public class WebSink implements IMetricsSink {
+  private static final Logger LOG = Logger.getLogger(WebSink.class.getName());
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private static final int HTTP_STATUS_OK = 200;
+
+  private static final String KEY_PORT = "port";
```
maosongfu: Is it possible to add a sample config for this sink in metrics_sinks.yaml (https://github.com/twitter/heron/blob/master/heron/config/src/yaml/conf/examples/metrics_sinks.yaml)?

maosongfu: Can you add comments for all these configs to explain what they are for?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final String KEY_PORT_FILE = "port-file";
+  private static final String KEY_PATH = "path";
+  private static final String KEY_FLAT_METRICS = "flat-metrics";
+  private static final String KEY_INCLUDE_TOPOLOGY_NAME = "include-topology-name";
+
+  protected ConcurrentHashMap<String, Object> metrics = new ConcurrentHashMap<>();
```
maosongfu: What is this variable for? Could we add comment for it?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+    } else {
+      source = includeServiceNamespace
+          ? String.format("%s/%s", topologyName, record.getSource())
+          : String.format("/%s", record.getSource());
+    }
+
+    if (isFlatMetrics) {
+      processMetrics(source + "/", record.getMetrics(), metrics);
+    } else {
+      Map<String, Object> sourceMap = new HashMap<>();
```
maosongfu: Can you tell me what are the differences between:
1. 
```
Map<String, Object> sourceMap = new HashMap<>();
processMetrics("", record.getMetrics(), sourceMap);
metrics.put(source, sourceMap);
```
2.
```
processMetrics("", record.getMetrics(), metrics);
```

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+          LOG.log(Level.INFO, "Received metrics request.");
+        }
+      });
+      httpServer.start();
+    } catch (IOException e) {
+      LOG.log(Level.SEVERE, "Could not create HttpServer on port " + port, e);
+    }
+  }
+
+  private static void processMetrics(String prefix, Iterable<MetricsInfo> metrics,
```
maosongfu: 1. Comment for this method? 

2.
Also, the argument `Iterable<MetricsInfo> metrics` will hide the field `protected ConcurrentHashMap<String, Object> metrics`, which is bad.

3. Why making it a static method? Also, why passing the out as a argument, which is pretty C/C++ style?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
maosongfu: After reviewing this sink, I found there is no limitation on the size of this Map `metrics`. In some cases, for instance, users export metrics with new name at an interval, it is possible the size of `metrics` will grow unboundedly and eventually lead to OOM, which is a bad design.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+        conf.getOrDefault(KEY_INCLUDE_TOPOLOGY_NAME, false));
+    topologyName = context.getTopologyName();
+
+    int port = TypeUtils.getInteger(conf.getOrDefault(KEY_PORT, 0));
+    if (port == 0) {
+      if (portFile != null) {
+        try {
+          port = TypeUtils.getInteger(Files.lines(Paths.get(portFile)).findFirst().get().trim());
+        } catch (IOException e) {
+          throw new IllegalArgumentException("Could not parse " + KEY_PORT_FILE + " " + portFile
```
fabianmenges: TypeUtils.getInteger already does that. However I should probably catch and re-throw that to add context.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
fabianmenges: Not sure if changing your metrics names over time is a great design either ;) but I will fix that.

heron/metricsmgr/src/java/BUILD
```diff
@@ -16,11 +16,12 @@ java_library(
         "//heron/proto:proto_metrics_java",
         "//heron/proto:proto_tmaster_java",
         "//heron/metricsmgr/src/thrift:thrift_scribe_java",
+        "//third_party/java:guava",
```
maosongfu: Can you add a comment here to indicate this library is used only for websink, since we would later split the all these sinks from the spi parts.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -0,0 +1,162 @@
...
+          httpExchange.sendResponseHeaders(HTTP_STATUS_OK, response.length);
+          OutputStream os = httpExchange.getResponseBody();
+          os.write(response);
+          os.close();
+          LOG.log(Level.INFO, "Received metrics request.");
+        }
+      });
+      httpServer.start();
+    } catch (IOException e) {
+      LOG.log(Level.SEVERE, "Could not create HttpServer on port " + port, e);
```
maosongfu: Given the RuntimeException contains the error msg, extra SEVERE LOG seems duplicated and unnecessary.

fabianmenges:
I don't think that failed integration check can be caused by my changes....

maosongfu:
The failed ci can be caused by some tricky test cases. @billonahill @objmagic Can you help on this?

This pull request looks good to me now. @nlu90 Can you check whether the new guava dependency looks ok to you?

objmagic:
restarted CI due to some flaky integration test

nlu90:
LGTM

Issue #1604, opened 2016-12-08
Move Scheduler related spi util classes into schedulers package


billonahill:
The following classes only seem to be required for schedulers and do not need to be in spi. Instead they could probably live in `c.t.h.scheduler.utils`:
```
c.t.h.spi.utils.Runtime
c.t.h.spi.utils.SchedulerConfig
c.t.h.spi.utils.Shutdown
c.t.h.spi.utils.LauncherUtils
```

`SchedulerConfig` should also not be a util.

mycFelix:
@billonahill - I send PR for this. Please review when you have a chance

mycFelix:
merged and closed

objmagic:
resolved via #1617

PR #1617, opened 2016-12-13
Move Scheduler related spi util classes into schedulers package

mycFelix:
As #1604 mentioned, this PR is trying to move some `spi-classes` into schedulers package. 

There are five src and two test classes included totally:
```
c.t.h.spi.utils.Runtime
c.t.h.spi.utils.SchedulerConfig
c.t.h.spi.utils.Shutdown
c.t.h.spi.utils.LauncherUtils
c.t.h.spi.utils.SchedulerUtils

c.t.h.spi.utils.SchedulerUtils.LauncherUtilsTest
c.t.h.spi.utils.SchedulerUtils.SchedulerUtilsTest
```

* move classes into `c.t.h.schedulers.utils`
* modify `c.t.h.schedulers.BUILD` and modify `c.t.h.scheduler-core.BUILD`
* modify related class invoke info 

I've no idea about how to refactor `SchedulerConfig` class. So this PR only move it into schedulers package without any change. Let me know if there has anything I could do and any comment is welcomed.

cc @billonahill


heron/schedulers/src/java/com/twitter/heron/scheduler/utils/SchedulerConfig.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.twitter.heron.spi.utils;
+package com.twitter.heron.scheduler.utils;
```
billonahill: This class is really a utility, not a config. We should probably name it `SchedulerConfigUtils`. Also all but the last method can be private.

mycFelix: Understood, I'll update it.

heron/spi/tests/java/BUILD
```diff
@@ -42,8 +42,6 @@ java_tests(
     test_classes = [
         "com.twitter.heron.spi.utils.ShellUtilsTest",
         "com.twitter.heron.spi.utils.NetworkUtilsTest",
-        "com.twitter.heron.spi.utils.SchedulerUtilsTest",
```
mycFelix: My bad, forgot doing this.

heron/scheduler-core/src/java/BUILD
```diff
@@ -18,10 +18,15 @@ spi_deps_files = [
     "//heron/spi/src/java:utils-spi-java",
 ]
 
+schedulers_deps_files = [
+    "//heron/schedulers/src/java:utils-schedulers-java",
```
mycFelix: Good Point!  Thank you for your advice. I'll try to move them into schedulers-core package

mycFelix:
Updated

* move classes into `schedulers-core`
* refactor `SchedulerConfig` class
* add `test_classes`

@billonahill Please take a look again when you have a chance.

billonahill:
👍

Thanks for the cleanup!

PR #1625, opened 2016-12-19
Consolidate more common failure tests in CommonPackingTests

billonahill:
Pull some failure scenario tests up to `CommonPackingTests` since they should apply to both packing impls. Also, expecting `PackingException` instead of `RuntimeException`.


objmagic:
👍

PR #1630, opened 2016-12-20
Check at launch that configured container count same as packing plan

### LINKED ISSUE #1623: Possible to submit topology with invalid packing plan
Version: 0.14.4

### To reproduce:
1. Specify a topology with 1 spout instance and 1 bolt instance
2. Set the `numWorker` to a number > 2
3. Deploy the topology

### Expected to see:
I expected to see the deploy fail because the packing plan would be invalid given that there are more workers requests than instances to fill up the space

### What happened:
The stream manager will fail at runtime when it tries to read the packing plan because some containers will not have a plan because there are no instances in that container. Specifically it fails on this line: https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py#L499

billonahill:
Fixes #1623.

If the number of containers requested is not the same as the packing plan says, throw an exception.


objmagic:
👍

PR #1635, opened 2016-12-21
cleanup protobuf third party build file

kramasamy:
This PR cleans up the protobuf BUILD file by moving include files, lib files and bin files into separate lists and use the lists where it is needed - instead of duplication.


billonahill:
👍

good one

nlu90:
LGTM 👍

congwang:
Looks good

PR #1636, opened 2016-12-21
cleanup glog build file

kramasamy:
Cleans up glog third party build to separate include files and lib files in a separate list. Appropriately name execution scripts

### DIFF
--- a/third_party/glog/BUILD
+++ b/third_party/glog/BUILD
@@ -38,12 +38,11 @@ mac_script = "\n".join(common_script + [
 ])
 
 linux_script = "\n".join(common_script + [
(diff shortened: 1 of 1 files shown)


third_party/glog/BUILD
```diff
@@ -40,36 +56,20 @@ genrule(
...
     name = "glog-cxx",
-    srcs = [
-        "empty.cc",
-        "lib/libglog.a",
-        "include/glog/log_severity.h",
-        "include/glog/raw_logging.h",
-        "include/glog/stl_logging.h",
-        "include/glog/vlog_is_on.h",
-    ],
+    srcs = ["empty.cc"] + lib_files,
```
nlu90: also + include_files?

kramasamy: The include files are in the hdrs now. Realized this is not needed in the srcs actually.

nlu90: kk, LGTM

PR #1637, opened 2016-12-22
cleanup of gflags to separate include files and lib files into lists

kramasamy:


third_party/gflags/BUILD
```diff
@@ -8,44 +8,44 @@ package_version = "2.0"
...
     name = "gflags-cxx",
     srcs = [
-        "include/gflags/gflags_declare.h",
         "empty.cc",
         "lib/libgflags.a",
     ],
-    hdrs = [
-        "include/gflags/gflags.h",
-    ],
+    hdrs = include_files,
```
nlu90: more header files are included than needed. Is this okay?

kramasamy: yes - to keep it simple in the grand scheme of things.

PR #1638, opened 2016-12-22
cleanup gtest BUILD file

kramasamy:


nlu90:
👍

congwang:
What's the difference between single-quote and double-quote in Bazel?

kramasamy:
@congwang - single quote takes precedence our double quotes. I am using a convention that the scripts use single quote - since some of them require it. I am trying to make it consistent - across all of them.

PR #1644, opened 2016-12-25
cleanup the warnings in stmgr

kramasamy:
Removed the warnings in stmgr during compilation.

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-client.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-client.cpp
@@ -199,7 +199,8 @@ void StMgrClient::SendStartBackPressureMessage() {
   REQID_Generator generator;
   REQID rand = generator.generate();
   // generator.generate(rand);
-  proto::stmgr::StartBackPressureMessage* message = acquire(message);
+  proto::stmgr::StartBackPressureMessage* message = acquire(
+    static_cast<proto::stmgr::StartBackPressureMessage *>(nullptr));
   message->set_topology_name(topology_name_);
   message->set_topology_id(topology_id_);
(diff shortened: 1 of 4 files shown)


heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -199,7 +199,8 @@ void StMgrClient::SendStartBackPressureMessage() {
   REQID_Generator generator;
   REQID rand = generator.generate();
   // generator.generate(rand);
-  proto::stmgr::StartBackPressureMessage* message = acquire(message);
+  proto::stmgr::StartBackPressureMessage* message = acquire(
+    static_cast<proto::stmgr::StartBackPressureMessage *>(nullptr));
```
congwang: I think it is cleaner to just use:

```
proto::stmgr::StartBackPressureMessage* message = nullptr;
message = acquire(message);
```

kramasamy: I don't mind this. If we use your suggestion, somebody later might think it is needs to be cleaned up since a single assignment will suffice.

kramasamy:
@congwang - any more comments?

heron/stmgr/src/cpp/manager/stmgr-client.cpp
congwang: People should read the definition of acquire() and then figure out the type deduction. So I don't think it is a problem.

srkukarni: Agree with @congwang

nlu90:
👍

congwang:
👍

Thanks for update!

PR #1646, opened 2016-12-29
Added a topology status to the list all topologies ui.

cckellogg:
topologies-with-status

### DIFF
--- a/heron/tools/tracker/src/python/handlers/stateshandler.py
+++ b/heron/tools/tracker/src/python/handlers/stateshandler.py
@@ -29,10 +29,10 @@ class StatesHandler(BaseHandler):
   {
     <cluster1>: {
       <environ1>: {
-        <top1>: {
+        <topology1>: {
           <the executionstate of the topology>
         },
-        <top2>: {...}
+        <topology2>: {...}
         ...
       },
       <environ2>: {...},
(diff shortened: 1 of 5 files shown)


billonahill:
While we're at it, could we also make it clear what the red font indicates?
Currently that's a bit of a mystery.

cckellogg:
Can we address the font colors in another patch? This was preexisting code and I'm not sure what it means either. Here the code for the colors (https://github.com/twitter/heron/blob/master/heron/tools/ui/resources/static/js/alltopologies.js) lines 20 - 31.

kramasamy:
@cckellogg - let us do another patch to appropriately report the status of the topologies in red

kramasamy:
@billonahill - the proposal is to change the color from RED to normal and display the state as 'ORPHANED'. @cckellogg is doing it in a different PR. Any suggestions for this state other than 'ORPHANED'?

billonahill:
@kramasamy the states we currently show align with the documented Heron states that users control via the Heron cli. Since 'ORPHANED' is not one of those, my concern is that it's meaning would also not be understood by a typical user. Somehow we need to convey exactly what is meant by orphaned or red-font. I'm open for UI suggestions for how to do that though, but I think it might need more of an explanation somewhere on the page.

PR #1658, opened 2017-01-10
Fix response of ``help'' and ``version'' subcommands

objmagic:
The boolean type response is invalid. We should use `Response` object instead.


kramasamy:
👍

PR #1665, opened 2017-01-13
Stream manager code cleanup and optimization

congwang:
Please review each commit separately.


heron/common/src/cpp/network/connection.cpp
```diff
@@ -62,25 +61,26 @@ Connection::~Connection() {
   delete mIncomingPacket;
   {
     for (auto iter = mOutstandingPackets.begin(); iter != mOutstandingPackets.end(); ++iter) {
-      delete iter->first;
+      delete *iter;
```
kramasamy: this does not seem to be correct. if you delete the iter, how can we do iter++

kramasamy: instead of for - you could use while similar to the mSentPackets (that follows)?

heron/common/src/cpp/network/connection.cpp
```diff
@@ -114,13 +114,13 @@ sp_int32 Connection::registerForBackPressure(VCallback<Connection*> cbStarter,
 sp_int32 Connection::writeIntoIOVector(sp_int32 maxWrite, sp_int32* toWrite) {
   sp_uint32 bytesLeft = maxWrite;
   sp_int32 simulWrites =
-      mIOVectorSize > mNumOutstandingPackets ? mNumOutstandingPackets : mIOVectorSize;
+      mIOVectorSize > mOutstandingPackets.size() ?: mIOVectorSize;
```
kramasamy: Is this standard c++? I thought it is GNU C++ extension. If this is part of the regular c++ standard, it is fine?

congwang: Good point. It does not seem to be standard even for C++.

heron/common/src/cpp/network/connection.cpp
congwang: iter points to an element which is a pointer, so *iter is the pointer, delete *iter means deleting the contents pointed by that pointer, it is fine.

heron/common/src/cpp/network/connection.cpp
```diff
@@ -75,12 +75,10 @@ Connection::~Connection() {
   delete[] mIOVector;
 }
 
-sp_int32 Connection::sendPacket(OutgoingPacket* packet) { return sendPacket(packet, NULL); }
```
huijunw: I think it is better keep the method with call back param to let the connection.h/cpp be general. There is method without call back param, which can be adopted when call back is not needed.

congwang:
@huijunw The benefits are their difference, that is the order, it speeds up the fast path, search, from O(logN) to O(1) (average case). You probably want to read their documents for details:http://www.cplusplus.com/reference/unordered_set/unordered_set/

heron/common/src/cpp/network/client.h
```diff
@@ -303,7 +274,6 @@ class Client : public BaseClient {
   std::unordered_map<REQID, std::pair<sp_string, void*> > context_map_;
 
   typedef std::function<void(IncomingPacket*)> handler;
-  std::unordered_map<std::string, handler> requestHandlers;
```
congwang: We can always add it back if we need it in the future, so I don't worry about it for now.

heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -177,7 +177,7 @@ void StMgrClient::SendTupleStreamMessage(proto::stmgr::TupleStreamMessage2& _msg
 }
 
 void StMgrClient::HandleTupleStreamMessage(proto::stmgr::TupleStreamMessage2* _message) {
-  release(_message);
```
congwang: Because it comes from std::allocator but returns to mempool so never goes back to std::allocator again.

heron/common/src/cpp/network/connection.cpp
congwang: It is unused, I don't know why an unused parameter contributes to generality. Mind to be more specific?
Again, we can always add it back if we need it in the future, it is pretty easy to do with git.

PR #1672, opened 2017-01-19
Change AuroraController to an interface

billonahill:
The current implementation of `AuroraController` shells out to the aurora cli and is intended to be run locally by a developer. Other implementations might call the aurora scheduler service directly. Changing `AuroraController` to be an interface allows other implementations to be used by `AuroraScheduler`.

The diff makes things appear messier than they are but basically `AuroraController` was renamed to `AuroraCLIController` and a new `AuroraController` interface was added.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
@@ -23,9 +23,15 @@ import com.twitter.heron.spi.packing.PackingPlan;
  */
 public interface AuroraController {
 
-  boolean createJob(Map<String, String> auroraProperties);
+  boolean createJob(Map<AuroraField, String> auroraProperties);
   boolean killJob();
-  boolean restartJob(int containerId);
+
+  /**
+   * Restarts a given container, or the entire job if containerId is null
+   * @param containerId ID of container to restart, or entire job if null
+   */
+  boolean restart(Integer containerId);
+
   void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove);
   void addContainers(Integer count);
 }
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
@@ -23,7 +23,7 @@ import com.twitter.heron.spi.packing.PackingPlan;
  */
 public interface AuroraController {
 
-  boolean createJob(String auroraFilename, Map<String, String> bindings);
+  boolean createJob(Map<String, String> auroraProperties);
   boolean killJob();
   boolean restartJob(int containerId);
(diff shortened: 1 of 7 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
```diff
@@ -11,144 +11,21 @@
...
-    }
+public interface AuroraController {
 
-    // Append batch size.
-    // Note that we can not use "--no-batching" since "restart" command does not accept it.
-    // So we play a small trick here by setting batch size Integer.MAX_VALUE.
-    auroraCmd.add("--batch-size");
-    auroraCmd.add(Integer.toString(Integer.MAX_VALUE));
-  }
+  boolean createJob(String auroraFilename, Map<String, String> bindings);
```
objmagic: Should we add some documentation here? What is `bindings` here? I need to read `aurora job create -h` to know what it is. Considering this is an argument of an interface function, I think it is necessary to add some docs here.

billonahill: Good catch. I refactored `createJob` to remove the `auroraFilename`, which only applies to the CLI impl. I also renamed `bindings` to `auroraProperties` and changed the key from a string to an enum to make it's usage clear.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
```diff
@@ -23,9 +23,15 @@
  */
 public interface AuroraController {
 
-  boolean createJob(Map<String, String> auroraProperties);
+  boolean createJob(Map<AuroraField, String> auroraProperties);
   boolean killJob();
-  boolean restartJob(int containerId);
+
+  /**
+   * Restarts a given container, or the entire job if containerId is null
```
objmagic: when `containerId` is `-1`, the entire job will also be restarted.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
```diff
@@ -23,9 +23,15 @@
...
+   * @param containerId ID of container to restart, or entire job if null
+   */
+  boolean restart(Integer containerId);
```
objmagic: when `containerId` is `-1`, the entire job will also be restarted according to [here](https://github.com/twitter/heron/pull/1672/commits/a89f379ddbd6a827ab3dbca014ef8a92c985ee92#diff-c41f12543c6d4d61074654ddf05c9c51R87)

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java
billonahill: Although not the preferred approach I kept that for backward compatibility, but there's no need to do that since the code will all change in lock step. I've made the change to match behavior to docs.

objmagic:
LGTM.

PR #1674, opened 2017-01-21
[travis-ci] make log detailed, and dump fail log

objmagic:
also break some long lines.

Reference: https://www.bazel.io/docs/command-line-reference.html


objmagic:
close since I merged #1629 with this. It is a trivial change and we already did this in internal CI months ago.

PR #1675, opened 2017-01-22
[dry-run] print with UTF-8 encoding and consume stream with UTF-8 decoding

objmagic:
Since we are using UTF-8 character when rendering dry-run response, we need to take care of some UTF-8 locale issue. For example, in CI Ubuntu environment, the Java program does not print with UTF-8 encoding or consume input stream with UTF-8 decoding, which could cause at least two problems:
1. CLI outputs question marks instead of proper UTF-8 character
2. Unit test fails because input stream of sample output file is not decoded with UTF-8

Test has also been done in internal Linux environment.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -15,6 +15,7 @@
 package com.twitter.heron.scheduler;
 
 import java.io.IOException;
+import java.io.PrintStream;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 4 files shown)


kramasamy:
👍

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -318,7 +319,9 @@ public static void main(String[] args)
       // SUPPRESS CHECKSTYLE IllegalCatch
     } catch (UpdateDryRunResponse response) {
       LOG.log(Level.FINE, "Sending out dry-run response");
-      System.out.print(runtimeManagerMain.renderDryRunResponse(response));
+      // Output may contain UTF-8 characters, so we should print using UTF-8 encoding
+      PrintStream out = new PrintStream(System.out, true, "UTF-8");
```
billonahill: Instead of using strings, can we use `StandardCharsets.UTF-8` for a `Locale` or `StandardCharsets.UTF-8.name()`for a string representation? This is consistency with how locales are specified in the rest of the codebase. 

Also please allow at least a few business hours of time for people to review non-critical patches before merging.

objmagic: We can't use enum here since Java does not have such API. See: https://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html. However, I agree we can use the `.name()`.

And yes, I should allow some hours for reviewers to review. Sorry I've been slow and hurrying myself too much on dry-run.

PR #1676, opened 2017-01-24
[dry-run] use enum.name() instead of string for locale

objmagic:
A tiny PR. Follow up of https://github.com/twitter/heron/pull/1675#discussion_r97236313


billonahill:
👍

PR #1677, opened 2017-01-24
Drop corrupt packets to clients

congwang:
Instead of crashing stmgr, we can just ignore the corrupt packets.

### DIFF
--- a/heron/common/src/cpp/network/client.cpp
+++ b/heron/common/src/cpp/network/client.cpp
@@ -181,8 +181,10 @@ void Client::OnNewPacket(IncomingPacket* _ipkt) {
 
   if (_ipkt->UnPackString(&typname) != 0) {
     Connection* conn = static_cast<Connection*>(conn_);
-    LOG(FATAL) << "UnPackString failed from connection " << conn << " from hostport "
+    LOG(ERROR) << "UnPackString failed from connection " << conn << " from hostport "
                << conn->getIPAddress() << ":" << conn->getPort();
+    delete _ipkt;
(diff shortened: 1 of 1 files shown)


heron/common/src/cpp/network/client.cpp
```diff
@@ -181,8 +181,10 @@ void Client::OnNewPacket(IncomingPacket* _ipkt) {
 
   if (_ipkt->UnPackString(&typname) != 0) {
     Connection* conn = static_cast<Connection*>(conn_);
-    LOG(FATAL) << "UnPackString failed from connection " << conn << " from hostport "
+    LOG(ERROR) << "UnPackString failed from connection " << conn << " from hostport "
```
maosongfu: On the one hand, not sure whether it would affect our semantics: originally it is fail-fast but now it is dropping. On the other hand, we would also need to export metrics for this behavior; otherwise, it is hard to trace and debug.

Also, did we really witness this data corruption, since protobuf is pretty neat.

congwang: Yeah, it makes sense to add a counter for dropped packets in another PR.
There is a real issue in our DC, the cause of the corruption is unknown yet.

maosongfu: I am still not convinced by the idea of pull request. This method returns nothing. So this failure will be no way visible to higher layer's logic: we can not handle this dropping. 
I would still prefer the fail-fast.
Could you describe more on the scenarios of corruption issue?

PR #1678, opened 2017-01-24
Display error message for NetworkErrorCode

### LINKED ISSUE #1654: Improve log message for connection errors
Logging a number is confusing:
`W0103 20:25:17.030653 31955 stmgr-client.cpp:94] Could not connect to stmgr stmgr-147 running at <host>:<port> due to: 1`

congwang:
This is to resolve #1654.

### DIFF
--- a/heron/common/src/cpp/network/network_error.h
+++ b/heron/common/src/cpp/network/network_error.h
@@ -17,6 +17,9 @@
 #ifndef _CORE_SRC_CPP_NEWTORK_NETWORK_ERROR_H_
 #define _CORE_SRC_CPP_NEWTORK_NETWORK_ERROR_H_
 
+#include <ostream>
+#include <string>
+
 enum NetworkErrorCode {
   OK = 0,
   CONNECT_ERROR,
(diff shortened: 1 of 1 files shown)


heron/common/src/cpp/network/network_error.h
```diff
@@ -30,4 +33,12 @@ enum NetworkErrorCode {
   TIMEOUT
 };
 
+static inline std::ostream& operator<<(std::ostream& os, const NetworkErrorCode code) {
+  std::string strs[] = { "OK", "Connection error", "Read error", "Write error",
```
billonahill: Can these instead be bound directly to the enum values themselves, instead of managed separately?
http://en.cppreference.com/w/cpp/language/enum

congwang: I am not aware of any C++ syntax for this, we may archive it by a wrapper class or something like that, but I doubt it is worth any effort to improve it. Does it hurt readability to you?

objmagic: this looks useful https://aantron.github.io/better-enums/tutorial/Conversions.html#Strings and easy to include in, if we bother to use it at all...

congwang: There is always some way to do it. Again, why do you think it is worth our effort to do it? For me, we only have one case so far, no need to bother a complex way to make it unnecessarily elegant. If you believe it is worth, you are welcome to do it in a separate PR.

objmagic: just pointing out. i have no opinion since C++ language does not have the capability of allowing people to use enum reflection easily.

objmagic:
can you check out the CI?

congwang:
@objmagic IntegrationTest_MultiSpoutsMultiTasks fails.

congwang:
There is a new CI failure:

```
[ RUN      ] StMgr.test_tmaster_restart_on_new_address
src/central_freelist.cc:333] tcmalloc: allocation failed 24576 
Unknown exception caught in global handler
```

objmagic:
Thanks. I'll just restart the CI.

objmagic:
The PR CI is fine. Feel free to merge it.

objmagic:
The PR CI check passed. Merged.

Issue #1671, opened 2017-01-19
Add a metric to indicate average queue length in #packets pending between stmgr and each instance


avflor:
This metric can be useful to detect overloaded instances.

congwang:
@avflor By number of tuples, does it count all kinds of tuples? Ack tuples, fail tuples too?

avflor:
@congwang Just the number of data tuples (no ack tuples). What is fail tuples? If they have failed would they be in the queue?

maosongfu:
We have 3 metrics: # of data_tuple, ack_tuple and fail_tuple?

avflor:
That would be great! We might need these metrics in the future.

maosongfu:
Pull request is here: #1695

objmagic:
resolved via https://github.com/twitter/heron/pull/1695

PR #1397, opened 2016-09-16
Add metrics for stream manager queue size

congwang:
This adds metrics for stream manager queue size (in bytes) sent to each instance.

Sample output:
```
            "__queue_size_bytes_to_instance_by_compid/container_1_exclaim1_1": "50900906",
            "__queue_size_bytes_to_instance_by_compid/container_1_word_2": "0",
```

PR #1686, opened 2017-01-31
Add number of packets to the per instance queue metrics

congwang:
This PR first moves the metrics collecting code out of the hot path, then adds # packet to the per instance queue metrics as requested in issue #1671.

Sample output:
```
            "__queue_size_to_instance_by_compid/container_1_consumer_2/bytes": "86421995",
            "__queue_size_to_instance_by_compid/container_1_consumer_2/packets": "794",
            "__queue_size_to_instance_by_compid/container_1_word_1/bytes": "0",
            "__queue_size_to_instance_by_compid/container_1_word_1/packets": "0",
```

### DIFF
--- a/heron/common/src/cpp/network/connection.cpp
+++ b/heron/common/src/cpp/network/connection.cpp
@@ -44,7 +44,6 @@ Connection::Connection(ConnectionEndPoint* endpoint, ConnectionOptions* options,
   mOnNewPacket = NULL;
   mOnConnectionBufferEmpty = NULL;
   mOnConnectionBufferFull = NULL;
-  mOnConnectionBufferChange = NULL;
   mNumOutstandingPackets = 0;
   mNumOutstandingBytes = 0;
   mIOVectorSize = 1024;
@@ -84,10 +83,6 @@ sp_int32 Connection::sendPacket(OutgoingPacket* packet, VCallback<NetworkErrorCo
   mNumOutstandingPackets++;
   mNumOutstandingBytes += packet->GetTotalPacketSize();
 
-  if (mOnConnectionBufferChange) {
-    mOnConnectionBufferChange(this);
-  }
-
   if (!hasCausedBackPressure()) {
     // Are we above the threshold?
     if (mNumOutstandingBytes >= systemHWMOutstandingBytes) {
@@ -107,10 +102,6 @@ sp_int32 Connection::sendPacket(OutgoingPacket* packet, VCallback<NetworkErrorCo
 
 void Connection::registerForNewPacket(VCallback<IncomingPacket*> cb) {
   mOnNewPacket = std::move(cb);
-}
-
-void Connection::registerForBufferChange(VCallback<Connection*> cb) {
-  mOnConnectionBufferChange = std::move(cb);
 }
 
 sp_int32 Connection::registerForBackPressure(VCallback<Connection*> cbStarter,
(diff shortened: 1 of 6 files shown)


heron/stmgr/src/cpp/manager/stmgr-server.h
```diff
@@ -153,6 +158,9 @@ class StMgrServer : public Server {
   typedef std::map<sp_string, heron::common::AssignableMetric*> QueueMetricMap;
   QueueMetricMap queue_metric_map_;
 
+  typedef std::map<sp_string, heron::common::MultiAssignableMetric*> TuplesMetricMap;
```
huijunw: could you add comment to this field - what this field is for and what key/value in the map means

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -107,6 +110,11 @@ StMgrServer::StMgrServer(EventLoop* eventLoop, const NetworkOptions& _options,
   metrics_manager_client_->register_metric(METRIC_TIME_SPENT_BACK_PRESSURE_INIT,
                                            back_pressure_metric_initiated_);
   spouts_under_back_pressure_ = false;
+
+  // Update queue related metrics every 10 seconds
+  CHECK_GT(eventLoop_->registerTimer([this](EventLoop::Status status) {
+    this->UpdateStMgrServerMetrics(status);
+  }, true, QUEUE_METRICS_FREQUENCY), 0);
```
huijunw: is 10 seconds proper?

heron/common/src/cpp/network/connection.cpp
```diff
@@ -44,7 +44,6 @@ Connection::Connection(ConnectionEndPoint* endpoint, ConnectionOptions* options,
   mOnNewPacket = NULL;
   mOnConnectionBufferEmpty = NULL;
   mOnConnectionBufferFull = NULL;
-  mOnConnectionBufferChange = NULL;
```
huijunw: why do we remove this callback?

heron/stmgr/src/cpp/manager/stmgr-server.h
congwang: Will add it.

heron/stmgr/src/cpp/manager/stmgr-server.cpp
congwang: It is just a copy of PROCESS_METRICS_FREQUENCY. I understand you may suggest to expose it in config file, but we already have too many config's...

heron/common/src/cpp/network/connection.cpp
congwang: Because we don't need it any more, as I already describe in this PR, we move the code out of hot path

avflor:
@congwang The code looks fine to me. I have some questions.

1) It seems that every time a new tuple for a particular instance arrives, the counter that corresponds to that instance is increased. Then every time the updatestrmmetrics is called the number of tuples for this instance is set to the value of the counter. Is it possible that the tuples do not exist any more in the queue at that point and thus the value of the metric is incorrect?

2)Does the queue size metric show the total bytes corresponding to the data tuples in the queue or it also includes other data/metadata?

avflor:
Two more comments:

1) The counter is updated based on each tuple's size. Can we also have the a similar metric for number of tuples?

2) it seems that the instance_info_[task_id] keeps increasing as we receive more tuples. Is that right? Since it represents the queue size at a particular time, it should also decrease as the tuples get processed by the corresponding instances.

congwang:
@avflor 

Answer your four questions together:

1) I use active_instances_, so these instances should be alive. But yeah, I should cleanup the metrics map in HandleConnectionClose() too, will add it.

2) You probably want to look at PR #1397, where I added it.

3) `tuples_size()` should return the number of tuples, not bytes.

4) This is exactly why I asked you if this is what you want. If you just want to count the number of tuples sending to each instance, this PR already does so. If you want the number of tuples pending in the queue, it is harder, because conceptually it does not fit: tuples are packed into packets before putting into queue, there are no longer tuples in concept at the lower level, so harder to count the number of tuples leaving the queue. One solution would be packing "number of tuples" into packets too, but again, it needs more work.

objmagic:
resolved in #1695

PR #1688, opened 2017-02-01
Standardize Java formatter string.

objmagic:
#### Problem
The current argument index is invalid, which will be rejected by some Java compiler (`jdk1.8.0_77` in internal OS X CI environment). See: http://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#syntax. Even though some compilers do not reject this syntax, it is not valid according to language spec anyway.

#### Solution
Switch to ``%s``.


billonahill:
👍

PR #1689, opened 2017-02-02
Clean up config visiblity and dead code

billonahill:
First commit of hopefully many to simplify config code.

- Removing dead config code
- Reducing scope of config methods to be as limited as it can be
- Simplifying config loading so `RuntimeManagerMain` and `SubmitterMain` don't need to load defaults and then load configs like this:

```
Config.Builder defaultsConfig = Config.newBuilder()
       .putAll(ClusterDefaults.getDefaults())
       .putAll(ClusterConfig.loadConfig(heronHome, configPath, releaseFile));
```

Instead, abstracting that in the API by having `loadConfig` properly initialize with defaults first.


ashvina:
LGTM 👍 . This was needed, thanks !

I find the `runtime` config object management confusing too. It is slightly related to this change, but could be addressed in a different PR. Currently two config objects are used in most components. My understanding is that the initial instance of config object will be built/loaded by the `ConfigConfig` class. The other instance, `runtime`, would be loaded by the `*Main` classes. If this is correct, should we remove any `runtime` specific keys from `ConfigKeys` and `keys.yaml`? For e.g. `COMPONENT_RAMMAP`. The user will never provide these values. Adding these strings to `keys.yaml` seems unnecessary and may become confusing later.

billonahill:
Thanks for the reviews. Yes @ashvina the config and runtime thing has also bothered me and I'd like to clean that up too. My main motivation for now is to reduce the amount of config code we have and to make for an easier API to work with. 

I've got a follow-up proposal in https://github.com/billonahill/heron/pull/5/files to remove keys.yaml and move into an enum. This is just a straight clean-up refactor, but we could also follow up with a better way to differentiate between cluster config and topology/runtime configs.

PR #1693, opened 2017-02-03
Replaces keys.yaml and Keys with Key enum

billonahill:
(Replaces prototype PR at billonahill#5 with the full refactor)

Refactoring to no longer read keys from `keys.yaml`. All key usage is via `Key` enum and removing `keys.yaml` and `Keys`.  Core classes in this review are `Key` and `Config`. All others are refactors to use changes to those two classes.

Benefits:
- compile time checks of invalid keys
- easier refactoring
- remove complexity of bundling and reading `keys.yaml`
- more concise usage
- extensibility of `Key` metadata (e.g., type, default, required, etc)

After this I'd like to also remove `defaults.yaml` in favor of the enum approach.


objmagic:
code LGTM. CI integration test fails.

heron/schedulers/src/java/com/twitter/heron/scheduler/local/LocalDefaults.java
```diff
@@ -33,8 +33,7 @@ private LocalDefaults() {
   static {
     try {
       defaults = Resource.load(
-          "com.twitter.heron.scheduler.local.LocalDefaults",
-          LocalConstants.DEFAULTS_YAML);
+          "com.twitter.heron.scheduler.local.LocalDefaults", DEFAULTS_YAML);
```
billonahill: This will be refactored out in a follow-up that addresses defaults.

objmagic:
That integration test is failing, but we should be able to merge this PR. I have some PRs waiting this to be merged.

PR #1694, opened 2017-02-03
Change dryrun to not include format args if dry run is false

billonahill:
I noticed when testing that dry run format args were being passed with empty values even when dry-run was not enabled. Change to not pass format if dry run is not passed.

Also SubmitterMain was using the wrong option char for dry run format.

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/utils/ReflectionUtils.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/utils/ReflectionUtils.java
@@ -28,6 +28,9 @@ public final class ReflectionUtils {
   @SuppressWarnings("unchecked") // we don't know what T is until runtime
   public static <T> T newInstance(ClassLoader classLoader, String className)
       throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+    if (className == null) {
+      throw new ClassNotFoundException("Can not instantiate class. className must not be null");
+    }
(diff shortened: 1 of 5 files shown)


heron/spi/src/java/com/twitter/heron/spi/utils/ReflectionUtils.java
```diff
@@ -28,6 +28,9 @@ private ReflectionUtils() {
   @SuppressWarnings("unchecked") // we don't know what T is until runtime
   public static <T> T newInstance(ClassLoader classLoader, String className)
       throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+    if (className == null) {
+      throw new ClassNotFoundException("Can not instantiate class. className must not be null");
+    }
     return (T) classLoader.loadClass(className).newInstance();
```
huijunw: loadClass() throws ClassNotFoundException. why do we throw it manually?

objmagic: Thanks. I didn't review carefully here. @billonahill do you have any particular reason to add this check?

billonahill: loadClass throws NullPointerException when passed null. The calling classes catch ClassNotFoundException but not NPE.

objmagic:
LGTM.

PR #1695, opened 2017-02-06
Support connection buffer metrics between stmgr and instances

maosongfu:
The patch supports connection buffer metrics between stmgr and instances.
It exports the connection buffer metrics in both # of packets and # of bytes.
It will sample the metrics at an interval and then export the average value during every metrics-collection-interval.

Metric name format:
1. packets count
``__connection_buffer_by_intanceid/{instance-id}/packets``
2. bytes
``__connection_buffer_by_intanceid/{instance-id}/bytes``

Examples:
``__connection_buffer_by_intanceid/container_2_word_6/bytes``
``__connection_buffer_by_intanceid/container_2_word_6/packets``

This bases on @congwang : #1686

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-server.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-server.cpp
@@ -73,8 +73,12 @@ const sp_string METRIC_TIME_SPENT_BACK_PRESSURE_INIT =
 // appended
 // to the string below
 const sp_string METRIC_TIME_SPENT_BACK_PRESSURE_COMPID = "__time_spent_back_pressure_by_compid/";
-// Queue size in bytes sent to each instance
-const sp_string METRIC_QUEUE_SIZE_TO_INSTANCE_COMPID = "__queue_size_bytes_to_instance_by_compid/";
+// Prefix for connection buffer's metrics
+const sp_string CONNECTION_BUFFER_BY_INSTANCEID = "__connection_buffer_by_intanceid/";
+
+// TODO(mfu): Read this value from config
+const sp_int64 SYSTEM_METRICS_SAMPLE_INTERVAL_MICROSECOND = 10 * 1000 * 1000;
+
 
 StMgrServer::StMgrServer(EventLoop* eventLoop, const NetworkOptions& _options,
                          const sp_string& _topology_name, const sp_string& _topology_id,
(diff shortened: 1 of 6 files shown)


avflor:
@maosongfu This looks good to me.

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -162,7 +173,18 @@ sp_string StMgrServer::MakeBackPressureCompIdMetricName(const sp_string& instanc
...
-  return METRIC_QUEUE_SIZE_TO_INSTANCE_COMPID + instanceid;
+  return CONNECTION_BUFFER_BY_INSTANCEID + instanceid;
+}
+
+void StMgrServer::UpdateQueueMetrics(EventLoop::Status) {
+  for (auto itr = active_instances_.begin(); itr != active_instances_.end(); ++itr) {
+    sp_int32 task_id = itr->second;
+    const sp_string& instance_id = instance_info_[task_id]->instance_->instance_id();
+    sp_int32 bytes = itr->first->getOutstandingBytes();
+    connection_buffer_metric_map_[instance_id]->scope("bytes")->record(bytes);
```
huijunw: record() increases denominator_ by 1. is it averaged by invocation count or by period 10 seconds?

maosongfu: sum of metrics value/ sum of invocation count

huijunwu: after long run, the mean metric becomes obtuse. 
for example, numerator_=10*1000, denominator_=1000, the mean is 10. 
then, 20 value pulse 100 times only lead to 0.9 increase on the mean. 
numerator_=10*1000+20*100, denominator_=1000+100, the mean is 10.9. 
is it the intention?

avflor: hey @huijunwu. What i'd like to have is the average queue length using the last N one second samples. For example If i want to ckeck this every minute, then I need the average over 60 samples (each sample taken every sec). Maybe I didn't understand the code correctly. Are you saying that this accumulates over time  instead of doing the above?

maosongfu: @huijunw At the start of every metrics-collection-cycle, it would reset the value as 0; then at the end of every metrics-collection-cycle, the metric value would be: sum of metrics value every time recorded/ sum of invocation count => the average value of this metric recorded during the cycle.

huijunw: i saw the reset 0 in MeanMetric::GetAndReset() and the cycle set in config 'heron.metrics.export.interval.sec: 60'. 
+1

objmagic:
I see this is part of 0.14.6 milestone. Any progress on this?

and @maosongfu, should I update viz code to reflect this metrics?

PR #1696, opened 2017-02-06
Remove defaults.yaml in favor of enum Key approach

billonahill:
Moving `defaults.yaml` functionality into `Key` enum and removing about 1k lines of code. This provides better typing, compile-time checks, elimination of resources, fewer config-related classes and fewer places to touch to add a config.

- The Defaults class is gone. That functionality is now handled by `ClusterConfig.loadConfig(String heronHome, String configPath, String releaseFile, String overrideConfigFile)` (used by production classes) and `Config.newBuilder(true)` (typically for unit tests).
- I still don't love the `ClusterConfig` class. I might refactor that in a follow-up PR if I can think of a cleaner way. 
- Now that a ton of cruft is gone, it would be good to refactor how we deal with sandbox configs to not be so clunky and repetitive.


objmagic:
PR is huge and has repetitive pattern so I just glanced through. LGTM. Nice use of enum to remove default configs.

huijunw:
I am curious why we used keys.yaml and defaults.yaml instead of enum at the very beginning.

billonahill:
Thanks @objmagic for the review. This one got huge indeed. Key classes  (no pun intended) are `Key`, `Config` and `ClusterConfig`. The rest is just the resulting fallout.

@huijunw not sure why this approach was taken initially.

PR #1687, opened 2017-02-01
[DO NOT MERGE] MultiSpoutsMultiTasks debugging

objmagic:
# DO NOT MERGE

Recently our Travis CI repeatedly fails on integration test `MultiSpoutsMultiTasks`.

Can't reproduce Travis CI integration test error locally, so open a PR here to really test Travis CI.

PR #1700, opened 2017-02-07
Temporarily disable MultiSpoutsMultiTasks test

objmagic:
continue investigation in #1687


billonahill:
👍 

Would you please rename #1687 to something like "Fix and re-enable MultiSpoutsMultiTasks test"

PR #1702, opened 2017-02-08
Removing sandbox enums that have duplicate concepts

billonahill:
First pass of sandbox cleanup. Removed a ton of repetitive code and sandbox enums that had duplicate non-sandbox concepts. See the `Key` file for the main changes, as well as `ClusterConfig`, `Misc` simplification and new tests in `ClusterConfigTests`.

More work to be done in follow-up PRs:
- Refactor our `HERON_SANDBOX_JAVA_HOME` and possibly `HERON_SANDBOX_CONF` and `HERON_SANDBOX_HOME` from Key.
- Update `AuroraField` enum values and aurora configs to not use SANDBOX_FOO fields.
- Possibly remove `Context` since it's an unnecessary place for new configs to also be added?
- Rename `Misc` class.


billonahill:
There's a flaw in this approach. `SubmitterMain` initializes configs using the local HERON_HOME (i.e.. ~/.) using [this method](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java#L305) when submitting:
```
    Config config = Config.expand(Config.newBuilder()
        .putAll(ClusterConfig.loadConfig(heronHome, configPath, releaseFile, overrideConfigFile))
        .putAll(commandLineConfigs(cluster, role, environ, dryRun, dryRunFormat, verbose))
        .putAll(topologyConfigs(topologyPackage, topologyBinaryFile, topologyDefnFile, topology))
.build());
```
This creates HERON_BIN as `${HERON_HOME}/bin` and HERON_SANDBOX_EXECUTOR as `${HERON_HOME}/bin/heron-executor` where `${HERON_HOME}=/var/lib/jenkins/.heron` which fails on aurora, because `${HERON_HOME}` is the absolute path of the heron install. What aurora needs to be set is `./heron-core/bin/heron-executor` based `${HERON_SANDBOX_HOME}/bin`.

For this reason trying to combine concepts like HERON_SANDBOX_BIN and HERON_BIN using smart wildcard substitution underneath for HERON_HOME, doesn't work in this case because the submitter needs both local configs (i.e. ${HERON_HOME}/lib) for submission and remote configs for running aurora tasks.

I need to think this through more, but suggestions welcome...

heron/spi/src/java/com/twitter/heron/spi/common/Config.java
```diff
@@ -44,17 +47,31 @@ public static Builder newBuilder(boolean loadDefaults) {
...
+        if (expandedValue.contains("${")) {
+          tokensCount++;
+        }
         cb.put(key, expandedValue);
       } else {
         cb.put(key, value);
       }
     }
-    return cb.build();
+    if (previousTokensCount != tokensCount) {
```
objmagic: what are the tokens and recursive expansion here...?

billonahill: Each time the method is invoked we iterate through each config value and do round of pattern substitution on the value, but substituting patterns can in the value can result in more tokens. For example `${HERON_BIN}/heron-executor` can be changed to `${HERON_HOME}/bin/heron-executor`, so after token substitution we can still have values with tokens. `tokensCount` counts the fields that still have tokens to know if we need to recurse for another pass.

I'm curious though about how this worked before without recursion so I'm going to investigate some more. If we do in fact need it, I'll provide more docs about what's happening.

billonahill:
I've come up with an approach to make the `Config` object hold on to the un-substituted configs (which I call the rawConfig) as well as the locally-substituted and remote-substituted configs. Consumers can toggle a config object back and forth with the `Config.toLocalMode(config)` and `Config.toRemoteMode(config)` methods as needed. Those methods return new cached config instances instead of mutating the one passed.

An example can be seen in how `AuroraScheduler` uses remote for setting the properties in `heron.aurora` (which is used on the server) but uses local for the heron config dir required to get the local path to `heron.aurora` used by the aurora CLI call.

Also note this PR has some duplicate unit test changes from #1704. Once #1704 ships and I merge this review will get smaller.

heron/spi/src/java/com/twitter/heron/spi/common/Misc.java
```diff
@@ -146,6 +67,23 @@ private static String substituteURL(Config config, String pathString) {
...
+  static {
+    SUBS.put("${HERON_HOME}", Key.HERON_HOME);
+    SUBS.put("${HERON_BIN}", Key.HERON_BIN);
+    SUBS.put("${HERON_CONF}", Key.HERON_CONF);
+    SUBS.put("${HERON_LIB}", Key.HERON_LIB);
+    SUBS.put("${HERON_DIST}", Key.HERON_DIST);
+    SUBS.put("${HERON_SANDBOX_HOME}", Key.HERON_HOME);
+    SUBS.put("${HERON_SANDBOX_BIN}", Key.HERON_BIN);
+    SUBS.put("${HERON_SANDBOX_CONF}", Key.HERON_CONF);
+    SUBS.put("${HERON_SANDBOX_LIB}", Key.HERON_LIB);
```
kramasamy: If sandbox is not needed any more, do we need this?

heron/spi/tests/java/com/twitter/heron/spi/common/ClusterDefaultsTest.java
```diff
@@ -30,13 +30,13 @@ public void initialize() {
   @Test
   public void testSandboxBinaries() throws Exception {
 
-    assertEquals(Key.SANDBOX_EXECUTOR_BINARY.getDefault(), Context.executorSandboxBinary(props));
-    assertEquals(Key.SANDBOX_STMGR_BINARY.getDefault(), Context.stmgrSandboxBinary(props));
-    assertEquals(Key.SANDBOX_TMASTER_BINARY.getDefault(), Context.tmasterSandboxBinary(props));
-    assertEquals(Key.SANDBOX_SHELL_BINARY.getDefault(), Context.shellSandboxBinary(props));
+    assertEquals(Key.EXECUTOR_BINARY.getDefault(), Context.executorBinary(props));
+    assertEquals(Key.STMGR_BINARY.getDefault(), Context.stmgrSandbox(props));
```
kramasamy: stmgrBinary?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerConfigUtils.java
```diff
@@ -78,7 +78,7 @@ public static Config loadConfig(
       Boolean verbose,
       TopologyAPI.Topology topology) {
 
-    return Config.expand(
+    return Config.toRemoteMode(
```
billonahill: Yeah, I don't love the terminology here of local, remote or sandbox and would like to change it. It's really about how Heron is installed. It's either installed on the system, or bundled in a distribution. I was thinking about "installed" and "bundled" modes instead, but those aren't great either. Suggestions welcome.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -280,7 +280,7 @@ public static void main(String[] args)
         .put(Key.TOPOLOGY_NAME, topologyName);
 
     // build the final config by expanding all the variables
-    Config config = Config.expand(Config.newBuilder()
+    Config config = Config.toLocalMode(Config.newBuilder()
```
billonahill: Agreed I don't like local mode. Client is also overloaded and could cause confusion. Let's discuss in the similar comment above.

heron/spi/src/java/com/twitter/heron/spi/common/ClusterConfig.java
```diff
@@ -25,137 +25,60 @@
   private ClusterConfig() {
   }
 
-  private static Config loadHeronHome(String heronHome, String configPath) {
-    Config.Builder cb = Config.newBuilder()
+  private static Config loadDefaults(String heronHome, String configPath) {
+    return Config.newBuilder(true)
```
billonahill: True says to initialize the Config with default values. See `Config.Builder.create(loadDefaults)` which this call delegates to.

billonahill:
@kramasamy the previous one-level expand method was broken. I think it only worked because callers happened to call expand more than once. I've added `ClusterConfigTest.assertConfig(..)` which now asserts that all tokens are substituted.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -204,38 +203,37 @@ private SchedulerUtils() {
     commands.add(FileUtils.getBaseName(Context.topologyDefinitionFile(config)));
     commands.add(Context.stateManagerConnectionString(config));
     commands.add(Context.stateManagerRootPath(config));
-    commands.add(Context.tmasterSandboxBinary(config));
-    commands.add(Context.stmgrSandboxBinary(config));
-    commands.add(Context.metricsManagerSandboxClassPath(config));
+    commands.add(Context.tmasterBinary(config));
+    commands.add(Context.stmgrSandbox(config));
```
billonahill: fixed

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -171,24 +172,24 @@ protected String formatJavaOpts(String javaOpts) {
...
     auroraProperties.put(AuroraField.TOPOLOGY_ID, topology.getId());
     auroraProperties.put(AuroraField.TOPOLOGY_DEFINITION_FILE,
         FileUtils.getBaseName(Context.topologyDefinitionFile(config)));
     auroraProperties.put(AuroraField.STATEMGR_CONNECTION_STRING,
         Context.stateManagerConnectionString(config));
     auroraProperties.put(AuroraField.STATEMGR_ROOT_PATH, Context.stateManagerRootPath(config));
-    auroraProperties.put(AuroraField.SANDBOX_TMASTER_BINARY, Context.tmasterSandboxBinary(config));
-    auroraProperties.put(AuroraField.SANDBOX_STMGR_BINARY, Context.stmgrSandboxBinary(config));
+    auroraProperties.put(AuroraField.SANDBOX_TMASTER_BINARY, Context.tmasterBinary(config));
+    auroraProperties.put(AuroraField.SANDBOX_STMGR_BINARY, Context.stmgrSandbox(config));
```
billonahill: fixed

heron/spi/src/java/com/twitter/heron/spi/common/Misc.java
billonahill: fixed

heron/spi/tests/java/com/twitter/heron/spi/common/ClusterDefaultsTest.java
billonahill: fixed

billonahill:
@kramasamy all comments addressed. Remaining open issue is that of terminology.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerConfigUtils.java
kramasamy: For remote - 'toClusterMode' is more appropriate I think.

billonahill:
Updated terminology to use LocalMode and ClusterMode. All identified issues have been addressed.

PR #1703, opened 2017-02-08
Upgrade packages

objmagic:


kramasamy:
👍

billonahill:
👍 assuming you've verified that the site can still be properly built on mac and internal/external CI and that the rendering is correct.

objmagic:
this seems to be the root cause: https://github.com/jakubpawlowicz/clean-css/issues/888

huijunw:
👍

objmagic:
CI is still using `node -v v0.10.36`. We need 0.12.x

PR #1704, opened 2017-02-08
Adding unit test for AuroraScheduler.createAuroraProperties(..)

billonahill:
- Moving some things around in `SubmitterMain` and `AuroraScheduler` to make it easier to test.
- Adding a unit test for `AuroraScheduler.createAuroraProperties(..)`. This is needed to validate work in #1702.
- Fixing how Misc gets `HOME` to be more idiomatic and to not rely on swing package, which is failing for me.
- Removing `heron/spi/tests/java/com/twitter/heron/spi/common/testdata/local/defaults.yaml` which is no longer needed.


billonahill:
Ping. Any comments on this one?

PR #1708, opened 2017-02-13
Rename Misc class to TokenSub class

billonahill:
Let me know if anyone has a better suggestion that is not `StringUtils`.


huijunw:
looks good to me

PR #1709, opened 2017-02-13
Removing SANDBOX_ from the AuroraField and aurora properties

billonahill:
No need for aurora properties to all have `SANDBOX_` prepended so removing it. This will also require and update of internal `heron.aurora`.


billonahill:
Ping

PR #1710, opened 2017-02-14
Use resources label

### LINKED ISSUE #1668: Use resources instead of data to load sample files in Java tests
See: https://groups.google.com/forum/#!topic/bazel-discuss/VUC8kTEXDYo

We also need to update the [java_tests.bzl](https://github.com/twitter/heron/blob/master/tools/rules/java_tests.bzl)

example: https://github.com/bazelbuild/bazel/blob/master/examples/java-native/src/main/java/com/example/myproject/Greeter.java#L36

objmagic:
use `resources` label as much as possible. Some places are still using `data` (see by `find heron/ -name "BUILD" -exec grep -RIn "data = " {} +`), but I think using `data` makes loading easier.

resolves #1668


heron/common/tests/java/com/twitter/heron/common/config/ConfigReaderTest.java
```diff
@@ -14,27 +14,56 @@
...
     Assert.assertEquals("role", props.get(Constants.ROLE_KEY));
     Assert.assertEquals("environ", props.get(Constants.ENVIRON_KEY));
     Assert.assertEquals("com.twitter.heron.scheduler.aurora.AuroraLauncher",
         props.get(Constants.LAUNCHER_CLASS_KEY));
-
     Assert.assertNull(props.get(Constants.USER_KEY));
   }
+
+  private InputStream loadResource() {
+    final String RESOURCE_LOC = "/heron/common/tests/resources/defaults.yaml";
```
billonahill: please change to a class-level psf String.

heron/common/tests/java/com/twitter/heron/common/config/ConfigReaderTest.java
```diff
@@ -14,27 +14,56 @@
...
+    if (inputStream == null) {
+      throw new RuntimeException("Sample output file not found");
+    }
+    return inputStream;
+  }
+
+  @Test
+  public void testLoadFile() throws Exception {
+    InputStream inputStream = loadResource();
+    final String PREFIX = "defaults_temp";
```
billonahill: ALL_CAPS is for class-level psf Strings. These should either be moved to class level or just passed directly to createTempFile without making a one-time use string variable.

heron/common/tests/java/com/twitter/heron/common/config/ConfigReaderTest.java
```diff
@@ -14,27 +14,56 @@
...
   }
+
+  private InputStream loadResource() {
+    final String RESOURCE_LOC = "/heron/common/tests/resources/defaults.yaml";
+    InputStream inputStream  = ConfigReaderTest.class.
+        getResourceAsStream(RESOURCE_LOC);
+    if (inputStream == null) {
+      throw new RuntimeException("Sample output file not found");
+    }
+    return inputStream;
```
billonahill: Does the caller ever need an InputStream for anything other than creating a Map<String, Object>? If not, let's just return a  Map<String, Object> here.

heron/common/tests/java/com/twitter/heron/common/config/Constants.java
```diff
@@ -23,9 +23,6 @@
 
   public static final String LAUNCHER_CLASS_KEY = "heron.launcher.class";
 
-  public static final String TEST_DATA_PATH =
```
billonahill: woot!

billonahill: While you're here would you please remove `public` from all these constants (and the class) and remove GROUP_KEY and VERSION_KEY, which are unused. #cruftslayer

heron/common/tests/java/com/twitter/heron/common/config/ConfigReaderTest.java
objmagic: because I also want to test loading from InputStream here: https://github.com/twitter/heron/pull/1710/files/4f721caaa991ff8c3cc156ebab9ce9b294993720#diff-535d4127d634ce5fd544311ba8e7fb2cR63

objmagic:
@billonahill comments addressed.

objmagic:
@billonahill any more feedback?

PR #1713, opened 2017-02-15
CentOS upgrade preparation

objmagic:
[CentOS 5 EOL is 03/31/2017](https://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d) and we are dropping support of CentOS 5. This PR is just path name change.

Will also do a PR on internal codebase to coordinate with this change. They'd better be merged at almost the same time.


kramasamy:
@objmagic - I don't think these files will be needed for centos7. You can build the binary scripts/packages:binpkgs. Can you verify and see if this necessary?

PR #1714, opened 2017-02-15
Fixing bug in WebSink metrics collector

fabianmenges:
I found a bug in the WebSink metrics collector, that I contributed a couple month ago.

If the metrics are getting exported in the hierarchical format, a new `MetricsRecord` would override and old `MetricsRecord` and the `MetricsInfo`s associated to it. Now a new `MetricsRecord` will merge the `MetricsInfo`s of an old `MetricsRecord`. A cache with expiration is used.

### DIFF
--- a/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
+++ b/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
@@ -84,6 +84,9 @@ public class WebSink implements IMetricsSink {
   private boolean isFlatMetrics = true;
   private boolean includeTopologyName = true;
   private String topologyName;
+  private long cacheMaxSize;
+  private long cacheTtlSec;
+
 
   @Override
   public void init(Map<String, Object> conf, SinkContext context) {
@@ -92,8 +95,8 @@ public class WebSink implements IMetricsSink {
     isFlatMetrics = TypeUtils.getBoolean(conf.getOrDefault(KEY_FLAT_METRICS, true));
     includeTopologyName = TypeUtils.getBoolean(
             conf.getOrDefault(KEY_INCLUDE_TOPOLOGY_NAME, false));
-    long cacheMaxSize = TypeUtils.getLong(conf.getOrDefault(KEY_METRICS_CACHE_MAX_SIZE, 1000000));
-    long cacheTtlSec = TypeUtils.getLong(conf.getOrDefault(KEY_METRICS_CACHE_TTL_SEC, 600));
+    cacheMaxSize = TypeUtils.getLong(conf.getOrDefault(KEY_METRICS_CACHE_MAX_SIZE, 1000000));
+    cacheTtlSec = TypeUtils.getLong(conf.getOrDefault(KEY_METRICS_CACHE_TTL_SEC, 600));
 
     metricsCache = CacheBuilder.newBuilder()
             .maximumSize(cacheMaxSize)
@@ -192,8 +195,20 @@ public class WebSink implements IMetricsSink {
     if (isFlatMetrics) {
       metricsCache.putAll(processMetrics(source + "/", record.getMetrics()));
     } else {
-      Map<String, Double> sourceMap = processMetrics("", record.getMetrics());
-      metricsCache.put(source, sourceMap);
+      Map<String, Double> sourceCache;
+      Object sourceObj = metricsCache.getIfPresent(source);
+      if (sourceObj instanceof Map) {
+        @SuppressWarnings("unchecked")
+        Map<String, Double> castObj =  (Map<String, Double>) sourceObj;
+        sourceCache = castObj;
+      } else {
+        sourceCache =  CacheBuilder.newBuilder()
+            .maximumSize(cacheMaxSize)
+            .expireAfterWrite(cacheTtlSec, TimeUnit.SECONDS)
+            .<String, Double>build().asMap();
+      }
+      sourceCache.putAll(processMetrics("", record.getMetrics()));
+      metricsCache.put(source, sourceCache);
     }
   }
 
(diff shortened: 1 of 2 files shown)


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -84,6 +84,9 @@
   private boolean isFlatMetrics = true;
   private boolean includeTopologyName = true;
   private String topologyName;
+  private long cacheMaxSize;
```
kramasamy: Some comments - will be nice.

fabianmenges: I would be just repeating this comment,

https://github.com/tc-dc/heron/blob/ed0e25cb71dd882c159ef4bbcf1fe634a891667c/heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java#L74

fabianmenges:
Thanks

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -192,8 +195,20 @@ public void processRecord(MetricsRecord record) {
...
       metricsCache.putAll(processMetrics(source + "/", record.getMetrics()));
     } else {
-      Map<String, Double> sourceMap = processMetrics("", record.getMetrics());
-      metricsCache.put(source, sourceMap);
+      Map<String, Double> sourceCache;
+      Object sourceObj = metricsCache.getIfPresent(source);
+      if (sourceObj instanceof Map) {
+        @SuppressWarnings("unchecked")
+        Map<String, Double> castObj =  (Map<String, Double>) sourceObj;
+        sourceCache = castObj;
```
maosongfu: Can we merge L202 and L203 into one line:
sourceCache = (Map<String, Double>) sourceObj;

fabianmenges: Sadly we cannot. Apparently you can not annotate a definition if the variable is not also assigned.
e.g. 
```
Map<String, Double> sourceCache;
Object sourceObj = metricsCache.getIfPresent(source);
if (sourceObj instanceof Map) {
    sourceCache =  (Map<String, Double>) sourceObj;
...
```
breaks with `warning: [unchecked] unchecked cast` and
```
sourceCache =  (Map<String, Double>) sourceObj;
```
is not legal syntax

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
```diff
@@ -192,8 +195,20 @@ public void processRecord(MetricsRecord record) {
...
+      } else {
+        sourceCache =  CacheBuilder.newBuilder()
+            .maximumSize(cacheMaxSize)
+            .expireAfterWrite(cacheTtlSec, TimeUnit.SECONDS)
+            .<String, Double>build().asMap();
+      }
+      sourceCache.putAll(processMetrics("", record.getMetrics()));
```
huijunw: does this override the old metric of the same name in the sourceCache?

fabianmenges: Yes, it does. Its the standard `.putAll` method from the Map interface. 
```
public void putAll(Map<? extends K,? extends V> m)
Copies all of the mappings from the specified map to this map. These mappings will replace any mappings that this map had for any of the keys currently in the specified map.
```

huijunwu: so it ignores the metric type and caches only the latest one metric

fabianmenges: I'm not sure what you mean with ignoring the metric type. 

Lets say we have two records `record_1` and `record_2` with this yaml representation:
```
record_1:
  source: source_1
  metrics: 
    metric_1: 1
    metric_2: nop
    metric_3: 3
record_2:
  source: source_1
  metrics:
    metric_1: 2
    metric_4: 4
```
And they are getting send to the WebSink in order.
The WebSink will publish these metrics on a REST endpoint returning json. 
After "record_1" is processed the endpoint will return
```
"source_1": {
    "metric_1": 1.0
    "metric_3": 3.0
}
```
As you can see only numeric metrics are published, "metric_2" is omitted.
After "record_2" is processed the endpoint will return:
```
"source_1": {
  "metric_1": 2.0,
  "metric_3": 3.0,
  "metric_4": 5.0,
}
```
"metric_1" was updated, "metric_3" was not changed, "metric_4" was added by the 2nd record.
The old behavior, before this change, would have returned:
```
"source_1": {
  "metric_1": 2.0,
  "metric_4": 5.0,
}
```

The point of this metrics sink is, that an external process queries periodically (once a min) the 
REST endpoint to collect the metrics. If the two records would get processed right after one another,
there would be the chance that the metrics collector would never see "metric_3".

huijunw: i see. thank you

maosongfu:
looks good to me. 👍

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/WebSink.java
maosongfu: sounds good.

objmagic:
note to team: this PR is merged into master, but we will also incorporate it in 0.14.6. I'll do that.

Issue #970, opened 2016-06-22
Provide default values for configs


maosongfu:
We need to uniform strategy to provide default values for configs in heron-internals. 
Otherwise, topology can throw null pointer exception and crash if someone has bundled an older version of heron_internals.yaml with their new heron release, which depending on new config values.

PR #1717, opened 2017-02-15
Change SystemConfig to use SystemConfigKey instead of constants

billonahill:
Refactor to use enum keys for `SystemConfig` class. Introduce a builder so `SystemConfig` can become immutable, and refactor `Simulator` and test classes to use it for custom settings.

Will follow up with refactoring other consumers to use builder and get rid of the use of `public SystemConfig(String configFile, boolean mustExist)` constructor.


billonahill:
I went ahead and removed usage of `public SystemConfig(String configFile, boolean mustExist)` constructor in favor of the Builder approach since it wasn't that big of a change.

heron/common/src/java/com/twitter/heron/common/config/SystemConfigKey.java
```diff
@@ -0,0 +1,396 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import com.twitter.heron.common.basics.ByteAmount;
+
+/**
+ * Keys to be used in the SystemConfigs
+ */
+// TODO(mfu): Provide default values for all configs uniformly
+// TODO(mfu): https://github.com/twitter/heron/issues/970
```
objmagic: is this TODO addressed?

nlu90:
👍

heron/common/src/java/com/twitter/heron/common/config/SystemConfigKey.java
billonahill: We now have a mechanism for defaults and we use it for 2 config keys. @maosongfu do you think that's enough for #970 or should we also be adding defaults for other key values. You comment was initially above keys that now have a default, so maybe we're ok now?

PR #1725, opened 2017-02-19
Remove special case CentOS build files

objmagic:
for CentOS 7 build and general clean-up and refactoring

An internal PR (D28289) has also been submitted.


kramasamy:
@objmagic - have you tested the package for centos7 before removing these files?

huijunw:
what were these 2 files originally for?

objmagic:
@huijunw because of Bazel's own issue, we have to write a separate BUILD file for CentOS 5 (and cannot reuse the one for Mac). This issue does not happen on CentOS 7. Since we are migrating to CentOS 7, we can remove these two historical files.

huijunw:
thanks @objmagic

Issue #1699, opened 2017-02-07
More info-level logging for CLI


objmagic:

objmagic:
display of internal uploader command, scheduler command under info mode. Also consider improve ShellUtils.

objmagic:
related https://github.com/twitter/heron/issues/1588

PR #1726, opened 2017-02-19
Real time I/O redirection

### LINKED ISSUE #1698: real-time redirection of I/O of shelled-out Java program

objmagic:
resolves #1698 and #1699 

Using while-loops instead of for-loops because of some implementation issue of Python2.7


heron/tools/cli/src/python/opts.py
```diff
@@ -21,6 +21,8 @@
 verbose_flag = False
 trace_execution_flag = False
 
+cleaned_up_files = []
```
huijunw: why not put it in main.py?

objmagic: because this variable is also used in other modules.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
+              break
+          else:
+            print >> sys.stdout, stdout_line[:-1]
+        else:
+          print >> sys.stderr, stderr_line[:-1]
+      proc.wait()
+      resp.status = status_type(proc.returncode)
     else:
-      raise RuntimeError("Unknown status type of value %d", resp.status)
+      assert resp.status is not None, "Unexpected break of invariant"
```
huijunw: can we check the invariant in the constructor?

objmagic: makes sense. will do.

objmagic:
https://github.com/twitter/heron/pull/1726/commits/4c46bb49aeee15225a0a16b6fec5dc14037b5ceb addressed https://github.com/twitter/heron/issues/1699

objmagic:
still want to talk to @billonahill about this.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -109,14 +108,14 @@ private static int runSyncProcess(
...
-    LOG.log(Level.FINE, "Process command: `$ {0}`", Arrays.toString(cmdline));
+    LOG.log(Level.INFO, "Process command: ``{0}''", String.join(" ", cmdline));
     ProcessBuilder pb = getProcessBuilder(isInheritIO, cmdline, workingDirectory, envs);
 
     Process process;
     try {
       process = pb.start();
     } catch (IOException e) {
-      LOG.log(Level.SEVERE, "Failed to run Sync Process ", e);
+      LOG.log(Level.SEVERE, "Failed to run sync process ", e);
```
billonahill: In this and other severe logs (below) can we include the process string in the message.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -183,7 +182,7 @@ public static Process runASyncProcess(String command) {
 
   private static Process runASyncProcess(String[] command, File workingDirectory,
       Map<String, String> envs, String logFileUuid, boolean logStderr) {
-    LOG.log(Level.FINE, "$> {0}", Arrays.toString(command));
+    LOG.log(Level.INFO, "Run async process: ``{0}''", String.join(" ", command));
```
billonahill: s/Run/Running/

heron/spi/src/java/com/twitter/heron/spi/utils/TopologyUtils.java
```diff
@@ -50,7 +50,7 @@ private TopologyUtils() {
```
billonahill: `verifyTopology` should really be fixed to return `void` and throw a better exception that `RuntimeException` (with a better message) but that's a fix for another PR...

heron/tools/cli/src/python/execute.py
```diff
@@ -147,8 +133,11 @@ def heron_pex(topology_pex, topology_class_name, args=None):
...
+      return Response(status=Status.Ok)
     except Exception as ex:
       Log.debug(traceback.format_exc())
-      err_str = "Topology %s failed to be loaded from the given pex" % topology_class_name
-      return Response(Status.HeronError, err_str, str(ex))
+      err_str = "Topology %s failed to be loaded from the given pex: %s" %\
+                (topology_class_name, ex)
+      resp = Response(status=Status.HeronError)
+      resp.add_context(err_str)
+      return err_str
```
billonahill: Why does this return the error string and not the response object containing the error string?

heron/tools/cli/src/python/main.py
```diff
@@ -125,7 +127,10 @@ def cleanup(files):
   :return:
   '''
   for cur_file in files:
-    shutil.rmtree(os.path.dirname(cur_file))
+    if os.path.isdir(cur_file):
+      shutil.rmtree(cur_file)
+    else:
+      shutil.rmtree(os.path.dirname(cur_file))
```
billonahill: Why add the conditional? Does `shutil.rmtree(os.path.dirname(cur_file))` fail if `cur_file` is a directory?

heron/tools/cli/src/python/response.py
```diff
@@ -52,10 +54,10 @@ class Response(object):
      If the response object is generated by a statement that
      shells out program, `msg` is stdout, `detailed_msg` is stderr
   """
-  def __init__(self, status_code, msg=None, detailed_msg=None):
-    self.status = status_type(status_code)
-    self.msg = msg
-    self.detailed_msg = detailed_msg
+  def __init__(self, proc=None, status=None):
+    # invariant: only one of proc and status is None
```
billonahill: Is there some python magic enforcing this? If not, let's explicitly check the state.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
+def log_context(resp):
+  status = resp.status
+  if status == Status.Ok or status == Status.DryRun:
+    do_log(Log.info, resp.succ_context)
+  elif status == Status.HeronError:
+    do_log(Log.error, resp.err_context)
+  elif status == Status.InvocationError:
+    pass
+  else:
+    raise RuntimeError("Unknown status type of value %d", status)
```
billonahill: can we also log `resp` here?

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
+    log(msg)
+
+def log_context(resp):
+  status = resp.status
+  if status == Status.Ok or status == Status.DryRun:
+    do_log(Log.info, resp.succ_context)
+  elif status == Status.HeronError:
+    do_log(Log.error, resp.err_context)
+  elif status == Status.InvocationError:
+    pass
```
billonahill: should include a comment re why we don't log here.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
     self.err_context = err_context
     self.succ_context = succ_context
 
+def do_log(log, msg):
```
billonahill: do_log and log_context seem like they are intended to be private. If that's the case please move them to the bottom and prepend `_` to the method name to make that clear.

heron/tools/cli/src/python/response.py
billonahill: Also please provide more context in the error message.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
+    do_log(Log.info, resp.succ_context)
+  elif status == Status.HeronError:
+    do_log(Log.error, resp.err_context)
+  elif status == Status.InvocationError:
+    pass
+  else:
+    raise RuntimeError("Unknown status type of value %d", status)
+
+# # pylint: disable=too-many-branches,unreachable
 def render(resp):
```
billonahill: Why is render a static method and not an instance method on Response? The latter seems more natural.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
-  def do_log(log, msg):
-    if msg:
-      log(msg)
+  """ Render output of the process until it exits
+
+  :param resp: Response
+  :return:
+  """
   if isinstance(resp, list):
```
billonahill: This `isinstance` of list or Response seems like some python jank. Are there places where lists of Response objects are being passed in? If not, we should remove this.

heron/tools/cli/src/python/response.py
```diff
@@ -68,33 +70,55 @@ def add_context(self, err_context, succ_context=None):
...
-    elif resp.status == Status.DryRun:
-      do_log(Log.info, resp.succ_context)
-      # No need to prefix [INFO] here. We want to display dry-run response in a clean way
-      print resp.msg
-      do_log(Log.debug, resp.detailed_msg)
+    if resp.proc:
+      proc = resp.proc
+      # use while loop here because of a bug of Python2.7
+      while True:
+        stderr_line = proc.stderr.readline()
```
billonahill: In this method you're reading stderr, polling if None and then readying stdout, then polling if None. Are their chances for raceconditions here due to state changing on proc between calls? Also are there edge conditions possible when both stderr and stdout are set? I don't know the specifics of the proc API in this sense to say for sure.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: what is process string? do you mean the command? we already do info-level logging of command [here](https://github.com/twitter/heron/pull/1726/files/4c46bb49aeee15225a0a16b6fec5dc14037b5ceb#diff-49b803240c7fae81aca5e36ac0559735R111).

heron/tools/cli/src/python/main.py
objmagic: if it is a directory, we remove the directory. If it is a temporary file, we remove the entire temporary directory that contains this temporary file.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
billonahill: Yes the command. This is always good to log in the exception so you don't have to dig through previous logs to find all others commands run, and then guess which one failed.

heron/tools/cli/src/python/main.py
billonahill: I follow the desired functionality, just not the need for the if/else. Without it would we not get the desired functionality?

objmagic: > Does shutil.rmtree(os.path.dirname(cur_file)) fail if cur_file is a directory

Yes. See below for API documentation. For `/usr/lib`, you'll get `/usr`.

> `os.path.dirname(path)`
> Return the directory name of pathname path. This is the first element of the pair returned by passing path to the function split().

heron/tools/cli/src/python/response.py
objmagic: most functions return list of responses because for some reason I don't know we [can launch multiple topologies](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/submit.py#L131). [help](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/help.py) and [version](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/version.py) return a single response object.

We can just make them all to return a list of response objects.

heron/tools/cli/src/python/response.py
objmagic: no particular reason. I'll make it an instance method.

heron/tools/cli/src/python/response.py
billonahill: Forcing all to use a list is also clunky. I'd say leave it as is for now.

heron/tools/cli/src/python/response.py
objmagic: i don't understand. you mean log `succ_context` and `err_context`? why?

billonahill: Note necessarily, no. Just something to give the user more actionable info than what they'd have with a message like "Unknown status type of value foo". If they got that message and reported it to us we'd have no idea what that mean and or what triggered it.

heron/tools/cli/src/python/response.py
objmagic: `render` could also render a list of responses so it should be static method.

heron/tools/cli/src/python/response.py
objmagic: The reason I implemented in this way is that all Java logging go to stderr and exception message/dry-run response go to stdout. That said, stderr always go before stdout (I am using `always` but not 100% sure this is the case). I may follow [this answer](http://stackoverflow.com/questions/375427/non-blocking-read-on-a-subprocess-pipe-in-python/1810703#1810703) to make some improvement.

objmagic:
@billonahill most reviews about Python programs are addressed. Working on tuning ShellUtils now.

heron/tools/cli/src/python/response.py
objmagic: i'm sticking with my original approach.

objmagic:
@billonahill updated. We can discuss about the output in-person tomorrow (don't want to upload terminal recording clip since it contains internal service info)

heron/tools/cli/src/python/main.py
objmagic: if there is no further question, I'll mark this as resolved.

heron/tools/cli/src/python/opts.py
objmagic: mark as resolved.

heron/tools/cli/src/python/result.py
```diff
@@ -0,0 +1,188 @@
...
+    self._log_context()
+
+
+class ProcessResult(Result):
+  """Process result: a wrapper of result class"""
+  def __init__(self, proc):
+    super(ProcessResult, self).__init__()
+    self.proc = proc
+
+  def renderProcessStdErr(self, stderr_line):
```
objmagic: This function needs very careful review... @billonahill

heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
```diff
@@ -376,6 +376,11 @@ public SubmitterMain(
    *
    */
   public void submitTopology() throws TopologySubmissionException {
+    Config primaryRuntime = Config.newBuilder()
+          .putAll(LauncherUtils.getInstance().createPrimaryRuntime(topology)).build();
+    if (Context.dryRun(config)) {
+      callLauncherRunner(primaryRuntime);
```
billonahill: Funny I just made the same change on my branch. :) You need a return statement here.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -281,7 +281,7 @@ public static boolean setSchedulerLocation(
 
     Scheduler.SchedulerLocation location = builder.build();
 
-    LOG.log(Level.INFO, "Setting SchedulerLocation: {0}", location);
+    LOG.log(Level.INFO, "Setting Scheduler locations:\n{0}", location);
```
billonahill: Inserting line breaks in log entries can cause pains when reading the logs. Can we remove the \n?

heron/schedulers/src/java/com/twitter/heron/scheduler/slurm/SlurmController.java
```diff
@@ -140,7 +137,7 @@ public boolean killJob(String jobIdFile) {
     List<String> jobIdFileContent = readFromFile(jobIdFile);
     if (jobIdFileContent.size() > 0) {
       String[] slurmCmd = new String[]{"scancel", jobIdFileContent.get(0)};
-      return runProcess(null, slurmCmd, new StringBuilder(), new StringBuilder());
+      return runProcess(null, slurmCmd, new StringBuilder());
```
billonahill: Why does this method call require passing a StringBuilder that the caller never uses? Can we simplify the API to not require this here, maybe with another signature?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
-  private static Thread asyncProcessStream(final InputStream input, final StringBuilder out) {
+  private static Thread asyncProcessStream(final InputStream input, final StringBuilder builder) {
     Thread thread = new Thread() {
       @Override
       public void run() {
-        try {
-          out.append(inputstreamToString(input));
-        } finally {
+        // do not buffer
+        System.err.println("Process output (stdout+stderr):");
```
billonahill: Why write directly to stderr here instead of LOG like we do below?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
-            input.close();
+            line = reader.readLine();
           } catch (IOException e) {
-            LOG.log(Level.WARNING, "Failed to close the input stream", e);
+            LOG.log(Level.SEVERE, "Error when reading line from subprocess", e);
+          }
+          if (line == null) {
+            break;
+          } else {
+            System.err.println(line);
```
billonahill: we should clarify in the javadocs how this method behaves w.r.t. logging and stderr.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
         }
+        try {
+          input.close();
+        } catch (IOException e) {
+          LOG.log(Level.WARNING, "Failed to close the input stream", e);
+        }
       }
     };
     thread.setDaemon(true);
-    thread.start();
```
billonahill: why do we no longer start the thread? This seems unexpected from the callers point of view. If we don't start it we need to clarify in the javadocs how to properly use this method.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
   private static int runSyncProcess(
-      boolean isInheritIO, String[] cmdline, StringBuilder stdout,
+      boolean isInheritIO, String[] cmdline,
       StringBuilder stderr, File workingDirectory, Map<String, String> envs) {
-    final StringBuilder pStdOut = stdout == null ? new StringBuilder() : stdout;
     final StringBuilder pStdErr = stderr == null ? new StringBuilder() : stderr;
 
     // Log the command for debugging
-    LOG.log(Level.FINE, "Process command: `$ {0}`", Arrays.toString(cmdline));
+    LOG.log(Level.INFO, "Running synced process:\n``{0}''''", String.join(" ", cmdline));
```
billonahill: can we not inject \n here?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
     ProcessBuilder pb = getProcessBuilder(isInheritIO, cmdline, workingDirectory, envs);
+    /* combine input stream and error stream because
+       1. this preserves order of process's stdout/stderr message
+       2. there is no need to distinguish stderr from stdout
+       3. follow one basic pattern of the design of Python<~>Java I/O redirection:
+          stdout contains useful message Java program needs to propagate back, stderr
```
billonahill: s/message/messages/

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,63 +60,80 @@ static String inputstreamToString(InputStream is) {
...
+      boolean isInheritIO, String[] cmdline,
       StringBuilder stderr, File workingDirectory, Map<String, String> envs) {
-    final StringBuilder pStdOut = stdout == null ? new StringBuilder() : stdout;
     final StringBuilder pStdErr = stderr == null ? new StringBuilder() : stderr;
 
     // Log the command for debugging
-    LOG.log(Level.FINE, "Process command: `$ {0}`", Arrays.toString(cmdline));
+    LOG.log(Level.INFO, "Running synced process:\n``{0}''''", String.join(" ", cmdline));
     ProcessBuilder pb = getProcessBuilder(isInheritIO, cmdline, workingDirectory, envs);
+    /* combine input stream and error stream because
```
billonahill: pls clarify that we're combining the two into stderr, which I think is the case.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -125,39 +142,23 @@ private static int runSyncProcess(
     // stream is read while waiting for the process to complete. If either buffer becomes full, it
```
billonahill: this comment talks about multiple buffers, which we no longer have below. please update to reflect new behaviour

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -183,7 +184,7 @@ public static Process runASyncProcess(String command) {
 
   private static Process runASyncProcess(String[] command, File workingDirectory,
       Map<String, String> envs, String logFileUuid, boolean logStderr) {
-    LOG.log(Level.FINE, "$> {0}", Arrays.toString(command));
+    LOG.log(Level.INFO, "Running async process:\n``{0}''''", String.join(" ", command));
```
billonahill: pls remove \n

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -259,6 +260,7 @@ static Process establishSSHTunnelProcess(
   }
 
   static Process establishSocksProxyProcess(String proxyHost, int proxyPort) {
+    LOG.info("Establishing SOCKS proxy...");
```
billonahill: pls include host and port

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -278,8 +280,8 @@ public static boolean curlPackage(
 
     // using curl copy the url to the target file
     String cmd = String.format("curl %s -o %s", uri, destination);
-    int ret = runSyncProcess(isVerbose, isInheritIO,
-        splitTokens(cmd), new StringBuilder(), new StringBuilder(), parentDirectory);
+    int ret = runSyncProcess(isInheritIO,
+        splitTokens(cmd), new StringBuilder(), parentDirectory);
```
billonahill: here and below we're passing an unused StringBuilder. can we refactor to not require that?

heron/tools/cli/src/python/result.py
billonahill: This logic seems sound if it works in your tests.

In the future, best to avoid making code changes _and_ doing a file rename at the same time, since there's no longer a diff to review.

heron/schedulers/src/java/com/twitter/heron/scheduler/slurm/SlurmController.java
objmagic: only internal uploader tool needs saves all redirected I/O using string builder.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: you are right. we should use LOG here.

heron/schedulers/src/java/com/twitter/heron/scheduler/slurm/SlurmController.java
objmagic: adding another signature actually means we will add three or four more signatures... https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java#L63

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: it will be started later.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: this is used for internal uploader tool. if you ``--json`` to internal uploader tool, it will return a JSON result and we want to catch that

heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
objmagic: fixed.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: fixed

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -281,7 +281,7 @@ public static boolean setSchedulerLocation(
 
     Scheduler.SchedulerLocation location = builder.build();
 
-    LOG.log(Level.INFO, "Setting SchedulerLocation: {0}", location);
+    LOG.log(Level.INFO, "Setting Scheduler locations:{0}", location);
```
billonahill: space after :

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -60,104 +60,112 @@ static String inputstreamToString(InputStream is) {
...
-   * Start a daemon thread to read data from "input" to "out".
+   * Read and print line from input, and save each line in a string builder
+   * line read from input is printed to stderr directly instead of using LOG (which will
+   * add redundant timestamp and class name info)
+   * @param input input stream
+   * @param builder string builder used to save input lines
+   * @return
    */
-  private static Thread asyncProcessStream(final InputStream input, final StringBuilder out) {
+  private static Thread asyncProcessStream(final InputStream input, final StringBuilder builder) {
```
billonahill: can we name builder stdout or output or something more descriptive?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
billonahill: please clarify in the javadocs that this method doesn't actually run anything. We should rename the method to make that very obvious too actually. Maybe createAsyncProcessThread?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -183,7 +191,7 @@ public static Process runASyncProcess(String command) {
 
   private static Process runASyncProcess(String[] command, File workingDirectory,
       Map<String, String> envs, String logFileUuid, boolean logStderr) {
-    LOG.log(Level.FINE, "$> {0}", Arrays.toString(command));
+    LOG.log(Level.INFO, "Running async process:``{0}''''", String.join(" ", command));
```
billonahill: space after :

objmagic: https://github.com/twitter/heron/pull/1726/commits/ee0b4f2d4ca72eb0a133dcf67670399f50a17287

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: https://github.com/twitter/heron/pull/1726/commits/ee0b4f2d4ca72eb0a133dcf67670399f50a17287

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
objmagic: https://github.com/twitter/heron/pull/1726/commits/ee0b4f2d4ca72eb0a133dcf67670399f50a17287

Issue #1727, opened 2017-02-20
Upgrade YarnLauncher driverMemory use ByteAmount


mycFelix:
As #1707 reported, currently, the YarnLauncher class member `driverMemory` use `int`. We should refactor to instead use `ByteAmount`. I will send PR for this.

PR #1728, opened 2017-02-20
upgrade yarn scheduler driver memory to ByteAmount

mycFelix:
As #1727 mentioned and #1707 refactored, this PR upgraded driver memory to use ByteAmount.

* add getByteAmountValueMB() in Config class
* change `driverMemory` from int to ByteAmount
* refactor YarnKey class

cc: @billonahill @ashvina

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/basics/TypeUtils.java
+++ b/heron/common/src/java/com/twitter/heron/common/basics/TypeUtils.java
@@ -87,6 +87,17 @@ public final class TypeUtils {
     }
   }
 
+  public static ByteAmount getByteAmountMB(Object o) {
+    if (o != null && o instanceof ByteAmount) {
+      return (ByteAmount) o;
+    }
+    try {
+      return ByteAmount.fromMegabytes(getLong(o));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Don't know how to convert " + o + " to ByteAmount", e);
+    }
+  }
+
   public static Boolean getBoolean(Object o) {
     if (o instanceof Boolean) {
       return (Boolean) o;
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnContext.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnContext.java
@@ -14,6 +14,8 @@
 
 package com.twitter.heron.scheduler.yarn;
 
+import com.twitter.heron.common.basics.ByteAmount;
+import com.twitter.heron.common.basics.TypeUtils;
 import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.Context;
 
@@ -27,8 +29,8 @@ public final class YarnContext extends Context {
         YarnKey.HERON_SCHEDULER_YARN_QUEUE.getDefaultString());
   }
 
-  public static int heronDriverMemoryMb(Config cfg) {
-    return cfg.getIntegerValue(YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.value(),
-        YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.getDefaultInt());
+  public static ByteAmount heronDriverMemoryMb(Config cfg) {
+    return cfg.getByteAmountValueMB(YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.value(),
+        (ByteAmount) YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.getDefault());
   }
 }
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnLauncher.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnLauncher.java
@@ -34,6 +34,7 @@ import org.apache.reef.tang.Tang;
 import org.apache.reef.tang.annotations.Unit;
 import org.apache.reef.tang.exceptions.InjectionException;
 
+import com.twitter.heron.common.basics.ByteAmount;
 import com.twitter.heron.scheduler.yarn.HeronMasterDriver.ContainerAllocationHandler;
 import com.twitter.heron.scheduler.yarn.HeronMasterDriver.FailedContainerHandler;
 import com.twitter.heron.scheduler.yarn.HeronMasterDriver.HeronSchedulerLauncher;
@@ -63,7 +64,7 @@ public class YarnLauncher implements ILauncher {
   private String role;
   private String env;
   private String queue;
-  private int driverMemory;
+  private ByteAmount driverMemory;
   private ArrayList<String> libJars = new ArrayList<>();
 
   @Override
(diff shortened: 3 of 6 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnContext.java
```diff
@@ -27,8 +28,8 @@ public static String heronYarnQueue(Config cfg) {
         YarnKey.HERON_SCHEDULER_YARN_QUEUE.getDefaultString());
   }
 
-  public static int heronDriverMemoryMb(Config cfg) {
-    return cfg.getIntegerValue(YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.value(),
-        YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.getDefaultInt());
+  public static ByteAmount heronDriverMemoryMb(Config cfg) {
```
mycFelix: Agree

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnContext.java
```diff
@@ -27,8 +28,8 @@ public static String heronYarnQueue(Config cfg) {
...
+    return cfg.getByteAmountValueMB(YarnKey.YARN_SCHEDULER_DRIVER_MEMORY_MB.value(),
```
mycFelix: will do

heron/common/src/java/com/twitter/heron/common/basics/TypeUtils.java
```diff
@@ -87,6 +87,17 @@ public static ByteAmount getByteAmount(Object o) {
     }
   }
 
+  public static ByteAmount getByteAmountMB(Object o) {
```
mycFelix: Good point~

mycFelix: @billonahill - It seems we don't have `ByteAmountUnit` class currently. Do you think it is a good idea if we add a new class in `c.t.h.common.basics` in this PR?

objmagic: @mycFelix Bill is on vacation so he may not have spare time to reply you. We can definitely add `ByteAmountUnit` as a enum class to keep it simple. Another approach is to make [`2048`](https://github.com/twitter/heron/pull/1728/files#diff-7a5b4ac4a4ebcf917aaa8c34416e02f5R27) into [`2048 * 1024`]. We sacrifice some clarity here but can avoid extending current APIs. Thoughts? @ashvina

objmagic: @mycFelix actually, now I think @ashvina's suggestion is better. It would be nice to add a enum class `ByteAmountUnit` in the same style of [`SystemConfigKey`](https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/config/SystemConfigKey.java) to expose [`MB`](https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java#L22) and [`GB`](https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java#L23) in a typed manner.

mycFelix: @objmagic - Understood. Thank you for your suggestion.

mycFelix:
@ashvina @objmagic I've addressed all comments. Would you please take a look again?

objmagic:
@mycFelix Looks good to me. I think putting `ByteAmountUnit` inside of `ByteAmount` might be better, but it's totally up to you. We can merge it if @ashvina gives a ship-it.

mycFelix:
@objmagic - Thanks for reviewing. I'm agree with you. At first, I think putting `ByteAmountUnit` inside of `ByteAmount` is a better approach and using enum to instead static `MB` and `GB` inside of `ByteAmount` seems nice too. But It might trigger a large refactor in `ByteAmount`. So, I decided putting `ByteAmountUnit` outside of `ByteAmount` temporarily.

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/YarnLauncher.java
```diff
@@ -161,7 +162,7 @@ Configuration getHMDriverConf() {
         .set(HeronDriverConfiguration.HTTP_PORT, 0)
         .set(HeronDriverConfiguration.VERBOSE, false)
         .set(YarnDriverConfiguration.QUEUE, queue)
-        .set(DriverConfiguration.DRIVER_MEMORY, driverMemory)
+        .set(DriverConfiguration.DRIVER_MEMORY, driverMemory.asMegabytes())
```
objmagic: I'm confused here. Shouldn't `DRIVER_MEMORY` accept value of type int, according to [reef](https://github.com/apache/reef/blob/master/lang/java/reef-common/src/main/java/org/apache/reef/client/DriverConfiguration.java#L59)?

mycFelix: To be honest, it make me confuse too. I didn't notice that...

mycFelix: @objmagic - cast it to int

objmagic: I don't think you can just cast here...?

objmagic: @mycFelix I now think I am not the right person to review this PR. You can think about [Bill's review](https://github.com/twitter/heron/pull/1728#pullrequestreview-23143032) and wait for @ashvina's comment.

mycFelix: @objmagic - Sure~

mycFelix:
Considering @billonahill 's comment,  creating `ByteAmountUnit` enum class might be not a  proper way.

@ashvina @billonahill - How about we make [2048](https://github.com/twitter/heron/pull/1728/files#diff-7a5b4ac4a4ebcf917aaa8c34416e02f5R27) into 2048 * 1024 here？

heron/common/src/java/com/twitter/heron/common/basics/TypeUtils.java
```diff
@@ -81,7 +81,28 @@ public static ByteAmount getByteAmount(Object o) {
...
     }
     try {
-      return ByteAmount.fromBytes(getLong(o));
+      return getByteAmount(o, ByteAmountUnit.BYTE);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Don't know how to convert " + o + " to ByteAmount", e);
+    }
+  }
+
+  public static ByteAmount getByteAmount(Object o, ByteAmountUnit unit) {
```
ashvina: I think it would be useful to add javadocs for these overloaded methods.

objmagic:
is there any update on this? seems we have very few things left to do for this PR? @mycFelix @billonahill

PR #1743, opened 2017-03-03
Modifying configs at submit if numContainers is off

### LINKED ISSUE #1742: Submit failure when packing plan numContains different from configured numContainers
`IPacking` algorithms might ignore the configured instance/container resource settings or the configured number of containers. `RoundRobinPacking` ignores the former, `ResourceCompliantRRPacking` ignores the later.

Currently if the packing plan calls for a different number of containers than what was configured, and exception is throw upon submit. Instead we should log a warning and update the configs to match the packing plan. (Even better would be to leave the configs and update the framework to look exclusively at the packing plan, but that's a larger effort.)

billonahill:
Fixes #1742 by updating configs and the topology to match the number of containers specified by the packing plan. The key parts to look at is the `SubmitterMain.updateNumContainersIfNeeded` method.

Also moves this validation into `SubmitterMain` so it can happen before uploading the package. @ashvina I know we refactored to not build the packing plan in both `SubmitterMain` and `LaunchRunner`, but I think it makes sense to do this logic as early as possible, since we're mutating the user-provided topology and configs.

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/builder/PackingPlanBuilder.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/builder/PackingPlanBuilder.java
@@ -124,7 +124,7 @@ public class PackingPlanBuilder {
           instanceId, instanceResource, containerId), e);
     }
 
-    LOG.fine(String.format("Added to container %d instance %s", containerId, instanceId));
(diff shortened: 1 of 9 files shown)


objmagic:
NPE in CI

PR #1751, opened 2017-03-07
Use PLOG() to interpret errno

### LINKED ISSUE #1732: Improve error message in stmgr log
`E0222 21:45:36.585096 44403 packet.cpp:168] Something really bad happened while reading 104
`

The errno 104 is actually ECONNRESET on Linux.

congwang:
This resolves #1732.

### DIFF
--- a/heron/common/src/cpp/network/packet.cpp
+++ b/heron/common/src/cpp/network/packet.cpp
@@ -22,7 +22,6 @@
 #include <arpa/inet.h>
 #include <google/protobuf/message.h>
 #include <string>
-#include <cstring>
 #include "glog/logging.h"
 
 // PacketHeader static methods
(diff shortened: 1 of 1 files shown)


heron/common/src/cpp/network/packet.cpp
```diff
@@ -165,7 +166,7 @@ sp_int32 IncomingPacket::InternalRead(sp_int32 _fd, char* _buffer, sp_uint32 _si
       } else {
         // something really bad happened. Bail out
         // try again
-        LOG(ERROR) << "Something really bad happened while reading " << errno << "\n";
```
kramasamy: Instead of LOG(ERROR), use PLOG(ERROR) and you won't need strerror(errno)

congwang: Good point. Never know this API before...

kramasamy:
👍

PR #1752, opened 2017-03-07
Fix flaky ShellUtils test --- use `waitFor` and increase timeout

### LINKED ISSUE #1269: Possible flaky test: ShellUtilsTest
`ShellUtilsTest` under `heron/spi/tests/java` is possibly flaky.

objmagic:
resolves #1269

### DIFF
--- a/heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
+++ b/heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
@@ -23,7 +23,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


heron/spi/tests/java/com/twitter/heron/spi/utils/ShellUtilsTest.java
```diff
@@ -125,7 +117,11 @@ public void testGetProcessBuilder() throws IOException {
 
     // Process running normally
     Process p = pb.start();
-    wait(10, TimeUnit.MILLISECONDS);
+    try {
+      p.waitFor(2, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
```
billonahill: It would be better to| declare this in the throws clause and not handle it, since it should trigger a test failure.

PR #1755, opened 2017-03-08
update doc

leonardgithub:
update doc


website/content/docs/concepts/design-goals.md
```diff
@@ -27,7 +27,7 @@ between systems.
 In a distributed system like Heron, there are no guarantees that all system
 components will execute at the same speed. Heron has built-in [back pressure
 mechanisms]({{< relref "architecture.md#stream-manager" >}}) to ensure that
-topologies can self-adjust in case components lag.
+topologies can self-adjust in case of components lag.
```
billonahill: adding 'of' would mean we should make components singular. Instead though, we can change this to "can self-adjust in the event that a component lags."

leonardgithub:
After a deep thought, I think no need to change. PR closed now

PR #1756, opened 2017-03-08
Read back pressure configs from config file

### LINKED ISSUE #1750: Backpressure configs are hard coded
The [documentation](http://twitter.github.io/heron/docs/operators/configuration/stmgr/) for the stream manage mentions the ability to configure specific different parameters related to the backpressure algorithm however it appears that the actual values are hard coded (see https://github.com/twitter/heron/blob/master/heron/common/src/cpp/network/connection.cpp#L27-L38).

It looks like the [methods](https://github.com/twitter/heron/blob/master/heron/common/src/cpp/network/connection.cpp#L27-L38) for accessing the config values exist but they are not being used. Heron should use the supplied values instead of having them hard coded.

congwang:
heron.streammgr.network.backpressure.highwatermark.mb and heron.streammgr.network.backpressure.lowwatermark.mb are already available in heron_internals.yaml, but we never actually read them from it. This PR makes stream manager reading these configs from heron_internals.yaml. Note, other non-stmgr connections still use the default values in the original code.

This resolves #1750.

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
@@ -37,13 +37,16 @@ const sp_string METRIC_STMGR_NEW_CONNECTIONS = "__stmgr_new_connections";
 StMgrClientMgr::StMgrClientMgr(EventLoop* eventLoop, const sp_string& _topology_name,
                                const sp_string& _topology_id, const sp_string& _stmgr_id,
                                StMgr* _stream_manager,
-                               heron::common::MetricsMgrSt* _metrics_manager_client)
+                               heron::common::MetricsMgrSt* _metrics_manager_client,
+                               sp_int64 _high_watermark, sp_int64 _low_watermark)
     : topology_name_(_topology_name),
       topology_id_(_topology_id),
       stmgr_id_(_stmgr_id),
       eventLoop_(eventLoop),
       stream_manager_(_stream_manager),
-      metrics_manager_client_(_metrics_manager_client) {
+      metrics_manager_client_(_metrics_manager_client),
+      high_watermark_(_high_watermark),
+      low_watermark_(_low_watermark) {
   stmgr_clientmgr_metrics_ = new heron::common::MultiCountMetric();
   metrics_manager_client_->register_metric("__clientmgr", stmgr_clientmgr_metrics_);
 }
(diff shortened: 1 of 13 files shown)


heron/stmgr/src/cpp/server/stmgr-main.cpp
```diff
@@ -70,8 +70,15 @@ int main(int argc, char* argv[]) {
     LOG(FATAL) << "Corrupt topology defn file" << std::endl;
   }
 
+  sp_int64 high_watermark = heron::config::HeronInternalsConfigReader::Instance()
+                              ->GetHeronStreammgrNetworkBackpressureHighwatermarkMb() *
+                                1024 * 1024;
```
kramasamy: Instead of multiplying by 1024* 1024 - every time, can you define a constant in one of the include file called 

const int MB = 1024 * 1024;

somewhere in the include file

heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
```diff
@@ -1445,11 +1440,6 @@ TEST(StMgr, test_tmaster_restart_on_same_address) {
 
   int num_msgs_sent_by_spout_instance = 100 * 1000 * 1000;  // 100M
```
kramasamy: This is not 100 MB it should be 100 * 1024 * 1024

congwang: This is not my code, it shows up as diff context... Also, the 100M there might mean 100 million rather than 100 megabytes.

heron/stmgr/src/cpp/server/stmgr-main.cpp
congwang: Yeah, we can do it in a separate PR, because there are multiple places using the same.

heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
kramasamy: got it. my bad.

heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
```diff
@@ -114,6 +117,8 @@ StMgrClient* StMgrClientMgr::CreateClient(const sp_string& _other_stmgr_id,
   options.set_max_packet_size(config::HeronInternalsConfigReader::Instance()
                                   ->GetHeronStreammgrNetworkOptionsMaximumPacketMb() *
                               1024 * 1024);
+  options.set_high_watermark(high_watermark_);
```
maosongfu: Should we pass high_watermark and low_watermark from constructor, or similar to max_packet_size, reading config inside CreateClient(...) method?

On the one hand, it is not necessary to change the signature of a class's constructor. On the other hand, It will match the style of max_packet_size.

congwang: We have to, because the stmgr unit test cases don't want to read from config, they set these values by themselves.

maosongfu: Just wondering: how does the "max_packet_size" config work?

congwang: 1) StMgr() doesn't accept NetworkOptions as parameter, max_packet_size is passed in NetworkOptions for those who accept it.

2) stmgr unit tests want 2 different values in different tests, so you have to provide at least 2 different configs if you really want to read from config. Not to mention other config's.

maosongfu: OK Sounds good to me.

heron/stmgr/src/cpp/server/stmgr-main.cpp
kramasamy: can you please open an issue to track this?

congwang: Done: https://github.com/twitter/heron/issues/1758

PR #1761, opened 2017-03-13
Restore the default values for backpressure watermarks

congwang:
Previously before we read config's from heron_internals.yaml, we use 100M and 50M as the hard-coded values for  backpressure watermarks, but in heron_internals.yaml, the default values are 50M and 30M, we silently change the default values after we are able to read the config's. Restore the previous default values.


billonahill:
👍

PR #1762, opened 2017-03-13
Define suffix for KB, MB, s, ms

### LINKED ISSUE #1758: Unify the constants for KB, MB in stmgr code
For example,

`sp_int64 Connection::systemHWMOutstandingBytes = 1024 * 1024 * 100;  // 100M`

Fortunately, C++11 allows us to define our own literals:
http://en.cppreference.com/w/cpp/language/user_literal

congwang:
Since C++11, we are allowed to define our own suffix for constants like KB, MB etc., use this new feature to replace hard-coded "1024 * 1024".

This resolves #1758.

### DIFF
--- a/heron/common/src/cpp/basics/spconsts.h
+++ b/heron/common/src/cpp/basics/spconsts.h
@@ -23,6 +23,8 @@
 #if !defined(HERON_CONSTS_H_)
 #define HERON_CONSTS_H_
 
+#include "basics/sptypes.h"
+
 // constants related to logs
 constexpr auto constMaxNumLogFiles = 5;
 constexpr auto constLogsDirectory = "log-files";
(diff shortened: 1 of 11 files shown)


heron/common/src/cpp/basics/spconsts.h
```diff
@@ -38,4 +40,20 @@ constexpr auto constTcpKeepAliveSecs = 300;          // 5 min
 constexpr auto constTcpKeepAliveProbeInterval = 10;  // 10 seconds
 constexpr auto constTcpKeepAliveProbes = 5;
 
+constexpr sp_int64 operator"" _KB(unsigned long long bytes) { // NOLINT
+    return (sp_int64)bytes * 1024;
+}
+
+constexpr sp_int64 operator"" _MB(unsigned long long bytes) { // NOLINT
```
objmagic: If I understand correctly, using `amount` for argument name instead of `bytes` makes better sense? (`100_MB` means _one hundreds_ megabytes. )

congwang: Yeah, or change it to 'megabytes'.

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -313,7 +313,7 @@ void TMaster::GetPhysicalPlanDone(proto::system::PhysicalPlan* _pplan,
   stats_options.set_port(stats_port_);
   stats_options.set_max_packet_size(config::HeronInternalsConfigReader::Instance()
                                         ->GetHeronTmasterNetworkStatsOptionsMaximumPacketMb() *
-                                    1024 * 1024);
+                                    1_MB);
```
huijunw: 'xxxxMb() * y_MB' looks like 'byte * byte = byte^2'

congwang: If you want to interpret literally, yeah.

PR #1765, opened 2017-03-14
Use std::to_string() instead of std::ostringstream

congwang:
Since C++11 std::to_string() is available to convert numbers to strings.

### DIFF
--- a/heron/common/src/cpp/metrics/time-spent-metric.cpp
+++ b/heron/common/src/cpp/metrics/time-spent-metric.cpp
@@ -21,7 +21,7 @@
 // Please see time-spent-metric.h for details
 //////////////////////////////////////////////////////////////////////////////
 #include "metrics/time-spent-metric.h"
-#include <sstream>
+#include <string>
 #include "basics/basics.h"
 #include "errors/errors.h"
 #include "metrics/imetric.h"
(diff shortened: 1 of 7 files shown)


heron/common/src/cpp/metrics/time-spent-metric.cpp
```diff
@@ -65,12 +65,11 @@ void TimeSpentMetric::GetAndReset(const sp_string& _prefix,
...
     start_time_ = now;
   }
-  std::ostringstream o;
-  o << total_time_msecs_;
+  std::string val = std::to_string(total_time_msecs_);
   total_time_msecs_ = 0;
   proto::system::MetricDatum* d = _message->add_metrics();
   d->set_name(_prefix);
-  d->set_value(o.str());
+  d->set_value(val);
```
huijunw: ? d->set_value(std::to_string(total_time_msecs_));

huijunw:
what c++ version does Heron require?

objmagic:
@huijunw we support C++11

heron/common/src/cpp/metrics/time-spent-metric.cpp
congwang: We could do that but need to move 'total_time_msecs_ = 0' down.

PR #1768, opened 2017-03-20
Use variadic templates and std::bind to rewrite callback

congwang:
C++11 supports variadic templates, we don't have to play with the old template tricks for arbitrary  number of arguments any more. The only thing needs to note is that template parameter pack must be the last so we have to reorder the arguments for callback1.


heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -302,7 +302,7 @@ void EventLoopImpl::handleInstantCallback(Status) {
   mListInstantCallbacks.erase(mListInstantCallbacks.begin(), ++enditr);
 }
 
-void EventLoopImpl::handleReadCallback(sp_int32 fd, sp_int16 event) {
+void EventLoopImpl::handleReadCallback(sp_int16 event, sp_int32 fd) {
```
kramasamy: Why the need to switch args for this one and several subsequent functions?

heron/common/src/cpp/zookeeper/zkclient.cpp
```diff
@@ -38,7 +38,7 @@ struct ZKClientGetChildrenStructure {
   CallBack1<sp_int32>* cb_;
 };
 
-void RunUserCb(VCallback<sp_int32> cb, sp_int32 rc) { cb(rc); }
+void RunUserCb(sp_int32 rc, VCallback<sp_int32> cb) { cb(rc); }
```
kramasamy: Again switching of arguments?

kramasamy:
@congwang - good PR. Cleans a lot of code. Just gave a few comments

heron/common/src/cpp/network/event_loop_impl.cpp
congwang: Please read the description of this PR: "The only thing needs to note is that template parameter pack must be the last so we have to reorder the arguments for callback1."

kramasamy: Got it!

PR #1770, opened 2017-03-20
[do not merge] Auto restart the container that has back pressure

huijunw:
1. Motivation:
One of the most common alerts we get is back pressure, and one of the most typical resolutions is to restart the containers that have trouble and wait for the topology to come back.
This feature will automate the above 'detecting back pressure and restarting container' procedures.

2. Life cycle:
This feature is a temporary solution before the comprehensive SLA/HealthMgr is available. This feature heavily depend on TMetricsCollector and should retire along with TMetricsCollector.

3. User interface:
The 'heron.config.auto_heal_window' and 'heron.config.auto_heal_interval' config-propertys are required for 'submit' command:
```
heron submit ... [--config-property heron.config.auto_heal_window=<back pressure time window, integer, in minutes>] [--config-property heron.config.auto_heal_interval=<min restart interval, integer, in minutes>] ...
```
For example:
```
~/bin/heron submit --config-property heron.config.auto_heal_window=10 --config-property heron.config.auto_heal_interval=20 local ~/.heron/examples/heron-examples.jar com.twitter.heron.examples.ExclamationTopology ExclamationTopology --verbose
```
**heron.config.auto_heal_window**
- The feature is disabled, when:
a. the 'heron.config.auto_heal_window' config is missing, or
b. the `<back pressure time window>` is less than or equals to zero.
- if the `<back pressure time window>` is greater than zero, the back pressure metrics are monitored. If any container has been with back pressure for more than `<back pressure time window>` period, the container will be restarted.

**heron.config.auto_heal_interval**
- 'heron.config.auto_heal_interval' is to address the uncontrolled data loss, so that it is valid only if 'heron.config.auto_heal_window' is enabled.
- 'heron.config.auto_heal_interval' is 20 minutes by default, if it is missing.
- The elapsed time between any container restarting is at least `<min restart interval>`.

4. Solution description:
a. TMaster monitors the back pressure metrics.
b. If there has been the back pressure in the container X for more than `<back pressure time window>` minutes, TMaster sends a command to the heron-shell in the container X.
c. The heron-shell kills the heron-executor and the Aurora will re-schedule the container X.


congwang:
@huijunw The motivation is not clear. Our current back pressure algorithm inside stmgr is **kinda** auto-healing too: it throttles the spouts in hope to get recovered from back pressure situation (and unthrottles them after recover). I **guess** you are trying to handle some unrecoverable situation like network partition? But I think in this situation Aurora (or other scheduler) should be able to detect and do its own auto-restart. So, why do we **have** to do it inside Heron?

Also, it needs to mention that our back pressure algorithm is improved a lot: https://github.com/twitter/heron/commits/cwang/backpressure, it should do better than before.

huijunw:
@congwang actually, the motivation is a jira ticket assigned to me month ago ..
Ideally, the backpressure is supposed to auto-heal

maosongfu:
Hi, Did we verify whether it works in local scheduler and aurora scheduler?
One simple way is to make bolt.execute() slow (sleep a few seconds every time processing a tuple). Then the bolt will trigger back-pressure soon.

billonahill:
Note that I commented on the twitter jira about taking steps to assert that the symptom still occurs when the effected topologies are running on OSS Heron, before deciding to ship this. Let's discuss that there.

maosongfu:
The motivation of this pull request is to handle some unrecoverable situations. There are a lot of in-application unrecoverable situations, which is transparent to aurora so aurora has no ways to handle them. For instance, slow hosts (symptoms can be high execute-latency; building up lag; unrecoverable back-pressure due to the whole topology consumes slower than the data incoming rate; and so on.)

It is expected that we could detect all cases mentioned above but it is good to start from a concrete case. We discussed with a lot of Twitter internal consumers and considered "an instance causing back-pressure for a long period of time without any indication to recover" is a good case to start. 

@congwang The back-pressure algorithm did improve a lot. But in most cases the back-pressure is caused by slow hosts; we can't do much without moving the containers.

congwang:
@maosongfu Yeah, there are certainly many unrecoverable cases, but I am still wondering how restart could help in such situations. Auto-restart sounds like saying "we are not lucky this time, let's try it again", it can never guarantee we could totally solve the problem. IOW, this does not look like a real solution, it looks more like a workaround. For me, I think the users should think about how to adjust their logic in such situations rather than just restarting, which means for Heron we probably need to provide some API(s) to users in back pressure situation and let them to decide what to do in their logic.

maosongfu:
@congwang We want to provide a mechanism to restart containers in unrecoverable bad situations. And it is pluggable what situations to track and how to determine whether a situation is recoverable or not. 

Here back-pressure is a case to start with since:
1. it can benefit customers (customers want it to reduce ongoing operation cost)
2. customers can use it in practice so it is easier to verify whether this restarting mechanism does help

congwang:
@maosongfu We all want to help customers to solve their problem, no doubt, but auto-restart is not solving it. They want auto-**heal**, you give auto-**restart**, auto-restart can't guarantee a healing.

avflor:
I agree with @congwang that restarting the container does not necessarily mean that backpressure will disappear. However, restart can be helpful in some cases (e.g., when backpressure is caused by a slow container/machine).  In the context of Healthmanager, restarting a container is one resolution mechanism that we have but there are others as well depending on the cause of backpressure.

kramasamy:
@congwang @maosongfu - we need some reliable information the container is indeed in some kind irrecoverable situation before we can start the container. If we restart the container without not much evidence, it could lead to inaccurate results.  If @congwang feels that the new back pressure algorithm has improved, it might be worth a try before adding autorestart. One fear is that if this code (which is kind of a hacky solution), it might not come out later.

kramasamy:
@avflor - wondering what is the ETA for the SLA/Health Manager so that we can do clean integration?

avflor:
@kramasamy We will submit  a PR with the health manager APIs tomorrow for review. Once we agree on the APIs, we will start checking in the code for the policies.

pankajroark:
Sounds like there is no limit on how many containers are killed in a period of time, this can result in uncontrolled data loss. Is this a stop-gap implementation? Will this be replaced/enhanced soon after?

huijunwu:
@pankajroark make sense. will do

congwang:
@avflor It helps only because you are luck enough after a restart, no other reasons. We don't want our solution to depend on luck, do we?

congwang:
@kramasamy You do need a lot more information to decide where is a good place to reschedule the container if you still prefer this way, and I doubt you could obtain enough information from Aurora, instead of relying on the container scheduler, why not do it by ourselves? IOW, why not consider adding more resources for slow containers to speed them up? This is a good direction, we already have scale up/down in our code (thanks for Bill).

My back pressure could improve the current algorithm, but can't solve all of the situations, for example if the resources are naturally the bottleneck back pressure will consistently happen no matter how we improve the algorithm (unless we drop).

objmagic:
My two cents:
> One of the most common alerts we get is back pressure, and one of the most typical resolutions is to restart the containers that have trouble and wait for the topology to come back.

Based on my on-call experience, this is not always true. I'd say around half of the time restarting does not resolve back-pressure. 

To make this PR more convincing, we should at least have
1.  detailed analysis of what restart **can** really solve (e.g. bad host), and what restart **cannot** solve (e.g. long de/serialization time). 
2. how often does this each scenario happen? some statistics. IMHO **at least people who work on this PR should all have decent enough on-call experience**
3. measure if auto-restart can help users in most of the scenarios. 

Also I'm wondering if heron will fall into auto-restart loop hell if auto-restart does not solve the problem. Do we have any measure to prevent this from happening?

objmagic:
Most importantly, as far as I know, no one has tested this auto-restart feature in Twitter internal environment, not even with long enough time. 

I cannot believe anyone here wants to merge this PR without testing it and measure its performance **thoroughly in Twitter internal production environment.**

That's crazy, man.

maosongfu:
@objmagic 
There is an internal ticket with action items (including test plans) and ETA. 
Simply saying, this pull request will not be merged unless:
1. It is verified working in Twitter distributed env
2. It is verified by Twitter internal customers (Prasad) that it does provide values
3. Oss heron is in production inside Twitter.

huijunwu:
beg for review

kramasamy:
Not convinced this is the right approach for dealing with the back pressure - container. Did we collect data points about any other issues that leads to it?

maosongfu:
@kramasamy 
Definitely we need to figure out the root cause for a back-pressure.  But this pull request does not prevent that and it can benefit topology developers:
1. It provides a mechanism to trigger handlers when encountering abnormal cases. We investigated and discussed with a lot of customers at Twitter. They mentioned the most common abnormal case is back-pressure and the it is always caused by slow hosts - so they recommended the handlers  could be "restarting that container on another host". Notice, it does not prevent other handlers; the reason we provide this handler is basing on customers' feedback.
2. This feature is not enabled by default: only when people provide specific flags explicitly and manually can they use this feature. So by default, it will not hurt. Furthermore, restarting a container does not prevent further investigation: we have logs; we have metrics. In future, we could also add something like "heapdump", "snapshot" before restarting to help debug. Besides those, keeping an instance running will not help debug too much, since we have limited interactive runtime debugging solutions.
3. It will reduce the down-time of a topology, consider at Twitter env, most topologies are actually in "down-time" if they can not recover in certain mins (that's also what this pull request designs for). It will also reduce the maintenance cost. Customers want this feature very much.

huijunw:
@kramasamy pls see the above comment

kramasamy:
@huijunw - for a stop gap implementation, this PR seems to touching a lot of code. It is going to be hard to take it out later. Also, I would recommend to add a configuration - which enable this feature. by default that value should be turned off.

kramasamy:
@huijunw - instead of adding parameters in heron submit, can you use the 

--config-property option for these values

This is advantageous
* already the mechanism is in place to pass the config values
* no need to introduce a specific feature like command line args
* you can ship with internally with specific values to enable this feature.
* it will reduce a lot of code of passing these parameters around

kramasamy:
@huijunw - do we have to implement this logic in Tmaster? Why can't we do a separate process - this allows for a clean migration to Dhalion as well - when it becomes available.

srkukarni:
This pr breaks so many abstractions all across the board(too many to list). Please _do not_ merge it. Thanks!

heron/executor/src/python/heron_executor.py
```diff
@@ -55,7 +55,9 @@ def print_usage():
       " <cluster> <role> <environ> <instance_classpath> <metrics_sinks_config_file>"
       " <scheduler_classpath> <scheduler_port> <python_instance_binary>"
       " <metricscachemgr_classpath> <metricscachemgr_masterport> <metricscachemgr_statsport>"
-      " <is_stateful> <ckptmgr_classpath> <ckptmgr_port> <stateful_config_file>")
+      " <is_stateful> <ckptmgr_classpath> <ckptmgr_port> <stateful_config_file>"
+      " <auto_restart_backpressure_sandbox_time_window>"
```
billonahill: Why is 'sandbox' in the new arg names? Can we remove that?

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
```diff
@@ -153,7 +153,8 @@ public void startExecutor() {
         topologyJar,
         topologyDefFile,
         verboseMode,
-        topology);
+        topology,
+        "0", "20");
```
billonahill: why are these being hard coded here and why use strings to represent integers?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerConfigUtils.java
```diff
@@ -58,12 +58,15 @@ private static Config topologyConfigs(String topologyBinaryFile,
    * @return config, the command line config
    */
   private static Config commandLineConfigs(String cluster, String role,
-                                           String environ, Boolean verbose) {
+                                           String environ, Boolean verbose,
+                                           String autoHeal, String autoHealInterval) {
```
billonahill: What are the acceptable values for autoHeal and autoHealInterval and can these be Duration?

heron/tmaster/src/cpp/manager/tmetrics-collector.cpp
```diff
@@ -74,6 +91,64 @@ void TMetricsCollector::Purge(EventLoop::Status) {
   auto cb = [this](EventLoop::Status status) { this->Purge(status); };
 
   CHECK_GT(eventLoop_->registerTimer(std::move(cb), false, interval_ * 1000000), 0);
+
+  // auto restart backpressure container feature: share the same timer with Purge
+  if (auto_restart_window_ > 0) {
```
billonahill: please add unit tests for the new logic in this class

heron/executor/src/python/heron_executor.py
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/yarn/HeronExecutorTask.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerConfigUtils.java
huijunw: updated

huijunw:
updated to '--config-property'

heron/config/src/yaml/conf/test/test_override.yaml
```diff
@@ -0,0 +1,2 @@
+heron.config.auto_heal_window: 10
```
kramasamy: Please move this to heron_internals.yaml.

```
# To autorestart containers, if they go into unrecoverable back pressure

# Time duration the container should be in back pressure
heron.container.backpressure.duration.min: 10

# Time interval between contiguous restarts
heron.container.backpressure.interval.min: 20
```

heron/executor/src/python/heron_executor.py
```diff
@@ -224,6 +224,10 @@ def __init__(self, args, shell_env):
 
     # Read the heron_internals.yaml for logging dir
     self.log_dir = self._load_logging_dir(self.heron_internals_config_file)
+    [self.auto_restart_backpressure_container_time_window, \
+     self.auto_restart_backpressure_container_min_interval] = \
+        self._load_auto_restart_backpressure_container_config( \
+            self.heron_internals_config_file.replace("heron_internals.yaml", "override.yaml"))
```
kramasamy: This block of code is not needed.

heron/executor/src/python/heron_executor.py
```diff
@@ -335,6 +339,15 @@ def _load_logging_dir(self, heron_internals_config_file):
...
 
+  def _load_auto_restart_backpressure_container_config(self, override_config_file):
+    try:
+      with open(override_config_file, 'r') as stream:
+        override_config = yaml.load(stream)
+      return (override_config['heron.config.auto_heal_window'], \
+              override_config['heron.config.auto_heal_interval'])
+    except:
+      return (0, 0)
+
```
kramasamy: This is not needed.

heron/executor/src/python/heron_executor.py
```diff
@@ -429,7 +442,9 @@ def _get_tmaster_processes(self):
         ','.join(self.stmgr_ids.values()),
         self.heron_internals_config_file,
         self.metrics_sinks_config_file,
-        self.metricsmgr_port]
+        self.metricsmgr_port,
+        str(self.auto_restart_backpressure_container_time_window),
+        str(self.auto_restart_backpressure_container_min_interval)]
```
kramasamy: Not needed.

heron/executor/src/python/heron_executor.py
```diff
@@ -668,6 +683,9 @@ def _get_heron_support_processes(self):
         '--port=%s' % self.shell_port,
         '--log_file_prefix=%s/heron-shell.log' % self.log_dir]
 
+    if self.auto_restart_backpressure_container_time_window > 0:
+      retval[self.heron_shell_ids[self.shard]].append('--secret=%s' % self.topology_id)
+
```
kramasamy: If clause is not needed since the secret could be enabled by default in all cases.

heron/executor/tests/python/heron_executor_unittest.py
```diff
@@ -134,7 +135,7 @@ def get_expected_instance_command(component_name, instance_id, container_id):
                   'tmaster_binary %s master_port '
                   'tmaster_controller_port tmaster_stats_port '
                   'topname topid zknode zkroot stmgr-1,stmgr-7 '
-                  '%s metrics_sinks_config_file metricsmgr_port' % (HOSTNAME, INTERNAL_CONF_PATH )),
+                  '%s metrics_sinks_config_file metricsmgr_port 10 20' % (HOSTNAME, INTERNAL_CONF_PATH )),
```
kramasamy: this is not needed.

heron/shell/src/python/main.py
```diff
@@ -34,15 +34,18 @@
     (r"^/filedata/(.*)", handlers.FileDataHandler),
     (r"^/filestats/(.*)", handlers.FileStatsHandler),
     (r"^/download/(.*)", handlers.DownloadHandler),
+    (r"^/killexecutor", handlers.KillExecutorHandler),
 ])
 
 
 if __name__ == '__main__':
   define("port", default=9999, help="Runs on the given port", type=int)
+  define("secret", default='', help="Shared secret with TMaster", type=str)
```
kramasamy: Shared secret with HeathManager

heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
```diff
@@ -190,7 +190,8 @@ void StartTMaster(EventLoopImpl*& ss, heron::tmaster::TMaster*& tmaster,
   tmaster =
       new heron::tmaster::TMaster(zkhostportlist, topology_name, topology_id, dpath, stmgrs_id_list,
                                   tmaster_controller_port, tmaster_port, tmaster_stats_port,
-                                  metrics_mgr_port, metrics_sinks_config_filename, LOCALHOST, ss);
+                                  metrics_mgr_port, metrics_sinks_config_filename, LOCALHOST, ss,
+                                  0, 20);
```
kramasamy: 0, 20 is not needed.

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -55,7 +55,8 @@ TMaster::TMaster(const std::string& _zk_hostport, const std::string& _topology_n
                  const std::vector<std::string>& _stmgrs, sp_int32 _controller_port,
                  sp_int32 _master_port, sp_int32 _stats_port, sp_int32 metricsMgrPort,
                  const std::string& _metrics_sinks_yaml, const std::string& _myhost_name,
-                 EventLoop* eventLoop) {
+                 EventLoop* eventLoop, sp_int32 auto_restart_window,
+                 sp_int32 auto_restart_interval) {
```
kramasamy: No change required.

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -69,9 +70,9 @@ TMaster::TMaster(const std::string& _zk_hostport, const std::string& _topology_n
   eventLoop_ = eventLoop;
   metrics_collector_ =
       new TMetricsCollector(config::HeronInternalsConfigReader::Instance()
-                                    ->GetHeronTmasterMetricsCollectorMaximumIntervalMin() *
-                                60,
-                            eventLoop_, _metrics_sinks_yaml);
+                                    ->GetHeronTmasterMetricsCollectorMaximumIntervalMin() * 60,
+                            eventLoop_, _metrics_sinks_yaml, auto_restart_window,
+                            auto_restart_interval, this);
```
kramasamy: no change required.

heron/tmaster/src/cpp/manager/tmaster.h
```diff
@@ -46,7 +46,8 @@ class TMaster {
           const std::string& _topology_id, const std::string& _topdir,
           const std::vector<std::string>& _stmgrs, sp_int32 _controller_port, sp_int32 _master_port,
           sp_int32 _stats_port, sp_int32 metricsMgrPort, const std::string& metrics_sinks_yaml,
-          const std::string& _myhost_name, EventLoop* eventLoop);
+          const std::string& _myhost_name, EventLoop* eventLoop, sp_int32 auto_restart_window,
+          sp_int32 auto_restart_interval);
```
kramasamy: No change here also.

heron/tmaster/src/cpp/manager/tmetrics-collector.cpp
```diff
@@ -18,7 +18,10 @@
 #include <iostream>
 #include <list>
 #include <map>
+#include <set>
```
kramasamy: No changes required in the file

heron/tmaster/src/cpp/manager/tmetrics-collector.cpp
```diff
@@ -74,6 +91,64 @@ void TMetricsCollector::Purge(EventLoop::Status) {
...
+          }
+        } else {
+          LOG(INFO) << "Auto restart feature: not found in badSandbox " << st.id();
+        }
+      }
+    } else {
+      LOG(INFO) << "skip auto-restart-backpressure checking, last " << auto_restart_last_
+          << " interval " << auto_restart_interval_;
+    }
+  }  // auto restart backpressure container feature: end
```
kramasamy: Logic should move to Health Manager process.

heron/tmaster/src/cpp/manager/tmetrics-collector.cpp
```diff
@@ -83,6 +158,43 @@ void TMetricsCollector::AddMetricsForComponent(const sp_string& component_name,
...
+          << metrics_data.instance_id() << " val: " << val;
+      if (last_timestamp_backpressure_instance.find(metrics_data.instance_id())
+          == last_timestamp_backpressure_instance.end()) {
+        LOG(INFO) << "Auto restart feature (add metrics): instance backpressure recorded "
+            << metrics_data.instance_id() << " timestamp: " << metrics_data.timestamp();
+        last_timestamp_backpressure_instance.insert(
+            std::make_pair(metrics_data.instance_id(), metrics_data.timestamp()));
+      }
+    }
+  }  // auto restart backpressure container feature: end
```
kramasamy: Logic should move to HealthManager process

heron/tmaster/src/cpp/manager/tmetrics-collector.h
```diff
@@ -29,14 +29,18 @@
 
 namespace heron {
 namespace tmaster {
+
```
kramasamy: No changes in this file.

heron/tmaster/src/cpp/server/tmaster-main.cpp
```diff
@@ -26,12 +26,14 @@
 #include "config/heron-internals-config-reader.h"
 
 int main(int argc, char* argv[]) {
-  if (argc != 13) {
+  if (argc != 15) {
```
kramasamy: no changes in this file

heron/tmaster/src/cpp/server/tmaster-main.cpp
```diff
@@ -66,7 +73,8 @@ int main(int argc, char* argv[]) {
 
   heron::tmaster::TMaster tmaster(zkhostportlist, topology_name, topology_id, topdir, stmgrs,
                                   controller_port, master_port, stats_port, metrics_manager_port,
-                                  metrics_sinks_yaml, myhost, &ss);
+                                  metrics_sinks_yaml, myhost, &ss, backpressue_window,
```
kramasamy: No changes in this file as well.

heron/tmaster/tests/cpp/server/tmaster_unittest.cpp
```diff
@@ -194,7 +194,8 @@ void StartTMaster(EventLoopImpl*& ss, heron::tmaster::TMaster*& tmaster,
   tmaster =
       new heron::tmaster::TMaster(zkhostportlist, topology_name, topology_id, dpath, stmgrs_id_list,
                                   tmaster_controller_port, tmaster_port, tmaster_port + 2,
-                                  tmaster_port + 3, metrics_sinks_config_filename, LOCALHOST, ss);
+                                  tmaster_port + 3, metrics_sinks_config_filename, LOCALHOST, ss,
+                                  0, 20);
```
kramasamy: No changes in this file

huijunw:
This feature will not merge. Instead, it will be moved to Dhalion

huijunw:
This PR is moved to #2230

PR #1771, opened 2017-03-20
Use template for PCQueue

congwang:
Make it type-safe.


kramasamy:
👍  - once the CI passes.

PR #1772, opened 2017-03-21
Improve some tmaster logs

### LINKED ISSUE #1740: TMaster log improvement "another stmgr exists at xxx..."
```
I0303 00:47:57.679550  2739 tmaster.cpp:434] Got a register stmgr request from stmgr-43
E0303 00:47:57.679597  2739 tmaster.cpp:442] Another stmgr exists with the same id and it hasn't timed out
```

would be better if the log tells where the old stmgr is, like ``Another stmgr exists AT HOSTNAME with the same id and it hasn't timed out``

congwang:
This resolves #1740.

### DIFF
--- a/heron/tmaster/src/cpp/manager/tmaster.cpp
+++ b/heron/tmaster/src/cpp/manager/tmaster.cpp
@@ -364,7 +364,10 @@ proto::system::Status* TMaster::RegisterStMgr(
     // First check to see if that other guy has timed out
     if (!stmgrs_[stmgr_id]->TimedOut()) {
       // we reject the new guy
-      LOG(ERROR) << "Another stmgr exists with the same id and it hasn't timed out" << std::endl;
+      LOG(ERROR) << "Another stmgr exists at "
+                 << stmgrs_[stmgr_id]->get_connection()->getIPAddress() << ":"
+                 << stmgrs_[stmgr_id]->get_connection()->getPort()
+                 << " with the same id and it hasn't timed out";
       proto::system::Status* status = new proto::system::Status();
       status->set_status(proto::system::DUPLICATE_STRMGR);
       status->set_message("Duplicate StreamManager");
@@ -375,7 +378,10 @@ proto::system::Status* TMaster::RegisterStMgr(
       // for the stmgrs_ list. Which means this case will only happen
       // if the stmgr maintains connection but hasn't sent a heartbeat
       // in a while.
-      LOG(ERROR) << "Another stmgr exists with the same id but it has timed out" << std::endl;
+      LOG(ERROR) << "Another stmgr exists at "
+                 << stmgrs_[stmgr_id]->get_connection()->getIPAddress() << ":"
+                 << stmgrs_[stmgr_id]->get_connection()->getPort()
+                 << " with the same id but it has timed out";
       stmgrs_[stmgr_id]->UpdateWithNewStMgr(_stmgr, _instances, _conn);
       connection_to_stmgr_id_[_conn] = stmgr_id;
     }


heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -364,7 +364,10 @@ proto::system::Status* TMaster::RegisterStMgr(
     // First check to see if that other guy has timed out
     if (!stmgrs_[stmgr_id]->TimedOut()) {
       // we reject the new guy
-      LOG(ERROR) << "Another stmgr exists with the same id and it hasn't timed out" << std::endl;
+      LOG(ERROR) << "Another stmgr exists at "
```
maosongfu: Can we also log down the stmgr-id here? like "Another stmgr {stmgr-id} exists at ...."

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -375,7 +378,10 @@ proto::system::Status* TMaster::RegisterStMgr(
       // for the stmgrs_ list. Which means this case will only happen
       // if the stmgr maintains connection but hasn't sent a heartbeat
       // in a while.
-      LOG(ERROR) << "Another stmgr exists with the same id but it has timed out" << std::endl;
+      LOG(ERROR) << "Another stmgr exists at "
```
maosongfu: Same as the comment above.

heron/tmaster/src/cpp/manager/tmaster.cpp
congwang: Will add it.

congwang: Hmm, actually we already have a log for stmgr-id:

`LOG(INFO) << "Got a register stmgr request from " << stmgr_id << std::endl;`

So I don't think we need to log it again.

objmagic: @congwang agree with @maosongfu. add id so when we grep it we can also see the id

congwang: @objmagic Why not turn on -B for you grep? ;)

objmagic: ok..

maosongfu: @congwang It is better to have a log having all necessary info self-contained, than relying on earlier logs. It is possible that people mistakenly hurt this log without knowing it.

congwang: @maosongfu It is right before this log, I don't know why it is hard for you to find it... At least I don't rely on one-line log for debugging, I read its context too.

maosongfu: Also one mirror comment, (maybe not relate to this pr): when logging "Got a register stmgr request from ", it is better to log "Got a register stmgr request from {stmgr-id} at {host:port}"

maosongfu: @congwang Yeah we read context for debugging. My concern is: people can change/remove the log "Got a register stmgr request from ..." without knowing this change will affect another log.

objmagic: Also, shall we tell everyone there is `grep -B`? What if people don't know? Do you just tell them to RTFM? Just how hard is it to add a stream id? If you do not do it, I will do it in another PR.

objmagic:
LGTM. Feel free to merge.

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -364,7 +364,10 @@ proto::system::Status* TMaster::RegisterStMgr(
     // First check to see if that other guy has timed out
     if (!stmgrs_[stmgr_id]->TimedOut()) {
       // we reject the new guy
-      LOG(ERROR) << "Another stmgr exists with the same id and it hasn't timed out" << std::endl;
+      LOG(ERROR) << "Another stmgr exists at "
+                 << stmgrs_[stmgr_id]->get_connection()->getIPAddress() << ":"
+                 << stmgrs_[stmgr_id]->get_connection()->getPort()
+                 << " with the same id and it hasn't timed out";
```
kramasamy: @congwang - can you print the stmgr id as well?

heron/tmaster/src/cpp/manager/tmaster.cpp
kramasamy: @congwang - here also please print the stmgr id

PR #1774, opened 2017-03-21
Remove sptime.cpp and sptime.h

congwang:
Surprisingly, no one uses sptime. And C++11 already provides std::chrono library, so we can use standard time classes if we need.


huijunw:
standard library +1

maosongfu:
lgtm. 👍

PR #1778, opened 2017-03-25
Improve log in baseclient.cpp

huijunw:
1. errno =>  gai_strerror(error)
2. LOG(ERROR) << ... << errno << std::endl; => PLOG(ERROR)

### DIFF
--- a/heron/common/src/cpp/network/baseclient.cpp
+++ b/heron/common/src/cpp/network/baseclient.cpp
@@ -73,8 +73,8 @@ void BaseClient::Start_Base() {
     struct sockaddr_in t;
     int error = IpUtils::getAddressInfo(t, options_.get_host().c_str(), PF_INET, SOCK_STREAM);
     if (error) {
-      LOG(ERROR) << "getaddrinfo failed in Client " << options_.get_host().c_str()
-          << ": "<< gai_strerror(error) << "\n";
+      LOG(ERROR) << "getaddrinfo failed in Client " << options_.get_host()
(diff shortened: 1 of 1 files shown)


heron/common/src/cpp/network/baseclient.cpp
```diff
@@ -72,7 +73,8 @@ void BaseClient::Start_Base() {
     struct sockaddr_in t;
     int error = IpUtils::getAddressInfo(t, options_.get_host().c_str(), PF_INET, SOCK_STREAM);
     if (error) {
-      LOG(ERROR) << "getaddrinfo failed in Client " << errno << "\n";
+      LOG(ERROR) << "getaddrinfo failed in Client " << options_.get_host().c_str()
```
congwang: No need c_str() here, operator<< works well with string

heron/common/src/cpp/network/baseclient.cpp
```diff
@@ -72,7 +73,8 @@ void BaseClient::Start_Base() {
...
+          << ": "<< gai_strerror(error) << "\n";
```
congwang: No need newline at the end.

heron/common/src/cpp/network/baseclient.cpp
huijunw: done

heron/common/src/cpp/network/baseclient.cpp
huijunw: done

PR #1781, opened 2017-03-26
[DO NOT MERGE] Initial Dhalion API for auto-scaling and tuning

ashvina:
This is initial review request for Dhalion. The aim of this PR is to review Dhalion API and the general direction. This is work in progress.

Dhalion is a toolkit for auto-scaling and auto-tuning a topology dynamically. Highlights:
1.  Dhalion will provide a set of algorithms to monitor a topology (`ISymptomDetector.java`), correlate the observations to identify the problem (`IDiagnoser.java`) and potentially resolve the issue (`IResolver.java`).
1. The algorithms above will become the building blocks for a policy, `IHealthPolicy.java`. An admin would compose a health policy using one or more algorithm implementations and provide configuration to customize the behavior of the policy. 
1. A `HealthManager` will periodically execute the health policies to keep a running topology healthy. The `HealthManager` process could manage the topology using one or more policies. 
1. Dhalion depends on metrics provided by `heron-tracker` or `metrics-cache` and employs `heron update` feature to manage the topology configuration.
1. Some of the algorithms are generic and could be integrated with other streaming systems, for e.g. Storm. With this goal, we plan to keep the code in a separate package and minimize dependency on Heron.

Code structure
1. The core api's are part of `org/apache/dhalion/api/` package.
1. Implementation of the core api specific to Heron are in `org/apache/dhalion/heron` package.

### DIFF
--- a/contrib/dhalion/src/main/java/org/apache/dhalion/api/IDiagnoser.java
+++ b/contrib/dhalion/src/main/java/org/apache/dhalion/api/IDiagnoser.java
@@ -1,0 +1,41 @@
+// Copyright 2016 Twitter. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package org.apache.dhalion.api;
+
+import org.apache.dhalion.symptom.Diagnosis;
+import org.apache.dhalion.symptom.Symptom;
+
+/**
+ * A {@link IDiagnoser} evaluates one or more {@link Symptom}s and produces a {@link Diagnosis}, if
+ * any, representing a possible problem responsible for the observed {@link Symptom}s.
+ */
+public interface IDiagnoser<T extends Symptom> extends AutoCloseable {
+  /**
+   * Initializes this instance and should be invoked once by the system before its use.
+   */
+  void initialize();
+
+  /**
+   * Evaluates available {@link Symptom}s and determines if a problem exists
+   *
+   * @return a {@link Diagnosis} instance representing a problem
+   */
+  Diagnosis<T> diagnose();
+
+  /**
+   * Release all acquired resources and prepare for termination of this instance
+   */
+  default void close() {
+  }
+}
--- a/contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
+++ b/contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
@@ -1,0 +1,36 @@
+//  Copyright 2017 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License
+package org.apache.dhalion.api;
+
+import java.util.Collection;
+
+import org.apache.dhalion.symptom.Symptom;
+
+public interface ISymptomDetector<T extends Symptom> extends AutoCloseable {
+  /**
+   * Initializes this instance and should be invoked once by the system before its use.
+   */
+  void initialize();
+
+  /**
+   * Detects a problem or issue with the distributed application
+   */
+  Collection<T> detect();
+
+  /**
+   * Release all acquired resources and prepare for termination of this instance
+   */
+  default void close() {
+  }
+}
--- a/contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
+++ b/contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
@@ -1,0 +1,38 @@
+// Copyright 2016 Twitter. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package org.apache.dhalion.api;
+
+import org.apache.dhalion.metrics.MetricsInfo;
+import org.apache.dhalion.symptom.ComponentInfo;
+import org.apache.dhalion.symptom.InstanceInfo;
+
+/**
+ * A {@link MetricsProvider} implementation will fetch and provide metrics to the consumers. For
+ * e.g. a {@link ISymptomDetector} may use it to get execute latency for a component.
+ */
+public interface MetricsProvider extends AutoCloseable {
+  <T extends ComponentInfo> MetricsInfo getComponentMetric(String metricName,
+                                                           T component,
+                                                           long duration);
+
+  <T extends InstanceInfo> MetricsInfo getInstanceMetric(String metricName,
+                                                         T instance,
+                                                         long duration);
+
+  /**
+   * Release all acquired resources and prepare for termination of this instance
+   */
+  default void close() {
+  }
+}
--- a/contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronDiagnoser.java
+++ b/contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronDiagnoser.java
@@ -1,0 +1,35 @@
+//  Copyright 2017 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License
+package org.apache.dhalion.heron.common;
+
+
+import javax.inject.Inject;
+
+import org.apache.dhalion.api.IDiagnoser;
+import org.apache.dhalion.symptom.ComponentSymptom;
+
+import com.twitter.heron.spi.common.Config;
+
+public abstract class BaseHeronDiagnoser implements IDiagnoser<ComponentSymptom> {
+  @Inject
+  protected Config config;
+  @Inject
+  protected Config runtimeConfig;
+  @Inject
+  protected TopologyProvider topologyProvider;
+
+  @Override
+  public void initialize() {
+  }
+}
--- a/contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronPolicy.java
+++ b/contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronPolicy.java
@@ -1,0 +1,35 @@
+//  Copyright 2017 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License
+package org.apache.dhalion.heron.common;
+
+import javax.inject.Inject;
+
+import org.apache.dhalion.api.IHealthPolicy;
+import org.apache.dhalion.api.IResolver;
+import org.apache.dhalion.symptom.ComponentSymptom;
+
+import com.twitter.heron.spi.common.Config;
+
+public abstract class BaseHeronPolicy implements IHealthPolicy {
+  @Inject
+  protected Config config;
+  @Inject
+  protected Config runtimeConfig;
+  @Inject
+  protected TopologyProvider topologyProvider;
+
+  @Override
+  public void initialize() {
+  }
+}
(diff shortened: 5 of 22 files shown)


contrib/dhalion/src/main/java/org/apache/dhalion/api/IDiagnoser.java
```diff
@@ -0,0 +1,41 @@
...
+   * Evaluates available {@link Symptom}s and determines if a problem exists
+   *
+   * @return a {@link Diagnosis} instance representing a problem
+   */
+  Diagnosis<T> diagnose();
+
+  /**
+   * Release all acquired resources and prepare for termination of this instance
+   */
+  default void close() {
```
huijunw: why is it 'default'?

contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
```diff
@@ -0,0 +1,36 @@
...
+public interface ISymptomDetector<T extends Symptom> extends AutoCloseable {
+  /**
+   * Initializes this instance and should be invoked once by the system before its use.
+   */
+  void initialize();
+
+  /**
+   * Detects a problem or issue with the distributed application
+   */
+  Collection<T> detect();
```
huijunw: Is the return value passed to IDiagnoser?

contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.dhalion.symptom.InstanceInfo;
+
+/**
+ * A {@link MetricsProvider} implementation will fetch and provide metrics to the consumers. For
+ * e.g. a {@link ISymptomDetector} may use it to get execute latency for a component.
+ */
+public interface MetricsProvider extends AutoCloseable {
+  <T extends ComponentInfo> MetricsInfo getComponentMetric(String metricName,
+                                                           T component,
+                                                           long duration);
```
huijunw: Does the 'duration' mean time window [now-duration, now]?

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public MetricValue getValue() {
+    return value;
+  }
+
+  public static class MetricValue implements Comparable<MetricValue> {
+    Double value;
+
+    public int asInt() {
```
huijunw: why do we need asInt()?

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
```diff
@@ -0,0 +1,53 @@
...
+      return value.intValue();
+    }
+
+    @Override
+    public int compareTo(MetricValue o) {
+      throw new UnsupportedOperationException("Needs to be implemented");
```
huijunw: we may provide a default comparison for 'Double value;'

contrib/dhalion/src/main/java/org/apache/dhalion/resolver/Action.java
```diff
@@ -0,0 +1,24 @@
...
+package org.apache.dhalion.resolver;
+
+import org.apache.dhalion.api.IResolver;
+import org.apache.dhalion.symptom.Diagnosis;
+
+/**
+ * {@link Action} is a representation of a action taken by {@link IResolver} to fix a
+ * {@link Diagnosis}
+ */
+public class Action {
```
huijunw: who is responsible to execute the Action object? IHealthPolicy.execute()?

billonahill:
If this code is to be committed into the heron project, the package naming should start with ```com.twitter.heron```. Heron should not be a home for code from other projects. If the ```org.apache.dhalion``` packing naming is desired, the code should live in a separate repo IMO.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * An {@link MetricsInfo} holds information pertinent to a specific metric. For e.g. has latency
+ * metric of a bolt instance for a Storm or Heron topology.
+ */
+public abstract class MetricsInfo {
+  private String name;
+  private MetricValue value;
+  private long timestamp;
+  private long duration;
```
huijunw: It may be hard to use 'duration'. 
The metric values are rarely same between two consecutive observations.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package org.apache.dhalion.metrics;
+
+/**
+ * An {@link MetricsInfo} holds information pertinent to a specific metric. For e.g. has latency
+ * metric of a bolt instance for a Storm or Heron topology.
+ */
+public abstract class MetricsInfo {
```
huijunw: which part of this class is abstract?

contrib/dhalion/src/main/java/org/apache/dhalion/symptom/Diagnosis.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A {@link Diagnosis} instance is a representation of a possible causes of one or more
+ * {@link Symptom}s. A {@link Symptom} could result in creation of one or more {@link Diagnosis}.
+ * Similarly, correlated {@link Symptom}s can result in generation of a {@link Diagnosis} instance.
+ */
+public class Diagnosis<T extends Symptom> {
+  private Set<T> symptoms;
```
huijunw: I saw Collection used in the other files. Why Set here? Any reason?

contrib/dhalion/src/main/java/org/apache/dhalion/symptom/Symptom.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.dhalion.symptom;
+
+/**
+ * A {@link Symptom} identifies an anomaly or a potential health issue in a specific component or
+ * in the distributed application in general. For e.g. identification of irregular processing
+ * latency.
+ */
+public abstract class Symptom {
+  long observedAt;
+  long duration;
```
huijunw: ComponentSymptom/InstanceSymptom is  a Collection of MetricsInfo that has timestamp and duration. What is the relation between timestamp/duration in MetricsInfo and observedAt/duration here?

contrib/dhalion/src/main/java/org/apache/dhalion/symptom/InstanceSymptom.java
```diff
@@ -0,0 +1,53 @@
...
+    this.instanceInfo = instanceInfo;
+    this.metrics = metrics;
+  }
+
+  public InstanceInfo getInstanceInfo() {
+    return instanceInfo;
+  }
+
+  public Optional<MetricsInfo> hasMetric(String metricName) {
+    return hasMetric(metricName, null);
```
huijunw: Does it imply there should be a 'name'->'null' pair for each metric in the collection?

contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.dhalion.metrics.MetricsInfo;
+import org.apache.dhalion.symptom.ComponentInfo;
+import org.apache.dhalion.symptom.InstanceInfo;
+
+/**
+ * A {@link MetricsProvider} implementation will fetch and provide metrics to the consumers. For
+ * e.g. a {@link ISymptomDetector} may use it to get execute latency for a component.
+ */
+public interface MetricsProvider extends AutoCloseable {
+  <T extends ComponentInfo> MetricsInfo getComponentMetric(String metricName,
```
huijunw: What if there are several data points in the duration? I suggest it return a Collection of MetricsInfo

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/HeronInstanceInfo.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.twitter.heron.spi.packing.PackingPlan;
+
+/**
+ * An {@link HeronInstanceInfo} holds information pertinent to a instance of a Heron topology
+ * component. For e.g. has task id of a bolt instance.
+ */
+public class HeronInstanceInfo extends InstanceInfo {
+  private int containerId;
+  private PackingPlan.InstancePlan instance;
```
huijunw: why do we use PackingPlan rather than physical plan?

contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
```diff
@@ -0,0 +1,36 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License
+package org.apache.dhalion.api;
+
+import java.util.Collection;
+
+import org.apache.dhalion.symptom.Symptom;
+
+public interface ISymptomDetector<T extends Symptom> extends AutoCloseable {
```
huijunw: I assume 
step 1, ISymptomDetector => Symptom
step 2, IDiagnoser => Diagnosis (Set<Symptom>)

Why not put step 1 & 2 together? What do we separate the detecting process into 2 procedures?

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronDiagnoser.java
```diff
@@ -0,0 +1,35 @@
...
+
+import javax.inject.Inject;
+
+import org.apache.dhalion.api.IDiagnoser;
+import org.apache.dhalion.symptom.ComponentSymptom;
+
+import com.twitter.heron.spi.common.Config;
+
+public abstract class BaseHeronDiagnoser implements IDiagnoser<ComponentSymptom> {
+  @Inject
```
huijunw: how does BaseHeronDiagnoser get the Symptom from BaseHeronDetector?
should we inject a BaseHeronDetector obj?

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.inject.Inject;
+
+import org.apache.dhalion.api.IHealthPolicy;
+import org.apache.dhalion.api.IResolver;
+import org.apache.dhalion.symptom.ComponentSymptom;
+
+import com.twitter.heron.spi.common.Config;
+
+public abstract class BaseHeronPolicy implements IHealthPolicy {
+  @Inject
```
huijunw: should we inject a BaseHeronDiagnoser and a BaseHeronResolver here so that the BaseHeronPolicy.execute() can access them?

contrib/dhalion/src/main/java/org/apache/dhalion/api/IDiagnoser.java
ashvina: This is to avoid creation of "empty" methods. I have seen empty `close()` method all over heron code. Once java 8 is accepted, dead code like this can be avoided.

contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
ashvina: Yes, a `HealthPolicy` could collect the symptoms and invoke `Diagnoser` as needed.

contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
ashvina: Yes, that is the intention. Current tracker API can return aggregated values.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
ashvina: This pattern is inspired by `ByteAmount`. The intent is to avoid casting related errors when converting from one type to other. This class will get more methods in due course.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
ashvina: Yes, double could be a good starting point. Will add it.

contrib/dhalion/src/main/java/org/apache/dhalion/resolver/Action.java
ashvina: The `Action` object would be used for maintaining a audit log.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
ashvina: One of the use cases of the `duration` field would be for `executeCount`. The aggregate value of count is return by tracker today. This `duration` field will be used by the `Detector` or `Diagnoser` for accurate comparisons and calculations.

contrib/dhalion/src/main/java/org/apache/dhalion/metrics/MetricsInfo.java
ashvina: We expect Storm and Heron specific implementations for `MetricsInfo`. The `abstract` enforces the need to create one. Also this is work in progress and more methods and fields will be added to this class.

contrib/dhalion/src/main/java/org/apache/dhalion/symptom/Symptom.java
ashvina: We should use consistent names for timestamp, thanks for caching this. Will fix it. The semantics for duration is the same in both cases.

contrib/dhalion/src/main/java/org/apache/dhalion/api/MetricsProvider.java
ashvina: Interesting. Can you share an example where Heron will return multiple data points for the same metric?

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/HeronInstanceInfo.java
ashvina: My impression si that the `protobuf` based physical plan is more suitable for serializing the plan and the `PackingPlan` is to be used otherwise. For e.g. the `UpdateManager` also depends on `PackingPlan`. What do you think?

contrib/dhalion/src/main/java/org/apache/dhalion/api/ISymptomDetector.java
ashvina: The relation between a `symptom` and a `diagnoser` is not 1-1. Separating enables reuse and customization.

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronDiagnoser.java
ashvina: Currently we are thinking of invoking the detector from the diagnoser. Other option is to inject a `provider`. Provider can optimize the number of times a detector is invoked.

contrib/dhalion/src/main/java/org/apache/dhalion/heron/common/BaseHeronPolicy.java
ashvina: Yes, that seems ok. Though, it might complicate the way these components are configured.

ashvina:
As pointed out by @billonahill  and based on discussion on Slack, the `Dhalion` project api will be moved outside Heron. Any `Heron` specific Dhalion integration work will be moved to `heron` directory from the `contrib` directory. 

So I will close this PR once a new PR is created.

huijunwu:
Any updates on this PR?

PR #1788, opened 2017-03-31
Improve dry-run

### LINKED ISSUE #1783: Dry-run should display RAM usage in MB to avoid rounding

objmagic:
1. default formatter `table` does not render in rich format. This PR adds a new formatter `colored_table` which will render in previous rich format. Tests are addded
2. Use MB instead of GB for RAM and disk usage
3. Other minor improvements

resolves #1783


heron/scheduler-core/src/java/com/twitter/heron/scheduler/dryrun/FormatterUtils.java
```diff
@@ -63,6 +69,14 @@ private FormatterUtils() {
     STRIKETHROUGH
   }
 
+  public enum ContainerChange {
+    UNAFFECTED,
+    MODIFIED,
```
billonahill: What's modified and how is it different than NEW and/or REMOVED?

objmagic: it is from [here](https://github.com/twitter/heron/pull/1788/files/25a39b507aa8c58419658b335289711afdf7c513#diff-4850d7a660ab463553d7e473b7ee6142L42). Modified means there are instances added/removed from the container.

PR #1789, opened 2017-04-03
check if the treemap is empty during metircscache purge process

huijunw:
check if the treemap is empty during purge process

### DIFF
--- a/heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/metricscache/CacheCore.java
+++ b/heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/metricscache/CacheCore.java
@@ -482,9 +482,11 @@ public class CacheCore {
     long now = System.currentTimeMillis();
     synchronized (CacheCore.class) {
       // remove old
-      for (Long firstKey = cacheMetric.isEmpty() ? null : cacheMetric.firstKey();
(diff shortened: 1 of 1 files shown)


huijunwu:
if (!cacheMetric.isEmpty()) {
   for ( ;; /*this may throw exception*/ ) {
      // remove changes map
   }
}
i do not think puting the `for `loop inside `if `block can handle the exception throw from `for `loop.
the loop changes the collection and the exception may be thrown from inside `if `block.

billonahill:
Can you clarify the excepton and what causes it? It's not clear from your description, so based on the code I assumed it was caused when calling ```cacheMetric.firstKey()``` when ```cacheMetric.isEmpty() == true```. Putting that check outside the for loop would keep that from happening no?

huijunw:
The `for` loop changes the treemap. After entering the `if` block, the treemap may change to empty.

billonahill:
Could we do this then instead? I'm suggesting for readability. Trying to decipher the reasoning for the null checks make the code harder to understand.
```
while (!cacheMetric.isEmpty()) {
  Long firstKey = cacheMetric.firstKey()
  if (firstKey < now - maxIntervalMilliSecs) {
    break;
  }
  cacheMetric.remove(firstKey);
}
```

huijunw:
updated

Issue #1502, opened 2016-10-19
Show backpressure containers in heron-ui


kramasamy:
Show the back pressure containers in UI - even if there is no instance that triggered them. This could indicate any network issue.

Issue #1501, opened 2016-10-19
Show backpressure instance in heron ui


kramasamy:
Visually need to show which instance in a container initiated the back pressure.

kkdoon:
Yes, PR https://github.com/twitter/heron/issues/1791 solves this issue.

PR #1791, opened 2017-04-04
Add API support for backpressure metric (instance level) & display backpressure (instance level) in Heron UI

kkdoon:
a. heron_api.py:
 1. Added API support (`/topologies/metrics/timeline?metric=bp`) for new metric `back-pressure`. Created a new function `fetch_bp` since querying back-pressure data from tracker API is quite different from other metrics. Other metrics (gc, capacity, etc.) pass their component name and `*` as their instance name to query respective metric from Tracker API. On the other hand, bp metric is a part of `__stmgr__` component, which contains multiple instances (`__stmgr__1`, `__stmgr__2`, etc.).

Therefore, it is not possible to directly query back pressure data for each component present in the DAG. Instead, we need to query **each component's instance individually** for back pressure data `(DEFAULT(0, TS(__stmgr__,*,__time_spent_back_pressure_by_compid/{instance})))`.

2. Also added `get_instances` function to retrieve all instances from topology's physical plan. These instances names are used while querying back pressure data from Tracker API.
3. Added `compute_max` & `get_metric_response` functions which extract common LOC from `fetch` & `fetch_max` methods and are re-used by the `fetch_bp` method.
4. `fetch_max` method had a minor bug, where each component response was re-computed multiple times, since the logic for computing maximum value was included inside the for-loop block. This issue is also fixed in this commit.

b. plan-stats.js
1. Included back-pressure metric in the front-end. 
2. Currently, the following values have been configured for low, medium & high ranges: 3000, 12000, 30000. This means that if any instance of the topology is under back pressure for 0 - 3 seconds, then the health of that instance is marked as good. Similarly, if it is under back pressure for 3 - 12 seconds, then the health of that instance is marked as medium, and if it is between 12 - 30 seconds, then it is marked as bad.


objmagic:
I was working with @kkdoon on this so I will review later.

huijunw:
does this pr related to #1501 and #1502 ?

kkdoon:
Yes, This PR solves https://github.com/twitter/heron/issues/1501. However, the issue mentioned in https://github.com/twitter/heron/issues/1502 is not tackled by this PR. So if there is any instance in back-pressure then the UI will indicate that instance. However, if any container is in BP due to network issues, then the UI will not display that.

billonahill:
Thanks @kkdoon! Would you please include a screenshot of the proposed change and update the title of this pr to be more descriptive.

kkdoon:
No problem @billonahill. Updated the title. Please refer to the attached image below for more details:

1. Overall Heron UI page for any topology:
This image shows all the instances that are under back-pressure. User can hover over the instance for more details.
bp_overall

2. Back Pressure Metric:
This image shows inclusion of back pressure metric in Heron UI.
bp_metric

3. Max Back Pressure Metric:
This image shows that max back pressure data is plotted, across all instances in topology.
max_bp

billonahill:
Thanks @kkdoon for the screenshots, this looks awesome and is similar to what we've envisioned for this. I think one of the key things is to be able to easily visualize the containers/instances under backpressure on the physical plan view in the upper right, without having to mouse over anything. If those are somehow highlighted when in backpressure that would be huge. (Being able to link to a viz plot that shows which container is in BP over time would also be a big win, but this is beyond the scope of this PR.)

Also, just as an FYI, we've also discussed being able to highlight which containers have registered/not registered with TMaster in that same view during the submit process. This is of course also outside of the scope of this PR, but another use case to keep in mind when developing the highlighting functionality.

kramasamy:
Thanks @kkdoon - it looks great. A couple of questions 

* If the back pressure is initiated by an instance, will we able to see the instance that generated the back pressure visually, when we click on the back pressure metric

* If the back pressure is initiated by the stream manager, we should be able to see it visually as well. In this case, when we click on the back pressure metric, it could show the container shaded? - so we can see right away that it is a stream manager initiated back pressure.

kkdoon:
Hi @kramasamy ,

- Point 1 is implemented, i.e. all instances in back pressure are shown in the physical plan, once the bp metric is clicked. Also, the components in logical plan/DAG are highlighted.
- Point 2 is not implemented. I was wondering which metric to use to find out which container is under back pressure. I can see that metric "__server/__time_spent_back_pressure_initiated" as well as "__server/__time_spent_back_pressure_aggr" is available from stream manager instances. Could you direct me to relevant API or metric? I can try to add this feature as well.

kramasamy:
@kkdoon - for point 2, please check with @congwang and see if there is any metrics for stream manager based back pressure.

objmagic:
Targeting 0.15.0+

Sorry for not reviewing this PR... Probably will do in May.

kramasamy:
@objmagic - any update on this?

objmagic:
probably merge this in later this week, or next week.

objmagic:
I've been testing this PR internally this week, and it functions well. I am in favor of merging this soon.

Does anyone here have any more comment?

heron/tools/common/src/python/access/heron_api.py
```diff
@@ -74,12 +74,15 @@
...
+bp = "DEFAULT(0, TS(__stmgr__,*,__time_spent_back_pressure_by_compid/{0}))"
+
 queries = dict(
     cpu=cpu,
     capacity=capacity,
     failures=failures,
     memory=memory,
-    gc=gc
+    gc=gc,
+    bp=bp
```
billonahill: s/bp/backpressure

heron/tools/common/src/python/access/heron_api.py
```diff
@@ -736,21 +753,91 @@ def fetch_max(self, cluster, metric, topology, component, instance, timerange, e
...
+    results = yield futures
+
+    data = self.compute_max(results)
+
+    result = self.get_metric_response(timerange, data, True)
+
+    raise tornado.gen.Return(result)
+
+  @tornado.gen.coroutine
+  def fetch_bp(self, cluster, metric, topology, component, instance, \
```
billonahill: s/fetch_bp/fetch_backpressure

heron/tools/common/src/python/access/query.py
```diff
@@ -42,3 +42,19 @@ def fetch_max(self, cluster, metric, topology, component, instance, timerange, e
     :return:
     '''
     pass
+
+  def fetch_bp(self, cluster, metric, topology, component, instance, \
```
billonahill: s/fetch_bp/fetch_backpressure

heron/tools/ui/resources/static/js/plan-stats.js
```diff
@@ -60,6 +60,16 @@
         scaleTicks: [0, 500, 1000, 1500, 2000],
         format: function (d) { return d.toFixed(0); }
       },
+      {
+        name: 'BP',
```
billonahill: Here and elsewhere, let's use backpressure instead of the abbreviated form of bp.

kkdoon: I intentionally named it BP in the heron/tools/ui/resources/static/js/plan-stats.js file as this name is too long and it gets cut while rendering, as shown below:
 
image

Instead of displaying Maximum Back Pressure it only displays Back Pressure. So I changed it to BP.

objmagic: @kkdoon ideally we should adjust CSS to make that column wider.

kkdoon: @objmagic ok I will try to adjust the CSS and try again with 'Back Pressure' label.

objmagic: @kkdoon We can include this in next release if we can do the CSS fix today or tomorrow.

objmagic:
Merged. Will fix the BP name and CSS change in another PR.

PR #1796, opened 2017-04-04
Adding support for Java 8 metaspace to fix #1790

### LINKED ISSUE #1790: Heron must set max metaspace size for Java 8 processes
We've seen java processes using up to .5GB more memory that allocated per ```Xmx``` settings. This is because in java 8 permgen is removed and metaspace is now used, which is unbounded by default (see https://dzone.com/articles/java-8-permgen-metaspace). The fix is to bound it like this if we're running java 8:
```
-XX:MaxMetaspaceSize=192m
```

For topologies running without this fix, the workaround is to do this for each component:
```
Config.setComponentJvmOptions(config, componentName, "-XX:MaxMetaspaceSize=192m")
```

billonahill:
See #1790 for summary of the issue. While testing I also found that there is java 8 code in ShellUtils that I fixed to be compatible with Java 7.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -209,6 +209,7 @@ class HeronExecutor(object):
     self.processes_to_monitor = {}
 
     self.state_managers = []
+    self.jvm_version = None
 
   @staticmethod
   def parse_args(args):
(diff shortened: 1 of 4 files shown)


heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
```diff
@@ -315,4 +315,15 @@ public static boolean extractPackage(
     return ret == 0;
   }
 
+  // java 7 compatible version of String.join(" ", array), available in java 8
+  private static String toString(String[] array) {
```
huijunw: I would suggest `joinString` name. The `toString` may be confusing with `Object.toString`.

heron/executor/src/python/heron_executor.py
```diff
@@ -466,6 +475,21 @@ def _get_java_instance_cmd(self, instance_info):
       retval[instance_id] = instance_cmd
     return retval
 
+  def _get_jvm_version(self):
+    if not self.jvm_version:
+      cmd = [os.path.join(self.heron_java_home, 'bin/java'),
+             '-cp', self.instance_classpath, 'com.twitter.heron.instance.util.JvmVersion']
+      process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
huijunw: why not call shell command 'java --version' from python?

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
objmagic: +1

heron/executor/src/python/heron_executor.py
objmagic: you still need to parse its output. This could be flaky compared with a call of `System.getProperty("java.version")` which is a stable API.

billonahill: Exactly. The string output is very different depending on the jvm. Try it with a default jvm vs a twitter one for example. Parsing those differences would be more brittle.

heron/spi/src/java/com/twitter/heron/spi/utils/ShellUtils.java
billonahill: Will do.

billonahill:
@objmagic I cache the result of ```_get_jvm_version``` so I don't need to keep spinning up new JVMs to check the version repeatedly. The cached value is ```self.jvm_version```.

objmagic:
@billonahill oh I misread and thought `self.jvm_version` was already added before. This is good choice 👍

heron/executor/src/python/heron_executor.py
kramasamy: Also, bin/java - you might not want to hard code it. You might want to do two path join - we can make it windows proof.

kramasamy: Another issue could be Twitter internal JVM version detection needs to be different since on the version string, it does something different - we have to look at the next line and parse it to determine the version.

billonahill:
@kramasamy the bin/java approach is what's used in a number of places in the existing code. I'd prefer for that change to be tackled in a PR of it's own, to keep the changes and scope clean.

Regarding the version string, yes ```java -version``` can output just about anything. That's why I wrote ```com.twitter.heron.instance.util.JvmVersion``` which just outputs the version string. This makes parsing much more resilient and reliable.

objmagic:
putting this into 0.14.7 for OSS testing. Is it OK? @billonahill

billonahill:
@objmagic yes. Just got the final approval and merges. Timur has identified other params that we might want to add as well fyi.

PR #1798, opened 2017-04-05
[Do not merge] change log FATAL to ERROR to avoid crashing stmgr

huijunw:
There were corrupt events in EventBus which were crashing stream manager. With errors such as this:
last message is : F0120 09:03:24.136044 55260 client.cpp:226] UnPackString failed from connection 0x28da580 from hostport 10.44.144.115:31661

Heron should be resilient to such issues. Corrupt packets should be dropped and not crash the Stream Manager.

### DIFF
--- a/heron/common/src/cpp/network/client.cpp
+++ b/heron/common/src/cpp/network/client.cpp
@@ -181,8 +181,10 @@ void Client::OnNewPacket(IncomingPacket* _ipkt) {
 
   if (_ipkt->UnPackString(&typname) != 0) {
     Connection* conn = static_cast<Connection*>(conn_);
-    LOG(FATAL) << "UnPackString failed from connection " << conn << " from hostport "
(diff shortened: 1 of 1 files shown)


maosongfu:
Previously we had discussion on this. We thought this error is super abnormal (not due to EventBus corrupt events) and we should LOG(FATAL) to exit the process rather than be resilient to it.

billonahill:
I'm not taking a stance on FATAL or not, as I don't have full context, but if we choose to not exit the process we must always take care not to log on error for high-throughput systems. This can quickly fill up disk and cause a cascade of problems. Error counters should be favored over error logging. Or some sort of configurable log throttling/sampling mechanism.

congwang:
@huijunw Don't waste your time on this... People do love crash here...

huijunw:
We do not have plan to merge this PR. Close it.

PR #1799, opened 2017-04-06
We need to unregister the metric and delete it before clearing the map

srkukarni:


congwang:
Looks good.

nlu90:
👍

maosongfu:
Looks good. I noticed that, during HandleConnectionClose(..), the instance_metric will not be cleaned (removed); will it be a problem too?

We need to verify the changes before merging it.

maosongfu:
It does not solve the issue. WILL NOT MERGE this pull request.

srkukarni:
@maosongfu Regardless of whether it fixes the bug that we saw, this itself is a memory leak that needs to be fixed.

maosongfu:
Close this pull request due to #1801

PR #1801, opened 2017-04-07
Fixes the SIGSEGV during UpdateQueueMetrics from a restarted instance

maosongfu:
This pull request fixes the SIGSEGV during UpdateQueueMetrics from a restarted instance

Current stmgr will clean only connection_buffer_metric_map_ during HandleConnectionClose.
However, during HandleRegisterInstanceRequest(), before adding necessary meta-info, stmgr will check the existence of instance basing on instance_info_, which is not cleaned. So connection_buffer_metric_map_ will no longer contain the entry of instance once it is restarted, since it will not be added. And accessing to that entry returns SIGSEGV.

Besides that, this pull request also fixes potential memory leaks.

Tested and verified on LocalScheduler.


nlu90:
👍

objmagic:
👍

PR #1805, opened 2017-04-10
One-liner to fix Heron Tracker when results aren't found

billonahill:
When components are not found, a None exception is returned. With this fix, a valid response is returned instead.


maosongfu:
lgtm. 👍

PR #1806, opened 2017-04-10
Throttle the amount of 'Stop reading due to not yet connected to Stream Manager' log messages emitted

billonahill:
We stream massive amounts of these to the logs upon startup, many per second. This changes throttles that log message.

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/network/StreamManagerClient.java
+++ b/heron/instance/src/java/com/twitter/heron/network/StreamManagerClient.java
@@ -254,7 +254,7 @@ public class StreamManagerClient extends HeronClient {
       }
     } else {
       long now = System.currentTimeMillis();
-      if (now - lastNotConnectedLogTime > lastNotConnectedLogThrottleSeconds * 5000) {
+      if (now - lastNotConnectedLogTime > lastNotConnectedLogThrottleSeconds * 1000) {
         LOG.info(String.format("Stop reading due to not yet connected to Stream Manager. This "
             + "message is throttled to emit no more than once every %d seconds.",
             lastNotConnectedLogThrottleSeconds));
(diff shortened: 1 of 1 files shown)


maosongfu:
lgtm. 👍

congwang:
I am a bit surprised glog doesn't provide an API for ratelimit logging.

objmagic:
@congwang this is Java client of stmgr.

congwang:
@objmagic Oops, I mean java.util.logging... My point is it is ugly to expose rate limit details to it caller.

billonahill:
@congwang yeah, this isn't pretty and maybe it could be put into a logging decorator perhaps if we want to do more of this, but `java.util.logging` doesn't do any of this for us AFAIK.

congwang:
@billonahill Agreed. My Java knowledge is almost none, so I don't know if it is hard to just extend java.util.logging.Logger.

heron/instance/src/java/com/twitter/heron/network/StreamManagerClient.java
```diff
@@ -249,7 +253,13 @@ private void readStreamMessageIfNeeded() {
         stopReading();
       }
     } else {
-      LOG.info("Stop reading due to not yet connected to Stream Manager.");
+      long now = System.currentTimeMillis();
+      if (now - lastNotConnectedLogTime > lastNotConnectedLogThrottleSeconds * 5000) {
```
objmagic: should it be `* 1000`?

objmagic: should it be `* 1000`?

billonahill:
Thanks @objmagic that was indeed a typo. Fixed.

PR #1807, opened 2017-04-10
DO NOT MERGE: Hackweek project to get Trident working in Heron

billonahill:
This is not to be merged. It's just to show the changes made for a prototype of how Trident might work on Heron. See https://github.com/billonahill/heron/blob/billg/trident_prototype/heron/storm/src/java/org/apache/storm/trident/README.md for specifics.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -276,6 +276,18 @@ maven_jar(
   artifact = "org.apache.reef:tang:" + reef_version
 )
 
+# added for trident prototype
+maven_jar(
+  name = "org_clojure_clojure",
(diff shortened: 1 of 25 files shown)


kramasamy:
@billonahill - for trident instead of using their stateful processing, we should map into our exactly once processing. This will be a cleaner implementation rather than implementing two stateful processing algorithms, IMO.

billonahill:
@kramasamy the point of this branch was just to see if we could get Trident to run on Heron, to understand how it's implemented and to see it work. I have not yet fully achieved all of those goals, but I recommend doing so before deciding to re-implement parts of Trident.

PR #1813, opened 2017-04-24
Clean up CustomGroupingTest

billonahill:
Clean up to improve readability, no no functionality here.


billonahill:
Sorry for piling on with my commits, but I made a large refactor to set these tests up better for extensibility. With this patch I can now follow up in another PR to make this class abstract and have `CustomGroupingTest` as the concrete subclass with only the content from below the `addComponent` method (I didn't want to do this in one PR since the class naming would make reading the diff hard). We can then add another subclass to test emit direct from the spout and another to do the same from a bolt.

Issue #1479, opened 2016-10-06
Erroneous MM response (PyHeron)


objmagic:
```
10/06/2016 14:45:55 -0700:INFO:spout_instance.py: Adding spout task...
10/06/2016 14:45:55 -0700:INFO:st_heron_instance.py: Started instance successfully.
10/06/2016 14:46:00 -0700:ERROR:event_looper.py: Error occured during _run_once(): Response from Metrics Manager not OK
10/06/2016 14:46:00 -0700:ERROR:event_looper.py: Traceback (most recent call last):
  File "heron/common/src/python/basics/event_looper.py", line 65, in _run_once
    self._trigger_timers()
  File "heron/common/src/python/basics/event_looper.py", line 146, in _trigger_timers
    task()
  File "heron/common/src/python/network/heron_client.py", line 213, in timeout_task
    self.handle_timeout(reqid)
  File "heron/common/src/python/network/heron_client.py", line 230, in handle_timeout
    self.on_response(StatusCode.TIMEOUT_ERROR, context, None)
  File "heron/instance/src/python/network/metricsmgr_client.py", line 80, in on_response
    raise RuntimeError("Response from Metrics Manager not OK")
RuntimeError: Response from Metrics Manager not OK
```

maosongfu:
Can you also check metrics-mgr's log?

objmagic:
turns out I found some err log in Stream Manager's log. They look helpful

```
E1006 14:35:00.763510 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.130671 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.138960 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.462795 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.472683 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.787544 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
E1006 14:35:01.794572 2053763072 server.cpp:170] Unknown type protobuf received heron.proto.stmgr.TupleMessage deleting...
[...]

PR #1814, opened 2017-04-25
Fix unexpected non-OK response in metricsmgr_client.py

Detoo:
metricsmgr_client.py did not return after a failed connection attempt (on_connect() & status != OK), which led to unwanted send_register_req() call and is likely the cause of issue #1479 since the request is sent without a proper connection.

In comparison, the Java counterpart does return:
https://github.com/twitter/heron/blob/0.14.6.3/heron/instance/src/java/com/twitter/heron/network/MetricsManagerClient.java#L139


kramasamy:
👍 looks good to me.

objmagic:
Thanks. Please sign CLA first: https://engineering.twitter.com/opensource/cla

Detoo:
Thanks, I have just signed it

kramasamy:
@Detoo - we will merge it once all the tests pass.

Issue #1561, opened 2016-11-14
Integration test topologies assert exactly once, but run with at most once.


billonahill:
We're seeing occasional failures like this in CI. This issue is that `IntegrationTestSpout` now decrements the `tuplesToAck` counter only when tuples are acked ([commit link](https://github.com/twitter/heron/commit/1123f731cb4a399ef4717e1c07dcbd22ccbab37d#diff-92e59551fb6a8e1002ba5c3f20f8f61aL127)). Previously it decremented that counter even when a `fail` call was received. What this implies is that our test topologies are dropping tuples and the spout is re-transmitting them. Why the test topologies are dropping tuples is concerning.

A few options we could pursue:

1. Why are tuples failing, causing them to be re-emitted? That shouldn't be happening in our tests
2. We could change `IntegrationTestSpout` back to decrement upon failure.
3. We could change test assertion to assert `atLeastOnce` instead of `exactlyOnce`.
```
ERROR: Actual result did not match expected result
INFO: Actual result ---------- 
['A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B']
INFO: Expected result ---------- 
['A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B']
[...]

Issue #1691, opened 2017-02-03
Fix and re-enable MultiSpoutsMultiTasks test


objmagic:
Recently we see constant failure of MultiSpoutsMultiTasks. Possibly related to #1561

objmagic:
working on #1687

objmagic:
resolved via https://github.com/twitter/heron/pull/1815

PR #1815, opened 2017-04-25
Fix flaky test IntegrationTest_MultiSpoutsMultiTasks

huijunwu:
Fix
https://github.com/twitter/heron/issues/1561
https://github.com/twitter/heron/issues/1691
https://github.com/twitter/heron/pull/1687
https://github.com/twitter/heron/pull/1700

The cause is:
tuple count includes 'termination tuple', which leads to multiple same acking tuples and stmgr failure to cancel out markers in xor_manager.
https://travis-ci.org/twitter/heron/builds/225447993

Fix:
count only normal tuples
https://travis-ci.org/twitter/heron/builds/225471892

### DIFF
--- a/integration-test/src/java/com/twitter/heron/integration_test/core/IntegrationTestBolt.java
+++ b/integration-test/src/java/com/twitter/heron/integration_test/core/IntegrationTestBolt.java
@@ -81,7 +81,6 @@ public class IntegrationTestBolt implements IRichBolt, IUpdatable {
 
   @Override
   public void execute(Tuple tuple) {
-    tuplesReceived++;
     String streamID = tuple.getSourceStreamId();
 
     LOG.info("Received a tuple: " + tuple + " ; from: " + streamID);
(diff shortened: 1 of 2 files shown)


kramasamy:
@huijunwu - wondering if you ran the tests a few times before the flakiness is gone.

objmagic:
Also enable https://github.com/twitter/heron/pull/1773 ?

huijunw:
@kramasamy Test was run hundred times; all passed
@objmagic IntegrationTest_OneSpoutBoltMultiTasks is different. More time needed.

billonahill:
Great find. I don't understand how this would make the test flaky. Why wouldn't the tests fail consistently without this fix?

huijunwu:
@billonahill 
The time when 'termination tuple' arrives the identitybolt is different everytime. If a 'termination tuple' arrives between two normal tuple, the count is one more than expected and the acking tuple is one more than expected. 
Usually, the 'termination tuple' is after all the normal tuples so there is no issue. 
There are 6 spouts in the test and if one of them runs fast and sends out 'termination tuple' before the other spouts sending out normal tuples, the wrong count appears.

PR #1820, opened 2017-04-26
Added Stateful spout/bolt interfaces and State interface for exactly once

srkukarni:


heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -147,6 +147,31 @@
    */
   public static final String TOPOLOGY_COMPONENT_RAMMAP = "topology.component.rammap";
   /**
+   * Is this topology stateful? The format of this flag is boolean
+   */
+  public static final String TOPOLOGY_STATEFUL = "topology.stateful";
```
nlu90: one general question:
1. should we provide the setters method like other configs listed here. e.g. setTopologyStateful();

srkukarni: Makes sense. I will add it and update

heron/api/src/java/com/twitter/heron/api/state/HashMapState.java
```diff
@@ -0,0 +1,25 @@
...
+
+package com.twitter.heron.api.state;
+
+import java.io.Serializable;
+import java.util.HashMap;
+
+/**
+ * A State backed up by HashMap
+ */
+public class HashMapState extends HashMap<Serializable, Serializable> implements State {
```
nlu90: The extends here looks odd to me. 

Could we make a hashmap member instead of using inheritance?

heron/api/src/java/com/twitter/heron/api/state/State.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.Serializable;
+import java.util.Map;
+
+/**
+ * State represents the state interface as seen by stateful bolts and spouts.
+ * In Heron, state gives a notional Key/Value interface along with the
+ * ability to iterate over the key/values
+ */
+public interface State extends Map<Serializable, Serializable> {
```
nlu90: is there any common operation across different type of states, like adding something into the current state? we could put them here.

srkukarni: The State as defined here is a very generic interface for spouts/bolts to keep their internal state. Any type of common operation will probably come in higher layers(like DSL)

heron/api/src/java/com/twitter/heron/api/topology/IStatefulComponent.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.twitter.heron.api.topology;
+
+import com.twitter.heron.api.state.State;
+
+/**
+ * Common methods for all possible components in a topology. This interface is used
+ * when defining topologies using the Java API.
+ */
+public interface IStatefulComponent extends IComponent {
```
srkukarni: Spouts/bolts should hold on to state variable to store their data. This is very consistent with OutputCollector in ISpout/IBolt

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: Done

nlu90:
👍

heron/api/src/java/com/twitter/heron/api/bolt/IStatefulBolt.java
```diff
@@ -0,0 +1,27 @@
...
+import com.twitter.heron.api.topology.IStatefulComponent;
+
+/**
+ * An IStatefulBolt represents a bolt that implements stateful functionality.
+ * In addition to the regular prepare/execute methods, it has initialization
+ * methods with state as defined in IStatefulComponent. An implementation
+ * of the stateful bolt saves state in the State interface and is able
+ * to restore its state from the State during init call.
+ */
+public interface IStatefulBolt extends IStatefulComponent, IRichBolt {
```
maosongfu: I would think having IStatefulComponent is good enough; no need IStatefulBolt or IStatefulSpout:
- IStatefulBolt does not add more methods than IStatefulComponent and IRichBolt. A bolt can simply implement IStatefulComponent if it wants Stateful feature; on the contrast, IUpdatable interface looks better (https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/topology/IUpdatable.java).
- IStatefulBolt extends IRichBolt. Then how about IBasicBolt? Need we add it too? And how about others, ? It can either lead to an explosion of APIs, which in fact IMO are duplicated; or it can lead to more refactoring on existing topologies to support these features (for instance, current bolt extends IBaseBasicBolt, which provides different APIs from IRichBolt).

In a word, I think this wrapper complicates things without bringing much benefits.

heron/api/src/java/com/twitter/heron/api/spout/IStatefulSpout.java
```diff
@@ -0,0 +1,30 @@
...
+ * An IStatefulSpout represents a spout that implements stateful functionality.
+ * In addition to the regular open/nextTuple methods, it has initialization
+ * methods with state as defined in IStatefulComponent. An implementation
+ * of the stateful spout saves state in the State interface and is able
+ * to restore its state from the State during init call. For example the
+ * Kafka spout could store the offset everytime it advances. If it dies
+ * the system restores the state and the spout is able to start from where
+ * it left of last.
+ */
+public interface IStatefulSpout extends IStatefulComponent, IRichSpout {
```
maosongfu: Similar to above.

heron/api/src/java/com/twitter/heron/api/state/HashMapState.java
huijunwu: interface State is already a map. If a hashmap member placed here, it leads to nested map.

heron/api/src/java/com/twitter/heron/api/bolt/IStatefulBolt.java
srkukarni: IStatefulSpout/IStatefulBolt are mimicking IRichSpout/IRichBolt. They are essentially a shorthand version for any kind of stateful component.
IRich/IBasic stuff are essentially more developer friendly versions of the basic ISpout stuff. Thus in the future, depending on requirements, we might have IBasicStatefulBolt/Spout as well. They make it easy for developers to quickly start writing bolts/spouts.

heron/api/src/java/com/twitter/heron/api/state/State.java
billonahill: Can we use generics in the State API for better type safety? (`Map<<K extends Serializable>, <V extends Serializable>>`)

heron/api/src/java/com/twitter/heron/api/Config.java
billonahill: Is this something the user needs to set, or the field that Heron internally sets when inspecting the DAG? (I would expect the latter, in which case can we document that or not expose it as a user setting?)

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -147,6 +147,31 @@
...
+  /**
+   * What's the checkpoint interval for stateful topologies in seconds
+   */
+  public static final String TOPOLOGY_STATEFUL_CHECKPOINT_INTERVAL =
```
billonahill: Can we include `seconds` in the variable and field naming to be explicit.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -147,6 +147,31 @@
...
+                             "topology.stateful.checkpoint.interval";
+  /**
+   * What's the provider for state? i.e. one where state is stored
+   */
+  public static final String TOPOLOGY_STATEFUL_PROVIDER_TYPE =
```
billonahill: Is this the implementation class? If so can we use `class` instead of type to be more explicit?

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -147,6 +147,31 @@
...
+                             "topology.stateful.provider.type";
+  /**
+   * What's the config for state provider?
+   */
+  public static final String TOPOLOGY_STATEFUL_PROVIDER_CONFIG =
```
billonahill: Is this a pointer to a config file? If so can we include `file` in the name and clarify in the comments?

heron/api/src/java/com/twitter/heron/api/bolt/IStatefulBolt.java
billonahill: +1 for `IStatefulComponent` interface with common methods. If we later find that bolts and spouts need specific methods we can subclass `IStatefulComponent`. Fewer interfaces and a shared top-level interface hierarchy will simplify the lives of both heron and user developers. There will be places where we will need to do common operations on stateful components and having to check both stateful bolt and spout will cause code duplication and headaches.

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: These are all user facing apis. Users are expected to set this(we of course take some default stuff)

billonahill: Why does the user need to set `topology.stateful`? Should using `IStatefulComponent` and configuring the stateful store be sufficient? Are you thinking they'll want a flag to turn it off?

heron/api/src/java/com/twitter/heron/api/state/State.java
srkukarni: We don't want to restrict State to a particular type of key/value. Thus say KafkaSpout can do state.put("offset", 111);
while some other type of bolt can do
state.put(120, "Sanjeev");
We should take any object as long as it is serializable.

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: Not really.
One could run a stateful topology in non stateful mode. i.e. I could write a IStatefulSpout/Bolt but may prefer to turn on and off the stateful config for maybe different use-cases or experimentation.

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: done

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: done

heron/api/src/java/com/twitter/heron/api/state/State.java
billonahill: `Map<<K extends Serializable>, <V extends Serializable>>` says that we should take any kind of state as long as it is serializable. The difference is how the users will write their code, not how the system deals with serializables. For the spout and bolt in your example, one would be able to work with `<String, Integer>` while the other would be able to work with `<String, String>`. This allows the user to write type safe code with compile time checks, while still allowing the system to persist any types of serializables.

To do this we would have something like this, where the user knows they can work with `<String, Integer>' for example:
```
interface IStatefulComponent<String, Integer> {
  initState(State<String, Integer> state);
}
```

heron/api/src/java/com/twitter/heron/api/Config.java
billonahill: Would we need this globally, or per component? Assuming globally, could we give a more descriptive name like `topology.component.state.enabled` or something like that?

srkukarni: This is a topology specific config. Thus it should always be topology.<feature>
All per component variables are prefixed by topology.component.

maosongfu: Agree it should be a topology specific config:
- if people want a component to be stateless, they can simply not to store any states, even with this config is set true
- it does not make sense to make this config per component; the whole topology DAG is a computing job and the global state will be correct only when all components have set `topology.component.state.enabled` true.

heron/api/src/java/com/twitter/heron/api/state/State.java
maosongfu: +1 `Map<<K extends Serializable>, <V extends Serializable>>`

heron/api/src/java/com/twitter/heron/api/topology/IStatefulComponent.java
maosongfu: @nlu90 Heron feeds the state during `void initState(State state);` so it always has the reference for the `state`.

heron/api/src/java/com/twitter/heron/api/state/State.java
objmagic: @billonahill 
> Can we use generics in the State API for better type safety? (Map<<K extends Serializable>, <V extends Serializable>>)

Out of curiosity, what is the meaning of "type safety" in this sentence? In what way could user implement the `State` interface so that certain "type safety" could be lost?

objmagic: @billonahill do you mean that `<K extends Serializable>` will simply put constraint on `K`, while `Map<Serializable, Serializable>` will cast key or value to `Serializable` and thus we will lose some type information?

billonahill: yes, but also if we had K, V users could write this:
```
  class MyStatefulBolt implements ITypedStatefulComponent<String, Integer> {
    @Override
    public void initState(TypedState<String, Integer> state) {

    }
  }
```
and work with the types that it expects, just like they would while working with a typed map like `Map<String, Integer>` (instead of `Map<Serializable, Serializable>`. This provides compile time checks in user code for their types and removed the runtime need to cast objects, which leads to more bug-free and easy to maintain code.

I would need to prototype though to see that we could in fact create an instance of `TypedState<String, Integer>` at runtime though, give persisted `Serializables` and an instance of `MyStatefulBolt<String, Integer>` to derive types from.

objmagic: @billonahill 
> This provides compile time checks in user code for their types

Sorry for bugging in this thread, but can you be more specific? Do you mean with K, V we can also provide compile-time type checks to make sure **both** that `String` is extending `Serializable` **and** that `K` is extending `String`?

billonahill: Correct. Here's some mock code to show what I mean. Note that the internals use serializable but the user land code can be typed.
```
public class StateExample {

  interface TypedState<K extends Serializable, V extends Serializable> {
    void put(K key, V value);
    V get(K key);
    Set<K> keySet();
  }

  interface ITypedStatefulComponent<K extends Serializable, V extends Serializable> {
    void initState(TypedState<K, V> state);
  }

  static class MyStateImpl implements TypedState {
    private Map<Serializable, Serializable> map = new HashMap<>();

    @Override
    public void put(Serializable key, Serializable value) {
      map.put(key, value);
    }

    @Override
    public Serializable get(Serializable key) {
      return map.get(key);
    }

    @Override
    public Set keySet() {
      return map.keySet();
    }
  }

  static class MyStatefulBolt implements ITypedStatefulComponent<String, Integer> {
    @Override
    public void initState(TypedState<String, Integer> state) {
      for (String key : state.keySet()) {
        System.out.printf("key: %s, value %d", key, state.get(key));
      }
    }
  }

  static TypedState loadStateFromDisk() {
    Serializable key = "String";
    Serializable value = 2;

    TypedState state = new MyStateImpl();
    state.put(key, value);
    return state;
  }

  public static void main(String[] args) {
    MyStatefulBolt bolt = new MyStatefulBolt();
    bolt.initState(loadStateFromDisk());
  }
}
```
```

**Comment by srkukarni** [comment]
I've removed IStatefulSpout/Bolt interfaces as requested. Also made State typed.
Please take a look again. Thanks!

**Comment by maosongfu** [comment]
LGTM.

**Comment by kramasamy** [comment]
@maosongfu @billonahill - is this ready to be merged?

**Comment by maosongfu** [comment]
@kramasamy This Pull Request looks good tome. Pending for @billonahill to see whether it contains all changes he requested.

**Comment by kramasamy** [comment]
@maosongfu - sounds good!

File: heron/api/src/java/com/twitter/heron/api/Config.java (hunk 1 above)
Comment (billonahill) [inline heron/api/src/java/com/twitter/heron/api/Config.java]: Can we name this in a way that implies that heron state management is enabled or disabled by this boolean flag? It should be very clear to the user the effect of using `IStatefulComponents` for example with this set to `false`. Or what this should be set to for an externally stateful topology that doesn't use `IStatefulComponents`. As named it sounds like it describes a characteristic of the topology, versus the behavior of the system.

Comment (srkukarni) [inline heron/api/src/java/com/twitter/heron/api/Config.java]: Changed to TOPOLOGY_STATEFUL_ENABLED

File: heron/api/src/java/com/twitter/heron/api/Config.java [hunk 10]
```diff
@@ -147,6 +147,31 @@
    */
   public static final String TOPOLOGY_COMPONENT_RAMMAP = "topology.component.rammap";
   /**
+   * Is this topology stateful? The format of this flag is boolean
```
billonahill: Instead of "Is this topology stateful?" can we clarify in the comments what happens when this is set to true/false? For example if I set it to false but use `IStatefulComponent`, I assume I can still be in-memory stateful, but the state will not persist upon instance failures, scaling, restarts, etc. since the framework will not manage the state transfer/persistence (even if I configure a non-memory state provider below). Would setting this to false with a persistent state provider configured be the equivalent to specifying this to be true while configuring an in-memory HashMap provider?

srkukarni: Added more comments to clarify

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -147,6 +147,42 @@
...
+   * When set to true, Heron will try to do a distributed checkpoint
+   * based on Lamport's idea of distributed snapshotting every
+   * TOPOLOGY_STATEFUL_CHECKPOINT_INTERVAL_SECONDS seconds. When any component
+   * of the topology dies, Heron initiates a recovery mechanism to restart
+   * the topology from the last globally consistent checkpoint.
+   * Submitting this flag to false does not initiate any checkpointing/recovery
+   * mechanism, and thus the topology is then run in either atmost/atleast once
+   * semantics as before. This means that even if the components themselves are
+   * IStatefulComponents their state is not saved/recovered. This allows topology
+   * developers to run the same topology in exacltly once or atleast/atmost once
```
billonahill: we're combing the behavior of state management with atleast/atmost/exactly once semantics, so I think more explanation is helpful. For exactly once semantics do we require acking to be enabled and topology.stateful.enabled, for example? We should clarify then what happens w.r.t. atleast/atmost/exactly once when these two settings are true/true, true/false and false/false.

also typo: exacltly

srkukarni: Thats a good point. I've separated out Stateful processing and exactly once semantics.

billonahill:
👍 looks good!

PR #1824, opened 2017-04-28
rename heron/common/src/cpp/config/config.h

kramasamy:
So that it does not clash with other include files in the upcoming PRs.


nlu90:
Is this config.h file included anywhere? Those places should also be updated.

kramasamy:
@nlu90 - it is not being include anywhere. all the compilation were successful and so did tests. It is a file that includes all the files in that directory.

huijunw:
i guess @nlu90 means https://github.com/twitter/heron/blob/master/heron/common/src/cpp/config/BUILD#L22

PR #1827, opened 2017-05-02
check server start() return value in stmgr unit test

huijunwu:
check if the server starts successfully

### DIFF
--- a/heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
+++ b/heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
@@ -228,7 +228,7 @@ void StartDummyStMgr(EventLoopImpl*& ss, DummyStMgr*& mgr, std::thread*& stmgr_t
 
   mgr = new DummyStMgr(ss, options, stmgr_id, LOCALHOST, stmgr_port, LOCALHOST, tmaster_port,
                        shell_port, instances);
-  mgr->Start();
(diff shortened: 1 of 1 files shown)


heron/stmgr/tests/cpp/server/stmgr_unittest.cpp
```diff
@@ -228,7 +228,7 @@ void StartDummyStMgr(EventLoopImpl*& ss, DummyStMgr*& mgr, std::thread*& stmgr_t
 
   mgr = new DummyStMgr(ss, options, stmgr_id, LOCALHOST, stmgr_port, LOCALHOST, tmaster_port,
                        shell_port, instances);
-  mgr->Start();
+  EXPECT_EQ(0, mgr->Start()) << stmgr_port;
```
kramasamy: What does << stmgr_port means? Does it just print stmgr_port when the test fails or succeeds?

huijunw: When the check fails, it prints stmgr_port.
The ci seems always failing after adding the checking. I doubt if this test ever run correctly.

kramasamy: @huijunw - can you add some meaningful message so that it says what it is printing?

huijunw: done.

kramasamy:
👍

PR #1828, opened 2017-05-02
return errno instead of -1

huijunwu:
return errno instead of -1, so that the test can tell why the start() fails

### DIFF
--- a/heron/common/src/cpp/network/baseserver.cpp
+++ b/heron/common/src/cpp/network/baseserver.cpp
@@ -101,7 +101,7 @@ sp_int32 BaseServer::Start_Base() {
 
   // Ask the EventLoop to deliver any read events
   if (eventLoop_->registerForRead(listen_fd_, on_new_connection_callback_, true) < 0) {
-    PLOG(ERROR) << "register for read of the socket failed in server";
+    LOG(ERROR) << "register for read of the socket failed in server";
     close(listen_fd_);
     return -1;
   }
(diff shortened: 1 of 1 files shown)


heron/common/src/cpp/network/baseserver.cpp
```diff
@@ -87,21 +87,21 @@ sp_int32 BaseServer::Start_Base() {
...
+    PLOG(ERROR) << "listen of a socket failed in server";
     close(listen_fd_);
-    return -1;
+    return errno;
   }
 
   // Ask the EventLoop to deliver any read events
   if (eventLoop_->registerForRead(listen_fd_, on_new_connection_callback_, true) < 0) {
-    LOG(ERROR) << "register for read of the socket failed in server\n";
+    LOG(ERROR) << "register for read of the socket failed in server";
```
objmagic: PLOG?

heron/common/src/cpp/network/baseserver.cpp
```diff
@@ -49,8 +49,8 @@ sp_int32 BaseServer::Start_Base() {
   errno = 0;
   listen_fd_ = socket(options_.get_socket_family(), SOCK_STREAM, 0);
   if (listen_fd_ < 0) {
-    LOG(ERROR) << "Opening of a socket failed in server " << errno << "\n";
-    return -1;
+    PLOG(ERROR) << "Opening of a socket failed in server";
+    return errno;
```
kramasamy: Are errno's negative? Otherwise it might not work. Instead I would suggest to return NOT_OK.

kramasamy: It should SP_NOTOK. Here is the file common/src/cpp/basics/sprcodes.h

huijunw:
lets keep what it is

PR #1829, opened 2017-05-02
Added proto files declaring all the messages transmitted across

srkukarni:
all different heron components for stateful processing and
exactly once semantics

### DIFF
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -5,12 +5,15 @@ option java_outer_classname = "CheckpointManager";
 import "common.proto";
 import "physical_plan.proto";
 
+// Message representing information about a globally consistent
+// checkpoint
+message StatefulConsistentCheckpoint {
+  required string checkpoint_id = 1;
+  // One can add more meta data about this ckpt later
+}
+
 // message stored in the state manager by the tmaster
 message StatefulConsistentCheckpoints {
-  message StatefulConsistentCheckpoint {
-    required string checkpoint_id = 1;
-    // One can add more meta data about this ckpt later
-  }
   // An ordered list of the globally consistent checkpoints
   // that have been snapshotted and can be recovered from.
   // 0th element is the most recent one, 1st is the next recent
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -141,7 +141,7 @@ message StartStmgrStatefulProcessing {
 // This is the message that a tmaster sends
 // when it wants to register itself
 // with checkpoint manager
-message RegisterTMasterRequest{
+message RegisterTMasterRequest {
   required string topology_name = 1;
   required string topology_id = 2;
 }
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -5,15 +5,17 @@ option java_outer_classname = "CheckpointManager";
 import "common.proto";
 import "physical_plan.proto";
 
-// message stored in zk by the tmaster
+// message stored in the state manager by the tmaster
 message StatefulConsistentCheckpoints {
-  // This is the most recent checkpoint that we believe exists
-  required string most_recent_checkpoint_id = 1;
-  // It might be possible that for various reasons(fs corruption, etc)
-  // the last checkpoint is not accessible. This is the set of
-  // other globally consistent checkpoints that can be recovered from.
-  // The order of preference(and recency) is from beginning to end
-  repeated string backup_checkpoint_ids = 2;
+  message StatefulConsistentCheckpoint {
+    required string checkpoint_id = 1;
+    // One can add more meta data about this ckpt later
+  }
+  // An ordered list of the globally consistent checkpoints
+  // that have been snapshotted and can be recovered from.
+  // 0th element is the most recent one, 1st is the next recent
+  // and so on
+  repeated StatefulConsistentCheckpoint consistent_checkpoints = 1;
 }
 
 // tmaster -> stmgr messages
@@ -27,6 +29,11 @@ message StartStatefulCheckpoint {
 // to this checkpoint
 message RestoreTopologyStateRequest {
   required string checkpoint_id = 1;
+  // Every Attempt made by Tmaster to restore to a globally consistent
+  // checkpoint is tagged with a unique id. The stmgrs include this in their
+  // RestoreTopologyStateResponse below. This way tmaster could
+  // try restoring to the same checkpoint_id multiple times and still
+  // keep track of for which attempt the restore was in the responses
   required int64 restore_txid = 2; 
 }
 
@@ -55,6 +62,9 @@ message InstanceStateStored {
 }
 
 // Message sent by tmaster to stmgr to start processing
+// For exactly once semantics, all the topology components
+// should be rolled back to a consistent state before they can
+// start processing.
 message StartStmgrStatefulProcessing {
   required string checkpoint_id = 1;
 }
@@ -87,6 +97,7 @@ message CleanStatefulCheckpointRequest {
 // Message sent by ckptmgr to tmaster about the cleanup request
 message CleanStatefulCheckpointResponse {
   required heron.proto.system.Status status = 1;
+  repeated string cleaned_checkpoints = 2;
 }
 
 // stmgr -> ckptmgr messages
@@ -97,7 +108,7 @@ message CleanStatefulCheckpointResponse {
 message RegisterStMgrRequest{
   required string topology_name = 1;
   required string topology_id = 2;
-  required string stmgr = 3;
+  required string stmgr_id = 3;
 }
 
 // This is the message that checkpoint manager
@@ -163,18 +174,22 @@ message RestoreInstanceStateRequest {
 // This is the message that instance sends to its stmgr
 // when done with restoring the state
 message RestoreInstanceStateResponse {
-  required string checkpoint_id = 1;
+  required heron.proto.system.Status status = 1;
+  required string checkpoint_id = 2;
 }
 
 // Message sent by stmgr to Instance to start processing
+// This is related to StartStmgrStatefulProcessing message above
+// but between stmgr -> instance.
 message StartInstanceStatefulProcessing {
   required string checkpoint_id = 1;
 }
 
 // stmgr -> stmgr messages
 
-// This is the message that stmgr sends to other stmgr
-// to send it to downstream task
+// Once an instance is done checkpointing, this message is sent
+// to all its downstream components so that they could checkpoint
+// their state. In that sense this is the marker message
 message DownstreamStatefulCheckpoint {
   required int32 origin_task_id = 1; 
   required int32 destination_task_id = 2; 
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -25,7 +25,7 @@ import "physical_plan.proto";
 // From the StatefulConsistentCheckpoints message stored in the
 // state manager, Tmaster picks the latest checkpoint_id and
 // sends RestoreTopologyStateRequest to all stmgrs. Each stmgr
-// sends does a GetInstanceStateRequest/GetInstanceStateResponse
+// does a GetInstanceStateRequest/GetInstanceStateResponse
 // dance with its local checkpoint manager for all of its local
 // instances to fetch the state. Upon fetch, stmgr does a
 // RestoreInstanceStateRequest/RestoreInstanceStateResponse
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -12,20 +12,16 @@ import "physical_plan.proto";
 // On Startup
 //   For topologies needing exact once, the startup sequence of
 // tmaster/stmgr/instance is a little different. Normally, upon
-// all stmgrs connecting to tmaster, tmaster computes pplan
-// and distributes to stmgrs, which inturn send it to all their
+// startup, all stmgrs connect to tmaster and then tmaster computes pplan
+// and distributes to stmgrs, which in turn send it to all their
 // local instances. As soon as instances get their pplan, they can
 // start processing(spouts emitting tuples, etc). However for
-// exactly once semantics, this is a little different. First
+// stateful topologies, this is a little different. First
 // in addition to the pplan, tmaster needs to send the last
 // consistent checkpoint id(if any) to recover from as well.
-// Secondly because instances are all busy recovering their state,
-// any tuples emitted before they recover tend to start causing
-// backpressure right away. To avoid this(as well as for having
-// the same mechanism during failure recovery), after all instances
-// recover their state, the tmaster sends out a Start processing
-// message and only then all the components start processing.
-// WRT messages below, the sequence is the following.
+// Secondly instances need to recover their prev state before
+// they can process any tuples.
+// WRT messages below, the startup sequence is the following.
 // From the StatefulConsistentCheckpoints message stored in the
 // state manager, Tmaster picks the latest checkpoint_id and
 // sends RestoreTopologyStateRequest to all stmgrs. Each stmgr
@@ -36,7 +32,7 @@ import "physical_plan.proto";
 // dance with the corresponding instances to restore the
 // state. Completion of which, stmgr sends out the 
 // RestoreTopologyStateResponse message to the tmaster.
-// Once the tmaster recieves this from all stmgrs, it sends
+// Once the tmaster receives this from all stmgrs, it sends
 // out the StartStmgrStatefulProcessing message to all
 // stmgrs which inturn send out StartInstanceStatefulProcessing
 // to all their local instances. This gets the topology
@@ -260,9 +256,9 @@ message StartInstanceStatefulProcessing {
 
 // stmgr -> stmgr messages
 
-// Once an instance is done checkpointing, this message is sent
-// to all its downstream components so that they could checkpoint
-// their state. In that sense this is the marker message
+// Once stmgr receives StoreInstanceStateCheckpoint from an instance,
+// it sends this message to all of that instance's 
+// downstream components' stmgrs to propagate the checkpoint marker
 message DownstreamStatefulCheckpoint {
   required int32 origin_task_id = 1; 
   required int32 destination_task_id = 2; 
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -125,7 +125,7 @@ message InstanceStateStored {
 }
 
 // Message sent by tmaster to stmgr to start processing
-// For exactly once semantics, all the topology components
+// For stateful processing, all the topology components
 // should be rolled back to a consistent state before they can
 // start processing.
 message StartStmgrStatefulProcessing {
@@ -166,8 +166,8 @@ message CleanStatefulCheckpointResponse {
 // stmgr -> ckptmgr messages
 
 // This is the message that a stmgr sends
-// when it wants to establish the connection
-// with checkpoint manager
+// when it wants to register itself with
+// the checkpoint manager
 message RegisterStMgrRequest {
   required string topology_name = 1;
   required string topology_id = 2;
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -5,10 +5,74 @@ option java_outer_classname = "CheckpointManager";
 import "common.proto";
 import "physical_plan.proto";
 
+// The messages in this file are for stateful processing and providing
+// exactly once semantics. Below are the various sequence of events that
+// happen in a few(non exhaustive) scenarios
+//
+// On Startup
+//   For topologies needing exact once, the startup sequence of
+// tmaster/stmgr/instance is a little different. Normally, upon
+// all stmgrs connecting to tmaster, tmaster computes pplan
+// and distributes to stmgrs, which inturn send it to all their
+// local instances. As soon as instances get their pplan, they can
+// start processing(spouts emitting tuples, etc). However for
+// exactly once semantics, this is a little different. First
+// in addition to the pplan, tmaster needs to send the last
+// consistent checkpoint id(if any) to recover from as well.
+// Secondly because instances are all busy recovering their state,
+// any tuples emitted before they recover tend to start causing
+// backpressure right away. To avoid this(as well as for having
+// the same mechanism during failure recovery), after all instances
+// recover their state, the tmaster sends out a Start processing
+// message and only then all the components start processing.
+// WRT messages below, the sequence is the following.
+// From the StatefulConsistentCheckpoints message stored in the
+// state manager, Tmaster picks the latest checkpoint_id and
+// sends RestoreTopologyStateRequest to all stmgrs. Each stmgr
+// sends does a GetInstanceStateRequest/GetInstanceStateResponse
+// dance with its local checkpoint manager for all of its local
+// instances to fetch the state. Upon fetch, stmgr does a
+// RestoreInstanceStateRequest/RestoreInstanceStateResponse
+// dance with the corresponding instances to restore the
+// state. Completion of which, stmgr sends out the 
+// RestoreTopologyStateResponse message to the tmaster.
+// Once the tmaster recieves this from all stmgrs, it sends
+// out the StartStmgrStatefulProcessing message to all
+// stmgrs which inturn send out StartInstanceStatefulProcessing
+// to all their local instances. This gets the topology
+// humming
+//
+// Periodic checkpoints
+//  Every so often(as dictated by TOPOLOGY_STATEFUL_CHECKPOINT_INTERVAL)
+// tmaster sends out StartStatefulCheckpoint message to all
+// stmgrs. Each stmgr sends out InitiateStatefulCheckpoint
+// message to each of its local spout instances. Because of the best effort
+// nature, this is delivered as a message. After instance does its
+// checkpoint, it sends StoreInstanceStateCheckpoint to stmgr to store
+// its checkpoint. Stmgr passes this to ckptmgr via
+// SaveInstanceStateRequest/SaveInstanceStateResponse dance with its
+// local checkpoint manager. Upon getting a successful
+// SaveInstanceStateResponse message, stmgr sends out a
+// InstanceStateStored message notifying tmaster that the
+// state of a particular instance has been saved. Parallelly
+// stmgr also sends out DownstreamStatefulCheckpoint message
+// to each of the downstream components to do the distributed
+// Lamport style checkpointing. Meanwhile at tmaster, when it
+// receives InstanceStateStored from all instances for a particular
+// checkpoint_id, it has reached globally consistent checkpoint
+// and it stores it in the state manager as StatefulConsistentCheckpoint
+//
+// Failure Recovery
+//  Either when stmgr dies, or when an instance dies(which is notified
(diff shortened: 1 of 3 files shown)


heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+// This is the message that stmgr sends to its instace
+// asking them to restore their state
+message RestoreInstanceStateRequest {
+  required InstanceStateCheckpoint state = 1;
+}
+
+// This is the message that instance sends to its stmgr
+// when done with restoring the state
+message RestoreInstanceStateResponse {
+  required string checkpoint_id = 1;
```
nlu90: add `required heron.proto.system.Status status`?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Message sent by stmgr to Instance to start processing
+message StartInstanceStatefulProcessing {
+  required string checkpoint_id = 1;
```
nlu90: the same question as `StartStmgrStatefulProcessing `

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// stmgr -> stmgr messages
+
+// This is the message that stmgr sends to other stmgr
+// to send it to downstream task
+message DownstreamStatefulCheckpoint {
```
nlu90: is this message to handle the case where a downstream instance has multiple updtream instances?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+option java_package = "com.twitter.heron.proto.ckptmgr";
+option java_outer_classname = "CheckpointManager";
+
+import "common.proto";
+import "physical_plan.proto";
+
+// message stored in zk by the tmaster
+message StatefulConsistentCheckpoints {
+  // This is the most recent checkpoint that we believe exists
+  required string most_recent_checkpoint_id = 1;
```
billonahill: Instead of storing an string id and a string[] backupId, could we instead store an ordered array of Checkpoint objects, most recent first? Using an object instead of a string would allow for extensibility to bind more metadata to a checkpoint than just id (e.g., creation time, location).  We could adjust how many checkpoints we keep via config (either time based or absolute count), which becomes the size of the array.

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+// Message sent to stmgrs by the tmaster to initiate checkpointing
+message StartStatefulCheckpoint {
+  required string checkpoint_id = 1;
+}
+
+// Message sent by tmaster to stmgr asking them to reset their instances
+// to this checkpoint
+message RestoreTopologyStateRequest {
+  required string checkpoint_id = 1;
+  required int64 restore_txid = 2;
```
srkukarni: Added more comments to explain this. Let me know if it makes more sense now :)

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+// Message sent by stmgr to tmaster informing it about
+// the fact that we stored a checkpoint belonging
+// to the instance
+message InstanceStateStored {
+  required string checkpoint_id = 1;
+  required heron.proto.system.Instance instance = 2;
+}
+
+// Message sent by tmaster to stmgr to start processing
+message StartStmgrStatefulProcessing {
```
srkukarni: Added comments.

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+// from tmaster
+message RegisterTMasterResponse {
+  required heron.proto.system.Status status = 1;
+}
+
+// Message sent by tmaster to ckptmgr to cleanup
+// old checkpoint state.
+message CleanStatefulCheckpointRequest {
+  // Any checkpoints older than this can be cleaned
+  optional string oldest_checkpoint_preserved = 1;
```
srkukarni: This is just the oldest checkpoint that needs to be preserved. All older ones can be cleaned

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+  optional bool clean_all_checkpoints = 2 [default = false];
+}
+
+// Message sent by ckptmgr to tmaster about the cleanup request
+message CleanStatefulCheckpointResponse {
+  required heron.proto.system.Status status = 1;
```
srkukarni: added

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// stmgr -> ckptmgr messages
+
+// This is the message that a stmgr sends
+// when it wants to establish the connection
+// with checkpoint manager
+message RegisterStMgrRequest{
+  required string topology_name = 1;
+  required string topology_id = 2;
+  required string stmgr = 3;
```
srkukarni: done

heron/proto/ckptmgr.proto
srkukarni: added

heron/proto/ckptmgr.proto
srkukarni: Explained earlier

heron/proto/ckptmgr.proto
srkukarni: This is the marker message. Added comments to explain

heron/proto/ckptmgr.proto
srkukarni: Made changes

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,182 @@
+package heron.proto.ckptmgr;
+option java_package = "com.twitter.heron.proto.ckptmgr";
+option java_outer_classname = "CheckpointManager";
+
+import "common.proto";
+import "physical_plan.proto";
+
+// message stored in zk by the tmaster
```
srkukarni: changed

nlu90:
LGTM

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,197 @@
+package heron.proto.ckptmgr;
+option java_package = "com.twitter.heron.proto.ckptmgr";
+option java_outer_classname = "CheckpointManager";
+
+import "common.proto";
+import "physical_plan.proto";
+
+// message stored in the state manager by the tmaster
+message StatefulConsistentCheckpoints {
+  message StatefulConsistentCheckpoint {
```
billonahill: Why an inner class? If we make this a top level class it can be used on its own without the collection, like in the RPC for example while working with a specific checkpoint.

srkukarni: Done

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+  // that have been snapshotted and can be recovered from.
+  // 0th element is the most recent one, 1st is the next recent
+  // and so on
+  repeated StatefulConsistentCheckpoint consistent_checkpoints = 1;
+}
+
+// tmaster -> stmgr messages
+
+// Message sent to stmgrs by the tmaster to initiate checkpointing
+message StartStatefulCheckpoint {
```
billonahill: If this is an RPC request object should it be `StartStatefulCheckpointRequest`?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+message RestoreTopologyStateResponse {
+  required heron.proto.system.Status status = 1;
+  required string checkpoint_id = 2;
+  required int64 restore_txid = 3; 
+}
+
+// Message sent by stmgr to tmaster asking it to reset the topology
+// to some valid checkpoint. This is sent either if stmgr dies
+// and comes back up or if an instance dies.
+message ResetTopologyState {
```
billonahill: Is this the payload of an RPC request? If so it should follow the same standard and be `ResetTopologyStateRequest`.

heron/proto/ckptmgr.proto
srkukarni: We do follow that terminology where all request/responses are labelled as such.
Anything else that is not a request/response are just best effort messages that go from one component to another. For some, a response is not expected. For others it cannot be a request because in Heron c++ client/server communication, only clients can send a request and a server cannot.

heron/proto/ckptmgr.proto
srkukarni: Same as above

heron/proto/ckptmgr.proto
billonahill: Do the heron client/server model use a protobuf RPC interface or do they just send protobuf object bytes on raw sockets?

srkukarni: the latter(i.e. use protobuf object bytes)

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+  optional string dead_stmgr = 1;
+  optional int32 dead_taskid = 2;
+}
+
+// Message sent by stmgr to tmaster informing it about
```
billonahill: Why does the tmaster need to be aware of specific instance state? Why not have the stream manager manage the states of the instances under it's control?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+message InstanceStateStored {
+  required string checkpoint_id = 1;
+  required heron.proto.system.Instance instance = 2;
+}
+
+// Message sent by tmaster to stmgr to start processing
+// For exactly once semantics, all the topology components
+// should be rolled back to a consistent state before they can
+// start processing.
+message StartStmgrStatefulProcessing {
```
billonahill: When/how is this message used and how does the stream manager get into the state that is before or after this one (i.e., one where it's not started)?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+  required string checkpoint_id = 1;
+}
+
+// tmaster -> ckptmgr messages
+
+// This is the message that a tmaster sends
+// when it wants to establish the connection
+// with checkpoint manager
+message RegisterTMasterRequest{
```
billonahill: Why does TMaster register itself with the checkpoint manager instead of the other way around, like the way state manager registers with TMaster? TMaster is already discoverable in StateManager.

Also space before {

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+  required string topology_name = 1;
+  required string topology_id = 2;
```
billonahill: Why would checkpoint manager need to be told the topology name and id? This info is static and can be passed to checkpoint manager at startup. Same question for `RegisterStMgrRequest` below.

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+message RegisterTMasterResponse {
+  required heron.proto.system.Status status = 1;
+}
+
+// Message sent by tmaster to ckptmgr to cleanup
+// old checkpoint state.
+message CleanStatefulCheckpointRequest {
+  // Any checkpoints older than this can be cleaned
+  optional string oldest_checkpoint_preserved = 1;
+  optional bool clean_all_checkpoints = 2 [default = false];
```
billonahill: If these fields are both optional neither of them can be set, which I assume would be invalid. If that's the case can we enforce that one of them is set view `oneof`? That would assure that every possible field setting is valid.
https://developers.google.com/protocol-buffers/docs/proto#oneof

What's the expected behavior when `oldest_checkpoint_preserved` is unset and `clean_all_checkpoints=false`?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Message sent by ckptmgr to tmaster about the cleanup request
+message CleanStatefulCheckpointResponse {
+  required heron.proto.system.Status status = 1;
+  repeated string cleaned_checkpoints = 2;
```
billonahill: `cleaned_checkpoint_ids`

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// stmgr -> ckptmgr messages
+
+// This is the message that a stmgr sends
+// when it wants to establish the connection
```
billonahill: "establish the connection" should be "register itself" no?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+// with checkpoint manager
+message RegisterStMgrRequest{
```
billonahill: space before {

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+
+// stmgr -> Instance messages
+
+// This is the message that the stmgr sends to its
+// local tasks to begin initiating checkpoint
+message InitiateStatefulCheckpoint {
+  required string checkpoint_id = 1;
+}
+
+// This is the message that the instance sends
```
billonahill: Is this the message or is this the state info that's part of another message? Reading above it seems the latter but the description states the former. If it is the former, can we put a verb in this object, to make it more clear what action is being requested?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+// when it wants stmgr to store its state
+message InstanceStateCheckpoint {
+  required string checkpoint_id = 1;
+  required bytes state = 2;
+}
+
+// This is the message that stmgr sends to its instace
```
billonahill: typo: instace

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+// when done with restoring the state
+message RestoreInstanceStateResponse {
+  required heron.proto.system.Status status = 1;
+  required string checkpoint_id = 2;
+}
+
+// Message sent by stmgr to Instance to start processing
+// This is related to StartStmgrStatefulProcessing message above
+// but between stmgr -> instance.
+message StartInstanceStatefulProcessing {
```
billonahill: Same question as above w.r.t. state changes. If this request triggers processing to start, how does the instance get into the state where it's not processing?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+  required string checkpoint_id = 1;
+}
+
+// stmgr -> stmgr messages
+
+// Once an instance is done checkpointing, this message is sent
```
billonahill: I thought checkpointing was non-blocking, but this comment sounds like checkpointing happens serially from one component to the next. Why?

Also why does each instance broadcast this message to all instances of the downstream component?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
...
+// local tasks to begin initiating checkpoint
+message InitiateStatefulCheckpoint {
+  required string checkpoint_id = 1;
+}
+
+// This is the message that the instance sends
+// when it wants stmgr to store its state
+message InstanceStateCheckpoint {
+  required string checkpoint_id = 1;
+  required bytes state = 2;
```
billonahill: Can we more clearly specify state than an arbitrary array of bytes? What is this byte array and how does the StateStore implementor know how to read the bytes? How for example would I persist the byte array in a key value store in a way where I could inspect individual records?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,200 @@
+package heron.proto.ckptmgr;
+option java_package = "com.twitter.heron.proto.ckptmgr";
+option java_outer_classname = "CheckpointManager";
+
+import "common.proto";
+import "physical_plan.proto";
+
+// Message representing information about a globally consistent
+// checkpoint
+message StatefulConsistentCheckpoint {
```
billonahill: We need to associate the checkpoint with a packing plan or physical plan so we know it's valid to reuse upon topology restart.

heron/proto/ckptmgr.proto
srkukarni: This is purely for information sake only. Mostly used for printing/debugging

heron/proto/ckptmgr.proto
srkukarni: The idea is to use ckptmgr as some sort of service by the tmaster. TMaster upon startup would connect to it(which is also running in the same container) and register itself

heron/proto/ckptmgr.proto
srkukarni: This is for the sake of consistency wrt all other register requests. The original idea was sanity checking

heron/proto/ckptmgr.proto
srkukarni: Current phase only treats it as arbitrary bytes. The aim is to get the flow mechanism correct, before expanding this

heron/proto/ckptmgr.proto
srkukarni: distributed checkpointing is non-blocking. However the mechanism of how they are triggered are serialized. This is part of the Lamport's algorithm. To be part of a consistent checkpoint, an upstreamer needs to checkpoint first and then send these markers to all potential downstream components. Downstream components can checkpoint only after they received all their upstream components.

heron/proto/ckptmgr.proto
srkukarni: oneof is not supported by the protobuf version that we are using

heron/proto/ckptmgr.proto
srkukarni: Rather than do this in the design doc(which is seperate entity) I would rather include some sort of flow in the file itself

srkukarni: This is done. Please take a look

heron/proto/ckptmgr.proto
srkukarni: Added more in the header section

heron/proto/ckptmgr.proto
srkukarni: Done style fix

heron/proto/ckptmgr.proto
srkukarni: done

heron/proto/ckptmgr.proto
srkukarni: changed

heron/proto/ckptmgr.proto
srkukarni: fixed

heron/proto/ckptmgr.proto
srkukarni: Fixed this

heron/proto/ckptmgr.proto
srkukarni: Fixed

heron/proto/ckptmgr.proto
billonahill: Why is this at the instance level and not the component level? Couldn't "Once an instance is done checkpointing" be "Once a component is done checkpointing"? Also, if each instance is broadcasting to all of it's downstream instances, that implies that the downstream instances need to be aware of and track the id set of all of its' upstream instances, and which ones have sent the message? Pushing this responsibility onto each instance seems unnecessary. Instead could TMaster message all instances of Component B (via stream managers) that it's time to checkpoint, based on it's knowledge that all Component A instances are complete?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
+package heron.proto.ckptmgr;
+option java_package = "com.twitter.heron.proto.ckptmgr";
+option java_outer_classname = "CheckpointManager";
+
+import "common.proto";
+import "physical_plan.proto";
+
+// The messages in this file are for stateful processing and providing
```
billonahill: This text description is helpful, but it would still be great to have a sequence diagram of these flows. Visualizing the interactions between the many agents really helps to understand the system and to expose quirks. Here's an example that we created to clearly illustrate the Topology Submit Sequence (which is less complicated than these flow IMO), for example:
https://twitter.github.io/heron/docs/concepts/architecture/

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// exactly once semantics. Below are the various sequence of events that
+// happen in a few(non exhaustive) scenarios
+//
+// On Startup
+//   For topologies needing exact once, the startup sequence of
+// tmaster/stmgr/instance is a little different. Normally, upon
```
billonahill: upon startup

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// all stmgrs connecting to tmaster, tmaster computes pplan
```
billonahill: all stmgrs connect to

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// and distributes to stmgrs, which inturn send it to all their
```
billonahill: in turn

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// local instances. As soon as instances get their pplan, they can
+// start processing(spouts emitting tuples, etc). However for
+// exactly once semantics, this is a little different. First
```
billonahill: s/exactly once semantics/stateful topologies/

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// in addition to the pplan, tmaster needs to send the last
+// consistent checkpoint id(if any) to recover from as well.
+// Secondly because instances are all busy recovering their state,
```
billonahill: "because instances are all busy recovering their state" assumes that the reader knows the startup sequence of events, which is what we're trying to describe. Could you list the sequence of startup events for stateful topologies? It should be very clear to the reader the order of things like fetching the packing plan, initializing state, registering with stream managers, physical plan generation, etc.

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// state manager, Tmaster picks the latest checkpoint_id and
+// sends RestoreTopologyStateRequest to all stmgrs. Each stmgr
+// sends does a GetInstanceStateRequest/GetInstanceStateResponse
+// dance with its local checkpoint manager for all of its local
+// instances to fetch the state. Upon fetch, stmgr does a
+// RestoreInstanceStateRequest/RestoreInstanceStateResponse
+// dance with the corresponding instances to restore the
+// state. Completion of which, stmgr sends out the 
+// RestoreTopologyStateResponse message to the tmaster.
+// Once the tmaster recieves this from all stmgrs, it sends
```
billonahill: receives

heron/proto/ckptmgr.proto
srkukarni: Couple of clarifications:-
1. Just to be clear in Heron parlance component means one logical component(like spout/bolt). Instance is one running process of it. Thus there are 5 instances of 'wordSpout' component. In this case, it is instance level.
2. Chandy Lamport's algorithm necessitates that downstream instances are aware of all possible upstreamers. In practice this is performed at the stmgr so the real instance code is not aware of it. We cannot involve tmaster here because then a) it is not decentralized and b) only local state inside stmgr is sufficient

heron/proto/ckptmgr.proto
billonahill: Two questions:
1. Could deactivate/activate be used for this, or would that overload the activation semantics?
2. This message would be sent when recovering to a given checkpoint, during which processing is already started, so this message would also be used to reset a started topology to a given checkpoint, correct?

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,270 @@
...
+// Message sent by stmgr to tmaster informing it about
+// the fact that we stored a checkpoint belonging
+// to the instance
+message InstanceStateStored {
+  required string checkpoint_id = 1;
+  required heron.proto.system.Instance instance = 2;
+}
+
+// Message sent by tmaster to stmgr to start processing
+// For exactly once semantics, all the topology components
```
billonahill: s/exactly once semantics/stateful processing/

heron/proto/ckptmgr.proto
srkukarni: 1. activate/deactivate are meant for totally different things and imo not applicable.
2. yup. In future we could add some commands like that to the cli itself

heron/proto/ckptmgr.proto
srkukarni: fixed

heron/proto/ckptmgr.proto
srkukarni: fixed

heron/proto/ckptmgr.proto
srkukarni: reworded

heron/proto/ckptmgr.proto
srkukarni: fixed

heron/proto/ckptmgr.proto
nlu90: The instance itself doesn't need to be aware of the upstream/downstream task infromations, stmgr can manage this and decide when to ask a local instance to do the checkpoint by sending `message InitiateStatefulCheckpoint` I suppose.

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,266 @@
...
+// stateful topologies, this is a little different. First
+// in addition to the pplan, tmaster needs to send the last
+// consistent checkpoint id(if any) to recover from as well.
+// Secondly instances need to recover their prev state before
+// they can process any tuples.
+// WRT messages below, the startup sequence is the following.
+// From the StatefulConsistentCheckpoints message stored in the
+// state manager, Tmaster picks the latest checkpoint_id and
+// sends RestoreTopologyStateRequest to all stmgrs. Each stmgr
+// sends does a GetInstanceStateRequest/GetInstanceStateResponse
```
billonahill: omit: sends

srkukarni: removed

heron/proto/ckptmgr.proto
```diff
@@ -0,0 +1,266 @@
...
+// local tasks to begin initiating checkpoint
+message InitiateStatefulCheckpoint {
+  required string checkpoint_id = 1;
+}
+
+// This proto encapsulates the info associated with
+// state of an instance
+message InstanceStateCheckpoint {
+  required string checkpoint_id = 1;
+  required bytes state = 2;
```
billonahill: Why is state a byte array and not something more typed, like a map of kv pairs? How does the `StatefulStorage` impl know how to read the bytes?

heron/proto/ckptmgr.proto
billonahill: I think your comment was about my request for a sequence diagram. What about my comment about how we need to bind a checkpoint to a given packing plan or physical plan to know if it's valid upon startup?

heron/proto/ckptmgr.proto
billonahill: It's  listed in the version 2 spec:
https://developers.google.com/protocol-buffers/docs/reference/proto2-spec

What's the minimum version that supports it? We should consider upgrading, since this was included in 2.x.

heron/proto/ckptmgr.proto
billonahill: Can you add a TODO next to that then to clarify that we don't want to ship with this being a byte array?

PR #1839, opened 2017-05-05
Fail Heron-instance fast for errors

maosongfu:
Currently heron-instance will try best-effort to clean the resources,
when uncaught exceptions happen. However, if it meets an error, for instance,
"java.lang.OutOfMemoryError: Java heap space", any consequent operations are not guaranteed to be done,
and it can lead the process to a bad state, for instance, hanging forever.

This pull request fixes it by making heron-instance fail fast for errors. And this issue is raised by ADs team from Twitter.

Tested with LocalScheduler.

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
@@ -233,6 +233,15 @@ public class HeronInstance {
 
     // The actual uncaught exceptions handing logic
     private void handleException(Thread thread, Throwable exception) {
+      // We would fail fast when errors occur to avoid unexpected bad situations
+      if (exception instanceof Error) {
+        LOG.log(Level.SEVERE,
+            "Error caught in thread: " + thread.getName()
+                + " with thread id: " + thread.getId() + ". Process halting...",
+            exception);
+        Runtime.getRuntime().halt(1);
+      }
+
       LOG.log(Level.SEVERE,
           String.format("Exception caught in thread: %s with id: %d",
               thread.getName(), thread.getId()), exception);


kramasamy:
@maosongfu - this might be ok for unrecoverable exceptions. But for recoverable exceptions, you should not die, just like that. How can you distinguish between the two?

objmagic:
Shall we also update comments [here](https://github.com/twitter/heron/pull/1839/files#diff-cf2b0320c00a0fc85f039efe83786ecfR215)?

maosongfu:
@kramasamy It will fail fast for error(https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html). In JAVA, error is different from exception; errors means unrecoverable things.

maosongfu:
@objmagic What comments you think to update?

objmagic:
@maosongfu add something like "will fail fast if `Error` is encountered" at [here](https://github.com/twitter/heron/pull/1839/files#diff-cf2b0320c00a0fc85f039efe83786ecfR215)

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
```diff
@@ -233,6 +233,15 @@ public void uncaughtException(Thread thread, Throwable exception) {
 
     // The actual uncaught exceptions handing logic
     private void handleException(Thread thread, Throwable exception) {
+      // We would fail fast when errors occur to avoid unexpected bad situations
+      if (exception instanceof Error) {
+        LOG.log(Level.SEVERE,
+            "Error caught in thread: " + thread.getName()
+                + " with thread id: " + thread.getId() + ". Process halting...",
+            exception);
+        Runtime.getRuntime().halt(1);
```
huijunw: i would suggest throw it so that all `halt()` in one same place at line 230

maosongfu: I would prefer to halt() directly here, since:
1. It is more clear to know the process will fail fast without tracing other parts of the code.
2. It reduces extra operations, which are not guaranteed to execute during Error. For instance, catching exceptions will use memory, and it might not work if we encounter `java.lang.OutOfMemoryError: Java heap space`

maosongfu:
@objmagic I think it is same as the comment I did on Ln236?

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
huijunw: ok

objmagic:
@maosongfu oh ok then.

Issue #1653, opened 2017-01-02
Stream manager crash in stdlib/strtol_l.c


ajorgensen:
We noticed the stream manager continually crashing in 2 out of 11 containers running in aurora. This specific topology is running on heron version 0.14.4.

I don't know gdb very well but I was able to get the following information from the core dump:
```Program terminated with signal SIGABRT, Aborted.
#0  0x00007f302e629c37 in __GI_____strtoul_l_internal (nptr=0x7fff1cf06d20 "", endptr=0x5807b0, base=10, group=<optimized out>, loc=<optimized out>) at ../stdlib/strtol_l.c:365
365     ../stdlib/strtol_l.c: No such file or directory.
[Current thread is 1 (LWP 100082)]
(gdb) bt
#0  0x00007f302e629c37 in __GI_____strtoul_l_internal (nptr=0x7fff1cf06d20 "", endptr=0x5807b0, base=10, group=<optimized out>, loc=<optimized out>) at ../stdlib/strtol_l.c:365
#1  0x0000000000000000 in ?? ()
(gdb) where
#0  0x00007f302e629c37 in __GI_____strtoul_l_internal (nptr=0x7fff1cf06d20 "", endptr=0x5807b0, base=10, group=<optimized out>, loc=<optimized out>) at ../stdlib/strtol_l.c:365
#1  0x0000000000000000 in ?? ()
(gdb) thread apply all bt

Thread 3 (LWP 100084):
#0  0x00007f302ede7404 in __pthread_mutex_unlock_usercnt (mutex=0x1d, decr=<optimized out>) at pthread_mutex_unlock.c:55
#1  0x00007f302cb599c0 in ?? ()
#2  0x00007f302cb59700 in ?? ()
#3  0x25fc65e0e2e8d055 in ?? ()
#4  0x25fc61379db2d055 in ?? ()
#5  0x0000000000000000 in ?? ()

Thread 2 (LWP 100083):
#0  0x00007f302e6dffdd in parse_param (quoted=0, ifs_white=0x0, ifs=0xdb9c3c8bff28d055 <error: Cannot access memory at address 0xdb9c3c8bff28d055>, pwordexp=0x7f302d35a700, flags=0, offset=<optimized out>, words=<optimized out>,
    max_length=0x20000000e, word_length=<optimized out>, word=0x486656 <do_io+294>) at wordexp.c:1628
[...]

PR #1840, opened 2017-05-05
Stream subprocess stdout and stderr in heron-executor

ajorgensen:
The logs for subprocesses launched by the heron-executor were
not actually streamed to the heron-executor log file. Using
`communicate` to read the output of a subprocess will wait until the
process is closed before actually reading the logs. In the non-blocking
case the stdout and stderr were sent to a pipe but were not actually
read from which results in a deadlock if the subprocess logs enough
information into the stderr buffer
([python docs on subprocess and
pipes](https://docs.python.org/2/library/subprocess.html#subprocess.check_call)).

> Do not use stdout=PIPE or stderr=PIPE with this function as that can
deadlock based on the child process output volume.

Heron uses glog to log messages to a file however zookeeper defaults to
using stderr with no easy way to tell it to use glog instead. As such
whenever zookeeper logged a message it would get written to the
subprocess stderr buffer and fill it up ultimately causing a deadlock in
the zookeeper thread in the tmaster trying to write to stderr. This then
manifests itself by the zookeeper client not being able to heartbeat and
having its session expire without being able to callback and restart the
process.

With `subprocess.communicate` the data is [buffered in
memory](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.communicate) until end-of-line is reached
and the subprocess is terminated so it is not suitable for long running processes.

One design consideration here for simplicity is the redirection of the
subprocess stderr to stdout. This is to make reading the data easier.
The downside here is that it is not possible to tell if a log message
came from stdout or stderr but I think that information will be clear
with the content of the message. Also some dependencies like zookeeper
log all of their messages on stderr anyway regardless of the message
type.

Fixes https://github.com/twitter/heron/issues/1653

### DIFF
--- a/heron/common/src/python/utils/log.py
+++ b/heron/common/src/python/utils/log.py
@@ -44,7 +44,10 @@ def stream_process_stdout(process, log_fn):
   :param log_fn: a function that will be called for each log line
   :return: None
   """
-  for line in iter(process.stdout.readline, b''):
+  while 1:
+    line = process.stdout.readline()
+    if not line:
+      break
     log_fn(line)
 
 def configure(level=logging.INFO, logfile=None):
--- a/heron/common/src/python/utils/log.py
+++ b/heron/common/src/python/utils/log.py
@@ -34,6 +34,10 @@ def async_stream_process_stdout(process, log_fn):
   :return: None
   """
   logging_thread = Thread(target=stream_process_stdout, args=(process, log_fn, ))
+
+  # Setting the logging thread as a daemon thread will allow it to exit with the program
+  # rather than blocking the exit waiting for it to be handled manually.
+  logging_thread.daemon = True
   logging_thread.start()
 
   return logging_thread
--- a/heron/common/src/python/utils/log.py
+++ b/heron/common/src/python/utils/log.py
@@ -15,6 +15,7 @@
 import logging
 from logging.handlers import RotatingFileHandler
 import colorlog
+from threading import Thread
 
 # Create the logger
 # pylint: disable=invalid-name
@@ -25,6 +26,26 @@ Log = logging.getLogger()
 # e.g. "08/16/1988 21:30:00 +1030"
 # see time formatter documentation for more
 date_format = "%Y-%m-%d %H:%M:%S %z"
+
+def async_stream_process_stdout(process, log_fn):
+  """ Stream the stdout and stderr for a process out to display async
+  :param process: the process to stream the log for
+  :param log_fn: a function that will be called for each log line
+  :return: None
+  """
+  logging_thread = Thread(target=stream_process_stdout, args=(process, log_fn, ))
+  logging_thread.start()
+
+  return logging_thread
+
+def stream_process_stdout(process, log_fn):
+  """ Stream the stdout and stderr for a process out to display
+  :param process: the process to stream the logs for
+  :param log_fn: a function that will be called for each log line
+  :return: None
+  """
+  for line in iter(process.stdout.readline, b''):
+    log_fn(line)
 
 def configure(level=logging.INFO, logfile=None):
   """ Configure logger which dumps log on terminal
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -30,6 +30,7 @@ import threading
 import time
 import yaml
 import socket
+import traceback
 
 from functools import partial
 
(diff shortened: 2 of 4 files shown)


heron/executor/src/python/heron_executor.py
```diff
@@ -610,24 +619,37 @@ def _untar_if_tar(self):
...
 
   def _run_process(self, name, cmd, env_to_exec=None):
     Log.info("Running %s process as %s" % (name, ' '.join(cmd)))
-    return subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                            env=env_to_exec)
+    try:
+      # stderr is redirected to stdout so that it can more easily be logged. stderr has a max buffer
+      # size and can cause the child process to deadlock if it fills up
+      process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+                                 env=env_to_exec)
```
objmagic: How about adding a [bufsize=1](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/execute.py#L71) here for better performance?

heron/common/src/python/utils/log.py
```diff
@@ -26,6 +27,26 @@
...
+
+  return logging_thread
+
+def stream_process_stdout(process, log_fn):
+  """ Stream the stdout and stderr for a process out to display
+  :param process: the process to stream the logs for
+  :param log_fn: a function that will be called for each log line
+  :return: None
+  """
+  for line in iter(process.stdout.readline, b''):
```
objmagic: I use while loop because of [this issue](http://bugs.python.org/issue3907). It seems that Python 2.7 fixed this issue but as you can see in the thread some people are still reporting seeing this issue in 2.7

objmagic:
Nice work.

(actually I am happy to see at the first time that not properly piping I/O from subprocess could lead to such bad error...)

ajorgensen:
Thanks for reviewing. Yeah this was an interesting one to track down, learned a lot about python and its interaction with subprocesses and also about stderr vs stdout buffering.

objmagic:
@ajorgensen last comment: why [this dependency](https://github.com/twitter/heron/pull/1840/files#diff-a39e34d854c69ea9a9bdee31d6947b20R188)?

ajorgensen:
@objmagic ah sorry I just copied another pex test block and didnt look too closed. I'll remove that dependency if its not actually used.

objmagic:
please fix CI.

kramasamy:
@ajorgensen - @srkukarni also encountered this when he was working with C++ api for Heron.

ajorgensen:
@objmagic I ran the tests locally and they all pass, the failures looked to be caused by 

```
Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000eab00000, 40894464, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 40894464 bytes for committing reserved memory.
# An error report file with more information is saved as:
```

Looks like the ci machines may need some more memory or is there some other recourse?

objmagic:
@ajorgensen 

I just read this PR again but I failed to understand why memory usage issue could happen. However, I observed that integration tests were running fine for first 3~4 tests, but started having problem afterwards.

Let's try this:
1. sleep 5 seconds after an integration test finishes
2. run `ps` and dump its output on terminal. Just want to make sure processes spawned by test have exited.
3. sleep another 5 seconds
4. run next integration test

What do you think?

ajorgensen:
@objmagic yeah I wonder if these changes impacted the interaction in the integration tests. I'll take a look and see if something jumps out. All of the other prs look like they have passing tests so I think its somewhat safe to say this change introduced some difference in the way the integration tests work.

objmagic:
@ajorgensen Agree. I am quite lost here since I did not expect such non-invasive change could blow up memory... We have to fix this issue to get this important PR merged.

ajorgensen:
@objmagic I see whats happening, the thread that is streaming the logs is not properly getting closed which is causing the process remain running which is causing the memory pressure. I'll add some better thread management to make sure the threads are getting closed when the process is.

ajorgensen:
@objmagic it looks like the processes are no longer being properly cleaned up during the integration tests which leaves them around ultimately consuming all the memory available. I'll look into why this is happening.

ajorgensen:
Latest commit should fix the issue, going to wait for it to go through ci to confirm but I was able to see the issue and the resolution locally as well, it just turned out my machine has more memory than the ci machines did :)

objmagic:
@ajorgensen you are right. nice observation.

objmagic:
This reminds me that I may need to use thread as well in [result.py](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/result.py) to prevent deadlock.

objmagic:
CI is good now. Feel free to merge @ajorgensen

objmagic:
CI is good now. Merged. Thanks! @ajorgensen

Issue #1841, opened 2017-05-06
Storm Config should not extends heron.api.Config


nlu90:
Currently we have `public class Config extends com.twitter.heron.api.Config` in `storm/src/java/org/apache/storm/Config.java`.

We met `java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer` error when call `conf.setMessageTimeoutSecs()` in topology code and submit the topology.

The cause of the problem is the following method in `heron.api.Config` is called to set message timeout

```
public static void setMessageTimeoutSecs(Map<String, Object> conf, int secs) {
  conf.put(Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS, Integer.toString(secs));
}
```
While the correct method should be called is
```
public static void setMessageTimeoutSecs(Map conf, int secs) {
  conf.put(Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS, secs);
}
```

The `secs` in converted to String by the first method and thus caused the cast error.

billonahill:
Why not just change that line in the setter to the following? Why do we even convert seconds to strings?
```
conf.put(Config.TOPOLOGY_MESSAGE_TIMEOUT_SECS, secs);
```

[Note: This is an example of why it's generally preferable to use custom types (e.g., `Duration`) over primitives, similar to the refactor we did to use `ByteAmount` over primitives. By doing that we don't run into issues like these.]

PR #1842, opened 2017-05-06
Make storm.Config don't extend heron.api.Config

nlu90:
#1841 

Tested with LocalScheduler.

### DIFF
--- a/heron/examples/src/java/com/twitter/heron/examples/AckingTopology.java
+++ b/heron/examples/src/java/com/twitter/heron/examples/AckingTopology.java
@@ -57,11 +57,11 @@ public final class AckingTopology {
     conf.setMaxSpoutPending(1000 * 1000 * 1000);
 
     // To enable acking, we need to setEnableAcking true
-    com.twitter.heron.api.Config.setEnableAcking(conf, true);
+    conf.setNumAckers(1);
     conf.put(Config.TOPOLOGY_WORKER_CHILDOPTS, "-XX:+HeapDumpOnOutOfMemoryError");
 
     // Set the number of workers or stream managers
-    com.twitter.heron.api.Config.setNumStmgrs(conf, 1);
+    conf.setNumWorkers(1);
     StormSubmitter.submitTopology(args[0], conf, builder.createTopology());
   }
 
--- a/heron/examples/src/java/com/twitter/heron/examples/AckingTopology.java
+++ b/heron/examples/src/java/com/twitter/heron/examples/AckingTopology.java
@@ -57,11 +57,11 @@ public final class AckingTopology {
     conf.setMaxSpoutPending(1000 * 1000 * 1000);
 
     // To enable acking, we need to setEnableAcking true
-    conf.setEnableAcking(true);
+    com.twitter.heron.api.Config.setEnableAcking(conf, true);
     conf.put(Config.TOPOLOGY_WORKER_CHILDOPTS, "-XX:+HeapDumpOnOutOfMemoryError");
 
     // Set the number of workers or stream managers
-    conf.setNumStmgrs(1);
+    com.twitter.heron.api.Config.setNumStmgrs(conf, 1);
     StormSubmitter.submitTopology(args[0], conf, builder.createTopology());
   }
 
--- a/heron/examples/src/java/com/twitter/heron/examples/ComponentJVMOptionsTopology.java
+++ b/heron/examples/src/java/com/twitter/heron/examples/ComponentJVMOptionsTopology.java
@@ -48,19 +48,19 @@ public final class ComponentJVMOptionsTopology {
     Config conf = new Config();
     conf.setDebug(true);
     conf.setMaxSpoutPending(10);
-    conf.setComponentRam("word", ByteAmount.fromMegabytes(500));
-    conf.setComponentRam("exclaim1", ByteAmount.fromGigabytes(1));
+    com.twitter.heron.api.Config.setComponentRam(conf, "word", ByteAmount.fromMegabytes(500));
+    com.twitter.heron.api.Config.setComponentRam(conf, "exclaim1", ByteAmount.fromGigabytes(1));
 
     // TOPOLOGY_WORKER_CHILDOPTS will be a global one
     conf.put(Config.TOPOLOGY_WORKER_CHILDOPTS, "-XX:+HeapDumpOnOutOfMemoryError");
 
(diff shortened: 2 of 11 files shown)


maosongfu:
I saw there are:
`com.twitter.heron.api.Config.setEnableAcking(conf, true);` 
and 
`com.twitter.heron.api.Config.setNumStmgrs(conf, 1);`

For those storm samples, do you want to use storm traditional configs:
`conf.setNumAckers(1);`
and
`conf.setNumWorkers(1);`

maosongfu:
LGTM. 👍

kramasamy:
@nlu90 - this limits some of the resource specifications where you could specify componentRam etc - which makes Heron tuneable?

maosongfu:
@kramasamy People can still access Heron specific config, for instance componentRam, like: https://github.com/twitter/heron/pull/1842/files#diff-b3d9497f7486aba1bca1cc5387948820R60

kramasamy:
@maosongfu - but that exposes the underlying heron api which we should not do. Heron API will undergo a lot of changes as other higher level apis are being developed.

maosongfu:
@kramasamy 
If people want heron features, they should use heron API. These interfaces do not exist in Apache Storm, and people should use these interfaces explicitly with heron Config to reflect they are using some features on Heron, rather than Storm. 
Mixture of these two makes things ugly and hard to trace, i,.e, Hard to reason and hard to debug.
Also, even if we expose heron API features under storm namespace, it is in fact still **heron-api**, since: 
-  we own and maintain this part of code. 
- people need to bundle their projects with heron's jars.  (When people develop project with apache-storm, they will have no ways to access these APIs.)

On the one hand, it can not provide more benefits than simply calling heron api. On the other hand, it increases maintainence cost.
Heron-storm-compatiblity is designed for easy migration from storm to heron. It is not for providing more features: new features are in heron.

Furthermore, lets distinguish high level APIs and storm-extension level APIs:
- APIs in this pull request are storm-extension and are **low-level**. Definitely we can have another API jar called storm-api-extension providing these interfaces. But then why not just call heron? 
- For high level APIs. I would favor for another high level APIs jar, which is on top of heron, providing features to make develpers life easier, for instance, window supports.

objmagic:
This issue this PR addresses is not really urgent, so we can leave this PR open until Monday when @billonahill can give his comments. Still targeting this in 0.14.7.2 though.

billonahill:
-1
This PR seems like we're taking a step backwards with our improvements to types, like `ByteAmount`, to work around another problem regarding seconds (#1841), which could be improved with better typing (i.e. `Duration`). With this change we're no longer storing `ByteAmount` objects, but we're instead storing string representations of integer representations of bytes (or MB, GB?). Even worse, we provide a static helper on `c.t.heron.api.Config` to take a well typed object like `ByteAmount` and then swallow it into a string? Instead, why not fix the seconds use a new `Duration` object and move towards better typing, not away from it.

nlu90:
@billonahill 

This PR keeps storing `ByteAmount` objects for ram related settings if you take a closer look at the changes. E.g, https://github.com/twitter/heron/pull/1842/files#diff-932b57c1a82df361bb7446cf9380e87cR58. 

The only difference is now user needs to call the static method provided in `heron.api.Config` to set them. This is reasonable because those ram settings(container ram, component ram) are more concepts in Heron not in Storm. The previous way of the config works just mix concepts from heron and storm together, while now the config for storm and config for heron are separated by breaking this inheritance.

Better typing is a good idea and we do need to move toward to it. But given the fact that our users are blocked by this bug for migration now and `Duration` object introduces a lot of changes in other places, I would prefer we do those changes in another separate PR. This could benefit our users' migration progress as well as us to keep issues separate.

billonahill:
I've discussed with @nlu90 about this extensively. I still don't like the approach because I think that from a user's perspective this API:
```
     Config conf = new o.a.s.Config();
    com.twitter.heron.api.Config.setComponentRam(conf, "word", ByteAmount.fromMegabytes(500));
    com.twitter.heron.api.Config.setComponentRam(conf, "exclaim1", ByteAmount.fromGigabytes(1));
```
is a lot more clunky and less intuitive[1]  to a user than this API, which is the easier way to migrate to Heron IMO:
```
    Config conf = new o.a.s.Config();
    conf.setComponentRam("word", ByteAmount.fromMegabytes(500));
    conf.setComponentRam("exclaim1", ByteAmount.fromGigabytes(1));
```
1 - When migrating or writing my job, doing code inspection on the setters on `o.a.s.Config` is easier than somehow knowing that I need to now look up static setters on an otherwise unknown `c.t.h.api.Config` class.

Also, I think we have way too many config-related classes already and this approach does not help with consolidation. Also I generally don't like static helper methods, especially to do OO things like building a config.

Also if the reason for this refactor is to fix the issue that was described in #1841, we should fix our setters to not call `toString` on numerical values. As a result I'm still unconvinced why breaking inheritance is the best thing to do here.

That said I don't want to block this release so I'll state my opinions and step out of the way on this review.

nlu90:
@billonahill Thanks for the idea and the discussion. I've created two issues #1845, #1846  to tracking the problems we found during our discussion.

I'll merge this PR to unblock the release process for now.

kramasamy:
-1 - even I am not happy with this PR - since it is exposing Heron API, which is something that will change extensively. Instead a storm extensions config - extensions to the Config of Storm is a good enough. Not sure why this was needed in the first place.

nlu90:
@kramasamy 

If those apis will change extensively, keep them out of users' reach will be the best strategy. But Heron APIs are also exposed through Inheritance. Users can access those public methods/apis in Heron Config through a Storm Config object. 

Also, in the first place in our internal heron, the storm config doesn't extend the heron.api.Config at all. It's `Config extends HashMap<String, Object>`, and our users are all using the provided static methods setting those heron unique confs.

maosongfu:
@kramasamy 
Extensions to the Config of Storm by adding more interfaces explicitly, is a bit different from making Config of Storm inheriting from Config of Heron directly.
I saw @billonahill comments, and it did mention some good points.

Currently the Config of Storm inherits from Config of Heron directly, which is error-prone (Acutally, it does break current master and a lot of topologies can not submit due to bugs caused by this). 
So even if we want to add extensions to the Config of Storm, it would be better to add them as methods explicitly. And I saw @nlu90 has created an issue for it: https://github.com/twitter/heron/issues/1845

PR #1833, opened 2017-05-04
replace hard coded port to getFreePort() in switch_unittest.cpp

huijunwu:
fix #1810: reduce the chance for port to conflict

add `getFreePort()` in C++ code
similar to `heron/common/src/java/com/twitter/heron/common/basics/SysUtils.java:getFreePort()`

PR #1847, opened 2017-05-09
use bind port 0 to solve port conflict in switch_unittest.cpp

huijunwu:
This PR is to replace #1833

The port returned by getFreePort() may be occupied before Start().
This PR let server bind port 0 and get the actual port after binding.

### DIFF
--- a/heron/common/src/cpp/network/baseserver.cpp
+++ b/heron/common/src/cpp/network/baseserver.cpp
@@ -107,7 +107,7 @@ sp_int32 BaseServer::Start_Base() {
   }
 
   // fetch the port after bind/listen port 0
-  if (options_.get_sin_family() == AF_INET) {
+  if (AF_INET == options_.get_sin_family() && 0 == options_.get_port()) {
     if (getsockname(listen_fd_, serv_addr, &sockaddr_len) != 0)  {
       LOG(ERROR) << "getsockname() error: " << strerror(errno);
       close(listen_fd_);
--- a/heron/common/src/cpp/network/baseserver.cpp
+++ b/heron/common/src/cpp/network/baseserver.cpp
@@ -106,6 +106,17 @@ sp_int32 BaseServer::Start_Base() {
     return -1;
   }
 
+  // fetch the port after bind/listen port 0
+  if (options_.get_sin_family() == AF_INET) {
+    if (getsockname(listen_fd_, serv_addr, &sockaddr_len) != 0)  {
+      LOG(ERROR) << "getsockname() error: " << strerror(errno);
+      close(listen_fd_);
+      return SP_NOTOK;
+    } else {
+      options_.set_port(ntohs(in_addr.sin_port));
+    }
+  }
+
   return 0;
 }
 
--- a/heron/common/tests/cpp/network/switch_unittest.cpp
+++ b/heron/common/tests/cpp/network/switch_unittest.cpp
@@ -145,6 +145,7 @@ void start_test(sp_int32 nclients, sp_uint64 requests) {
             << std::endl;
 
   delete server_;
+  delete latch;
 }
 
 TEST(NetworkTest, test_switch_1) { start_test(1, 100); }
--- a/heron/common/tests/cpp/network/switch_unittest.cpp
+++ b/heron/common/tests/cpp/network/switch_unittest.cpp
@@ -33,8 +33,6 @@
 #include "threads/modinit.h"
 #include "network/modinit.h"
 
-static sp_uint32 SERVER_PORT;
-
 class Terminate : public Client {
  public:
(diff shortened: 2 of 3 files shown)


heron/common/src/cpp/network/baseserver.cpp
```diff
@@ -106,6 +106,17 @@ sp_int32 BaseServer::Start_Base() {
     return -1;
   }
 
+  // fetch the port after bind/listen port 0
```
kramasamy: Can you add comments in the Start_Base method and also in options that new port is set when port 0 is provided.

heron/common/tests/cpp/network/switch_unittest.cpp
```diff
@@ -26,13 +26,14 @@
...
 #include "threads/threads.h"
+#include "threads/spcountdownlatch.h"
 #include "network/network.h"
 #include "basics/modinit.h"
 #include "errors/modinit.h"
 #include "threads/modinit.h"
 #include "network/modinit.h"
 
-static const sp_uint32 SERVER_PORT = 61000;
+static sp_uint32 SERVER_PORT;
```
kramasamy: Do we need to need to have a global variable for saving the port? Can we add something in the class itself?

heron/common/src/cpp/network/baseserver.cpp
huijunw: comments added in the header.

heron/common/tests/cpp/network/switch_unittest.cpp
huijunw: removed global variable

heron/common/src/cpp/network/baseserver.cpp
huijunw: comments added in the header.

objmagic:
ping @kramasamy. Any more comments?

heron/common/src/cpp/network/baseserver.cpp
```diff
@@ -106,6 +106,17 @@ sp_int32 BaseServer::Start_Base() {
...
+  if (AF_INET == options_.get_sin_family() && 0 == options_.get_port()) {
+    if (getsockname(listen_fd_, serv_addr, &sockaddr_len) != 0)  {
```
kramasamy: This code is being repeated in the other PRs as well. Can you please abstract it to a function in sock utils.cpp and call that function here.

huijunwu: This code appears only once.
This PR is the prerequisite of the other PRs. If this PR is merged, this code will not appear in the other PRs.

heron/common/tests/cpp/network/switch_unittest.cpp
```diff
@@ -108,15 +110,20 @@ void terminate_server(sp_uint32 port) {
 }
 
 void start_test(sp_int32 nclients, sp_uint64 requests) {
+  sp_uint32 server_port = 0;
+  CountDownLatch* latch = new CountDownLatch(1);
+
   // start the server thread
-  std::thread sthread(start_server, SERVER_PORT);
+  std::thread sthread(start_server, &server_port, latch);
```
huijunwu: The latch solves the race condition.
The server thread updates server_port, then latch is released, then the client thread can read server_port.

heron/common/tests/cpp/network/switch_unittest.cpp
```diff
@@ -68,16 +67,19 @@ class Terminate : public Client {
 
 static TestServer* server_;
 
-void start_server(sp_uint32 port) {
+void start_server(sp_uint32* port, CountDownLatch* latch) {
```
huijunwu: This function is run as thread with no return value.
I tried reference, but compiler complained. I suggest pointer.

kramasamy: weird error though - let us stick with pointer

heron/common/tests/cpp/network/switch_unittest.cpp
kramasamy: Ok got it.

heron/common/src/cpp/network/baseserver.cpp
huijunw: It is better not to put this piece of code in a separate function:
1. This code is not re-used in the other places.
2. Making it separate breaks the same style in the Start_Base()
3. Many variables involve in this code. It is not good tradeoff to pass many variables to a simple separate function.
4. If this code was in a separate function, why would not the other parts of the same Start_Base() be refactored into separate functions?

objmagic:
@huijunwu LGTM. Feel free to merge this.

huijunw:
Thanks @objmagic

Issue #1835, opened 2017-05-04
AuroraScheduler doesn't request max container memory correctly


nlu90:
In the code,

https://github.com/twitter/heron/blob/master/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java#L96

when creating `containerResource`, we should use `updatedPackingPlan` instead of the `packing`. 

`updatedPackingPlan` aligns containers' memory to the max one required among all containers while `packing` only contains the memory requirement to run all instance in that container.

objmagic:
closed via #1836

nlu90:
#1850

PR #1850, opened 2017-05-10
fix aurora to use maximum resources to start containers

nlu90:
`getRequiredResource()` returns the amount of minimum resource needed by each container generated by packing plan, while `getScheduledResource()` returns the maximum resource among all of them.

To start a aurora container with enough resource, `scheduled resource` should be used. If `required resource` is used, there's a chance that aurora container doesn't reserve enough rams. Containers get  out of memory error and are restarted constantly.

This change is tested with actual topologies within twitter environment.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
@@ -30,6 +30,7 @@ import com.google.common.base.Optional;
 
 import com.twitter.heron.api.generated.TopologyAPI;
 import com.twitter.heron.common.basics.FileUtils;
+import com.twitter.heron.packing.builder.Container;
 import com.twitter.heron.proto.scheduler.Scheduler;
 import com.twitter.heron.scheduler.UpdateTopologyManager;
 import com.twitter.heron.scheduler.utils.Runtime;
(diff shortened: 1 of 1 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -92,13 +92,16 @@ public boolean onSchedule(PackingPlan packing) {
...
     PackingPlan updatedPackingPlan = packing.cloneWithHomogeneousScheduledResource();
     SchedulerUtils.persistUpdatedPackingPlan(Runtime.topologyName(runtime), updatedPackingPlan,
         Runtime.schedulerStateManagerAdaptor(runtime));
 
+    // Use the ScheduledResource to create aurora properties
+    // the ScheduledResource is guaranteed to be set after calling
+    // cloneWithHomogeneousScheduledResource in the above code
     Resource containerResource =
-        updatedPackingPlan.getContainers().iterator().next().getRequiredResource();
+        updatedPackingPlan.getContainers().iterator().next().getScheduledResource().get();
```
huijunw: is it possible that get() throws NoSuchElementException

nlu90: see my comments above. It's guaranteed the scheduledresource is set

huijunw: 👍

nlu90:
#1835

PR #1855, opened 2017-05-12
Add support for emit direct

billonahill:
My experimentation with Trident required this, so adding it. Verified in an example trident topology.


heron/instance/src/java/com/twitter/heron/instance/AbstractOutputCollector.java
```diff
@@ -93,23 +94,30 @@ public long getTotalTuplesEmitted() {
...
     // set the key. This is mostly ignored
     builder.setKey(0);
 
     List<Integer> customGroupingTargetTaskIds = null;
-    if (!helper.isCustomGroupingEmpty()) {
+    if (taskId != null) {
+      // TODO: somehow assert that the input stream of the downstream bolt was configured
+      // with directGrouping
+
+      customGroupingTargetTaskIds = new ArrayList<>();
```
nlu90: can we use a separate `directGroupingTragetTaskIds` to handle direct grouping cases? Also instead of using `if... else if`, can we separate the handling of direct grouping and custom grouping?

billonahill:
Nothing's been decided about Trident at this point. This was the main feature that was blocking it, and after discussions with @mfu we realized it was easy to implement so I added it. There's a not a lot to supporting direct grouping, since it's very similar to custom grouping, so now that it's been developed we might as well add it IMO, regardless of Trident.

heron/instance/src/java/com/twitter/heron/instance/AbstractOutputCollector.java
billonahill: @nlu90 I tried splitting up the code paths and it got pretty redundant and hard to follow. 

To remove the if/then I'd need to do the emit direct check at a higher level when creating the tuple builder. But the tuple builder is created in both SpoutOutputCollectorImpl.admitSpoutTuple and BoltOutputCollectorImpl.admitBoltTuple after doing other validation. To do the emit direct check separately, I'd need to have 4 more implementations of the same validation (with/without emit direct ids for Bolt/Spout collectors), and then support two similar code paths in AbstractOutputCollector to support emit direct, and non-emit direct. The latter impl would also have the custom grouping checks at 2 levels lower than the emit direct checks.

I can show you what I mean if it's not clear, but if you have a specific suggestion for how to avoid what I'm describing please let me know.

nlu90: kk, I see what you mean

nlu90:
👍

PR #1856, opened 2017-05-13
add 'heron' prefix to all heron client optional args

nlu90:
One of our user failed to deploy his topology with the following cmd

```
heron submit dc/role/env ${TOPO_FILE} ${TOPO_CLASS} --config ${TOPO_CONF} --env staging --topology_id 1
```

The error message shows:
```
usage: heron submit [options] cluster/[role]/[env] topology-file-name topology-class-name [topology-args]
heron submit: error: ambiguous option: --config could match --config-path, --config-property
```

It turns out the user's topology takes an argument named "config" which caused the client confusion and failed the submission. 

Since this use case is common, the "heron" prefix is added to optional argument to avoid such conflicts. And this is also how the optional arguments named in the old internal heron client.

### DIFF
--- a/heron/tools/cli/src/python/args.py
+++ b/heron/tools/cli/src/python/args.py
@@ -35,7 +35,7 @@ def add_verbose(parser):
   :return:
   '''
   parser.add_argument(
-      '--verbose',
+      '--heron-verbose',
       default=False,
       help='Verbose mode. Increases logging level to show debug messages')
   return parser
@@ -107,12 +107,12 @@ def add_config(parser):
   default_config_path = config.get_heron_conf_dir()
 
   parser.add_argument(
-      '--config-path',
+      '--heron-config-path',
       default=os.path.join(config.get_heron_dir(), default_config_path),
       help='Path to cluster configuration files')
 
   parser.add_argument(
-      '--config-property',
+      '--heron-config-property',
       metavar='PROPERTY=VALUE',
       action='append',
       default=[],
@@ -126,7 +126,7 @@ def add_system_property(parser):
   :return:
   '''
   parser.add_argument(
-      '--topology-main-jvm-property',
+      '--heron-topology-main-jvm-property',
       metavar='PROPERTY=VALUE',
       action="append",
       default=[],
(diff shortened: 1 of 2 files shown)


maosongfu:
This change can break the interfaces provided by Heron. Are there ways we could:
- precisely match the name of optional args' key, or
- distinguish the scope heron itself's configs from the scope of user's configs, just like bazel using `--` as a separator?

kramasamy:
Let use -- since it is standard in python/gnu/hazel etc.

maosongfu:
+1 for @kramasamy suggestion if it can work

objmagic:
Unfortunately, this does not solve the root problem at all. What if user passes some argument with prefix `heron`? You will see the same error. And more annoyingly, this problem is not due to our mistake. 

The root cause is that **Python 2.7 argparser module is full of crazy shit**. Excuse me again for the foul language, but if you have experience with using other command line parsers in other languages, and you have ever read through the entire documentation of argparser APIs, you have no choice but wholeheartedly agree with me. Python 3 fixed this problem by allowing a `(allow_abbrev=False)` which can give you **exact** match. 

I have no opinion with @kramasamy's suggestion of using `--`, as long as you can implement it and add comprehensive tests! Another way of solving this problem is [here](http://stackoverflow.com/a/10751356)

objmagic:
Also, I prefer to solve this issue by not requesting changes from users' side (or you can say by not breaking Heron client's current behavior). Imagine users come to ask us "hey my script does not work. what has happened?", and you have to say "we did the following **breaking** change and you need to change accordingly"... 

That said, `--` solution is not as good as [this one](http://stackoverflow.com/questions/10750802/disable-abbreviation-in-argparse/10751356#10751356). Let us solve this issue as transparent as possible.

objmagic:
Also, your PR forgot to change all the places that access argument values. For example, [here](https://github.com/twitter/heron/blob/master/heron/tools/cli/src/python/submit.py#L106).

kramasamy:
@objmagic already -- works in heron client and it is a consistent behavior across the language.

objmagic:
ok. whatever is fine to me.

nlu90:
As a discussion result, we will ask our users to add "--" to fix the problem. And this solution is verified working internall. So I'll close this pr

PR #1831, opened 2017-05-04
Added Checkpoint Manager for stateful topologies

srkukarni:

PR #1857, opened 2017-05-15
Add functions/handlers in statemgrs to deal with StatefulCheckpoints

srkukarni:


maosongfu:
Plz fix eariler broken unit tests in https://github.com/twitter/heron/pull/1831 first rather than hurry for new codes. Otherwise it will block and slow down the whole heron migration process at Twitter; and consequently slow down the whole stateful processing features.

objmagic:
@maosongfu @srkukarni I agree we need to fix failed unit tests. But please do not worry about our internal migration process since currently release is always cherry-picked. That said, we can always control what we want to put into our next release.

nlu90:
Split this PR into 2 separate smaller ones: #1861 #1862, for easier and fast reviewing

PR #1860, opened 2017-05-15
Change SystemConfig and HeronSocketOptions to use ByteAmount

billonahill:
Instead of using ints and longs of bytes, MBs, GBs and doing math to convert between units, it's better to use the `ByteAmount` typed object. This change does that. The core changes are to `HeronSocketOptions`,  `SystemConfig` and `SystemConfigKey`. The rest is the refactor fallout of those changes. Also making `HeronSocketOptions` immutable.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
+++ b/heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Queue;
 import java.util.logging.Logger;
 
+import com.twitter.heron.common.basics.ByteAmount;
 import com.twitter.heron.common.basics.Constants;
 import com.twitter.heron.common.basics.ISelectHandler;
 import com.twitter.heron.common.basics.NIOLooper;
@@ -71,9 +72,9 @@ public class SocketChannelHelper {
   private long totalBytesRead;
   private long totalBytesWritten;
   // System Config related
-  private long writeBatchSizeInBytes;
+  private ByteAmount writeBatchSize;
   private long writeBatchTimeInNs;
-  private long readBatchSizeInBytes;
+  private ByteAmount readBatchSize;
   private long readReadBatchTimeInNs;
 
   public SocketChannelHelper(NIOLooper looper,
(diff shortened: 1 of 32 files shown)


heron/simulator/src/java/com/twitter/heron/simulator/instance/SpoutInstance.java
```diff
@@ -78,7 +79,7 @@ protected void produceTuple() {
       }
 
       if (collector.getTotalDataEmittedInBytes() - totalDataEmittedInBytesBeforeCycle
```
objmagic: Just curious since I don't have a strong opinion here: in this kind of scenario, shall we use [greaterThan](https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/basics/ByteAmount.java#L179) API instead?

heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
```diff
@@ -71,9 +72,9 @@
   private long totalBytesRead;
```
objmagic: also `ByteAmount` here?

heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
```diff
@@ -134,7 +135,7 @@ public boolean sendPacket(OutgoingPacket outgoingPacket) {
     // 1. We spent too much time
     // 2. We have read large enough data
     while ((System.nanoTime() - startOfCycle - readReadBatchTimeInNs) < 0
-        && (bytesRead < readBatchSizeInBytes)) {
+        && (bytesRead < readBatchSize.asBytes())) {
```
objmagic: like other questions, shall we also do `ByteAmount` here? I assume you are following some principle here: what are converted to ByteAmount, what are converted. Can you elaborate a bit?

objmagic:
Also: CI.

nlu90:
LGTM, but please fix the failed test before merge the changes.

heron/simulator/src/java/com/twitter/heron/simulator/instance/SpoutInstance.java
billonahill: The collector and the simulator both still track how many bytes they've emitted as a long, so I convert to that and do the comparison. I've mainly just been converting the static `ByteAmounts` that are read from the configs, but not the dynamic ones that get frequently updated at runtime. I'd like to also update the latter but I need to look at that more closely to make sure we can always do that efficiently without too much object creation, versus incrementing a long. I'd like to hold on that for a follow-up PR.

heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
billonahill: Same comment as above. For now, holding off on updating dynamic byte amounts.

heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
billonahill: See my response above. Holding off on highly dynamic byte counters for now.

billonahill:
@objmagic we're not changing any user-facing APIs, just the internal `SystemConfig` getters used by internal heron classes.

PR #1861, opened 2017-05-16
add stateful checkpoints support in cpp statemgr

nlu90:
This is the cpp changes for state manager to support stateful processing


heron/statemgrs/src/cpp/statemgr/heron-localfilestatemgr.cpp
```diff
@@ -55,6 +55,9 @@ void HeronLocalFileStateMgr::InitTree() {
   path = dpath;
   path += "/executionstate";
   FileUtils::makeDirectory(path);
+  path = dpath;
+  path += "/statefulcheckpoint";
```
billonahill: We want to use plural here for consistency, right? `statefulcheckpoints`

nlu90: good catch. I'll update them all to `statefulcheckpoints`

PR #1862, opened 2017-05-16
add stateful checkpoints support for java statemgrs

nlu90:
This is the java state manager changes to support stateful processing

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
@@ -19,6 +19,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.twitter.heron.api.generated.TopologyAPI;
 import com.twitter.heron.classification.InterfaceAudience;
 import com.twitter.heron.classification.InterfaceStability;
+import com.twitter.heron.proto.ckptmgr.CheckpointManager;
 import com.twitter.heron.proto.scheduler.Scheduler;
 import com.twitter.heron.proto.system.ExecutionEnvironment;
 import com.twitter.heron.proto.system.PackingPlans;
@@ -276,4 +277,27 @@ public interface IStateManager extends AutoCloseable {
    */
   ListenableFuture<PhysicalPlans.PhysicalPlan> getPhysicalPlan(
       WatchCallback watcher, String topologyName);
+
+  /**
+   * Set the Stateful Checkpoint
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<Boolean> setStatefulCheckpoint(
+      CheckpointManager.StatefulConsistentCheckpoints checkpoint, String topologyName);
+
+  /**
+   * Get the Stateful Checkpoint
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<CheckpointManager.StatefulConsistentCheckpoints> getStatefulCheckpoint(
+      WatchCallback watcher, String topologyName);
+
+  /**
+   * Delete the stateful checkpoint for the given topology
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<Boolean> deleteStatefulCheckpoint(String topologyName);
 }
--- a/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
@@ -85,6 +85,13 @@ public interface IStateManager extends AutoCloseable {
   void close();
 
   /**
+   * Is the given topology in RUNNING state?
+   *
+   * @return Boolean
+   */
+  ListenableFuture<Boolean> isTopologyRunning(String topologyName);
+
+  /**
    * Return a lock object backed by the state manager store.
    *
    * @param topologyName the name of the topology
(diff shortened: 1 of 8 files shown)


heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -276,4 +277,27 @@ public String getName() {
...
+   */
+  ListenableFuture<Boolean> setStatefulCheckpoint(
+      CheckpointManager.StatefulConsistentCheckpoints checkpoint, String topologyName);
+
+  /**
+   * Get the Stateful Checkpoint
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<CheckpointManager.StatefulConsistentCheckpoints> getStatefulCheckpoint(
```
billonahill: The object contains multiple checkpoints so we should put an 's' on the method name. Also can it take some identifier of the currently topology like the packing plan id? For example upon restarting a topology we need to know if the previous checkpoints are still valid for the new plan, if it's been changed between deploys.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -276,4 +277,27 @@ public String getName() {
...
   ListenableFuture<PhysicalPlans.PhysicalPlan> getPhysicalPlan(
       WatchCallback watcher, String topologyName);
+
+  /**
+   * Set the Stateful Checkpoint
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<Boolean> setStatefulCheckpoint(
+      CheckpointManager.StatefulConsistentCheckpoints checkpoint, String topologyName);
```
billonahill: This method should set a single checkpoint, so `StatefulConsistentCheckpoint` not `StatefulConsistentCheckpoints`. It will also need to set them in a way that we can a.) delete a single checkpoint and b.) identify the packing plan that it's bound to.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: added the 's'.

Checkpoint state is bounded to an instance. And instance is a logical concept. Where is the instance running should not affect what the instance checkpoint state is. So the mapping of checkpoints and packing plan id is not needed to me.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: If only a single checkpoint is allowed to set, then the topology is not able to recovery from some early states.

nlu90:
the last commit includes 
1. some code refactoring in the `IStateManager.java` for easier reading
2. updated `getStatefulCheckpoint ` to `getStatefulCheckpoints`

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: The packing plan id is already packed in the message `StatefulConsistentCheckpoint`
```
message StatefulConsistentCheckpoint {
  required string checkpoint_id = 1;
  required string packing_plan_id = 2;
  // One can add more meta data about this ckpt later
}
```

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -228,52 +248,56 @@ public String getName() {
...
-  ListenableFuture<PhysicalPlans.PhysicalPlan> getPhysicalPlan(
+  ListenableFuture<CheckpointManager.StatefulConsistentCheckpoints> getStatefulCheckpoints(
       WatchCallback watcher, String topologyName);
+
+  /**
+   * Delete the stateful checkpoint for the given topology
+   *
+   * @return Boolean - Success or Failure
+   */
+  ListenableFuture<Boolean> deleteStatefulCheckpoint(String topologyName);
```
billonahill: The deleteStatefulCheckpoint method needs to take an identifier to make it clear which checkpoint is to be deleted. The state server stores multiple checkpoints per topology

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
```diff
@@ -228,52 +248,56 @@ public String getName() {
...
   /**
-   * Set the packing plan for the given topology
+   * Set the Stateful Checkpoint
    *
    * @return Boolean - Success or Failure
    */
-  ListenableFuture<Boolean> setPackingPlan(
-      PackingPlans.PackingPlan packingPlan, String topologyName);
+  ListenableFuture<Boolean> setStatefulCheckpoint(
+      CheckpointManager.StatefulConsistentCheckpoints checkpoint, String topologyName);
```
billonahill: setStatefulCheckpoint sets a specific checkpoint, not a collection.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
billonahill: Multiple checkpoints must be persisted and managed by the state server, but they're set one by one, each time a new one is created. Think of it as a Heap, where new ones are pushed to the head and old ones are periodically deleted from the tail. At any time the entire set could be fetched.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
billonahill: We should store the checkpoints in a way that allows us to easily delete (or manually browse when troubleshooting) an entire set by the packing plan hash and checkpoint id, so we should considering including those in the path in the directory tree.

Consider the use case where the topology gets redeployed with the same parallelism and packing plan hash. We can reuse that state with the new topology. If instead the topology is deployed with different parallelism (i.e., a new packing plan), we can not reload that old state directly. We also might want to easily prune all old checkpoints bound to old packing plans. Hence, being able to delete an entire file system branch would be ideal.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
srkukarni: The proto structure thats stored in the statemgr is the actual list. Not just the last one.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
srkukarni: All manipulation of the proto structure happens inside the relevant servers. This class purely is for interfacing with the statemgr to store/retrieve/delete

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: Storing a list of checkpoints achieves the same result as the solution you proposed here. Plus it's much more simpler in terms of implementation and management.

To manage those checkpoints, the TMaster could fetch all of them in the state managers and then modifies the fetched states and then write it back.

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: The actual checkpoints are stored in the stateful storages not in the state manager. State manager only stores the ckpt id (which reference the actual checkpoint) and pp_id(which corresponds the ckpt_id).

heron/spi/src/java/com/twitter/heron/spi/statemgr/IStateManager.java
nlu90: maybe `setStatefulCheckpoints` makes more sense

billonahill:
In the future please don't reorder unrelated methods in a feature PR, since it makes it hard to read the diff of functional changes.

PR #1863, opened 2017-05-16
Use Duration to represent durations of time instead of primatives

billonahill:
Instead of using ints and long to represents nanoseconds, milliseconds, seconds and minutes, we should use Duration. This makes units clear and removes the need for unit conversion math.

Apologies for the large review but there were a number of cascading effects. The main goal was to:

- Convert all time duration configs in `SystemConfig` to use `java.time.Duration`
- Remove all possible date math and the conversion constants in `c.t.h.basics.Constants`

There are still places where we use `long` to represent seconds since the epoch for some local calculations, which is fine. From an API perspective though, wherever we pass around time-based durations we should use the `Duration` class. This make unit unambiguous.


huijunw:
do we plan to use `Instant` to represent an instantaneous point on the time-line in future refactoring?

billonahill:
@huijunw yes, I think it would be good for us to use `Instant` to represent points in time in a follow up PR.

PR #1864, opened 2017-05-17
heron-shell refactor

objmagic:
1. split handlers in `heron-shell` into separate files. Put them under `handlers` directory
2. increase timeout of HTTP request, since `jmap` and `jstack` could run longer than the default `20ms`

This is a prerequisite patch for fixing `jmap/jstack` operations on UI in Twitter internal environment. This means that we will provide our all jmap/jstack handlers for Twitter internal environment.


objmagic:
Thanks. Merged!

Issue #1559, opened 2016-11-14
order_unittest sometimes timeout


objmagic:
``//heron/common/tests/cpp/network:order_unittest``

objmagic:
`//heron/common/tests/cpp/network:switch_unittest` also sometimes times out. Maybe mark all tests under `//heron/common/tests/cpp/networ` as flaky so that Bazel could rerun them?

huijunw:
made an issue to separate order_unittest and switch_unittest in #1810

PR #1865, opened 2017-05-17
improve order_unittest

huijunwu:
Prerequisite: #1847
Fix: #1559

1. SERVER_PORT = 59000; => bind port 0
2. thread.join(); => thread.wait_for(std::chrono::seconds(5))


heron/common/tests/cpp/network/order_unittest.cpp
```diff
@@ -68,16 +68,20 @@ class Terminate : public Client {
 
 static OrderServer* server_;
 
-void start_server(sp_uint32 port) {
+void start_server(sp_uint32* port, CountDownLatch* latch) {
```
kramasamy: @huijunw - instead of * can you use reference like sp_uint32& port?  It is much more easy to read.

heron/common/tests/cpp/network/oserver_unittest.cpp
```diff
@@ -75,5 +75,5 @@ void OrderServer::Terminate() {
 
 void OrderServer::HandleTerminateMessage(Connection* _connection __attribute__((unused)),
                                          TerminateMessage* _message __attribute__((unused))) {
-  AddTimer([this]() { this->Terminate(); }, 1);
+  AddTimer([this]() { std::cout << "OrderServer:Terminate"; this->Terminate(); }, 1);
```
kramasamy: @huijunw - do you need this message?

heron/common/tests/cpp/network/switch_unittest.cpp
```diff
@@ -68,16 +67,19 @@ class Terminate : public Client {
 
 static TestServer* server_;
 
-void start_server(sp_uint32 port) {
+void start_server(sp_uint32* port, CountDownLatch* latch) {
```
kramasamy: use reference instead of pointer for both port and latch?

heron/common/tests/cpp/network/oserver_unittest.cpp
huijunwu: This printout is added for future diagnosis. 
The timeout issue was not reproduced. I suspect the timeout issue may be due to this timer action. 
It does not print a lot since it is supposed to be called only once.

heron/common/tests/cpp/network/order_unittest.cpp
huijunwu: The compiler complained on the reference

heron/common/tests/cpp/network/switch_unittest.cpp
huijunwu: The compiler complained on the reference

heron/common/tests/cpp/network/order_unittest.cpp
```diff
@@ -108,27 +112,38 @@ void terminate_server(sp_uint32 port) {
...
+  std::vector<std::future<void>> cthreads;
   for (sp_int32 i = 0; i < nclients; i++) {
-    cthreads.push_back(std::thread(start_client, SERVER_PORT, requests));
+    cthreads.push_back(std::async(std::launch::async, start_client, server_port, requests));
   }
 
   // wait for the client threads to terminate
   for (auto& thread : cthreads) {
-    thread.join();
+    if (thread.wait_for(std::chrono::seconds(5)) == std::future_status::timeout) {
```
kramasamy: Should we use constants instead of hard coded 5 secs?

heron/common/tests/cpp/network/order_unittest.cpp
```diff
@@ -108,27 +112,38 @@ void terminate_server(sp_uint32 port) {
...
+      GTEST_FAIL() << "timeout for client thread to join in 5 seconds";
```
kramasamy: Update the print out if changed to constant

heron/common/tests/cpp/network/order_unittest.cpp
huijunw: done

heron/common/tests/cpp/network/order_unittest.cpp
huijunw: done

PR #1866, opened 2017-05-17
package ckptmgr and statefulstorages into heron-core

nlu90:
add ckptmgr and statefulstorages into heron-core for distribution. They are not used for now, and won't affect any production code.


huijunw:
do we have a switch to turn this feature on or off ?

nlu90:
@huijunw This only adds the binary into the heron-core.tar.gz. They will not be started at all.

PR #1852, opened 2017-05-11
replace hard coded stmgr port in stmgr_unittest

huijunwu:
Prerequisite: #1827, #1832, #1847

The hard coded ports sometimes conflict in the continuous integration test. This PR replaces hard coded stmgr ports in stmgr_unittest.

PR #1868, opened 2017-05-17
replace hard coded tmaster ports in tmaster_unittest

huijunwu:
Prerequisite: #1852
Fix: #1838

1. Start 3 servers at the beginning of TMaster startup
2. Set port 0 in tmaster_unittest


heron/common/src/cpp/network/httpserver.cpp
```diff
@@ -47,13 +47,25 @@ sp_int32 HTTPServer::Start() {
             << "0.0.0.0"
             << ":" << port;
   // Bind to INADDR_ANY instead of using the hostname
-  sp_int32 retval = evhttp_bind_socket(http_, "0.0.0.0", port);
-  if (retval == 0) {
+  auto* retval = evhttp_bind_socket_with_handle(http_, "0.0.0.0", port);
+  if (retval != NULL) {
```
kramasamy: use nullptr?

heron/common/src/cpp/metrics/metricsmgr-client.cpp
```diff
@@ -63,6 +63,12 @@ void MetricsMgrClient::HandleClose(NetworkErrorCode) {
 
 void MetricsMgrClient::ReConnect() { Start(); }
 
+void MetricsMgrClient::SetPublisherPort(const sp_int32 _port) {
+  if (0 == port_ && 0 != _port) {
```
kramasamy: How can you guarantee that port_ is initialized to 0?

kramasamy: It is the same for other PRs as well.

huijunw: The `port_` may be 0 or not be 0.
In the unit test, it is initialized as 0 through constructor parameter.
If it is not initialized as 0, it will not be set to other value during its lifetime, which is the same behavior as the before.

heron/common/src/cpp/network/httpserver.cpp
huijunw: done

heron/common/src/cpp/network/httpserver.cpp
```diff
@@ -47,13 +47,25 @@ sp_int32 HTTPServer::Start() {
             << "0.0.0.0"
             << ":" << port;
   // Bind to INADDR_ANY instead of using the hostname
-  sp_int32 retval = evhttp_bind_socket(http_, "0.0.0.0", port);
-  if (retval == 0) {
+  auto* retval = evhttp_bind_socket_with_handle(http_, "0.0.0.0", port);
```
kramasamy: if you are using auto there is no need for auto*

huijunw: done

objmagic:
are we good on this? @kramasamy

huijunw:
This PR has been refactored. Ping @srkukarni @nlu90 @kramasamy for review

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -106,6 +104,12 @@ TMaster::TMaster(const std::string& _zk_hostport, const std::string& _topology_n
   do_reassign_ = false;
 
   master_establish_attempts_ = 0;
+
+  // start servers to get the actual ports before `tmaster_location_` related operation
```
srkukarni: this breaks one fundamental invariant in tmaster design. No operations before verifying that we are the master.

huijunwu: What is the reason for 'No operations before verifying that we are the master'?

huijunw: @srkukarni ^^

srkukarni: One of the fundamental assumptions in tmaster when it is processing requests is that it is already established itself as the master. This is to prevent multiple tmasters trampling on each other. Thats the reason, we do StartServer only after TMaster Establishment is complete.
This change starts the TMasterServer which is now ready to accept connections, even before it is established as master. Thus all sorts of bad issues can happen

Issue #1859, opened 2017-05-15
add test utility method for heron server


nlu90:
Currently many tests shares the pattern that starting a heron server and waiting for it to start and then sending test requests. We could create some utility method for the setup of the heron server thread to simplify the test workload

billonahill:
See additional discussion and an alternate approach in PR #1858.

PR #1870, opened 2017-05-17
Adding new HeronServerTester to simplify testing servers

billonahill:
Adding `HeronServerTester` to simplify writing unit tests against Heron server. Much of what's in `HeronServerTester` is code that's copied in multiple tests.

Provides the following benefits:

- Consolidates common boilerplate client/server code that each unit test shouldn't have to deal with (loopers, threadpools, etc)
- Removes the use of timers and uses countdown latches instead. This reduces this test from 13 seconds to 1 second.
- Reduces possibility of flaky tests due to mis-implementation of sleep logic.
- Makes test logic easier to read and understand since other clutter is removed.

Review includes one test to show how `HeronServerTester` is used. I've upgraded one other that I'll do in a follow-up review. There will be a few more to upgrade after that.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
+++ b/heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
@@ -1,0 +1,257 @@
+//  Copyright 2017 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.common.network;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+import com.google.protobuf.GeneratedMessage;
+import com.google.protobuf.Message;
+
+import com.twitter.heron.common.basics.ByteAmount;
+import com.twitter.heron.common.basics.Communicator;
+import com.twitter.heron.common.basics.NIOLooper;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Class to help simplify testing HeronServer implementations by consolidating common test
+ * functionality. A new instance of this class should be used for each test. After initializing the
+ * class, tests should call start(), which does the following:
+ * <ol>
+ * <li>starts the server and waits for it to come up</li>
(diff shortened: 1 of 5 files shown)


heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
```diff
@@ -0,0 +1,367 @@
...
+  /**
+   * Wait for a multiCountMetric to reach or exceed a specific value before returning. Assumes that
+   * values are being increased over time by the implementor.
+   * @param multiCountMetric the multiCountMetric to check
+   * @param key the key of the value to be checked
+   * @param expected the minimum expected value to observe before returning
+   * @param timeout how long to wait for the expected value before failing
+   * @throws InterruptedException
+   */
+  public static void awaitMultiCountMetric(MultiCountMetric multiCountMetric, String key,
```
nlu90: This method seems not generic for most test cases to utilize. Can we put it back to the MetricsManagerServerTest?

heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
```diff
@@ -0,0 +1,367 @@
...
+                                           Long expected, Duration timeout)
+      throws InterruptedException {
+    Instant expiration = Instant.now().plus(timeout);
+    Long found = multiCountMetric.scope(key).getValue();
+    while (Instant.now().isBefore(expiration)) {
+      if (found >= expected) {
+        return;
+      }
+      Thread.sleep(5);
```
nlu90: can we avoid `Thead.sleep` here？

heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
```diff
@@ -0,0 +1,367 @@
...
+ * functionality. A new instance of this class should be used for each test. After initializing the
+ * class, tests should call start() which guarantees the following:
+ * <ol>
+ * <li>the server has started</li>
+ * <li>the client has been launched</li>
+ * </ol>
+ * Upon test completion, the stop() method should be closed to properly release resources. This
+ * method is safe to be called on an instance that was never properly started.
+ */
+public class HeronServerTester {
```
nlu90: this class packs mant interfaces, abstract classes and classes. How about creating separate files for them and only keep those needed here?

kramasamy:
@billonahill - great pr. Wondering if there are other tests that require changes as well to take advantage of the framework.

heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
billonahill: Yeah good call @nlu90 . I moved `TestCommunicator` into `c.t.h.common.basics` to be along side `Communicator`. I also moved `awaitMultiCountMetric` into the test that needs it (not yet put up for PR).

The remaining inner classes are now more closely coupled with `HeronServerTester`. They're basically the handlers and the client helper classes. I'm normally not that big a fan of inner classes but in this case I think it makes sense to keep them as inner classes for now since a.) their impl is closely coupled to `HeronServerTester` and, b.) I don't want to pollute the `c.t.h.common.network` package space with too many test-related classes.

billonahill:
@kramasamy yeah, a bunch. :) I've got others queued up and will submit in separate PRs, to keep them small.

heron/common/src/java/com/twitter/heron/common/network/HeronServerTester.java
nlu90: 👍

billonahill:
Fixes #1859.

PR #1872, opened 2017-05-18
add tmaster zk watch on tmaster location

nlu90:
Fix the issue that tmaster keeps running after zk state manager lost the tmaster location.

The fix is: add a watch, which triggers tmaster to quit immediately, in tmaster.

Tested with local scheduler and zk statemanager.

### DIFF
--- a/heron/tmaster/src/cpp/manager/tmaster.cpp
+++ b/heron/tmaster/src/cpp/manager/tmaster.cpp
@@ -142,6 +142,14 @@ void TMaster::EstablishTMaster(EventLoop::Status) {
   auto cb = [this](proto::system::StatusCode code) { this->SetTMasterLocationDone(code); };
 
   state_mgr_->SetTMasterLocation(*tmaster_location_, std::move(cb));
+
+  // if zk lost the tmaster location, tmaster quits to bail out and re-establish its location
+  auto cb2 = [this]() {
+    LOG(ERROR) << " lost tmaster location in zk state manager. Bailing out..." << std::endl;
+    ::exit(1);
+  };
+  state_mgr_->SetTMasterLocationWatch(tmaster_location_->topology_name(), std::move(cb2));
+
   master_establish_attempts_++;
 }
 


heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -142,6 +142,14 @@ void TMaster::EstablishTMaster(EventLoop::Status) {
   auto cb = [this](proto::system::StatusCode code) { this->SetTMasterLocationDone(code); };
 
   state_mgr_->SetTMasterLocation(*tmaster_location_, std::move(cb));
+
+  // if zk lost the tmaster location, tmaster quits to bail out and re-establish its location
+  auto cb2 = [this]() {
+    LOG(ERROR) << " lost tmaster location in zk state manager. Bailing out..." << std::endl;
```
billonahill: Could we instead attempt to set the location again in ZK if it's lost?

srkukarni: According to zk semantics, this situation should never happen. If it happens it certainly is a zk client/server bug and we are already in a bad state. better to die and restart fresh

nlu90: 1. IMO, Fail fast in this case is better. TMaster only knows the fact that its location is lost but doesn't know the exact cause. So cleaning itself up and then resetting its location from scratch is cleaner.
2. From implementation side, "set the location again" is not that convenient to implement. It requires a lot of changes in other parts(stmgr, tmaster).

billonahill: Is it possible that a lost ZK location gets recovered or once it's lost is it lost for good? Is there a scenario where brief network faults would cause this to be lost and the quickly recovered? I suspect it's lost for good, but I don't know for sure.

nlu90: Based on the problematic topologies we investigate, if the location is lost, then it's just lost for good. Even the connection between the tmaster container and zk is alive, the location is still not recovered.

huijunw: shall we reuse the existing code
```
master_establish_attempts_ = 0;
EstablishTMaster(EventLoop::TIMEOUT_EVENT);
```

huijunw: Tmaster trys to reconnect to zookeeper in `SetTMasterLocationDone()` for `master_establish_attempts_` times. Shall we somehow reuse that piece of code?

nlu90: @huijunw killing and restarting the tmaster achieve the same result. In addition, it's cleaner and simplier

PR #1874, opened 2017-05-18
Marathon does not allow upper case applications

srkukarni:


srkukarni:
@jrcrawfo please review

PR #1880, opened 2017-05-21
start ckptmgr if topology is stateful-enabled

nlu90:
This PR finishes the following tasks:
1. update config file for stateful processing according to the current Config design
2. start ckptmgr process if the topology is stateful-enabled; won't start it if unspecified or disabled
3. add necessary Key and util methods

Tested with local scheduler. Ckptmgr is started correctly if stateful enabled; no ckptmgr is started otherwise


nlu90:
two separate PRs will be sent instead:
1. config updates
2. schedulers and executor chagnes to start ckptmgr

PR #1882, opened 2017-05-22
Further improve heron-shell

objmagic:
1. Use thread to pipe stdout/stderr from subprocess to prevent deadlock
2. in `chain`, close previous process's stdout to allow it to receive SIGPIPE


heron/shell/src/python/utils.py
```diff
@@ -141,34 +143,40 @@ def read_chunk(filename, offset=None, length=None):
...
+  def add(self, line):
+    self.str += line
+
+  def result(self):
+    return self.str
 
-def pipe(in_stream, to_cmd):
+def pipe(prev_proc, to_cmd):
   """
   Pipes in_stream from output of previous pipe into to_cmd.
```
nlu90: update docstring

heron/shell/src/python/utils.py
```diff
@@ -141,34 +143,40 @@ def read_chunk(filename, offset=None, length=None):
...
   Runs the command and returns its stdout and stderr.
   """
-  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-  (stdout, stderr) = (stream_to_string(p.stdout), stream_to_string(p.stderr))
+  proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                          stderr=subprocess.PIPE, cwd=cwd, env=env)
+  stdout_builder, stderr_builder = StringBuilder(), StringBuilder()
+  log.async_stream_process_stdout(proc, stdout_builder.add)
+  log.async_stream_process_stderr(proc, stderr_builder.add)
+  proc.wait()
```
nlu90: Will this be blocked forever?

objmagic: why? and what is "this"?

nlu90: If the `proc.wait()` call is blocked forever, then it will never return. And the method is blocked here. Just want to make sure this case won't happen.

objmagic: no, none of the subprocess will block in heron-shell. And even if it blocks, server will return a HTTP timeout

heron/shell/src/python/utils.py
```diff
@@ -155,7 +155,7 @@ def result(self):
 
 def pipe(prev_proc, to_cmd):
   """
-  Pipes in_stream from output of previous pipe into to_cmd.
+  Pipes output of prev_proc into to_cmd.
   Returns stdout stream of to_cmd
```
nlu90: return doc also needs to be updated

heron/shell/src/python/utils.py
nlu90: kk

PR #1893, opened 2017-05-24
Add SlaveTester to centralize duplicate slave, looper, communicator test logic

billonahill:
These tests all have to duplicate the same verbose boilerplate. Centralizing that in `SlaveTester`.

### DIFF
--- a/heron/instance/tests/java/com/twitter/heron/instance/spout/SpoutInstanceTest.java
+++ b/heron/instance/tests/java/com/twitter/heron/instance/spout/SpoutInstanceTest.java
@@ -18,29 +18,22 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import com.google.protobuf.ByteString;
 
 import org.junit.After;
-import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
-import org.junit.BeforeClass;
 import org.junit.Test;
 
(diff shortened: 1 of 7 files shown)


billonahill:
I broke out the `CommunicatorTester` from the `SlaveTester` since we have two use cases:

1. The test requires a `Slave` with `Communicators` that use a `SlaveLooper`. This is the case for the bolt/spout tests in this review.
2. The test requires only `Communicators` that use an `NIOLooper`. This is the case for all tests under `heron/instance/tests/java/com/twitter/heron/network` which I'll submit ion a follow up review.

billonahill:
Ping. Any takers on this review?

objmagic:
This PR is quite large but I'll find some time today to review.

heron/instance/tests/java/com/twitter/heron/instance/spout/SpoutInstanceTest.java
```diff
@@ -70,106 +63,51 @@
...
   private static final String SPOUT_INSTANCE_ID = "spout-id";
-  private static IPluggableSerializer serializer;
-  private WakeableLooper testLooper;
-  private SlaveLooper slaveLooper;
+  private static IPluggableSerializer serializer = new JavaSerializer();
 
   // Singleton to be changed globally for testing
   private AtomicInteger ackCount;
   private AtomicInteger failCount;
   private PhysicalPlans.PhysicalPlan physicalPlan;
```
nlu90: for the `physicalPlan` seems each test has its own value, can we make it a local variable?

billonahill:
Thanks @objmagic. To make the review easier, it's really just centralizing copied logic in 2 new classes, `SlaveTester` and `CommunicatorTester`. The rest is just removing the code from the other classes that can now leverage them.

PR #1894, opened 2017-05-24
Remove heron core fetching for MarathonScheduler.

cckellogg:
The heron core will be installed within the docker image.

### DIFF
--- a/website/content/docs/operators/deployment/schedulers/dcos.md
+++ b/website/content/docs/operators/deployment/schedulers/dcos.md
@@ -93,7 +93,8 @@ heron.class.launcher:                        com.twitter.heron.scheduler.maratho
 # location of java - pick it up from shell environment
 heron.directory.sandbox.java.home:          $JAVA_HOME
 
-heron.directory.home:                       "./heron-core/"
+# location of heron - pick it up from shell environment
+heron.directory.home:                       $HERON_HOME
 
 heron.directory.conf:                       "./heron-conf/"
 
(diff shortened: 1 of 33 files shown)


cckellogg:
@jrcrawfo what do you think of this approach?

jrcrawfo:
In general, I don't think we should do this unless the Docker build process for adding those core binaries into the image are made available. Otherwise, it's difficult to test any changes to a new build since the Dockerfile is not available

cckellogg:
What about adding a flag in the config on whether or not to download the heron core into the docker container?

jrcrawfo:
I really like the idea of including the the core binaries in the image, but the Dockerfile you used to put the libraries in there should be made available to the user and built into the build process. Otherwise, it just seems very disconnected and when a new release is cut, there's no process for getting that new release built into a new docker image to be ran.

srkukarni:
+1

huijunw:
Will this PR be included in the 0.14.8?

objmagic:
This is not for 0.14.8. And please fix merge conflict. Also someone please review it.

kramasamy:
Ideally it will be nice to merge this - but we have some additional work to get this completed.

cckellogg:
I added a pull request to compile and build a docker image for running heron. https://github.com/twitter/heron/pull/1934.

website/content/docs/operators/deployment/schedulers/dcos.md
```diff
@@ -98,11 +99,8 @@ heron.marathon.scheduler.uri: "<core.dcos_url>/service/marathon"
 # Invoke the IScheduler as a library directly
 heron.scheduler.is.service:                  False
 
-# location of the core package
-heron.package.core.uri:                      https://github.com/twitter/heron/releases/download/0.14.7/heron-core-0.14.7-ubuntu.tar.gz
-
-# docker repo for executor
-heron.executor.docker.image: 'ndustrialio/heron-executor:jre8'
+# docker repo for heron with core packages installed
+heron.executor.docker.image: 'streamlio/heron:latest-ubunut14.04'
```
jrcrawfo: With the tag, the only comment i have is to be sure we have tags corresponding to versions number. So, ideally, you'd be referencing something like `streamlio/heron:0.14.8`

jrcrawfo:
Other than the Docker tagging comment above, which is something to be done outside of this anyway, this looks good 👍

website/content/docs/operators/deployment/schedulers/dcos.md
cckellogg: We are planning on publishing images for different operating systems so we came up with the convention heron:<heron-version>-<os-version>. So would publish a release image like this streamlio/heron:0.14.8-ubuntu14.04. What do you think?

PR #1898, opened 2017-05-25
Added helper methods for stateful operations

srkukarni:


heron/common/src/cpp/config/topology-config-helper.cpp
```diff
@@ -197,5 +197,60 @@ sp_int64 TopologyConfigHelper::GetContainerRamRequested(const proto::api::Topolo
       (total_parallelism / nstmgrs) + (total_parallelism % nstmgrs);
   return max_components_per_container * 1073741824l;
 }
+
+bool TopologyConfigHelper::IsTopologyStateful(const proto::api::Topology& _topology) {
```
billonahill: The 3 bool methods have identical implementations except for the `TopologyConfigVars` object checked. Can you have each method instead delegate to an internal helper that centralizes this logic that takes a `TopologyConfigVars` object?

heron/common/src/cpp/config/topology-config-helper.h
```diff
@@ -74,6 +75,22 @@ class TopologyConfigHelper {
 
   // Gets the per container ram requested by this topology
   static sp_int64 GetContainerRamRequested(const proto::api::Topology& _topology);
+
+  // Is the topology stateful
+  static bool IsTopologyStateful(const proto::api::Topology& _topology);
+
+  // Is the topology exactly once
+  static bool IsTopologyExactlyOnce(const proto::api::Topology& _topology);
```
billonahill: Why is this method required?

heron/common/src/cpp/config/topology-config-vars.cpp
```diff
@@ -39,6 +39,11 @@ const sp_string TopologyConfigVars::TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS =
 const sp_string TopologyConfigVars::TOPOLOGY_ENABLE_ACKING = "topology.acking";
 const sp_string TopologyConfigVars::TOPOLOGY_CONTAINER_CPU_REQUESTED = "topology.container.cpu";
 const sp_string TopologyConfigVars::TOPOLOGY_CONTAINER_RAM_REQUESTED = "topology.container.ram";
+const sp_string TopologyConfigVars::TOPOLOGY_STATEFUL_ENABLED = "topology.stateful.enabled";
+const sp_string TopologyConfigVars::TOPOLOGY_STATEFUL_CHECKPOINT_INTERVAL_SECONDS =
+                            "topology.stateful.checkpoint.interval.seconds";
+const sp_string TopologyConfigVars::TOPOLOGY_STATEFUL_START_CLEAN = "topology.stateful.start.clean";
+const sp_string TopologyConfigVars::TOPOLOGY_EXACTLYONCE_ENABLED = "topology.exactlyonce.enabled";
```
billonahill: Where does this field get set? Is this controlled by the user? The topology sets acking enabled or not for at least and at most once. How does exactly once get enabled and how does that relate to acking settings?

heron/common/src/cpp/config/topology-config-helper.h
srkukarni: This will be required in stmgrs to see whether marker propagation needs to be done

heron/common/src/cpp/config/topology-config-vars.cpp
nlu90: It's set in the topology config: https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/Config.java#L472

I think we need to clarify the relations of acking and exactly once and educate our users to set them properly. 

Also, I'll suggest if a topology is set to exactly once, then this semantic will override the acking setting.

billonahill: The user has 3 options: exactly once, at most once and at least once. Can this Config setting take an enum of one of those 3 and do the right thing re acking etc? If we have a simple approach like that, there is no user education required about the various combinations of setting required for the different semantics.

srkukarni: I feel that we kind of made that choice some time back. IMO it belongs to a different discussion and not really releveant to this pr.

heron/common/src/cpp/config/topology-config-helper.cpp
billonahill: Can IsTopologyStateful also call GetBooleanConfigValue?

srkukarni: Sorry, missed that. Fixed

heron/common/src/cpp/config/topology-config-vars.cpp
billonahill: This PR is adding support for a boolean for exactly once, which I don't think we should do, so it is relevant. Doing so will create a poor user experience/design where the right combination of isExactlyOnce and acking needs to be set to for people to get what they want, which is exactly once, at most once or at least once, which is how users think about semantics. Can we design this properly and take an enum? What's the reasoning to not do so now? The enum can only have ExactlyOnce for now if you don't want to fix the other parts. If we did that we at least wouldn't be introducing a new config API that should be deprecated upon launch.

srkukarni: This pr is not adding any new user facing config variables. It is just the implementation part of the already existing api(ACKING ENABLED, EXT1_ONCE, etc) that already exists in the Config.java which is the main user api. As such this api just addresses the implementation part.
There might be a new pr needed to clean up the user facing api(which is really is Config.java). But that is not this pr. And that has wider implications. If/once we do that, the underlying implementation like this one, can also change at that point

billonahill: Apologies, I missed that `Config.setTopologyExactlyOnceEnabled(boolean)` was already committed. That should be fixed before we ship. We've discussed the need for better support for how users specify semantics in the past, and the desire to get rid of the acking setting required by the user (acking should be an internal implementation detail used to support at least once). We should do that refactor before exactly once ships so we don't perpetuate a confusing design. But that can be done in another PR. Can we follow up with that after this one so we don't write more code that works with a boolean instead of an enum? The further down that path we go the more painful the refactor will be.

srkukarni: We could certainly clean that up in a later pr. However one good thing is that all of these methods are abstracted out by the config helpers. So when we make that change, its just a matter of making the change in this class rather than in any other component.

srkukarni: BTW I also feel that the other change has wider implications. It would need a careful migration policy since existing topologies will have to be recompiled.
One good thing however is that heron api itself is not widely used, but the wrapper apis(Storm, etc) are. So even when we change it later, it would still have very little impact on other users.

billonahill: I would envision we deprecate the `setAckingEnabled` method for a few releases while updating it's implementation to set the semantics enum to atLeastOnce. Then we can migrate before ultimately removing the acking setter.

billonahill: Opened issue #1904 to track refactoring out `setTopologyExactlyOnceEnabled`

PR #1901, opened 2017-05-25
add ckptmgr client in tmaster

nlu90:


kramasamy:
@srkukarni - Should we merge this?

PR #1914, opened 2017-05-27
Removed per client/server specific acquire/release and consolidated them in one place

srkukarni:
In-order to remove the excessive cost of new/deletes of the google protobuf objects, Heron introduced protobuf pool in https://github.com/twitter/heron/pull/1421. This implementation relied on a per client/server proto pool implemented as part of the Server/Client classes. While this improved performance, it's drawback was that the proto pools were per client/server specific. Thus if say a stmgr connects with two remote stmgrs, each client would have its own proto pool. This is a) unnecessary and b) suboptimal because varying traffic patterns means that while some protopools could grow fast and could remain unused later on, others might see traffic increases and might have to resort to doing new/deletes cater for the increased traffic. This results in suboptimal performance and increased memory usage. This pr solves that by making a single unified protopool that everyone shares across.


heron/common/src/cpp/network/mempool.cpp
```diff
@@ -0,0 +1,25 @@
+/*
```
nlu90: should we put this `mempool` into `common/src/cpp/basics` dir?

heron/common/src/cpp/network/server.h
```diff
@@ -205,18 +205,6 @@ class Server : public BaseServer {
   // Called when the connection is closed
   virtual void HandleConnectionClose_Base(BaseConnection* connection, NetworkErrorCode _status);
 
-  MemPool<google::protobuf::Message> _heron_message_pool;
```
nlu90: the method `InstallRequestHandler` and `InstallMessageHandler` on line 150 and 160 are also using the `new` and `delete` operator manage message memory, should we also update them？

heron/common/src/cpp/network/mempool.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// Just defines the static member of Mempool class
+////////////////////////////////////////////////////////////////////////////////
+#include "network/mempool.h"
+#include "basics/basics.h"
+
+MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
+                                   new MemPool<google::protobuf::Message>();
+std::mutex __global_protobuf_pool_mutex__;
```
nlu90: the process is single-thread, any specific reason to use mutex to guard the access to the pool?

heron/common/src/cpp/network/mempool.h
```diff
@@ -16,8 +16,10 @@
 #ifndef MEM_POOL_H
 #define MEM_POOL_H
 
+#include <google/protobuf/message.h>
 #include <vector>
 #include <unordered_map>
+#include <mutex>
 #include <typeindex>
 
 template<typename T>
```
nlu90: line 59 mentioned memory leak, can we solve the issue also?

heron/common/src/cpp/network/mempool.cpp
srkukarni: Makes sense. Done

heron/common/src/cpp/network/mempool.cpp
srkukarni: Made it generic so that even multi threaded processes can use the interface.

heron/common/src/cpp/network/server.h
srkukarni: The Install methods are only in the control plane. Not worth using pool imo.

heron/common/src/cpp/network/mempool.h
srkukarni: Not sure what the memory leak here is. Ideally would have had more comments.
I want to keep this change as logic less as possible so maybe we can find the leak and fix it in a separate pr.

maosongfu:
Can you add more context in the description? 
For instance:
- How current heron works
- What is the problem of current way
- How this pull request works
- What will change?

Thanks

srkukarni:
@maosongfu updated pr.

PR #1917, opened 2017-05-30
update STATEFUL_YAML to STATEFUL_CONFIG_YAML

nlu90:
The field for stateful config file defined in AuroraField.java is `STATEFUL_CONFIG_YAML`, so update the heron.aurora file to use it correctly.


objmagic:
LGTM. See https://github.com/twitter/heron/issues/1918. We should really break this long line into separate lines so that diff could make better sense.

nlu90:
@objmagic the issue you mentioned will be fixed in another PR

PR #1925, opened 2017-06-01
Adding sleep to TestSpout to throttle down cpu usage

billonahill:
This issue came up in the forums:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/heron-users/xxFlKHCAUU4/5WFt8s0zAwAJ

It's annoying because it cranks the cpu, which is noisy and kills laptop batteries when not plugged in. :)

### DIFF
--- a/heron/examples/src/java/com/twitter/heron/examples/ExclamationTopology.java
+++ b/heron/examples/src/java/com/twitter/heron/examples/ExclamationTopology.java
@@ -14,6 +14,7 @@
 
 package com.twitter.heron.examples;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


objmagic:
sometimes we use it to benchmark in cluster so maybe we don't want to throttle it all the time?

billonahill:
@objmagic it's easy to disable by using the default constructor. I recommend we create our benchmark topologies that way, but the tutorial topology we reference in the quickstart shouldn't drain the cpu.

objmagic:
ok. lgtm

nlu90:
@billonahill 

This integration test: "BasicTopologyOneTaskScaleUp" is flaky. I've seen it fails the internal CI several times. Any ideas about this test?

billonahill:
Not yet, but I'm trying to reproduce. I hadn't seen this one in a while, so I thought it was ok...

nlu90:
I also saw this test failed internally. If you like, I can share the failed build number and log with you

billonahill:
Yeah, I saw that. I was also able to reproduce locally. I think there's a potential race condition with when the physical plan gets updated vs when the stream managers get restarted, but still investigating.

PR #1926, opened 2017-06-02
When the python pex file is copied over to the container, only base name should be refered

srkukarni:
We copy the pex file into the ./ directory of the container. Thus if I do
heron submit ... <path/to/my/pex/topology.pex> only topology.pex should be referenced as the topology bin

### DIFF
--- a/heron/tools/cli/src/python/submit.py
+++ b/heron/tools/cli/src/python/submit.py
@@ -94,7 +94,7 @@ def launch_a_topology(cl_args, tmp_dir, topology_file, topology_defn_file, topol
       "--release_file", release_yaml_file,
       "--topology_package", topology_pkg_path,
       "--topology_defn", topology_defn_file,
-      "--topology_bin", topology_file   # pex file if pex specified
+      "--topology_bin", os.path.basename(topology_file)   # pex file if pex specified
   ]
 
   if Log.getEffectiveLevel() == logging.DEBUG:


billonahill:
I suspect there are reasons why the `SubmitterMain` would want the entire path. Should we truncate it lower down in the call stack? Would this work in cases where the defn file isn't in the rood directory where the client is invoked? I'd expect the submitter needs the full path to the file right?

srkukarni:
The submitter main actually does nothing with this argument. All the packing of the defn/bin happens inside the cli before calling the submitter. The submitter merely copies this as the arg to the executor.

srkukarni:
Disagree with that. Infact this looks like a bug to me. 
Schedulers should not be doing this truncating business. It should be done at a base layer(which is cli). Otherwise we will have the same situation like the one we have today where some schedulers handle it properly and others dont. For instance, aurora handles it fine(does a basename, and thats pretty much the only thing that it does). The others(including LocalScheduler) do not do that.
Unless there is a strong reason why schedulers would want the absolute filename, I would vote for merging this change.

billonahill:
I agree it would be best to handle it centrally, but it should be in the java code of scheduler-core somewhere. The `SubmitterMain` and the rest of the scheduler-core code might need the full path. Also the contract `SubmitterMain` provides to it's callers is that of a full path, not the name of a file at some unknown location on the filesystem:
```
    Option topologyJar = Option.builder("f")
        .desc("Topology jar/pex file path")
        .longOpt("topology_bin")
        .hasArgs()
        .argName("topology binary file")
        .required()
        .build();
```
The python wrapper should be as thin as possible, just passing through the args passed to it in this case. Best to let the scheduler-core core use the local FQ path to determine and set the relative filename that the scheduler impl will find on the remote scheduler.

srkukarni:
I believe that comment needs to be changed. And there should be no reason why the scheduler needs fq path. Please let me know why would the scheduler would need it.

srkukarni:
Ultimately the scheduler abstraction should be as focused as possible. It needs to get bare minimum things a) topology contents and b) parameters affecting scheduler like number of containers and container description(ram/cpu, etc). Ultimately the scheduler's main job is to just act as the interface to the actual scheduler implementation to fire off the containers and start the executor on each container. The rest should be handled outside.

objmagic:
So I checked our schedulers. Some of them does truncation using `FileUtils.getBaseName`, some of them invoke [`executorCommandArgs`](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java#L189) to invoke `getBaseName` indirectly. I think we should do truncation for all schedulers and do it once at one place only.

srkukarni:
Great that we agree that we need to do this at a central place :)
Only debate now is where.

billonahill:
The submitter takes a number of (valid) local file paths and uses them to build configs to represent what will be the remote settings on the cluster. I recommend we stick with that pattern and do that transformation where we make the remote configs. That's here in `SubmitterMain`:
```
  protected static Config topologyConfigs(
      String topologyPackage, String topologyBinaryFile, String topologyDefnFile,
      TopologyAPI.Topology topology) {
    PackageType packageType = PackageType.getPackageType(topologyBinaryFile);

    return Config.newBuilder()
        .put(Key.TOPOLOGY_ID, topology.getId())
        .put(Key.TOPOLOGY_NAME, topology.getName())
        .put(Key.TOPOLOGY_DEFINITION_FILE, topologyDefnFile)
        .put(Key.TOPOLOGY_PACKAGE_FILE, topologyPackage)
        .put(Key.TOPOLOGY_BINARY_FILE, topologyBinaryFile)
        .put(Key.TOPOLOGY_PACKAGE_TYPE, packageType)
        .build();
}
```
https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java#L81

I don't love the inconsistency of mixing remote and local paths in `topologyDefnFile`, `topologyPackage`, `topologyBinaryFile` here as doing so has bit us in the past, but we can at least centralize the fix for this field here. When doing so we should remove all downstream `FileUtils.getBaseName` calls since we're changing the contract to where the impl already provides that now, which we should also document in `Key.TOPOLOGY_BINARY_FILE`.

srkukarni:
So here is the interesting part. The cli already does a basename conversion while preparing the topology.tar.gz. See https://github.com/twitter/heron/blob/master/heron/tools/common/src/python/utils/config.py#L61
Moreoever the submitter main at any point does not require fq path of either the defn file/topology binary. How about consolidating all of that within cli and cleaning all basename stuff in all schedulers?

srkukarni:
Also note that none of the config files are really available to the scheduler. The cli bundles all of them inside the topology.tar.gz.

PR #1929, opened 2017-06-03
add stateful checkpointer into tmaster

nlu90:
The tmaster stateful utils that issue checkpoint requests to stmgrs and tracking if the a global consistent checkpoint is done.

No logic changes in tmaster code, so won't affect any production performance now.

### DIFF
--- a/heron/tmaster/src/cpp/manager/stateful-checkpointer.cpp
+++ b/heron/tmaster/src/cpp/manager/stateful-checkpointer.cpp
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2015 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "manager/stateful-checkpointer.h"
+#include <iostream>
+#include <sstream>
+#include <chrono>
+#include <string>
+#include <vector>
+#include "config/physical-plan-helper.h"
+#include "manager/tmaster.h"
+#include "manager/stmgrstate.h"
+#include "errors/errors.h"
+
+namespace heron {
+namespace tmaster {
+
+StatefulCheckpointer::StatefulCheckpointer(
+  std::chrono::high_resolution_clock::time_point _tmaster_start_time)
+  : tmaster_start_time_(_tmaster_start_time) {
+  // do nothing
+}
+
+StatefulCheckpointer::~StatefulCheckpointer() { }
+
+sp_string StatefulCheckpointer::GenerateCheckpointId() {
+  // TODO(skukarni) Should we append any topology name/id stuff?
+  std::ostringstream tag;
+  tag << tmaster_start_time_.time_since_epoch().count()
+      << "-" << time(NULL);
+  return tag.str();
+}
+
+void StatefulCheckpointer::StartCheckpoint(const StMgrMap& _stmgrs) {
+  // Generate the checkpoint id
+  sp_string checkpoint_id = GenerateCheckpointId();
+
+  // Send the checkpoint message to all active stmgrs
+  LOG(INFO) << "Sending checkpoint message with id: " << checkpoint_id
+            << " to all stmgrs";
+  for (auto iter = _stmgrs.begin(); iter != _stmgrs.end(); ++iter) {
+    proto::ckptmgr::StartStatefulCheckpoint request;
+    request.set_checkpoint_id(checkpoint_id);
+    iter->second->NewStatefulCheckpoint(request);
+  }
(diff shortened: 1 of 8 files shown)


heron/tmaster/src/cpp/manager/tmaster.h
```diff
@@ -37,6 +37,10 @@ class StatsInterface;
 class TMasterServer;
 class TMetricsCollector;
 
+typedef std::map<std::string, StMgrState*> StMgrMap;
+typedef StMgrMap::iterator StMgrMapIter;
+typedef StMgrMap::const_iterator StMgrMapConstIter;
```
billonahill: Why is `StMgrMapConstIter` needed? I don't see it used anywhere.

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
```diff
@@ -0,0 +1,71 @@
...
+namespace heron {
+namespace proto {
+class PhysicalPlan;
+}
+}
+
+namespace heron {
+namespace tmaster {
+
+class StatefulCheckpointer {
```
billonahill: please add class-level docs describing what this class does and what it's responsibilities are.

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
```diff
@@ -0,0 +1,71 @@
...
+  explicit StatefulCheckpointer(std::chrono::high_resolution_clock::time_point _tmaster_start_time);
+  virtual ~StatefulCheckpointer();
+  void RegisterNewPhysicalPlan(const proto::system::PhysicalPlan& _pplan);
+
+  void StartCheckpoint(const StMgrMap& _stmgrs);
+
+  // Called when a InstanceStateStored message is received
+  // Return true if this completes a globally consistent checkpoint
+  // for this _checkpoint_id
+  bool HandleInstanceStateStored(const std::string& _checkpoint_id,
```
billonahill: who calls `HandleInstanceStateStored`? Why does this need to be public?

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
```diff
@@ -0,0 +1,71 @@
...
+#include <unordered_set>
+#include "network/network.h"
+#include "proto/tmaster.pb.h"
+#include "manager/tmaster.h"
+#include "basics/basics.h"
+
+
+namespace heron {
+namespace proto {
+class PhysicalPlan;
```
billonahill: Why is this redefined here?

heron/tmaster/src/cpp/manager/stateful-checkpointer.cpp
```diff
@@ -0,0 +1,98 @@
...
+  : tmaster_start_time_(_tmaster_start_time) {
+  // do nothing
+}
+
+StatefulCheckpointer::~StatefulCheckpointer() { }
+
+sp_string StatefulCheckpointer::GenerateCheckpointId() {
+  // TODO(skukarni) Should we append any topology name/id stuff?
+  std::ostringstream tag;
+  tag << tmaster_start_time_.time_since_epoch().count()
```
billonahill: How do we know that a checkpoint is valid for a given packing plan? Should we build this into the id or bind it some other way? We will need this to know if a checkpoint can be assigned to a redeployed topology with a different packing plan.

heron/tmaster/src/cpp/manager/tmaster.h
nlu90: removed. Will add it when used in the future code

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
nlu90: This is not redefining but forward declaration.  Details about forward declaration could be found here: https://stackoverflow.com/questions/4757565/what-are-forward-declarations-in-c

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
nlu90: added

heron/tmaster/src/cpp/manager/stateful-checkpointer.h
nlu90: When tmaster received the InstanceStateStored message, it calls this method to notify the checkpointer. And also, checkpointer will let tmaster know if all states for the checkpoint_id are stored.

I'll update the comment

heron/tmaster/src/cpp/manager/stateful-checkpointer.cpp
nlu90: This is not a checkpointer's job. tmaster will be responsible for it.

nlu90:
unit tests added

heron/tmaster/tests/cpp/server/stateful_checkpointer_unittest.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include "errors/modinit.h"
+#include "threads/modinit.h"
+#include "network/modinit.h"
+#include "config/topology-config-vars.h"
+#include "config/topology-config-helper.h"
+#include "config/physical-plan-helper.h"
+#include "manager/stateful-checkpointer.h"
+
+// Generate a dummy topology
+static heron::proto::api::Topology* GenerateDummyTopology(
```
billonahill: I've seen this method in a number of test classes. If you have ideas for how to effectively share test code like this that would be awesome.

PR #1937, opened 2017-06-07
UpdateTopologyManager using stale topology

geoffret:
Hi all, not sure if I'm doing this right, but open for comments/improvements. 

#1924 fixes the Update mechanism using a stale copy of the Topology for determining whether to deactivate the topology before scaling.

This fixes another similar issue. It looks like the UpdateTopologyManager pulls the Topology before deactivating the running topology. UpdateTopologyManager then updates this stale copy (that still shows the topology as running) and pushes it to the state manager.

This fix just pulls an updated Topology from the state manager after deactivating.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -151,10 +151,8 @@ public class UpdateTopologyManager implements Closeable {
 
     // deactivate and sleep
     if (initiallyRunning) {
-      deactivateTopology(stateManager, updatedTopology);
       // Update the topology since the state should have changed from RUNNING to PAUSED
-      updatedTopology =
-        getUpdatedTopology(topologyName, proposedPackingPlan, stateManager);
+      updatedTopology = deactivateTopology(stateManager, updatedTopology, proposedPackingPlan);
     }
 
     // request new resources if necessary. Once containers are allocated we should make the changes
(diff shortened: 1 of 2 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -152,6 +152,9 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
     // deactivate and sleep
     if (initiallyRunning) {
       deactivateTopology(stateManager, updatedTopology);
+      // Update the topology since the state should have changed from RUNNING to PAUSED
+      updatedTopology =
```
billonahill: Great catch! Could we instead change `deactivateTopology` to return the new topology, and push the fetch into that method?

geoffret: Thanks, and sure I'll take a look at it this week. I think that'd require modifying the argument list for deactivateTopology so that the new packing plan is available to construct the final Topology. Is that okay? Another way to do it would be to return a fresh Topology and merge it with the new packing plan in updateTopology().

I first noticed it while reviewing logs for one of my spouts. I noticed that the spout would be activated right after receiving a NewInstanceAssignmentMessage. I then followed it back to the scheduler logs where I saw that the reactivation task would fail because the topology was already running.

billonahill: Ah right, we'd have to pass in proposedPackingPlan to the deactivateTopology method. That's unfortunate, but I think that's our best option. That way we can at least encapsulate the deactivateTopology logic to return a new valid topology that aligns with the one passed in.

billonahill:
👍 once CI passes.

geoffret:
Okay, sounds good! 

I had to made changes to the UpdateTopologyManagerTest to allow everything to build since the method signature changed, but didn't change any logic there since I'm not familiar with Mockito.

PR #1938, opened 2017-06-07
Add support for src task id in the storm api

srkukarni:
In Storm, the bolts can use the tuple.src_task_id() to find out where this tuple originated from. However currently stmgr batches tuples from different local source task_ids and thus the destination stmgrs do not have any idea of where a particular tuple came from. Several of current storm users use this feature and since Heron claims to be api compliant, we need to support this. This change makes the stmgr break tuple batches sent to other stmgrs based on the src task id so that each tuple batch sent to the remote stmgr is always coming from the same src task id. Thus stmgrs know exactly where a tuple originated and can pass this information to the instances.


billonahill:
Why do tuple batches need to be grouped by source task id? Since source task id is tracked at the tuple level, couldn't batches contain tuples from a mix of sources?

srkukarni:
We do not track src task id at tuple level but at batch level for performance reasons.

maosongfu:
> couldn't batches contain tuples from a mix of sources

-- I think this is due to the exactly-once algorithm. We need to be able to distinguish and control tuples from different sources, so we can control the markers.

PR #1939, opened 2017-06-07
Update documentation regarding topology_bin in the SubmitterMain

srkukarni:
Cli passes only the file name of the topology_bin.
Thus it's not necessary for the schedulers to get the basename.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
@@ -189,10 +189,10 @@ public class SubmitterMain {
         .build();
 
     Option topologyJar = Option.builder("j")
-        .desc("user heron topology jar/tar/pex file")
+        .desc("The filename of the heron topology jar/tar/pex file to be run by the executor")
         .longOpt("topology_bin")
         .hasArgs()
-        .argName("topology binary file")
+        .argName("topology binary filename on the cluster")
         .required()
         .build();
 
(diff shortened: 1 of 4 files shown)


objmagic:
Also [here](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java#L206), which is used by other scheduler indirectly.

objmagic:
also please fix style error in CI.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
```diff
@@ -189,7 +189,7 @@ private static Options constructOptions() {
         .build();
 
     Option topologyJar = Option.builder("j")
-        .desc("user heron topology jar/pex file path")
+        .desc("user heron topology jar/tar/pex file")
         .longOpt("topology_bin")
         .hasArgs()
         .argName("topology binary file")
```
billonahill: Is there a way to covey that this doesn't actually represent a file or path on the local disk, but the name of the file that will be invoked on the executor?

srkukarni:
@objmagic We cannot remove that since a) that's topology defn file and b) scheduler currently needs to read topology_defn. 
Fixed style error.

objmagic:
@srkukarni yes you are right.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
srkukarni: Suggestions please?

billonahill: .desc("The remote filename of the heron topology jar/tar/pex file to be run by the executor.")
.argName("topology binary filename on the cluster")

srkukarni:
Made the changes wrt arg description based on feedback. Please take a look again. Thanks!

PR #1940, opened 2017-06-07
add stateful restorer

nlu90:
The stateful restorer component. No logic changes in the tmaster code.


heron/tmaster/src/cpp/manager/stateful-restorer.h
```diff
@@ -0,0 +1,56 @@
...
+#include "manager/tmaster.h"
+
+namespace heron {
+namespace tmaster {
+
+class StatefulRestorer {
+ public:
+  StatefulRestorer();
+  virtual ~StatefulRestorer();
+  // Start a new 2PC with this checkpoint_id
```
billonahill: Is 2PC 2 Phase Commit? If so can we expand that here and elsewhere in this patch. That abbreviation is not obvious to the code reader.

heron/tmaster/src/cpp/manager/stateful-restorer.cpp
```diff
@@ -0,0 +1,81 @@
...
+StatefulRestorer::StatefulRestorer()
+  : in_progress_(false),
+    restore_txid_(0) {
+}
+
+StatefulRestorer::~StatefulRestorer() { }
+
+void StatefulRestorer::StartRestore(const std::string& _checkpoint_id, const StMgrMap& _stmgrs) {
+  if (in_progress_) {
+    LOG(WARNING) << "Starting a Restore for checkpoint "
```
billonahill: Can you clarify what happens in the log message (e.g., do we ignore the request or start another one?). If we're able to start and do another one, why is it a warning and how does shared state of this class in_progress_, restore_txid_, etc) not collide?

heron/tmaster/src/cpp/manager/stateful-restorer.h
nlu90: good point. will update it

heron/tmaster/src/cpp/manager/stateful-restorer.cpp
nlu90: basically what happens here is if a new start restore requested, the previous one will be ignored and a new storing is started.

nlu90: The warning is to indicated that a new restoring is started.

Given that only a single thread executing the tmaster code, everything will be ordered. So there's no race conditions here. Either the states remains unchanged or they are all updated to next states.

And we are checking if the checkpoint_id and restore_txid are expected in the `HandleStMgrRestored` callback, so only the response with latest ids will be used to update the `unreplied_stmgrs`.

billonahill: In that case can you include in the log that the restore with id checkpoint_id_in_progress_ will be abandoned?

nlu90: sure

heron/tmaster/src/cpp/manager/stateful-restorer.cpp
```diff
@@ -0,0 +1,83 @@
...
+                 << _checkpoint_id << " when we are already busy"
+                 << " within Restore of " << checkpoint_id_in_progress_
+                 << ". The Restore result of " << checkpoint_id_in_progress_
```
billonahill: s/Restore/restore

nlu90: fixed

heron/tmaster/src/cpp/manager/stateful-restorer.cpp
```diff
@@ -0,0 +1,83 @@
...
+    restore_txid_(0) {
+}
+
+StatefulRestorer::~StatefulRestorer() { }
+
+void StatefulRestorer::StartRestore(const std::string& _checkpoint_id, const StMgrMap& _stmgrs) {
+  if (in_progress_) {
+    LOG(WARNING) << "Starting a Restore for checkpoint "
+                 << _checkpoint_id << " when we are already busy"
+                 << " within Restore of " << checkpoint_id_in_progress_
```
billonahill: Please also replace the 2 other occurrences of Restore in this log message with restore.

billonahill:
👍

PR #1944, opened 2017-06-08
add a readme file for pyheron

kramasamy:
When we package pyheron for publishing to pypi, it expects a readme file. This readme file should be either README or README.txt or README.rst

### DIFF
--- a/scripts/packages/pyheron/README.txt
+++ b/scripts/packages/pyheron/README.txt
@@ -1,0 +1,2 @@
+PyHeron is the python API for Heron, the distributed streaming engine. The API is
+based on the Storm python API provided by Parsely.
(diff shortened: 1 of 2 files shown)


scripts/packages/pyheron/README.txt
```diff
@@ -0,0 +1,2 @@
+PyHeron is the python API for Heron, the distributed streaming engine. The API is
+based on the Storm python API provided by Parsely.
```
billonahill: Can you provide a link to Parsely for reference. Also should we change to README.md to be consistent with other readmes in the repo?

kramasamy: This is required for Python packaging - so that we can submit to pypi. When we package it, the packager specifies that the readme file should either README or README.txt or README.rst (not README.md)

kramasamy: Link is added.

PR #1945, opened 2017-06-08
Add note on hostname in /etc/hosts

lucperkins:
MacOS users may have trouble running topologies locally if they don't add their machine's hostname to `/etc/hosts`. This PR adds a note about that to the Getting Started doc.

I also fixed a persistent issue with shortcodes that have been affecting the website build process. It turns out that shortcodes must be HTML and cannot be in the Ace template format.


objmagic:
LGTM.

PR #1948, opened 2017-06-10
Scripts to build docker artifacts

kramasamy:
Several scripts and docker files to build docker image artifacts for release. These docker scripts expect the artifacts already to be built and available in a directory before they can package them into docker. Docker images are saved as tar.gz files that are easy to publish. Once these files are published the user can download them, load them and push into private registry, if needed. Public docker images will be pushed into docker hub.


cckellogg:
ship it

PR #1952, opened 2017-06-13
[DO NOT MERGE] Topology self-regulation using Heron

ashvina:
This PR is related to Heron and Dhalion integration. This work was briefly discussed earlier in #1781. Dhalion integration will enable self-regulation of Heron Topologies. For e.g. it could be used to auto-scale a under-provisioned topology. Another use case is auto-tuning of a topology to meet certain throughput SLO.

1. Dhalion collects metrics from `tracker` using `TrackerMetricsProvider`. Currently it consumes `back-pressure`, `execution-count` and `pending-buffer-counts` metrics.
2. Then Dhalion executes rules using `detectors`. For e.g. a rule could be there should not be any back-pressure. A `symptom` or trigger is created if a rule is violated. In this PR, the rules are implemented in `BackPressureDetector`, `DataSkewDetector`, `LargeWaitQueueDetector` and `WaitQueueDisparityDetector`.
1. Then Dhalion tries to analyze the root cause of the violation using `Diagnosers`. It correlates all the `symptoms` created by `Detectors` in the previous stage. The PR includes `UnderProvisioningDiagnoser`, `DataSkewDiagnoser` and `SlowInstanceDiagnoser`.
1. Once a root cause is know, Dhalion will try to resolve the issue using `Resolvers`. For e.g. `ScaleUpResolver` will be invoked to scale up the topology if `UnderProvisioning` is the root cause.
1. Dhalion runs in its own process. A new process will be launched for each topology. This process,`HealthManager` will read heron configuration, initialize state managers and invoke Dhalion.
1. The behavior of `HealthManager` can be customized using `healthMgr.yaml` configuration file. This is similar to the way other `yaml` files are used in heron.


WORKSPACE
```diff
@@ -336,3 +426,13 @@ maven_jar(
   name = "com_101tec_zkclient",
   artifact = "com.101tec:zkclient:0.3"
 )
+
+http_jar(
+  name = "com_microsoft_dhalion",
+  url = "https://github.com/ashvina/dhalion/releases/download/v0.0.1/dhalion-0.0.1.jar",
```
ashvina: 👍  
Fixed in #1980

heron/healthm/tests/java/com/twitter/heron/healthmgr/common/PackingPlanProviderTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.twitter.heron.healthmgr.common.HealthManagerEvents.TopologyUpdate;
+import com.twitter.heron.packing.roundrobin.RoundRobinPacking;
+import com.twitter.heron.proto.system.PackingPlans;
+import com.twitter.heron.spi.packing.PackingPlan;
+import com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor;
+import com.twitter.heron.spi.utils.PackingTestUtils;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
```
ashvina: Fixed. Somehow, checkstyle on my dev box did not catch this issue.

billonahill: I've had this issue too. For some reason checkstyle doesn't always seem to fire on mac and I haven't been able to figure out why.

PR #1955, opened 2017-06-14
Change TMaster to watch the packing plan for changes

### LINKED ISSUE #1932: Possible race condition during Heron update
During the Heron update command the following flow is orchestrated by the scheduler:

1. scheduler: deactivate
  a. tmaster: deactivated
1. scheduler: delete physical plan
1. scheduler: update packing plan in state manager
  a. executors: detects new packing plan
  b. executor0: refreshes process list, restarts TMaster
    1. tmaster: restarts with new packing plan
    1. tmaster: receive new strmgr registrations
    1. tmaster: satisfies new packing plan
    1. tmaster: creates/distributes new physical plan
1. scheduler: await new physical plan
1. scheduler: receive new physical plan
1. scheduler: activate
  a. tmaster: activated

The scheduler deletes the physical plan so that the TMaster will restart with a new packing plan, recognize stream manager registration requests and build a new physical plan representing the new packing plan.

Since the executors all refresh their packing plans in parallel, a potential race condition exists if executor0 takes a long time to restart the TMaster (i.e. step 3b) (this happened due to a python issue fixed by #1930).

In this case new stream managers will register themselves with the old TMaster. The TMaster will generate a new physical plan that matches of old packing plan (if it can) and update it in the state manager. The scheduler will detect the new packing plan and complete the update, but the topology will still be in the old packing plan state (new scale-out stream managers will go unrecognized). Finally the TMaster will get restarted by the executor with new packing plan. It will load the old physical plan and reject registration requests by the new stream managers.

What is needed is a guarenteed way for the scheduler to invalidate/update the packing plan on the TMaster that doesn't rely on an asynchronous change propagating through the stream manager.

billonahill:
Fixes #1932.

During a scaling event, instead of the scheduler deleting the packing plan and the executor restarting the tmaster with new stream manager ids, the TMaster will fetch the packing plan from the state manager and watch it. The packing plan is first fetched during initialization. Later if it updates, the TMaster updates it's state to reflect the new packing plan and stream manager list.

The main classes to review are `tmaster.cpp`, `UpdateTopologyMaster.java` and `heron_executor.py`. The rest of the classes are just incorporate the new tmaster changes.


heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -113,7 +108,7 @@ TMaster::TMaster(const std::string& _zk_hostport, const std::string& _topology_n
   tmaster_location_->set_master_port(master_port_);
   tmaster_location_->set_stats_port(stats_port_);
   DCHECK(tmaster_location_->IsInitialized());
-  EstablishTMaster(EventLoop::TIMEOUT_EVENT);
+  EstablishPackingPlan(EventLoop::TIMEOUT_EVENT);
```
srkukarni: Why Packing Plan is fetched ahead of TMaster set? One invariant is that we should be confirmed as Tmaster before doing anything

heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -138,6 +133,68 @@ TMaster::TMaster(const std::string& _zk_hostport, const std::string& _topology_n
...
+        absent_stmgrs_.insert("stmgr-" + std::to_string(packing_plan_->container_plans(i).id()));
+      }
+
+      // chain TMaster location setting to the completion of packing plan info above
+      EstablishTMaster(EventLoop::TIMEOUT_EVENT);
+    } else {
+      if (packing_plan_ != newPackingPlan) {
+        LOG(INFO) << "Packing plan changed. Deleting physical plan and restarting TMaster to "
+                  << "reset internal state. Exiting";
+        state_mgr_->DeletePhysicalPlan(tmaster_location_->topology_name(),
```
srkukarni: Are we sure we are the tmaster here?

heron/tmaster/src/cpp/manager/tmaster.cpp
billonahill: Fetching the packing plan is required for initialization so we can determine the stream manager id set (`absent_stmgrs_`). We need to be prepared with this info before TMaster registers it's location and starts receiving registration requests from stream managers.

heron/tmaster/src/cpp/manager/tmaster.cpp
billonahill: No, we're not actually. What is the deterministic way to know that? We could set a flag during a successful `SetTMasterLocationDone` call to indicate that we're TMaster since we successfully set our location.

billonahill: @srkukarni I added a flag to denote if we're master. LMKWYT

billonahill:
Ping. Any comments?

This approach still has some shortcomings which I suggest we iterate upon in follow-ups, but in it's current form it's is a big improvement over what we have now.

srkukarni:
Please look at my prev comment about the zk node hierarchy.

billonahill:
@srkukarni I see two previous comments of you, both of which I've addressed. What's the ZK node hierarchy comment?

srkukarni:
Sorry, looks like i hadn't submitted my comments(damn github).
Anyways, I really want the invariant to be maintained by the logic of the code rather than the is_master_ thing. I just checked the code, and it looks like we do not start serving until much later. Thus, I believe the following approach is better
1. Upon start, make sure you are the master(just like the current logic).(Btw this avoids the is_master_ thingie).
2. Upon becoming master, check for packing plan and make sure that it exists. 
3. Then continue with checking for physical plan and becoming master.
This way I believe you also reduce the size of this commit substantially

billonahill:
Why is it a bad thing to fetch the packing plan at initialization before being master? This is the safe thing to do. If we become master and start taking requests without initializing the TMaster we're going to have lots of failed RPCs from the stream managers since `absent_stmgrs_` won't be initialized. I'm not seeing the reasoning for why we'd want to do that.

Note that this is a totally different concern than deleting the physical plan upon update, for which we should assure we're the master somehow.

heron/tmaster/src/cpp/manager/tmaster.cpp
billonahill: Why is the proposed approach a better one? With the proposed approach tmaster would register itself as master, which would result in stream managers trying to connect to it. They will fail until tmaster continues it's initialization and opens its server socket, which seems undesirable.

srkukarni:
There are two steps here
1. Becoming the master(when we write the /tmasters zk node).
2. Serving as master(when we start the TMasterServer.
IMO, 1 should be the first step in the logic of tmaster. 2 can follow anytime. So as soon as you become tmaster, you can start the fetching packing plan and only after that you can start serving requests.
Right now we do not do 2 until much later(after physical plan fetch). So we will not get any requests from stmgrs.

srkukarni:
So, what's the issue with stmgrs trying to connect and retrying?

heron/tmaster/src/cpp/manager/tmaster.cpp
billonahill: If the intent is to not register the packing plan listener that handles updates (and physical plan deletion) until after we're fully up and running, that seems like a good call. We can differ the packing plan watch initialization until after we've become master, while still fetching the initial packing plan before registering location.

billonahill:
@srkukarni and I discussed on slack and have a plan, which I'll refactor to:

> we can address both of our concerns by fetching the initial packing plan before registering (when we start and it’s null), but then registering the packing plan watcher only once we’re tmaster. That way we can still front-load initialization, while guaranteeing that the packing plan watcher (which takes mutating actions on the topology as master during update) would only happen when it’s master.

PR #1959, opened 2017-06-14
Allow all file types to be downloaded from a container through the shell.

cckellogg:


objmagic:
LGTM in general. I feel there is some security issues here. For examples, it seems that you can use `../../../../..` to access files else where. This is not allowed in Twitter internal environment since info sec team has program which keeps doing such HTTP request to search for such security hole. And I guess if they find it they will ping our team.

cckellogg:
The tornado web server will take care of that it. The following request 
http://localhost:9999/../../../../../../../../../../etc/passwd. will be converted to /etc/pass and then fail because there is no handler to handle that route. This works the same as the file FileDataHandler in the shell (https://github.com/twitter/heron/blob/master/heron/shell/src/python/handlers/filedatahandler.py)

objmagic:
ok sounds good.

PR #1961, opened 2017-06-14
Streamline preparation of docker images for distribution

kramasamy:
- Add only minimum software needed to distribute and run in docker images
- Heron install should require only java run time checking, not javac checking
- Docker images distributed require for running in kubernetes. Including them in the docker dist file for all linux variations.


objmagic:
LGTM. Please go ahead to merge it.

PR #1969, opened 2017-06-16
Fix inaccurate execute-latency metric

maosongfu:
Currently in bolt instance, we would use an inaccurate start of time, causing the execute-latency higher than expected, especially when the actual execute-latency is small. For instance, the time spent in execute() could be 200+s in a minute. This is also an issue reported from Twitter internally.

This pull request fixes this issue by recording a more accurate start of time.

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
@@ -192,9 +192,8 @@ public class BoltInstance implements IInstance {
       int nValues = topologyContext.getComponentOutputFields(
           stream.getComponentName(), stream.getId()).size();
 
-      // Record current time at the start of processing a set of tuples
-      long currentTime = System.nanoTime();
       for (HeronTuples.HeronDataTuple dataTuple : tuples.getData().getTuplesList()) {
+        long startExecuteTuple = System.nanoTime();
         // Create the value list and fill the value
         List<Object> values = new ArrayList<>(nValues);
         for (int i = 0; i < nValues; i++) {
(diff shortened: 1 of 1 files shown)


objmagic:
LGTM. Some other customers also reported that execute latency is becoming red.

Targeting 0.14.9.1.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -193,6 +192,8 @@ public void readTuplesAndExecute(Communicator<HeronTuples.HeronTupleSet> inQueue
           stream.getComponentName(), stream.getId()).size();
 
       // We would reuse the System.nanoTime()
+      long startOfCycle = System.nanoTime();
```
nlu90: why not let `long currentTime = System.nanoTime()` and keep the `startOfCycle` unchanged? On line 225 or 226, there's a timeout check. If we move create the `startOfCycle` here, the semantics there is changed.

maosongfu: Make senses. Updated.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -192,8 +192,8 @@ public void readTuplesAndExecute(Communicator<HeronTuples.HeronTupleSet> inQueue
       int nValues = topologyContext.getComponentOutputFields(
           stream.getComponentName(), stream.getId()).size();
 
-      // We would reuse the System.nanoTime()
-      long currentTime = startOfCycle;
+      // Record current time at the start of processing a set of tuples
+      long currentTime = System.nanoTime();
```
billonahill: This logic makes sense, but what doesn't is what `currentTime` is expected to be after the loop on line 225. Because it gets reset on line 213, after the loop it seems to represent the time after the last tuple was deserialized but before it was handled. This seems really arbitrary. If that's really the intent could you add some comments below to clarify the time math that is occurring after the loop with `currentTime - startOfCycle - instanceExecuteBatchTime.toNanos()`.

maosongfu: There was an intent optimization to reduce one  `System.nanoTime()` call per tuple execution. In the new commit this optimization is removed to simplify the reasoning. 
Now it is much easier to understand.

billonahill: Can you clarigy (in comments) what `currentTime` is expected to reflect when it's used after the loop? It's not clear from reading the code.

maosongfu: Please check with latest code. No `currentTime` there now

PR #1970, opened 2017-06-16
Kubernetes Manual Scaling

jrcrawfo:
What?
Support for manual scaling of kubernetes-based topologies

Why?
Need to be able to scale our topologies via `update` command

TO-DO before merging:
- Add tests


heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -52,9 +60,12 @@ protected KubernetesController getController() {
 
   @Override
   public void initialize(Config aConfig, Config aRuntime) {
+    LOG.log(Level.INFO, aRuntime.toString());
```
billonahill: In the past we've had issues with credentials being logged when doing this. Can we change this to FINE or remove it. If we keep it, it's also good to add some sort of contextual logging to the message.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -253,4 +368,29 @@ protected String getFetchCommand() {
 
     return command;
   }
+
+  @Override
+  public void addContainers(Set<PackingPlan.ContainerPlan> containersToAdd) {
```
billonahill: Please add unit tests for add/remove containers and the spec builder methods.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+
+  /**
+   * Get information about a pod
+   *
+   * @return json object for pod
+   */
+  public JsonNode getBasePod(String podId) {
+
+    String podURI = String.format(
+        "%s/api/v1/namespaces/default/pods/%s",
```
billonahill: Is this path always this way for all kubernetes users? Is `default` ever something else for example?

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
       conn.disconnect();
     }
+  }
+
+  /**
+   * Get information about a pod
+   *
+   * @return json object for pod
+   */
+  public JsonNode getBasePod(String podId) {
```
billonahill: Could we reduce the visibility of many of these methods to package private?

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+
+    try {
+      // send get request
+      if (!NetworkUtils.sendHttpGetRequest(conn)) {
+        LOG.log(Level.SEVERE, "Failed to send GET to k8s deployment API");
+        return null;
+      }
+
+      // check the response
+      boolean success = NetworkUtils.checkHttpResponseCode(conn, HttpURLConnection.HTTP_OK);
```
billonahill: success can be removed and this can just be in the `if` clause. same for below.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+    String podURI = String.format(
+        "%s/api/v1/namespaces/default/pods/%s",
+        this.kubernetesURI,
+        podId);
+
+    // Get a connection
+    HttpURLConnection conn = NetworkUtils.getHttpConnection(podURI);
+    if (conn == null) {
+      LOG.log(Level.SEVERE, "Failed to find k8s deployment API");
+      return null;
```
billonahill: can we throw an exception here instead of returning null? same for other occurrences in this class.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+      LOG.log(Level.SEVERE, "Unable to parse pod JSON");
+      return null;
+    } finally {
+      conn.disconnect();
+    }
+  }
+
+  public boolean deployContainer(String deployConf) {
+    String deploymentURI = String.format(
+        "%s/api/v1/namespaces/default/pods",
```
billonahill: variants of this uri building are repeated in this class. could you centralize this logic into a shared method or member? During initialization of this class it seems we could replace set `kubernetesURI` as `"[kubernetesURI]/api/v1/namespaces/default/pods"` which would help.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+      }
+    } catch (IOException ioe) {
+      LOG.log(Level.SEVERE, "Unable to parse pod JSON");
+      return null;
+    } finally {
+      conn.disconnect();
+    }
+  }
+
+  public boolean deployContainer(String deployConf) {
```
billonahill: instead of returning a boolean to indicate success, can we return void and throw an exception (maybe `SchedulerException`) upon failure? Same elsewhere. See #1311.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+    String deploymentURI = String.format(
+        "%s/api/v1/namespaces/default/pods",
+        this.kubernetesURI);
+
+    // Get a connection
+    HttpURLConnection conn = NetworkUtils.getHttpConnection(deploymentURI);
+    if (conn == null) {
+      LOG.log(Level.SEVERE, "Failed to find k8s deployment API");
```
billonahill: in all of these messages we should include more contextual info to help the log reader understand the issue.. this one for example, should include the uri. The following one should include the uri and probably the json post body.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -79,8 +83,128 @@ public boolean killTopology() {
...
+    return true;
+  }
+
+  public boolean removeContainer(String podId) {
+    String podURI = String.format(
+        "%s/api/v1/namespaces/default/pods/%s",
+        this.kubernetesURI,
+        podId);
+
+    // Get a connection
```
billonahill: This same 6 lines of connection initialization logic are repeated in each method. could you put them into a shared helper that takes a uri, returns a connection and throws if it can't for any reason?

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -109,44 +233,13 @@ public boolean submitTopology(String[] appConfs) {
     boolean allSuccessful = true;
 
     for (int i = 0; i < appConfs.length; i++) {
```
billonahill:
```
for (String appConf : appConfs) {
```

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -109,44 +233,13 @@ public boolean submitTopology(String[] appConfs) {
...
-      LOG.log(Level.INFO, "Topology configuration is: " + appConfs[i]);
 
-      // Get a connection
-      HttpURLConnection conn = NetworkUtils.getHttpConnection(deploymentURI);
-      if (conn == null) {
-        LOG.log(Level.SEVERE, "Fauled to find k8s deployment API");
-        return false;
+      allSuccessful = deployContainer(appConfs[i]);
```
billonahill: You can remove the need for `allSuccessful` by doing this here and returning true at the end:
```
if (!deployContainer(appConfs[i])) {
  return false;
}
```

Or better yet, change this method to return void and throw an exception if one fails. It can have detailed info included about what went wrong.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
jrcrawfo: Good point. Should probably convert this into a configuration for the scheduler

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
```diff
@@ -397,6 +397,15 @@ public static boolean checkHttpResponseCode(HttpURLConnection connection, int ex
     }
   }
 
+  public static Integer getHttpResponseCode(HttpURLConnection connection) {
+    try {
+      return connection.getResponseCode();
+    } catch (IOException ex) {
+      LOG.log(Level.SEVERE, "Failed to get response code");
+      return null;
```
billonahill: I'm really not as fan of returning null to represent exceptions (see rant in #1311). If we can't get a response code it's an exceptional case, so I think we should throw an exception.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerJsonAPIUtils.java
```diff
@@ -0,0 +1,146 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import com.twitter.heron.scheduler.TopologyRuntimeManagementException;
+import com.twitter.heron.spi.utils.NetworkUtils;
+
+/**
+ *  Utility functions for talking to a JSON-based scheduler API
+ */
+public final class SchedulerJsonAPIUtils {
```
billonahill: I recommend we put this in `c.t.h.common.network` since it's generic and not specific to schedulers. Also instead of a bunch of static helpers, could follow an approach more like the java network libraries or `HeronClient`, where we initialize a `HttpJsonClient` object with a base uri, and then call one of get, post, put delete on it? That way we can provide better control/contact over the connection lifecycle. It also makes it easier to use in unit tests via mocks.

We could throw `IOException` instead of `TopologyRuntimeManagementException`.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -15,25 +15,36 @@
...
+  private final String kubernetesNamespace;
   private final String topologyName;
   private final boolean isVerbose;
 
   public KubernetesController(
       String kubernetesURI,
+      String kubernetesNamespace,
       String topologyName,
       boolean isVerbose
   ) {
```
billonahill: close parens or open curly don't get their own line in our styles.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -15,25 +15,36 @@
...
     this.kubernetesURI = kubernetesURI;
+
+    if (kubernetesNamespace == null) {
+      LOG.log(Level.INFO, "Namespace not provided in configuration. Using default.");
+      this.kubernetesNamespace = "default";
+    } else {
+      this.kubernetesNamespace = kubernetesNamespace;
```
billonahill: Instead of initializing namespace, can we initialize `"%s/api/v1/namespaces/%s/pods` as `baseUriPath`? Every method uses this as their base, so this would reduce duplicate logic.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -14,26 +14,34 @@
...
-import com.twitter.heron.spi.utils.NetworkUtils;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import com.twitter.heron.common.network.HttpJsonClient;
+import com.twitter.heron.scheduler.TopologySubmissionException;
 
 public class KubernetesController {
   private static final Logger LOG = Logger.getLogger(KubernetesController.class.getName());
 
   private final String kubernetesURI;
```
billonahill: Looks like kubernetesURI is only used in the constructor now so this can be removed.

heron/spi/src/java/com/twitter/heron/spi/utils/NetworkUtils.java
```diff
@@ -397,6 +397,10 @@ public static boolean checkHttpResponseCode(HttpURLConnection connection, int ex
     }
   }
 
+  public static Integer getHttpResponseCode(HttpURLConnection connection) throws IOException {
```
billonahill: can we remove this method since it's implementation is only a one-liner?

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -253,4 +370,46 @@ protected String getFetchCommand() {
...
+
+  @Override
+  public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove) {
+    for (PackingPlan.ContainerPlan container : containersToRemove) {
+      String podName = Runtime.topologyName(runtime) + "-" + container.getId();
+      try {
+        controller.removeContainer(podName);
+      } catch (IOException ioe) {
+        LOG.log(Level.SEVERE, "Problem removing a container", ioe);
+        throw new TopologyRuntimeManagementException("Unable to remove container for update");
```
billonahill: General comment about logging and exceptions... Typically we either log an error or throw an exception, but not both. Doing both tends to clutters up the logs with duplicate info. Let's instead just throw here, and include more contextual info in the exception message to clarify which container we couldn't remove. Also we should always chain the exceptions so we can get a good stack trace. So something like this:
```
throw new TopologyRuntimeManagementException("Unable to remove container with id " + container.getId(), ioe);
```
Comment applies here and elsewhere in the patch.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -253,4 +370,46 @@ protected String getFetchCommand() {
...
+      // deploy this new container
+      try {
+        controller.deployContainer(newContainer);
+      } catch (IOException ioe) {
+        LOG.log(Level.SEVERE, "Problem deploying new container", ioe);
+        throw new TopologyRuntimeManagementException("Unable to add container for update");
+      }
+
+    }
+
```
billonahill: pls remove empty lines between closing brackets.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -253,4 +370,46 @@ protected String getFetchCommand() {
...
+    try {
+      podConfig = controller.getBasePod(basePodName);
+    } catch (IOException ioe) {
+      LOG.log(Level.SEVERE, "Problem retrieving base pod configuration", ioe);
+      throw new TopologyRuntimeManagementException("Unable to add containers for update");
+    }
+
+    // iterate over the containers we need to add and rebuild the spec based on the new plan
+    ObjectMapper mapper = new ObjectMapper();
+    for (PackingPlan.ContainerPlan containerPlan : containersToAdd) {
```
billonahill: Does kubernetes provide the ability to add/remove multiple containers with one call? Doing it individually leaves us in a bad state if we fail midway through. If not we should clearly log that state upon failure (which ones succeeded, which failed or were never attempted).

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -162,6 +215,70 @@ protected ObjectNode getPodMetadata(ObjectMapper mapper, int containerIndex) {
...
+      // set the docker image to the same image as the existing container
+      containerInfo.put(KubernetesConstants.DOCKER_IMAGE,
+          existingContainer.get(KubernetesConstants.DOCKER_IMAGE).asText());
+
+      // Port info -- all the same
+      containerInfo.set(KubernetesConstants.PORTS, getPorts(mapper));
+
+      // Heron command for the new container
+      ArrayNode commandsArray = mapper.createArrayNode();
+      for (JsonNode cmd : existingContainer.get("command")) {
```
billonahill: please include comments about what's happening here with an example.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
jrcrawfo: Unfortunately not. I will make a clear log to cover this.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
jrcrawfo: Done

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -43,114 +49,110 @@ public KubernetesController(
    *
    * @return success
    */
-  public boolean killTopology() {
+  protected boolean killTopology() {
```
billonahill: It would be great to change the existing methods in this class to also return `void` and throw exceptions instead. I know that's scope creep on what you're adding in this PR though, so we could make that change in a follow up PR if you prefer.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -253,4 +374,52 @@ protected String getFetchCommand() {
...
+      try {
+        controller.deployContainer(newContainer);
+        deployedContainerCount++;
+        LOG.log(Level.INFO, "New container " + ++deployedContainerCount + "/"
+            + totalNewContainerCount + " deployed");
+      } catch (IOException ioe) {
+        LOG.log(Level.SEVERE, "Issue deploying a container! Previously added container count: "
+            + deployedContainerCount);
+        LOG.log(Level.INFO, "Configuration of problem container: " + newContainer);
+        throw new TopologyRuntimeManagementException("Problem adding container with id "
```
billonahill: can we remove the logging here and add any relevant info to the exception message? Instead of log info, log sever and throw exception, best to just throw the exception with a good message.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
jrcrawfo: Yeah, since I'll need to re-do the tests here, I'd prefer to do a separate PR here.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -358,11 +418,26 @@ protected ArrayNode getPorts(ObjectMapper mapper) {
     return ports;
   }
 
+  /**
+   * getFetchCommand
```
billonahill: Here and elsewhere can we not echo the method name in the javadocs, since it's redundant.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -358,11 +418,26 @@ protected ArrayNode getPorts(ObjectMapper mapper) {
...
+   *
+   * Get the command that will be used to retrieve the topology JAR
+   *
+   * @return String
```
billonahill: here and elsewhere let's not specify `@return [type]` unless we have something specific to say about what's returned. The return type will be included in the javadocs without specifying `@return String`.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -401,15 +487,23 @@ public void addContainers(Set<PackingPlan.ContainerPlan> containersToAdd) {
...
+  /**
+   * Remove containers for a scale-down event from an update command
+   *
+   * @param containersToRemove, the list of containers that need to be removed
+   *
+   * NOTE: Due to the mechanics of Kubernetes pod removal, each container must be removed on
+   * a one-by-one basis. If one container out of many containers to be removed failed, it will
+   * leave the topology in a bad state.
+   *
+   * TODO (jrcrawfo) -- Improve the handling of this issue
```
billonahill: please include a link to a github issue for this.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -14,143 +14,145 @@
...
 
-    try {
-      if (!NetworkUtils.sendHttpDeleteRequest(conn)) {
-        LOG.log(Level.SEVERE, "Failed to send delete request to k8s deployment API");
-        return false;
-      }
+  /**
+   * Get information about a pod
+   *
+   * @return
```
billonahill: here and elsewhere please remove empty returns.

billonahill:
There are checkstyle errors in CI FYI.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -460,13 +460,16 @@ public void addContainers(Set<PackingPlan.ContainerPlan> containersToAdd) {
...
-            + containerPlan.getId() + ". Deployed " + deployedContainerCount + " out of " + totalNewContainerCount + " containers", ioe);
+            + containerPlan.getId() + ". Deployed " + deployedContainerCount + " out of "
+            + totalNewContainerCount + " containers", ioe);
       }
     }
   }
 
   /**
-   * @param containersToRemove, the list of containers that need to be removed
+   * Remove containers for a scale-dowp event from an update command
```
billonahill: typo: dowp

billonahill:
👍

PR #1979, opened 2017-06-22
Rephrase references to Storm

billonahill:


website/content/docs/concepts/architecture.md
```diff
@@ -24,11 +24,10 @@ in the [Heron Topologies](../topologies) document.
...
 ## Relationship with Apache Storm
 
-Heron is the direct successor of [Apache Storm](http://storm.apache.org) but
-built with two goals in mind:
+Heron was initially developed at Twitter with two goals in mind:
 
-1. Overcoming Storm's performance, reliability, and other shortcomings by
-replacing Storm's thread-based computing model with a process-based model.
+1. Providing performance, reliability and ease of troubleshooting by leveraging a process-based
+computing model. Storm utilizes a thread-based model.
```
srkukarni: Why mention Storm at all here?

website/content/docs/concepts/architecture.md
```diff
@@ -2,7 +2,7 @@
 title: Heron Architecture
 ---
 
-Heron is the direct successor of [Apache Storm](http://storm.apache.org). From
+Heron provides compatibility with [Apache Storm](http://storm.apache.org). From
```
srkukarni: Can we say more along the lines of
Heron is a general purpose streaming engine with performance, low latency, isolation, reliability in mind. It also provides API compatibilty with Storm.

website/content/docs/concepts/architecture.md
billonahill: I had the same thought but kept it because the section title is Relationship with Storm, so the intent I believe was to clarify the differences. I'm fine with removing the Storm mention here if we think the section would still make sense without being explicit about the distinction.

billonahill:
Updated per your suggestion @srkukarni with some minor tweeks. LMKWYT.

PR #1980, opened 2017-06-22
Dhalion health manager

ashvina:
PART-1 of #1952 

A original PR related to Topology health manager, #1952, may be too big to review. This PR is "subset" of the original PR, and is created to review the following components of topology health manager:
1. HealthManager
1. Tracker metrics reader
1. Health policy config reader
1. Guice injector


billonahill:
Why `DO NOT MERGE` in the title? Is this ready for review to be merged?

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+    String configPath = cmd.getOptionValue("config_path");
+    String overrideConfigFile = cmd.getOptionValue("override_config_file");
+    String releaseFile = cmd.getOptionValue("release_file");
+    String topologyName = cmd.getOptionValue("topology_name");
+    String trackerUrl = cmd.getOptionValue("trackerURL", "http://localhost:8888");
+    Boolean verbose = cmd.hasOption("verbose");
+
+    // build the final config by expanding all the variables
+    Config config = Config.toLocalMode(Config.newBuilder()
+        .putAll(ConfigLoader.loadConfig(heronHome, configPath, releaseFile, overrideConfigFile))
```
huijunw: loadClusterConfig() may be simpler here.

heron/healthm/src/java/com/twitter/heron/healthmgr/common/HealthMgrConstants.java
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.twitter.heron.healthmgr.common;
+
+public interface HealthMgrConstants {
```
huijunw: why not enum?

heron/healthm/src/java/com/twitter/heron/healthmgr/common/HealthMgrConstants.java
```diff
@@ -0,0 +1,39 @@
...
+  String METRIC_BUFFER_SIZE_SUFFIX = "/packets";
+
+  String COMPONENT_STMGR = "__stmgr__";
+
+  String CONF_TRACKER_URL = "TRACKER_URL";
+  String CONF_TOPOLOGY_NAME = "TOPOLOGY_NAME";
+  String CONF_CLUSTER = "CLUSTER";
+  String CONF_ENVIRON = "ENVIRON";
+
+  // health policy configuration related keys
```
huijunw: They may be moved to com.twitter.heron.spi.common.Key

heron/healthm/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
```diff
@@ -0,0 +1,94 @@
...
+import com.twitter.heron.api.generated.TopologyAPI.Topology;
+import com.twitter.heron.healthmgr.common.HealthManagerEvents.TopologyUpdate;
+import com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor;
+
+/**
+ * A topology may be updated after initial deployment. This provider is used to provide the latest
+ * version to any dependent components.
+ */
+@Singleton
+public class TopologyProvider implements Provider<Topology>, EventHandler<TopologyUpdate> {
```
huijunw: How does HealthMgr use TopologyProvider?

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+ * -d ~/.heron -p ~/.heron/conf/local -c local -e default -r userName -t AckingTopology
+ */
+public class HealthManager {
+  private static final Logger LOG = Logger.getLogger(HealthManager.class.getName());
+  private final Config config;
+  private final String trackerURL;
+
+  private Config runtime;
+  private Injector injector;
+  private SchedulerStateManagerAdaptor stateMgrAdaptor;
```
ashvina: The `HealthManager` will fetch topology info and scheduler location from the state manager.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+  private ISchedulerClient schedulerClient;
+
+  private List<IHealthPolicy> healthPolicies = new ArrayList<>();
+  private HealthPolicyConfigReader policyConfigReader;
+
+  public HealthManager(Config config, String trackerUrl) {
```
ashvina: Agree, `MetricsCacheMgr` could be another provider of metrics. We should talk about it.  Currently `Dhalion` depends on the `timeline` api of the `Tracker`. When requested for metrics over 300 seconds, `Tracker` will provide 5 readings one for each of the 60 second intervals. Could you please point at some `MetricsCache` examples for this?

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+
+    options.addOption(help);
+    return options;
+  }
+
+  // Construct all required command line options
+  private static Options constructCliOptions() {
+    Options options = new Options();
+
+    Option cluster = Option.builder("c")
```
ashvina: The `Tracker` api needs cluster name.

heron/healthm/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
ashvina: The `HealthManager` will just `inject` instance of `TopologyProvider`. It will be used by `Dhalion Sensors`

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
huijunw: MetricsCacheMgr has the same http query interface as Tmaster. https://github.com/twitter/heron/blob/master/heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/MetricsCacheManagerHttpServer.java

kramasamy: +1 - let us use Metrics Cache Manager.

ashvina: Just to be clear, Dhalion should provide both implementations, one for `tracker` and one for `MetricsCacheManager`. It is possible that users have not configured `MetricsCacheManager` sinks or `tmaster` sink. We cannot replace one for other. I also think that this PR need not be blocked if this implementation is not available. A couple of PRs are blocked on this.

huijunw: Agree to let this PR go. We may add MetricsCache ingestor in later PR. @kramasamy @maosongfu

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+      throw new RuntimeException("Error parsing command line options: ", e);
+    }
+
+    String cluster = cmd.getOptionValue("cluster");
+    String role = cmd.getOptionValue("role");
+    String environ = cmd.getOptionValue("environment");
+    String heronHome = cmd.getOptionValue("heron_home");
+    String configPath = cmd.getOptionValue("config_path");
+    String overrideConfigFile = cmd.getOptionValue("override_config_file");
+    String releaseFile = cmd.getOptionValue("release_file");
```
ashvina: `release_file` is not needed. Will replace the only usage with a `null` value.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
ashvina: As per the javadocs of `loadClusterConfig`, the method assumes default home and config path: `Loads raw configurations using the default configured heronHome and configPath on the cluster`. This may not be the case here. The method is only useful for `scheduler` service only.

heron/healthm/src/java/BUILD
```diff
@@ -0,0 +1,71 @@
...
+    "@javax_inject_javax_inject//jar",
+    "@org_apache_thrift_libthrift//jar",
+    "@org_glassfish_jersey_core_jersey_client//jar",
+    "@org_glassfish_jersey_core_jersey_common//jar",
+    "@org_glassfish_jersey_ext_jersey_entity_filtering//jar",
+    "@org_glassfish_jersey_media_jersey_media_json_jackson//jar",
+    "@org_glassfish_jersey_bundles_repackaged_jersey_guava//jar",
+    "@org_glassfish_hk2_hk2_api//jar",
+    "@org_glassfish_hk2_hk2_utils//jar",
+    "@org_glassfish_hk2_hk2_locator//jar",
```
kramasamy: @ashvina - wondering why there are so many dependencies?

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
kramasamy: @ashvina - is the health manager one per topology or for all topologies?

kramasamy: If the health manager is per topology, we can enable MetricsCacheManager Sink by default, similar to Tmaster sink.

heron/healthm/tests/java/com/twitter/heron/healthmgr/common/PackingPlanProviderTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.twitter.heron.healthmgr.common.HealthManagerEvents.TopologyUpdate;
+import com.twitter.heron.packing.roundrobin.RoundRobinPacking;
+import com.twitter.heron.proto.system.PackingPlans;
+import com.twitter.heron.spi.packing.PackingPlan;
+import com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor;
+import com.twitter.heron.spi.utils.PackingTestUtils;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
```
billonahill: no star imports please, here and elsewhere

huijunw:
@ashvina any update on this pr?

ashvina:
Hi @huijunw I will soon push updates to resolve some of the review comments.

huijunw:
Thank you @ashvina

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Logger;
+
+import org.yaml.snakeyaml.Yaml;
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
ashvina: Fixed. The trackerUrl is replaced by a guice module. Alternatively, we could also opt for adding a new constructor once metrics cache implementation is added.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,355 @@
...
+
+    Option configFile = Option.builder("p")
+        .desc("Path of the config files")
+        .longOpt("config_path")
+        .hasArgs()
+        .argName("config path")
+        .required()
+        .build();
+
+    Option configOverrides = Option.builder("o")
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/common/HealthMgrConstants.java
ashvina: Fixed.
I have removed the interface. An enum was added where feasible. Guice `Named` qualifier annotation expects a constant String. Such constants cannot be replaced with enum.

heron/healthm/src/java/com/twitter/heron/healthmgr/common/HealthMgrConstants.java
ashvina: Guice `Named` qualifier annotation expects a constant String. Such constants cannot be replaced with enum.

heron/healthm/src/java/BUILD
```diff
@@ -0,0 +1,71 @@
+licenses(["notice"])
```
ashvina: Fixed

heron/config/src/yaml/conf/local/healthmgr.yaml
```diff
@@ -0,0 +1,10 @@
+## list of policies to be executed for self regulation
+#heron.class.health.policies:
+#  - dynamic-resource-allocation
+#
+## configuration specific to individual policies listed above
+#dynamic-resource-allocation:
+#  health.policy.class: "com.twitter.heron.healthmgr.policy.DynamicResourceAllocationPolicy"
+#  health.policy.interval.ms: "120000"
```
ashvina: I think a consistent value format, String in this case, is easier to create using scripts. Also the consumer gets to decides the format in which the value is used?

heron/healthm/src/java/BUILD
```diff
@@ -0,0 +1,71 @@
...
+    "//heron/api/src/java:classification",
+    "//heron/common/src/java:basics-java",
+    "//heron/common/src/java:config-java",
+    "//heron/common/src/java:utils-java",
+    "//heron/spi/src/java:common-spi-java",
+    "//heron/spi/src/java:packing-spi-java",
+    "//heron/spi/src/java:statemgr-spi-java",
+    "//heron/spi/src/java:utils-spi-java",
+    "//heron/spi/src/java:metricsmgr-spi-java",
+    "//heron/proto:proto_common_java",
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,386 @@
...
+ * <ul>
+ * <li>cluster name: <code>-c local</code>
+ * <li>role: <code> -r dev</code>
+ * <li>environment: <code> -e default</code>
+ * <li>heron home directory: <code> -d ~/.heron</code>
+ * <li>config directory: <code> -p ~/.heron/conf</code>
+ * <li>topology name: <code> -n AckingTopology</code>
+ * </ul>
+ * <p>
+ * Optional command line options for the {@link HealthManager} include
```
ashvina: Yes. it will be part of a new PR. User would provide the dry run configuration using policy config file. Dhalion policy executor will just log the diagnosis and skip execution of resolver if dry run is true.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,386 @@
...
+      }
+    };
+  }
+
+  @VisibleForTesting
+  SchedulerStateManagerAdaptor createStateMgrAdaptor() throws ReflectiveOperationException {
+    String stateMgrClass = Context.stateManagerClass(config);
+    IStateManager stateMgr = ReflectionUtils.newInstance(stateMgrClass);
+    stateMgr.initialize(config);
+    return new SchedulerStateManagerAdaptor(stateMgr, 5000);
```
ashvina: The state mgr needs to be initialized before it could be used in the adaptor. I can think of instantiating the stateMgr through guice, but not sure if adds any value. Using `ReflectionUtils` is consistent with the usage in other places.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -0,0 +1,386 @@
...
+  /**
+   * Load the config parameters from the command line
+   *
+   * @param cluster, name of the cluster
+   * @param role, user role
+   * @param environ, user provided environment/tag
+   * @param verbose, enable verbose logging
+   * @return config, the command line config
+   */
+  private static Config commandLineConfigs(String cluster,
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.twitter.heron.common.basics.TypeUtils;
+import com.twitter.heron.common.config.ConfigReader;
+import com.twitter.heron.healthmgr.common.HealthMgrConstants;
+
+public class HealthPolicyConfigReader {
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/common/HealthMgrConstants.java
```diff
@@ -0,0 +1,39 @@
...
+  String CONF_FILE_NAME = "healthmgr.yaml";
+  String HEALTH_POLICIES = "heron.class.health.policies";
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthPolicyConfig.java
```diff
@@ -0,0 +1,47 @@
...
+  public HealthPolicyConfig(Map<String, String> configs) {
+    this.configs = configs;
+    LOG.info("Health Policy Configuration:" + configs);
+  }
+
+  public String getPolicyClass(String policyId) {
+    return configs.get(HealthMgrConstants.HEALTH_POLICY_CLASS);
+  }
+
+  public long getInterval() {
```
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/common/PackingPlanProvider.java
```diff
@@ -0,0 +1,105 @@
...
+  @Inject
+  public PackingPlanProvider(SchedulerStateManagerAdaptor stateManagerAdaptor,
+                             EventManager eventManager,
+                             @Named(HealthMgrConstants.CONF_TOPOLOGY_NAME) String topologyName) {
+    this.stateManagerAdaptor = stateManagerAdaptor;
+    this.topologyName = topologyName;
+    eventManager.addEventListener(TopologyUpdate.class, this);
+  }
+
+  public String[] getBoltInstanceNames(String... boltComponents) {
```
ashvina: This method is used in classes in a followup PR.

heron/healthm/src/java/com/twitter/heron/healthmgr/common/PackingPlanProvider.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Invalidates cached packing plan on receiving update notification
+   */
+  public synchronized void onEvent(TopologyUpdate event) {
+    LOG.info("Received topology update event, invalidating cached PackingPlan: " + event);
+    this.packingPlan = null;
+  }
+
+  private synchronized void fetchLatestPackingPlan() {
```
ashvina: This method is used in classes in a followup PR.

heron/healthm/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
```diff
@@ -0,0 +1,94 @@
...
+    LOG.info("Received topology update event, invalidating cached topology: " + event);
+    this.topology = null;
+  }
+
+  /**
+   * A utility method to extract bolt component names from the topology.
+   *
+   * @return array of all bolt names
+   */
+  public String[] getBoltNames() {
```
ashvina: This method is used in classes in a followup PR.

heron/healthm/tests/java/com/twitter/heron/healthmgr/common/PackingPlanProviderTest.java
ashvina: Fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @Override
+  public Map<String, ComponentMetrics> getComponentMetrics(String metric,
+                                                           int startTimeSec,
+                                                           int durationSec,
+                                                           String... components) {
+    Map<String, ComponentMetrics> result = new HashMap<>();
+    for (String component : components) {
+      String response = getMetricsFromTracker(metric, component, startTimeSec, durationSec);
```
ashvina: No. It is assumed that the detectors will wait for the next execution cycle or retry.

heron/healthm/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
```diff
@@ -0,0 +1,149 @@
...
+    this.baseTarget = client.target(trackerURL)
+        .path("topologies/metricstimeline")
+        .queryParam("cluster", cluster)
+        .queryParam("environ", environ)
+        .queryParam("topology", topologyName);
+  }
+
+  @Override
+  public Map<String, ComponentMetrics> getComponentMetrics(String metric,
+                                                           int startTimeSec,
```
ashvina: Sure. I am assuming any of the java 8 features can be used henceforth? Including lambda?

huijunw:
@ashvina could you merge master that has ci fix
@billonahill could you review and give shipit

billonahill:
Yes, I should be able to review today.

heron/config/src/yaml/conf/local/healthmgr.yaml
billonahill: all other heron yaml configs in this directory and elsewhere leverage primitive ints so we should provide a consistent user experience. Actually we don't use quotes around strings or numbers elsewhere, so I think we should remove them here to not be the exception. Also asking users to use quotes where not necessary is cumbersome.

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthManager.java
billonahill: Ideally we could replace the reflection pattern with guice everywhere. The initialize() method is an artifact of the reflection approach. If there were a way to do this with guice that would be awesome, since it would show how this pattern could be replaced elsewhere. If you want to explore/consider that outside this PR that's fine though.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,95 @@
...
+    return new ArrayList<>(configs.keySet());
+  }
+
+  public Map<String, String> getPolicyConfig(String policyId) {
+    return configs.get(policyId);
+  }
+
+  @Override
+  public String toString() {
+    return "HealthPolicyConfigReader{"
```
billonahill: could you consolidate unnecessary string concatenation here pls. Much of this can fit on one line.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,95 @@
...
+  private String configFilename;
+
+  public HealthPolicyConfigReader(String filename) throws FileNotFoundException {
+    this.configFilename = filename;
+  }
+
+  @SuppressWarnings("unchecked")
+  public void loadConfig() {
+    Map<String, Object> ret = readConfigFromFile(configFilename);
+    for (String id : TypeUtils.getListOfStrings(ret.get(POLICY_CONF.HEALTH_POLICIES.key))) {
```
billonahill: could we use the enums as the config keys?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/PackingPlanProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+    PackingPlan packing = get();
+    ArrayList<String> boltInstanceNames = new ArrayList<>();
+    for (ContainerPlan containerPlan : packing.getContainers()) {
+      for (InstancePlan instancePlan : containerPlan.getInstances()) {
+        if (!boltComponentNames.contains(instancePlan.getComponentName())) {
+          continue;
+        }
+
+        String name = "container_" + containerPlan.getId()
```
billonahill: This assumes an instance naming convention declared elsewhere which could change. Are there other places in the java codebase where this logic exists which we could share somehow?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
```diff
@@ -0,0 +1,96 @@
...
+  public synchronized Topology get() {
+    if (topology == null) {
+      fetchLatestTopology();
+    }
+    return topology;
+  }
+
+  private synchronized void fetchLatestTopology() {
+    LOG.log(Level.INFO, "Fetching topology from state manager: {0}", topologyName);
+    this.topology = stateManagerAdaptor.getTopology(topologyName);
```
billonahill: This is the topology at the time of submission, not the latest one, which is what you want right?. The latest one can be gotten from the current `physicalPlan.getTopology()`, but even that one doesn't contain updates to the config (e.g. parallelism) since creation. If that's needed see `UpdateTopologyMaster`.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
```diff
@@ -0,0 +1,96 @@
...
+    if (topology == null) {
+      throw new RuntimeException("Failed to fetch topology from the State Manager");
```
billonahill: Can we throw an exception that is not RuntimeException? Also, let's include the topologyName in the message.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
```diff
@@ -0,0 +1,153 @@
...
+import com.microsoft.dhalion.api.MetricsProvider;
+import com.microsoft.dhalion.metrics.ComponentMetrics;
+import com.microsoft.dhalion.metrics.InstanceMetrics;
+
+import net.minidev.json.JSONArray;
+
+import static com.twitter.heron.healthmgr.HealthManager.CONF_TOPOLOGY_NAME;
+
+public class TrackerMetricsProvider implements MetricsProvider {
+  public static final String CONF_TRACKER_URL = "TRACKER_URL";
```
billonahill: @can these be package private and `@VisibleForTesting`? looks like they're only used by testing.

heron/healthm/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
billonahill: We've been using Instant and Duration yes, but not labdas AFAIK. Seems if we're using the former we should be able to use the latter. :)

heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
```diff
@@ -0,0 +1,153 @@
...
+    this.baseTarget = client.target(trackerURL)
+        .path("topologies/metricstimeline")
+        .queryParam("cluster", cluster)
+        .queryParam("environ", environ)
+        .queryParam("topology", topologyName);
+  }
+
+  @Override
+  public Map<String, ComponentMetrics> getComponentMetrics(String metric,
+                                                           int startTimeSec,
```
billonahill: Let's use Instance and Duration for these.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
ashvina: Ok. The code is generated using tools provided by java IDE. Is this change suggested for readability?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
ashvina: These variables are used outside the package by the `HealthManager`

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
ashvina: I am not sure I understand that. Do you mean the name of the enum should be same as the key in the config file?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
ashvina: I am looking at https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java?utf8=%E2%9C%93#L168. The topology is getting updated after a config change. Could you please help me understand why `stateManagerAdaptor` return stale value?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
billonahill: The IDE generated code is overly clunky at times. For readability and perf it's nicer to avoid this much string concat for what could be just "string" + object + "string".

heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
billonahill: got it.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,95 @@
...
+ * heron.class.health.policies:
+ *   - dynamic-resource-allocation
+ *
+ * dynamic-resource-allocation:
+ *   health.policy.class: "com.twitter.heron.healthmgr.policy.DynamicResourceAllocationPolicy"
+ *   health.policy.interval.ms: "120000"
+ * </pre>
+ */
+public class HealthPolicyConfigReader {
+  enum POLICY_CONF {
```
billonahill: Our standard for enums is camel case, so `PolicyConf`.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
billonahill: Can the map be `Map<PolicyConf, Object>`?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
billonahill: Look at the `getTopology` method:
https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java?utf8=%E2%9C%93#L337
```
stateManager.getPhysicalPlan(topologyName).getTopology();
```
The physical plan and the topology it holds gets updated during the licecycle of the running topology. `stateManager.getTopology(topologyName)` does not. It is the immutable topology from the time of submission.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/PackingPlanProvider.java
ashvina: The instance id `container_%s_%s_%d` seems to be set here https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py?utf8=%E2%9C%93#L568 and then used by `MetricsManagerServer` to be sent to tracker. I did not find any other user of this string.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
ashvina: Understood. Thanks! 

In my opinion there is redundancy and confusing implementation here. Both `stateManagerAdaptor.getTopology` and `stateManagerAdaptor.getPhysicalPlan.getTopology` should return the same state of the topology. WDYT?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/PackingPlanProvider.java
billonahill: That's what I suspected - only set elsewhere in python. Good to leave it as is here then. Maybe make a comment on both occurrences of it about the other for now.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
billonahill: I recently learned of this myself and I agree with you that it's confusing and error prone. I wonder if there is any value though in keeping the original topology at getTopology. @maosongfu @kramasamy @nlu90 @srkukarni thoughts?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
ashvina: The policy config map `configs` will contain system configs, for e.g. policy class name, and user component specific configs, for e.g. buffer size threshold. Hence a map with enum as the key cannot be created in this case.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
ashvina: Fixed

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
ashvina: Fixed

heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/TrackerMetricsProvider.java
ashvina: Fixed

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
ashvina: Fixed

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
```diff
@@ -0,0 +1,96 @@
...
+      fetchLatestTopology();
+    }
+    return topology;
+  }
+
+  private synchronized void fetchLatestTopology() {
+    LOG.log(Level.INFO, "Fetching topology from state manager: {0}", topologyName);
+    this.topology = stateManagerAdaptor.getPhysicalPlan(topologyName).getTopology();
+    if (topology == null) {
+      throw new RuntimeException("Failed to fetch topology from the State Manager");
```
billonahill: Can we create a different exception to use here that's not `RuntimeException` here and elsewhere in this review? Typically we should avoid throwing generic exceptions like this, except maybe in a `main(..)` method. Using a more specific exception allows callers to handle them specifically and provides the ability for us to add context-specific features to the custom exception as needed.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,93 @@
...
+ * heron.class.health.policies:
+ *   - dynamic-resource-allocation
+ *
+ * dynamic-resource-allocation:
+ *   health.policy.class: "com.twitter.heron.healthmgr.policy.DynamicResourceAllocationPolicy"
+ *   health.policy.interval.ms: "120000"
+ * </pre>
+ */
+public class HealthPolicyConfigReader {
+  enum PolicyConf {
```
billonahill: Could we name this `PolicyConfigKey` for consistency with `SystemConfigKey`?

heron/healthm/src/java/BUILD
ashvina: @kramasamy As of now Dhalion uses http client to query tracker and fetch metrics. These dependencies are related to the client and parsing the json returned by the tracker.

heron/config/src/yaml/conf/local/healthmgr.yaml
ashvina: fixed

heron/healthm/src/java/com/twitter/heron/healthmgr/HealthPolicyConfigReader.java
```diff
@@ -0,0 +1,64 @@
...
+  private final Map<String, Map<String, String>> configs = new HashMap<>();
```
ashvina: fixed

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/TopologyProvider.java
ashvina: Fixed by adding `MissingStateException`

heron/healthmgr/src/java/com/twitter/heron/healthmgr/common/MissingStateException.java
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.twitter.heron.healthmgr.common;
+
+/**
+ * This exception is generated when desired state information is not found
+ */
+public class MissingStateException extends RuntimeException {
```
billonahill: Could we change to something like `InvalidStateException` to make it slightly more generic? That way it could be used for any inconsistent or unexpected state.

ashvina: Fixed by renaming

heron/healthmgr/src/java/BUILDX
```diff
@@ -0,0 +1,65 @@
+licenses(["notice"])
```
kramasamy: Why this file is called BUILDX?

WORKSPACE
```diff
@@ -338,6 +428,11 @@ maven_jar(
   artifact = "com.101tec:zkclient:0.3"
 )
 
+#http_jar(
```
kramasamy: If you don't need this, please remove it.

ashvina: The Dhalion build files and the dependencies have been temporarily disabled till maven repo is enabled. I would suggest we keep it like this for a few more days.

heron/healthmgr/src/java/BUILDX
ashvina: The Dhalion build files and the dependencies have been temporarily disabled till maven repo is enabled. I would suggest we keep it like this for a few more days. Some CI systems are based on local maven repo only and cannot download jar files from the internet. The build will be enabled once maven repo is updated. Also, a couple of PRs related to Dhalion will follow this request. We split the PR in parts to simplify review.

huijunw:
@kramasamy what do you think on the above reply from @ashvina ? could you give a shipit

huijunw:
talked with @kramasamy on slack and we agreed:
we can give a ship it and later fix it

PR #1982, opened 2017-06-22
Native Python API for Heron.

srkukarni:
Currently we support Parsely Python API for running python topologies. However it suffers from two issues:-
1. For Java API, Heron supports a native Java API and other APIs(like the Storm API) are just wrappers around the native Heron Java api. The lack of a native Python API and going directly to an external API thus is inconsistent
2. The current Parsely stuff mixes apis and implementation much like Storm did. It will be very hard to add new features without this clear separation.

This pr adds a native Python API layer for Heron. Currently this is just a dangling code; no other component uses it. The idea is to morph the python instance to use this API next and then later on transition parsely API to be just a simple wrapper.
Much of this code is copied from the existing Parsely API. The api is identical, only difference is that we have borrowed any classes from heron.common directory that should belong to the api.


billonahill:
Can we also create/copy unit tests?

srkukarni:
Added unittests

PR #1983, opened 2017-06-22
[WIP] Heron client unit test

objmagic:
We test if we will `Popen` two Java program which writes definition file and calls submitter. I can add more tests but I'd like to have some opinions first.

### DIFF
--- a/heron/tools/cli/src/python/args.py
+++ b/heron/tools/cli/src/python/args.py
@@ -41,18 +41,6 @@ def add_verbose(parser):
   return parser
 
 
-def add_trace_execution(parser):
-  '''
-  :param parser:
-  :return:
-  '''
-  parser.add_argument(
-      '--trace-execution',
-      metavar='(a boolean; default: "false")',
-      default=False)
-  return parser
-
-
 def add_topology(parser):
   '''
   :param parser:
--- a/heron/tools/cli/src/python/execute.py
+++ b/heron/tools/cli/src/python/execute.py
@@ -59,7 +59,7 @@ def heron_class(class_name, lib_jars, extra_jars=None, args=None, java_defines=N
   all_args += [class_name] + list(args)
 
   # set heron_config environment variable
-  heron_env = {}
+  heron_env = os.environ.copy()
   heron_env['HERON_OPTIONS'] = opts.get_heron_config()
 
   # print the verbose message
--- a/heron/tools/cli/src/python/execute.py
+++ b/heron/tools/cli/src/python/execute.py
@@ -59,7 +59,7 @@ def heron_class(class_name, lib_jars, extra_jars=None, args=None, java_defines=N
   all_args += [class_name] + list(args)
 
   # set heron_config environment variable
-  heron_env = os.environ.copy()
+  heron_env = {}
   heron_env['HERON_OPTIONS'] = opts.get_heron_config()
 
   # print the verbose message
--- a/heron/tools/cli/tests/python/parse_unittest.py
+++ b/heron/tools/cli/tests/python/parse_unittest.py
@@ -1,0 +1,75 @@
+# Copyright 2017 Twitter. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+''' opts_unittest.py '''
+import unittest2 as unittest
+import mock
+from mock import call, patch, Mock, MagicMock
+import os
+import subprocess
+import sys
+import tempfile
+import heron.tools.cli.src.python.main as main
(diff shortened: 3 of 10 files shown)


heron/tools/cli/src/python/execute.py
```diff
@@ -59,7 +59,7 @@ def heron_class(class_name, lib_jars, extra_jars=None, args=None, java_defines=N
   all_args += [class_name] + list(args)
 
   # set heron_config environment variable
-  heron_env = os.environ.copy()
+  heron_env = {}
```
billonahill: this seems like a functional change beyond just what's required for the testing, no?

heron/tools/cli/src/python/args.py
```diff
@@ -41,18 +41,6 @@ def add_verbose(parser):
   return parser
 
 
-def add_trace_execution(parser):
-  '''
-  :param parser:
-  :return:
-  '''
-  parser.add_argument(
-      '--trace-execution',
```
billonahill: why is this being removed?

heron/tools/cli/tests/python/parse_unittest.py
```diff
@@ -0,0 +1,75 @@
...
+    result.SimpleResult.__init__ = Mock(return_value=None)
+    result.ProcessResult.__init__ = Mock(return_value=None)
+    def launch(cl_args, topology_file, tmp_dir):
+      return submit.launch_a_topology(cl_args, tmp_dir, topology_file, 'T.defn', 'WordCount')
+    submit.launch_topologies = Mock(side_effect=launch)
+    pass
+
+  def test(self):
+    subprocess.Popen = MagicMock()
+    args = ["heron", "submit", "local", "~/.heron/examples/heron-examples.jar",
```
billonahill: If we're going to copy this pattern for other calls, we might want to consider a generic method that takes a string of the heron command and a collection of strings of popen commands. Using strings will likely be easier than arrays. The shared method would deal with converting strings to arrays.

heron/tools/cli/tests/python/parse_unittest.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import subprocess
+import sys
+import tempfile
+import heron.tools.cli.src.python.main as main
+import heron.tools.cli.src.python.submit as submit
+import heron.tools.cli.src.python.result as result
+import heron.tools.common.src.python.utils.config as config
+
+class OptsTest(unittest.TestCase):
```
billonahill: `OptsTest` sounds like a test that verifies option parsing logic. Can we call this something more representative like `SubmitTest` or something (which extends a common `ClientCommandTest` or something like that)? We can then have a bunch of tests to represent how different args to submit result in different popens.

heron/tools/cli/tests/python/parse_unittest.py
```diff
@@ -0,0 +1,75 @@
...
+    config.parse_override_config = MagicMock(return_value='/heron/home/override.yaml')
+    # Mock result module
+    result.render = MagicMock(return_value=None)
+    result.is_successful = MagicMock(return_value=True)
+    result.SimpleResult.__init__ = Mock(return_value=None)
+    result.ProcessResult.__init__ = Mock(return_value=None)
+    def launch(cl_args, topology_file, tmp_dir):
+      return submit.launch_a_topology(cl_args, tmp_dir, topology_file, 'T.defn', 'WordCount')
+    submit.launch_topologies = Mock(side_effect=launch)
+    pass
```
billonahill: is `pass` necessary?

heron/tools/cli/src/python/args.py
objmagic: It is not used. Pretty sure it is some leftover when we move from close-sourced to open-sourced Heron.

heron/tools/cli/src/python/execute.py
objmagic: At first I thought we don't need to copy all the env variables, but now I think we do need all env vars. For example, Java program could fetch environmental variables.

billonahill: right, so should this change be reverted?

heron/tools/cli/tests/python/parse_unittest.py
objmagic: addressed in https://github.com/twitter/heron/pull/1983/commits/f18b0a417bb96f0b473db62eea39b12705a59f3f. Using string looked better at first. But when I was refactoring to use string, I realized the command string is too long and I have to break them into multiple lines, but I forgot to add **a space** at the end of some lines, which made me spend like 5 minutes comparing why assertion failed. I suggest we use **list**, which makes spacing clearer.
image

heron/tools/cli/src/python/execute.py
objmagic: yes later.

objmagic: https://github.com/twitter/heron/pull/1983/commits/4de4efbb19ade9fada1a6adbe6c9b6bd7e8cbfd7

heron/tools/cli/tests/python/parse_unittest.py
billonahill: IMO putting spaces at the end of multi-line strings is a pretty standard thing to do, so I would recommend that.

We could even evolve this to read the commands from a text file at some point if we wanted to remove the need for long strings in code.

objmagic:
@billonahill if this looks good now, we can add more tests now.

billonahill:
This looks good. Can you merge this and then add more tests in another PR?

Issue #1935, opened 2017-06-06
add pmap in heron-shell


huijunw:
[feature request]
pmap is helpful to diagnose memory for process. it is good to add support for command like 'pmap -x <pid>' in heron-shell.

bismoy2013:
@objmagic I think this issue can be marked as closed , I can see you've merged it already #1995 .

PR #1995, opened 2017-06-27
add pmap to shell

huijunwu:
For #1935 
This PR adds 'pmap' to shell following the same way as 'histo'
There will be another PR to add pmap to UI


objmagic:
have you tried locally, since we don't have unit test for heron-shell

maosongfu:
Agree with @objmagic . Please make sure it works in LocalScheduler before merging.

huijunw:
tested on local ubuntu

PR #1997, opened 2017-06-27
add kill-executor handler

huijunwu:
From #1770

Add kill-executor handler for shell.
We consider it as a 'write' operation for the container, which is different from the existing 'read' handlers.
The security check will be added in another PR.


heron/shell/src/python/handlers/killexecutorhandler.py
```diff
@@ -0,0 +1,38 @@
...
+  def post(self):
+    """ post method """
+    logger = logging.getLogger(__file__)
+    logger.info("Received 'Killing parent executor' request")
+    # todo(huijun): add security check
+    logger.info("Killing parent executor response 200")
+    self.set_status(200)
+    self.finish()
+    logger.info("Killing parent executor")
+    os.killpg(os.getppid(), signal.SIGTERM)
```
objmagic: Add newline at end of the file.

objmagic:
Also, have you tried locally that it will work?

heron/shell/src/python/handlers/killexecutorhandler.py
huijunwu: done

huijunw:
tested on local machine

objmagic:
@huijunw don't forget to PR twitter internal repo

PR #2003, opened 2017-06-27
changed travis scripts 

kramasamy:
This change takes an input argument which indicates whether the script is running in 
```
* travis 
* applatix
```
This is needed to identify the environment so that scripts can accordingly use the appropriate config files and toolchain.

### DIFF
--- a/scripts/travis/build.sh
+++ b/scripts/travis/build.sh
@@ -9,6 +9,11 @@ set -e
 DIR=`dirname $0`
 source ${DIR}/common.sh
 
+# check if the ci environ argument is provided
+if [ $# -eq 0 ]; then
+  echo "ci environ arg not provided (travis|applatix)"
+fi
+
 # verify that jars have not been added to the repo
 JARS=`find . -name "*.jar"`
 if [ "$JARS" ]; then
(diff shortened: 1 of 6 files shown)


scripts/travis/build.sh
```diff
@@ -9,6 +9,11 @@ set -e
 DIR=`dirname $0`
 source ${DIR}/common.sh
 
+# check if the ci environ argument is provided
+if [ $# -eq 0 ]; then
+  echo "ci environ arg not provided (travis|applatix)"
```
objmagic: `exit 1` here?

scripts/travis/ci.sh
```diff
@@ -3,12 +3,16 @@
 set -e
 
 DIR=`dirname $0`
-
 source ${DIR}/common.sh
 
+# check if the ci environ argument is provided
+if [ $# -eq 0 ]; then
+  echo "ci environ arg not provided (travis|applatix)"
```
objmagic: `exit 1` early here?

scripts/travis/build.sh
kramasamy: if the argument is not valid, we exit right away - since it is a mandatory argument

scripts/travis/ci.sh
kramasamy: Yes it is a mandatory argument

PR #2005, opened 2017-06-27
Enable heap prof of stmgr in heron-shell.

objmagic:
Tested locally by `curl` heron-shell.

### DIFF
--- a/heron/stmgr/src/cpp/server/stmgr-main.cpp
+++ b/heron/stmgr/src/cpp/server/stmgr-main.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <gperftools/heap-profiler.h>
 #include <iostream>
 #include <string>
 #include <vector>
@@ -55,6 +56,8 @@ int main(int argc, char* argv[]) {
   sp_int32 shell_port = atoi(argv[11]);
   sp_string heron_internals_config_filename = argv[12];
 
+  HeapProfilerStart("stmgr");
+
   EventLoopImpl ss;
 
   // Read heron internals config from local file
@@ -82,5 +85,6 @@ int main(int argc, char* argv[]) {
                           high_watermark, low_watermark);
   mgr.Init();
   ss.loop();
+  HeapProfilerStop();
   return 0;
 }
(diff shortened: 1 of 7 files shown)


srkukarni:
What are the performance implications of doing this? Do we want to do this only when DEBUG is set?

nlu90:
Before we merge this pr into master, let's test with some user topologies to see if there's any performance degradation. 

For debugging the issue, we can make a customized release based on the branch.

maosongfu:
We need to carefully evaluate the performance implications of this since stmgr is very performance-critical.

heron/stmgr/src/cpp/server/stmgr-main.cpp
```diff
@@ -55,6 +56,8 @@ int main(int argc, char* argv[]) {
   sp_int32 shell_port = atoi(argv[11]);
   sp_string heron_internals_config_filename = argv[12];
 
+  HeapProfilerStart("stmgr");
```
huijunw: The profiling is enabled during the whole stmgr life time.

What about adding two handlers to the stmgr server to control the profiling start & end:
1. HandleStartHeapProfilingMessage: HeapProfilerStart()
2. HandleStopHeapProfilingMessage: HeapProfilerStop()
In this way, the profiling is not enabled by default and there is no impact on performance.

objmagic:
I searched around the web but found no official comment on overhead. Several blog posts say enabling heap profiling introduces little overhead. I have also tested on @ttim's large devel topology and I see no performance degradation. 

This PR simply adds shell support. For tracker and UI support, we can do in another patch.

objmagic:
Any comment on this?

srkukarni:
@objmagic One easy check that you can do is run the exclamation topology both with and without this change and see what kind of impact you see.

objmagic:
@srkukarni Agree. I am on PTO and will find time to do so. Also, exclamation topology is throttled now so I will try word count one instead.

huijunw:
Any updates on this PR ? We saw stmgr out of memory on 0.15.1 version again. If the stmgr memory profiling is enabled, it should be easier to debug stmgr out of memory issue.

What about adding two handlers to the stmgr server to control the profiling start & end:
HandleStartHeapProfilingMessage: HeapProfilerStart()
HandleStopHeapProfilingMessage: HeapProfilerStop()
In this way, the profiling is not enabled by default and there is no impact on performance.

Besides, this profiling tool cannot tell which stmgr part code calls new_string(), new_string() occupied more than 60% of total stmgr memory if I remember correctly. @objmagic @srkukarni  thoughts?

PR #2007, opened 2017-06-27
rename tools/travis-ci to tools/travis 

kramasamy:
This is to move a consistent CI naming convention so that we can use scripts easily. Also updated the references to this directory as well.


objmagic:
lgtm.

PR #2011, opened 2017-06-27
limit the mem pool size

nlu90:
After investigations on an aurora container OOM issues inside Twitter, we found that stmgr consumes up to 3G memory in total. This is too large compared to normal 300~600MB usage.

This PR sets an soft limit for the memory pool when `release` is called.  
The mempool works as follows: 1. when `acquire` is called, the mempool will return an object already in the pool or new an object if the pool is empty regardless of the limit; 2. when `release` is called, the mempool will delete the object if the size of the pool >= limit of the pool; Or put it into pool otherwise.

The current limit is 50MB for each mempool given there're multiple type of messages. This ensures that eventually a mempool size won't be larger than 50M.

Also removed unused code `BaseMemPool`

### DIFF
--- a/heron/common/src/cpp/basics/mempool.cpp
+++ b/heron/common/src/cpp/basics/mempool.cpp
@@ -17,9 +17,9 @@
 ////////////////////////////////////////////////////////////////////////////////
 // Just defines the static member of Mempool class
 ////////////////////////////////////////////////////////////////////////////////
-#include "basics/mempool.h"
 #include "basics/basics.h"
 
+// TODO(nlu): get the pool size limit from config
 MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
-                                   new MemPool<google::protobuf::Message>();
+                                   new MemPool<google::protobuf::Message>(50 * 1024 * 1024);
 std::mutex __global_protobuf_pool_mutex__;
--- a/heron/common/src/cpp/basics/mempool.h
+++ b/heron/common/src/cpp/basics/mempool.h
@@ -21,34 +21,7 @@
 #include <unordered_map>
 #include <mutex>
 #include <typeindex>
-
-template<typename T>
-class BaseMemPool {
- public:
-  template<class... Args>
-  T* acquire(Args&&... args) {
-    if (pool_.empty()) {
-      return new T(std::forward<Args>(args)...);
-    }
-    T* t = pool_.back();
-    pool_.pop_back();
-    return t;
-  }
-  void release(T* t) {
-    pool_.push_back(t);
-  }
-  BaseMemPool() {
-  }
-  ~BaseMemPool() {
-      for (auto& p : pool_) {
-        delete p;
-      }
-      pool_.clear();
-  }
- private:
-  std::vector<T*> pool_;
-};
-
+#include "basics/basics.h"
 
 template<typename B>
 class MemPool {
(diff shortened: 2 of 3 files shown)


maosongfu:
@nlu90 Can you also explain what will happen when hitting the limit of mempool in the description?

nlu90:
@maosongfu  Updated the description.

huijunw:
__global_protobuf_pool_acquire__() is called with these types:
StartBackPressureMessage*
StopBackPressureMessage*
TupleStreamMessage2*
HeronTupleSet2*

__global_protobuf_pool_release__() is called with these types:
HeronTupleSet2*
StrMgrHelloResponse*
StartBackPressureMessage*
StopBackPressureMessage*
HeronTupleSet*
TupleStreamMessage2*

I doubt __global_protobuf_pool_release__() continues pushing objects into the vectors of some types but these cached objects are not re-used .?

nlu90:
@huijunw I checked the code, there's a `dispatchMessage` method in `client.h` amd `server.h` which acquires memory from the mempool for messages

srkukarni:
Just a note that in cases when there is heavy traffic, this will not stop the mempool from growing beyond a certain size. It will only ensure that once the traffic passes, the stmgr comes back to some limits rather than remain at the large size.

huijunwu:
@srkukarni is there a way to prevent stmgr heap growing beyond a certain size?

srkukarni:
setrlimit might be one way to go here.

nlu90:
@huijunw Simply limiting the resources used by stmgr to me is not a good solution. It may cause performance issues if the stmgr really needs so many resources.

huijunw:
As @maosongfu suggested, what about triggering back-pressure if heap reaches a certain size ? @nlu90

nlu90:
@huijunw This suggestion in general is good. But before we add this functionality, we need to first analyze what leads to the large heap and see if we bound the mempool will the problem still occur.

srkukarni:
The only reason stmgr uses large amounts of heap is when there is a mismatch of producer and consumer rates. Spout based backpressure is the only mechanism that we have right now to limit this, however there will still be cases when even after back pressure has been applied on the spouts, there are other bolts that are consuming the remaining buffer in the channel and producing more data after processing. However this should be a temporary phenomenon and should go away quickly. The change that @nlu90 did makes sure that when we dont need that much memory, we give it away.
IMO if we really see stmgr using Gigs of ram, we need to investigate what other problems besides this exist if any.

objmagic:
Seems it is not correct to use `sizeof` here. We printed out memory pool stat using [`getSpaceUsed`](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.SpaceUsed)

```
I0714 04:03:08.716549 63460 stmgr.cpp:131] N5heron5proto6system13HeronTupleSetE - nums: 1 bytes used: 2990914
I0714 04:03:08.716555 63460 stmgr.cpp:131] N5heron5proto5stmgr19TupleStreamMessage2E - nums: 1 bytes used: 131063
I0714 04:03:08.716559 63460 stmgr.cpp:131] N5heron5proto6system14HeronTupleSet2E - nums: 550 bytes used: 69140211
I0714 04:03:08.716563 63460 stmgr.cpp:131] N5heron5proto5stmgr24StartBackPressureMessageE - nums: 1 bytes used: 274
I0714 04:03:08.716567 63460 stmgr.cpp:131] N5heron5proto5stmgr23StopBackPressureMessageE - nums: 1 bytes used: 274
I0714 04:03:08.716570 63460 stmgr.cpp:131] N5heron5proto5stmgr19StrMgrHelloResponseE - nums: 1 bytes used: 72
```

Tuples of type TupleStreamMessage2 are using more than 60MB. This means using `sizeof` is wrong and should be removed.

However, notice that `getSpaceUsed` could be heavy, as said in protobuf computation.

objmagic:
I went through lots of discussion with @huijunw and @ttim (thanks them!). Here are some discussions result

1. Looping through vector and sum all `getSpaceUsed` calls results are quite heavy. If we want to do a cap here, how can we determine if we should put it back or delete in a *fast* way? I was thinking to use a thread to calculate mempool size constantly, so when every time we can release, we don't need to loop over to sum message size. Yes, this idea is naive because i did not consider thread-safety, blah blah. 

2. 50MB limit is probably not what we want here. As you can see, `N5heron5proto5stmgr19TupleStreamMessage2E` counts most of the usage. What we are doing here is equivalent to limit whole mempool to  around 50MB...

2. We managed to get a stacktrace. See it [here](https://gist.github.com/objmagic/13a59d0efcc7dcacd096dec59fd9d720). As we can see, `add_data_tuple` calls append, which calls reserve in string library, which calls `_S_create`. When we do heap profiling, we see `_S_create` is the largest source of heap memory usage. Putting this observation here if you are interested details of heap usage.

This observation leads to one point: Message object, which is being constantly put back and taken out from mempool, will keep growing because every time the number of tuples you put is more than ever before, we need to reserve more memory. But we believe it will stop at a certain size.

objmagic:
and to answer question why stmgr memory usage issue gets worse when we switch to 0.14.9.1 from 0.14.8. I think the reason seems to be https://github.com/twitter/heron/pull/1914 starts using mem pool more.

PR #2013, opened 2017-06-28
add security check for shell-kill-executor

huijunwu:
Follow #1997 

For some data center, there are security bots randomly walking, which may hit the heron-shell's kill-executor endpoint. This PR adds request validation.

Tested on local Mac:
`curl -X POST -d "secret=ExclamationTopology8c04f42b-e5f4-43b7-9674-8546697c3820" http://localhost:53969/killexecutor`


objmagic:
instead of POST, can we keep uniform to use GET? So URL should be `http://localhost:53969/killexecutor?secret=ExclamationTopology8c04f42b-e5f4-43b7-9674-8546697c3820`

huijunwu:
For http semantics, kill-executor updates/writes the container state rather than query/read/GET. POST is more proper.

objmagic:
@huijunw make sense. 

LGTM.

PR #2019, opened 2017-06-28
clean common dependency

nlu90:
This PR cleans the dependencies in heron/common component with the following changes:
1. Removed the google guava dependency.
2.Move all test related util classes into a separate directory and create a separate build target for it. Clean the mockito and junit dependencies.
3. Move the HttpJsonClient into K8s scheduler, to clean the fastxml and spi dependencies.


heron/common/src/java/BUILD
```diff
@@ -4,15 +4,15 @@ package(default_visibility = ["//visibility:public"])
 
 load("/tools/rules/heron_deps", "heron_java_proto_files")
 
+# basics-java only depends on jdk.
```
huijunw: Shall we add the design principle at the file beginning? Copied from slack:

```
The heron/common targets is divided into 3 levels by design: 
1. only depends on jdk; 
2. depends on jdk and heron internal code; 
3. depends on jdk, heron internal code and 3rdparty code. 
```

nlu90: good idea. updated

heron/common/src/java/com/twitter/heron/common/test/CommunicatorTestHelper.java
```diff
@@ -11,14 +11,16 @@
 //  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
-package com.twitter.heron.common.basics;
+package com.twitter.heron.common.test;
```
kramasamy: Can you make the package name as com.twitter.heron.common.testhelpers or something similar - you don't want to give the impression that it is some kind of test.

heron/metricsmgr/tests/java/BUILD
```diff
@@ -7,6 +7,7 @@ java_library(
         "//heron/api/src/java:api-java",
         "//heron/common/src/java:basics-java",
         "//heron/common/src/java:network-java",
+        "//heron/common/src/java:test-java",
```
kramasamy: similarly the target should //heron/common/src/java:testhelpers-java

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/HttpJsonClient.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.twitter.heron.common.network;
+package com.twitter.heron.scheduler.kubernetes;
```
kramasamy: I would suggest not to move the file - this is going to be used in DC/OS and ECS later. Hence the original place was more apt and we take a dependency on jackson and we shade jackson where it is being used (like instance).

nlu90: The `network` directory contains very fundamental network util classes for internal usage. Given this `HttpJsonClient` will be used in `schedulers` in different places only, I would suggest we put it into `schedulers-utils`.

Thus we can keep the `common` simple and clean. Let me know your idea, @kramasamy @jrcrawfo

objmagic:
LGTM. By the way, I just checked `heron-api.jar` and noticed Google Guava shaded inside of it. We don't want to do that either.

nlu90:
@objmagic Once this PR is merged, the `heron-api.jar` won't have Google Guava any more. Notice I removed the guava dependency from common's `basics-java` target

objmagic:
@nlu90 I'm wrong. `heron-storm.jar` still has guava dependency
g

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/HttpJsonClient.java
jrcrawfo: @nlu90 agreed, that seems like the best place for it

nlu90:
The `guava` comes from `heron-spi` component. And most of the case are using the annotation `@VisableForTesting`. 

Given it comes from a separate component and the current `common` component is blocking new production releases, I suggest we do a separate cleaning for the `heron-spi` component in a later PR.

objmagic:
Sure. Let's merge this one now.

billonahill:
Nice refactor @nlu90. Mixing the test helpers with the other code bugged me too.

Other changes I'd like for us to consider: 1.) removing the `util` package and pulling it's current sub-packages up to `c.t.h.common`. `util` is non-descriptive and unnecessary here. 2.) Removing `misc` (which is worse than `util` IMO) and finding it's current inhabitants better homes.

PR #2020, opened 2017-06-29
Export heron-instances' streams-aggregated metrics

maosongfu:
Currently heron-instances exports most of metrics scoped by different streams.

But a lot of customers expressed their interest on the value of all streams aggregated and for now they have to sum them up by themselves.
This pull request exports metrics aggregated on different streams too.
Tested with LocalScheduler.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/utils/metrics/ComponentMetrics.java
+++ b/heron/common/src/java/com/twitter/heron/common/utils/metrics/ComponentMetrics.java
@@ -13,16 +13,14 @@
 //  limitations under the License.
 package com.twitter.heron.common.utils.metrics;
 
-import com.twitter.heron.classification.InterfaceAudience;
-import com.twitter.heron.classification.InterfaceStability;
-
 /**
- * Interface for common metric actions that both spouts and bolts support
+ * Abstract Class for common metric actions that both spouts and bolts support
  */
-@InterfaceAudience.Private
-@InterfaceStability.Evolving
-public interface ComponentMetrics {
+public abstract class ComponentMetrics {
+  // Metric-name suffix reserved for value aggregating on all different streams
+  public static final String ALL_STREAMS_AGGREGATED = "__all-streams-aggregated";
 
-  void serializeDataTuple(String streamId, long latency);
-  void emittedTuple(String streamId);
+  public abstract void serializeDataTuple(String streamId, long latency);
+
+  public abstract void emittedTuple(String streamId);
 }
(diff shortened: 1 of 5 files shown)


heron/common/src/java/com/twitter/heron/common/utils/metrics/ComponentMetrics.java
```diff
@@ -13,16 +13,14 @@
...
 /**
- * Interface for common metric actions that both spouts and bolts support
+ * Abstract Class for common metric actions that both spouts and bolts support
  */
-@InterfaceAudience.Private
-@InterfaceStability.Evolving
-public interface ComponentMetrics {
+public abstract class ComponentMetrics {
+  // Metric-name suffix reserved for value aggregating on all different streams
+  public static final String ALL_STREAMS_AGGREGATED = "__all-streams-aggregated";
```
nlu90: can we put this constant string into a separate final class like `MetricsConstants`?

Making a variable public in interface or abstract class smells dangerous.

maosongfu: One the one hand, it seems we don't have such Constants class;
on the other hand, we have a class designed to abstract out the shared actions/implementations, which IMO is a good place to put some shared things, like a shared field.

nlu90: Making a field public while allowing the class to be extended is in general a bad design.

For the problem why it's an interface, @billonahill may have a better idea.

To land this metric fast, I'll approve it for now.

maosongfu: No rush for this pull request. Will also wait for Bill's comments.

ashvina:
@maosongfu Could you please elaborate on how the customers use the aggregated metrics. Also graphite and other tools provide easy mechanism to aggregate metrics. Do we really need to worry about the intent inside the code? In my opinion metrics aggregation is better handled by tools.

maosongfu:
@ashvina Sure. It is desired to handle by the tools. But:
1. Sometimes it is not easy or efficient for tools to do the aggregation. For example, at Twitter, we leverage a lot on wildcard to do pattern matching, and it does not work well on aggregation if there are metrics missing. Also, with wildcard pattern matching, the aggregation is inefficient.
2. For some metrics, people would care more on the value aggregated on all streams. For instance, "execute-ns", which indicates time spending on execute(). It is good to provide those metrics.

ashvina:
Thanks @maosongfu 
With the proposed change, every heron instance will be doing extra work for every tuple. Is this overhead less than the wildcard aggregate inefficiency? What if the aggregation takes place at container or topology level. For instance, the metrics cache could aggregate and push this information to sinks.

maosongfu:
From overhead point of view, doing aggregation on heron-instance side will be cheapest and most scalable: it will be an in-memory java hashmap get() and put(); also the aggregation will be done distributedly in different heron-instances.

PR #2023, opened 2017-06-29
Check if any unshaded third party libraries are introduced into `heron-instance.jar`

objmagic:


scripts/travis/check.sh
```diff
@@ -0,0 +1,20 @@
...
+# Check if any unshaded third party library is inside heron-instance.jar
+
+HERON_CORE="heron-core.tar.gz"
+HERON_CORE_LOC="bazel-bin/scripts/packages/${HERON_CORE}"
+HERON_INSTANCE="heron-core/lib/instance/heron-instance.jar"
+TMP_DIR=$(mktemp -d)
+
+tar -xf $HERON_CORE_LOC -C $TMP_DIR 2>/dev/null
+cd $TMP_DIR
+THIRD_PARTIES="$(jar -tf $HERON_INSTANCE | grep -v 'com/twitter' | grep -oE '(com|org)/[^/]+/' | uniq)"
```
nlu90: need to `sort` before `uniq`

objmagic: `jar tf` gives you sorted result, but just in case I can add a sort here.

objmagic: resolved via https://github.com/twitter/heron/pull/2023/commits/4e04b678ab612dd3b64ba100c993613c21ec1011

PR #2047, opened 2017-07-10
Added Checkpoint Gateway Layer while sending data to instances

srkukarni:
For Stateful Processing, the checkpoint gateway needs to potentially buffer the data before sending it to the local instances. This change makes the necessary changes inside stmgr to do this.


heron/stmgr/src/cpp/manager/stmgr-server.h
```diff
@@ -75,6 +76,9 @@ class StMgrServer : public Server {
   virtual void HandleConnectionClose(Connection* connection, NetworkErrorCode status);
 
  private:
+  void DrainTupleSet(sp_int32 _task_id, proto::system::HeronTupleSet2* _message);
```
huijunw: `DrainTupleSet2` to explicitly mark the function for protobuf tuple version 2

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -404,39 +419,64 @@ void StMgrServer::HandleTupleSetMessage(Connection* _conn,
...
+  TaskIdInstanceDataMap::iterator iter = instance_info_.find(task_id);
   if (iter == instance_info_.end() || iter->second->conn_ == NULL) {
-    LOG(ERROR) << "task_id " << _task_id << " has not yet connected to us. Dropping...";
-  } else {
-    SendMessage(iter->second->conn_, _byte_size, _type_name, _message);
+    LOG(ERROR) << "task_id " << task_id << " has not yet connected to us. Dropping...";
+    drop = true;
   }
+
+  if (!drop) {
```
huijunw: `bool drop` is not necessary. This branch can be in a `else`

srkukarni: Fixed

heron/stmgr/src/cpp/manager/stmgr-server.h
srkukarni: As part of my cleanup, I'm going to change the TupleSet2 protobufs to remove the 2 part in future changes.

huijunw: got it

billonahill:
👍

PR #2054, opened 2017-07-11
Shell refactor

objmagic:
Refactor handlers part so that twitter internal heron-shell can use most of the codes from the OSS side.

### DIFF
--- a/heron/shell/src/python/main.py
+++ b/heron/shell/src/python/main.py
@@ -21,25 +21,9 @@ import tornado.web
 from tornado.httpclient import AsyncHTTPClient
 from tornado.options import define, options, parse_command_line
 
-from heron.shell.src.python import handlers
+from heron.shell.src.python import utils
 
-AsyncHTTPClient.configure(None, defaults=dict(request_timeout=120.0))
-app = tornado.web.Application([
-    (r"^/jmap/([0-9]+$)", handlers.JmapHandler),
-    (r"^/histo/([0-9]+$)", handlers.MemoryHistogramHandler),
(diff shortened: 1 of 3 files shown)


billonahill:
@objmagic could you please provide a description explaining the motivation for the refactor?

heron/shell/src/python/main.py
```diff
@@ -21,25 +21,9 @@
...
-    (r"^/file/(.*)", handlers.FileHandler),
-    (r"^/filedata/(.*)", handlers.FileDataHandler),
-    (r"^/filestats/(.*)", handlers.FileStatsHandler),
-    (r"^/download/(.*)", handlers.DownloadHandler),
-    (r"^/killexecutor", handlers.KillExecutorHandler),
-])
-
-
-if __name__ == '__main__':
+def run(extra_handlers=[]):
```
billonahill: Could we instead allow the caller to pass the entire list of handlers, so they could either remove or add to the defaults?
```
def run(handlers=utils.default_url_to_handlers)
```

objmagic: i see little possibility that we will need to remove in the future. And if we want to enable remove here, we may want to switch to use dict. find and remove a tuple in a list is quite awkward.

huijunw:
shipit after ci pass

heron/shell/src/python/main.py
billonahill: I wouldn't expect someone to remove from the url_to_handlers list, but instead pass a new list, that might include some of those entries.

PR #2055, opened 2017-07-12
Adding debugging to integration test spout, increasing timeouts for get condition

billonahill:
Still seeing flaky scale up tests so adding some debugging to help troubleshoot. Also noticed that since the restart takes longer that the HttpGetCondition used by the spout to see if it should stop emitting now times out. This causes the spout to restart which I think could put it in a bad state. Hence increasing the timeout there. Also saw that we decrement tuplesToAck even if the ack is for a tuple that is not part of the test. That seems wrong.

Here are my notes from troubleshooting:
```
TMaster:
01:20:06 tmaster activates, 1 container 1 spout, 2 indentity bolts, 1 aggregator bolt
01:20:39 tmaster gets new packing plan of 2 containers, restarts without removing location
01:20:51 new tmaster starts, find existing location, dies at 01:21:11
01:21:30 tmaster activates, 2 container 2 spout, 2 indentity bolts, 1 aggregator bolt
01:26:51 shutdown

container_1_ab-spout_1.log.0:
01:19:50 emits 17 tuples, A_0..A_16
01:20:06 deactivated
01:20:47 connection to stream manager lost
01:21:30 reactivated, fails to get topology_updated state from http server, restarts <-- this is where we would lose state about which have been acked
01:21:50 starts, emits 10 tuples until 01:21:59
01:22:20 re-emitting failed tuple starts until shutdown at 01:26:50 <-- Why are acks not received for these?

container_2_ab-spout_5.log.0:
01:21:30 starts, emits 20 tuples until 01:21:51
01:21:51 emits terminal, posts tuples emitted to http server

container_1_identity-bolt_3.log
01:19:49 starts TerminalsToReceive: 1 
01:19:50 received and emits all even A tuples from 1..14 from container_1_ab-spout_1 until 01:20:04
01:20:06 paused
01:20:06 receives and emits tuple A_16 <-- is that this was received after the pause part of the issue?
01:20:47 connection to stream manager lost
01:21:30 reactivated, TerminalsToReceive: 2
01:21:34 recieves and re-emits same tuples reopeatedly until shutdown at 01:26:29

container_1_identity-bolt_4.log
[similar to bolt 3 except there is no tuple received after first pause]

container_1___integration_test_aggregator_bolt_2.log:
01:19:49 starts TerminalsToReceive: 2
01:19:50 receives tuples A_0..B_15 until 01:20:05
01:20:06 paused, updated TerminalsToReceive: 2
01:21:27 reconnect to stream manager TerminalsToReceive: 3
01:21:32 receives tuples A_0..B_19, A_0..B_9, A_0..B_9, A_0..B_9, ... until shutdown at 01:26:50 
```

### DIFF
--- a/integration_test/src/java/com/twitter/heron/integration_test/core/IntegrationTestSpout.java
+++ b/integration_test/src/java/com/twitter/heron/integration_test/core/IntegrationTestSpout.java
@@ -134,11 +134,12 @@ public class IntegrationTestSpout implements IRichSpout {
 
   @Override
   public void ack(Object messageId) {
+    tuplesToAck--;
+    LOG.info("Received an ack with MessageId: " + messageId + " tuplesToAck=" + tuplesToAck);
+
     if (!isTestMessageId(messageId)) {
       delegateSpout.ack(messageId);
     } else {
-      tuplesToAck--;
-      LOG.info("Received an ack with MessageId: " + messageId + " tuplesToAck=" + tuplesToAck);
       handleAckedMessage(messageId, pendingMessages.get(messageId));
     }
     emitTerminalIfNeeded();
--- a/integration_test/src/java/com/twitter/heron/integration_test/core/IntegrationTestSpout.java
+++ b/integration_test/src/java/com/twitter/heron/integration_test/core/IntegrationTestSpout.java
@@ -134,12 +134,11 @@ public class IntegrationTestSpout implements IRichSpout {
 
   @Override
   public void ack(Object messageId) {
-    LOG.fine("Received a ack with MessageId: " + messageId);
-
-    tuplesToAck--;
     if (!isTestMessageId(messageId)) {
       delegateSpout.ack(messageId);
     } else {
+      tuplesToAck--;
+      LOG.info("Received an ack with MessageId: " + messageId + " tuplesToAck=" + tuplesToAck);
       handleAckedMessage(messageId, pendingMessages.get(messageId));
     }
     emitTerminalIfNeeded();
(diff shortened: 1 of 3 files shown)


integration_test/src/java/com/twitter/heron/integration_test/core/IntegrationTestSpout.java
```diff
@@ -134,12 +134,11 @@ public void nextTuple() {
 
   @Override
   public void ack(Object messageId) {
-    LOG.fine("Received a ack with MessageId: " + messageId);
-
-    tuplesToAck--;
     if (!isTestMessageId(messageId)) {
       delegateSpout.ack(messageId);
     } else {
+      tuplesToAck--;
```
huijunw: why eliminate `count--` in the `if` branch

billonahill: tuplesToAck gets incremented when a test message is sent, but decremented when any ack is recieved, even if it's not a test message. This seems like a bug, so I changed decrement to only happen when acks for test messages are received.

I actually don't know why we need to track tuplesToAck, and why we can't just check that size of the pendingMessages map gets to 0, but I'm not ready to make that change yet.

billonahill: Realized this isn't the right change and reverted it, since there are cases where tuples are emitted with a pre-existing message id, in the case that they're not test messages.

PR #2057, opened 2017-07-13
Mempool acquire should always return a Clean protobuf

srkukarni:
Uncleared protobufs might have left over stuff that could cause memory corruption


objmagic:
`Clear` does not release memory back to memory allocator. See [documentation](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.Clear.details).

>Clear() avoids freeing memory, assuming that any memory allocated to hold parts of the message will be needed again to hold the next message. If you actually want to free the memory used by a Message, you must delete it.

FYI, @congwang and I just realized this on Friday.

srkukarni:
The semantics of clear makes sense. This change however was not intended for memory saving but with client semantics in mind. When I acquire a protobuf I expect it to start using it as if I got a clean data structure.

PR #2058, opened 2017-07-13
Added ability in tuple cache to take in checkpoint messages to be sent to downstream tasks

srkukarni:
For stateful processing, once a task takes its checkpoint, it propagates checkpoint markers to all its downstream tasks. Just like regular tuples, these markers flow thru tuple cache in the stmgr. This change adds the ability for tuple cache to accept these messages.

### DIFF
--- a/heron/stmgr/src/cpp/util/tuple-cache.cpp
+++ b/heron/stmgr/src/cpp/util/tuple-cache.cpp
@@ -41,8 +41,8 @@ TupleCache::TupleCache(EventLoop* eventLoop, sp_uint32 _drain_threshold)
 }
 
 TupleCache::~TupleCache() {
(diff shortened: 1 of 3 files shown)


heron/stmgr/src/cpp/util/tuple-cache.cpp
```diff
@@ -91,11 +98,20 @@ TupleCache::TupleList* TupleCache::get(sp_int32 _task_id) {
   return l;
 }
 
+void TupleCache::clear() {
```
billonahill: Can you clarify in comment what the difference is between draining and clearing?

srkukarni: Done

PR #2059, opened 2017-07-13
Add APIs to control packing/repacking algorithm and change instance resources

objmagic:
In Twitter, customers wants the ability to change packing algorithm, and tune instance ram usage without passing `--config-property` to Heron client. This PR adds API to change them.

Yes, the `PackingAlgorithmType.java` and `RepackingAlgorithmType.java` are pretty ugly. Please free feel to advise better solutions.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import javax.xml.bind.DatatypeConverter;
 
 import com.twitter.heron.common.basics.ByteAmount;
+import com.twitter.heron.common.basics.PackingAlgorithmType;
 import com.twitter.heron.common.basics.TypeUtils;
 
 /**
@@ -229,6 +230,12 @@ public class Config extends HashMap<String, Object> {
   public static final String TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS =
       "topology.update.reactivate.wait.secs";
 
+  /**
+   * Packing algorithm used to calculate packing plan
+   */
+  public static final String TOPOLOGY_PACKING_ALGORITHM =
+      "heron.class.packing.algorithm";
+
   private static final long serialVersionUID = 2550967708478837032L;
   // We maintain a list of all user exposed vars
   private static Set<String> apiVars = new HashSet<>();
@@ -265,6 +272,7 @@ public class Config extends HashMap<String, Object> {
     apiVars.add(TOPOLOGY_ADDITIONAL_CLASSPATH);
     apiVars.add(TOPOLOGY_UPDATE_DEACTIVATE_WAIT_SECS);
     apiVars.add(TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS);
+    apiVars.add(TOPOLOGY_PACKING_ALGORITHM);
   }
 
   public Config() {
@@ -452,6 +460,11 @@ public class Config extends HashMap<String, Object> {
     conf.put(Config.TOPOLOGY_EXACTLYONCE_ENABLED, String.valueOf(exactOnce));
   }
 
+  public static void setTopologyPackingAlgorithm(Map<String, Object> conf,
+                                                 PackingAlgorithmType type) {
+    conf.put(Config.TOPOLOGY_PACKING_ALGORITHM, type.toString());
+  }
+
   public void setDebug(boolean isOn) {
     setDebug(this, isOn);
   }
@@ -588,5 +601,9 @@ public class Config extends HashMap<String, Object> {
 
   public void setTopologyExactlyOnceEnabled(boolean exactOnce) {
     setTopologyExactlyOnceEnabled(this, exactOnce);
+  }
+
+  public void setTopologyPackingAlgorithm(PackingAlgorithmType type) {
+    setTopologyPackingAlgorithm(this, type);
   }
 }
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -25,6 +25,7 @@ import javax.xml.bind.DatatypeConverter;
 
 import com.twitter.heron.common.basics.ByteAmount;
 import com.twitter.heron.common.basics.PackingAlgorithmType;
+import com.twitter.heron.common.basics.RepackingAlgorithmType;
 import com.twitter.heron.common.basics.TypeUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


huijunwu:
These config are set in yaml as well. What happens if config in yaml conflicts with setting by topology writter's code?
Does this PR violate 'Neither system- nor component-level configurations can be overridden by topology developers.' in document https://twitter.github.io/heron/docs/operators/configuration/config-intro/

objmagic:
This argument sounds strange to me. If this needs to be true all the time, how shall we change packing algorithm? The only way is to pass `--config-property`?

billonahill:
@huijunw these are not system or component level configs as defined on that page. They're not set in `heron_internal.yaml`. These should be defaulted but overridable by the topology.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -229,6 +231,38 @@
...
+   */
+  public static final String TOPOLOGY_REPACKING_ALGORITHM =
+      "heron.class.repacking.algorithm";
+
+  /**
+   * Amount of RAM per instance to be reserved for this topology.
+   * In bytes.
+   */
+  public static final String TOPOLOGY_INSTANCE_RAM_REQUESTED =
+      "heron.resources.instance.ram";
```
billonahill: please append with `.bytes` to make units explicit. same for disk settings.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -452,6 +487,28 @@ public static void setTopologyExactlyOnceEnabled(Map<String, Object> conf, boole
...
+  public static void setTopologyRepackingAlgorithm(Map<String, Object> conf,
+                                                   RepackingAlgorithmType type) {
+    conf.put(Config.TOPOLOGY_REPACKING_ALGORITHM, type.toString());
+  }
+
+  public static void setInstanceCpuRequested(Map<String, Object> conf, float ncpus) {
+    conf.put(Config.TOPOLOGY_INSTANCE_CPU_REQUESTED, Float.toString(ncpus));
+  }
+
+  public static void setInstanceDiskRequested(Map<String, Object> conf, ByteAmount nbytes) {
```
billonahill: s/nbytes/byteAmount/g here and below.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -452,6 +487,28 @@ public static void setTopologyExactlyOnceEnabled(Map<String, Object> conf, boole
...
+                                                 PackingAlgorithmType type) {
+    conf.put(Config.TOPOLOGY_PACKING_ALGORITHM, type.toString());
+  }
+
+  public static void setTopologyRepackingAlgorithm(Map<String, Object> conf,
+                                                   RepackingAlgorithmType type) {
+    conf.put(Config.TOPOLOGY_REPACKING_ALGORITHM, type.toString());
+  }
+
+  public static void setInstanceCpuRequested(Map<String, Object> conf, float ncpus) {
```
billonahill: s/ncpus/cpus/g here and below

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -452,6 +487,28 @@ public static void setTopologyExactlyOnceEnabled(Map<String, Object> conf, boole
     conf.put(Config.TOPOLOGY_EXACTLYONCE_ENABLED, String.valueOf(exactOnce));
   }
 
+  public static void setTopologyPackingAlgorithm(Map<String, Object> conf,
+                                                 PackingAlgorithmType type) {
```
billonahill: The packing type is just a Class. It needs to be extensible so users can implement their own so an enum won't work.
```
public static void setTopologyPackingClass(Map<String, Object> conf,
    Class<? extends IPacking> packingClass);
public static void setTopologyRepackingClass(Map<String, Object> conf,
    Class<? extends IRepacking> repackingClass);
```

objmagic: I was worried if doing this can pull in lots of dependencies from schedulers. But I can try though.

billonahill: This would cause a problem, since IPacking and IRepacking are in spi, which depends on api, but not the other way around. We could considering moving these interfaces into api. @kramasamy do you have an opinion on this one?

objmagic: You are right. I was not clear in my message above: yes, I was worried about circular dependencies as well when reading BUILD files.

billonahill:
@srkukarni that dep already existed, so removing it should not be in scope for this issue, but it should not make it worse. Opened PR #2061 to remove dep on `TypeUtils` and #2062 to move `ByteAmount` into API.

srkukarni:
@billonahill This pr makes it worse by adding more dep from common.

billonahill:
@srkukarni which I'm saying it shouldn't do. :) That probably wasn't clear in my wording above...

objmagic:
depends on #2061 #2066

heron/api/src/java/com/twitter/heron/api/Config.java
ashvina: Regarding packaging of `IPacking` and `IRepacking`, is the distribution of interfaces between `api` and `spi` packages needed? Given the current structure, moving these interfaces may create confusion. These interfaces are closely linked to other components like `IScheduler`.

We could avoid future movements of interfaces used by users by keeping all the interfaces implemented by custom code in api package?

billonahill: I think the intent (which makes sense to me) is that `api` is where  topology-level APIs are that topology developers might implement or specify (i.e., IBolt, ISpout, etc). The `spi` package is where system-level APIs live that would be implemented by administrating teams for a new scheduler or state manager for example.

I think the packing algorithms could be implemented or at least specified by topology authors at the topology level. Hence it makes sense for those interfaces to move to `api`, but not others perse.

ashvina: Thanks @billonahill I agree with the interpretation of the `api` and `spi` packages. 

I think packing configuration is rarely provided by topology developer. It is rather commonly specified by the administrator or the member deploying the topology via `yaml` config files or command line arguments. To me this fits the definition of system-level API. WDYT?

billonahill: I think the packing impls should be defaulted at the system level, but I certainly see the need to be able to override them at the topology level. We've recommended to some topology owners to specify a different packing impl to change the way resources are allocated for their specific job, which seems reasonable to me.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -229,6 +231,38 @@
   public static final String TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS =
       "topology.update.reactivate.wait.secs";
 
+  /**
+   * Packing algorithm used to calculate packing plan
+   */
+  public static final String TOPOLOGY_PACKING_ALGORITHM =
```
nlu90: I really doubt if we need the packing algorithm to be configurable by users. By the original design, it's a system wide property which should only be set by the heron administrator.

Also, how are the `ram`, `cpu` and `disk` resource configs used? It seems these config doesn't tell the difference for different components. It applies all components' instances for a topology. Correct me if my understanding is wrong here.

billonahill: I could see having multiple packing implementation supported, where power users could specify a non-default based on their needs. Or even write one of their own.

objmagic:
It seems that we are not going to reach an agreement here soon. I have asked @ttim to add `--config-property` support in tsar config.

PR #2060, opened 2017-07-13
Add hooks in stmgr client/clientmgr to send checkpoint markers.

srkukarni:
Also add hook into stmgr to get notifications about dead client connections

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr.cpp
@@ -450,7 +450,7 @@ void StMgr::NewPhysicalPlan(proto::system::PhysicalPlan* _pplan) {
   delete pplan_;
   pplan_ = _pplan;
   neighbour_calculator_->Reconstruct(*pplan_);
-  clientmgr_->NewPhysicalPlan(pplan_);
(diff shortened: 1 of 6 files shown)


heron/stmgr/src/cpp/manager/stmgr.cpp
```diff
@@ -746,5 +746,15 @@ void StMgr::SendStartBackPressureToOtherStMgrs() {
 
 void StMgr::SendStopBackPressureToOtherStMgrs() { clientmgr_->SendStopBackPressureToOtherStMgrs(); }
 
+// Do any actions if a stmgr client connection dies
+void StMgr::HandleDeadStMgrConnection(const sp_string&) {
+  // TODO(srkukarni) For Stateful Topologies, we need to do a restore
```
huijunw: How does a restore help the stmgr resume from dead connection/connection lost?

srkukarni: This will become clearer in the following prs.
But essentially the intuition is that when a stmgr notices that its connection to another stmgr is gone, it should initiate certain actions. Like for instance, if it was running normally, it should ask the tmaster to start a recover topololgy from last sane checkpoint.

Issue #1795, opened 2017-04-04
KafkaSpout: java.nio.HeapByteBuffer is not serealizible for emission


smirnp:
The trouble is that KafkaSpout cannot serialize and emit java.nio.HeapByteBuffer, which was readed from Kafka via RawMultiScheme mode. 

My spout:

SpoutConfig spoutConfig = new SpoutConfig(hosts, kafkaTopic, "/" + kafkaTopic, UUID.randomUUID().toString());
spoutConfig.zkServers = Arrays.asList(new String[]{ zkServerHost });
spoutConfig.zkPort = 2181;
spoutConfig.scheme = new RawMultiScheme();
spoutConfig.startOffsetTime = kafka.api.OffsetRequest.EarliestTime();

Error trace:

SEVERE: Exception caught in thread: spout_spout_1 with id: 22
java.lang.IllegalArgumentException: Class is not registered: java.nio.HeapByteBuffer
Note: To register this class use: kryo.register(java.nio.HeapByteBuffer.class);
	at com.esotericsoftware.kryo.Kryo.getRegistration(Kryo.java:488)
	at com.esotericsoftware.kryo.util.DefaultClassResolver.writeClass(DefaultClassResolver.java:97)
	at com.esotericsoftware.kryo.Kryo.writeClass(Kryo.java:517)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:622)
	at org.apache.storm.serialization.HeronPluggableSerializerDelegate.serialize(HeronPluggableSerializerDelegate.java:47)
	at com.twitter.heron.instance.spout.SpoutOutputCollectorImpl.admitSpoutTuple(SpoutOutputCollectorImpl.java:241)
	at com.twitter.heron.instance.spout.SpoutOutputCollectorImpl.emit(SpoutOutputCollectorImpl.java:117)
	at com.twitter.heron.api.spout.SpoutOutputCollector.emit(SpoutOutputCollector.java:49)
	at org.apache.storm.spout.SpoutOutputCollectorImpl.emit(SpoutOutputCollectorImpl.java:43)
	at org.apache.storm.spout.SpoutOutputCollectorImpl.emit(SpoutOutputCollectorImpl.java:48)
	at org.apache.storm.kafka.PartitionManager.next(PartitionManager.java:171)
	at org.apache.storm.kafka.KafkaSpout.nextTuple(KafkaSpout.java:135)

[...]

PR #2063, opened 2017-07-14
fix kafka spout HeapByteBuffer not serializable problem

nlu90:
The problem is previously reported here: #1795 . And also reported internally today.

The scheme's `deserialize` method is used by kafka spout to read from kafak and generate tuples. If we wrap the ByteBuffer directly in tuple, it cause serialization problems during emit as follows. 

```
java.lang.IllegalArgumentException: Class is not registered: java.nio.HeapByteBuffer
Note: To register this class use: kryo.register(java.nio.HeapByteBuffer.class);
        ...
        at com.twitter.heron.instance.spout.SpoutOutputCollectorImpl.emit(SpoutOutputCollectorImpl.java:85)
        at com.twitter.heron.api.spout.SpoutOutputCollector.emit(SpoutOutputCollector.java:49)
        at org.apache.storm.spout.SpoutOutputCollectorImpl.emit(SpoutOutputCollectorImpl.java:43)
        at com.twitter.storm.spouts.kafka.KafkaSpout.nextTuple(KafkaSpout.java:121)
        at org.apache.storm.topology.IRichSpoutDelegate.nextTuple(IRichSpoutDelegate.java:72)
        ...
        at java.lang.Thread.run(Thread.java:748)
```

This PR fixes the issue. Once test/verification is finished, the PR will be updated for merge.


billonahill:
Wrapping byte[] makes sense but does that break existing consumers that expect ByteBuffer?

Also could you add `@Override` annotations to these method while you're here please.

nlu90:
Twitter internally verified that the fix works.

nlu90:
@billonahill 

This doesn't break existing customer. For the following reasons:
1. Anyone using `ByteBuffer` will facing the same problem here since `ByteBuffer` is not serializable. So there should be no existing case using it correctly with the topology.
2. The wrapped `tuple` is passed into heron internal system via `OutputCollector`, and it accepts the tuple of `List<Object>`. So the byte array is fine to be passed.

Also Added the `@Override` annotation to `RawScheme`'s methods

PR #2067, opened 2017-07-14
Added stateful processing message handlers in stmgr-server

srkukarni:
This change makes StreamManager Server accept all stateful processing messages from peer stmgrs and instances and pass them to the stmgr for actual handling.
The actual handling currently is not yet implemented and will come in future prs.

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-server.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-server.cpp
@@ -684,6 +684,7 @@ void StMgrServer::AttemptStopBackPressureFromSpouts() {
 
 void StMgrServer::InitiateStatefulCheckpoint(const sp_string& _checkpoint_tag) {
   for (auto iter = instance_info_.begin(); iter != instance_info_.end(); ++iter) {
+    // Checkpoint markers originate from spouts.
     if (iter->second->is_local_spout() && iter->second->conn_) {
       LOG(INFO) << "Propagating Initiate Stateful Checkpoint for "
                 << _checkpoint_tag << " to local spout "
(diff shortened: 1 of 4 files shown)


heron/stmgr/src/cpp/manager/stmgr.h
```diff
@@ -66,6 +67,11 @@ class StMgr {
                                proto::stmgr::TupleStreamMessage2* _message);
   void HandleInstanceData(sp_int32 _task_id, bool _local_spout,
                           proto::system::HeronTupleSet* _message);
+  // Called when an instance does checkpoint and sends its checkpoint
+  // to the stmgr to save it
+  void HandleStoreInstanceStateCheckpoint(sp_int32 _task_id,
```
billonahill: why are both instance and task id necessary? Doesn't the former contain the latter?

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -667,5 +682,117 @@ void StMgrServer::AttemptStopBackPressureFromSpouts() {
   }
 }
 
+void StMgrServer::InitiateStatefulCheckpoint(const sp_string& _checkpoint_tag) {
+  for (auto iter = instance_info_.begin(); iter != instance_info_.end(); ++iter) {
+    if (iter->second->is_local_spout() && iter->second->conn_) {
```
billonahill: What is a local spout? Is that a marker on spout instances that are on this container?

srkukarni: A stmgr services a bunch of tasks, some spouts and others bolts. local spouts are the former. This check is basically checking if this task is a local spout and only send the initiate checkpoint message to them. I will add a comment here.

heron/stmgr/src/cpp/manager/stmgr.h
srkukarni: good catch. Fixed.

srkukarni:
@billonahill Unfortunately, there is no unittest for stmgr-server.cpp(which this change targets). I'm planning to augment the stmgr_unittest.cpp which tests the stmgr.cpp in the future pr when stmgr TODO items are filled up.

PR #2069, opened 2017-07-14
Add Pulsar spout.

cckellogg:
A java spout to subscribe to a pulsar topic and receive messages.


cckellogg:
@kramasamy comments where? More javadocs?

huijunw:
There is a spout implementation in heron/contrib/kafka-spout/. Why do we put spouts in different directories?

cckellogg:
There will be some future patches to combine everything - possibly into a connectors directory that will contain spouts and sink bolts.

PR #2073, opened 2017-07-17
Limit number of messages in mempool, instead of total byte sizes

objmagic:
This PR limits size of memory pool by limiting number of messages in each type of memory pool, instead of bytes used.

### Motivation

In the past two weeks, team at Twitter was investigating huge memory usage of stream manager of one topology. We printed out number of messages in memory pool, and sum of messages' size **using [`SpaceUsed`](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.SpaceUsed.details) instead of `sizeof`**. Thanks @ttim for pointing this out and I've confirmed with @congwang as well. We saw memory pool usage of HeronTupleSet2 can reach more than 70MB. This confirms that **using `sizeof` is wrong**, and if we want to limit memory pool usage, we should use `SpaceUsed` instead. To be more specific here, `google::protobuf::Message` allocates memory on heap, and has a fairly complex structure, which is why we should use `SpaceUsed` provided by library to know its size, instead of `sizeof`.

### Why not using `SpaceUsed`

As [documentation of `SpaceUsed`](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.SpaceUsed.details) points out, each call can be quite expensive. Of course, one can argue to use a thread to update mempool usage continuously, but there will be thread-safety issue and seems unnecessarily complicated.

### Why use `size` directly

This PR simply limits number of messages in each memory pool. Advantage here is that implementation is simple and release call can finish in `O(1)`.

### Why set maximum to be 512

Because based on what observed from the topology at Twitter, we see max number of messages in memory pool can be around ~630. It is reasonable to limit pool size to be 512. I am thinking to limit the size more here. Maybe 256? Please comment.

### Side notes
1. `Message->Clear()` does not shrink `Message`. See [here](https://github.com/twitter/heron/pull/2057#issuecomment-315552964)
2. Why `Message` can become bloated gradually? See [here](https://github.com/twitter/heron/pull/2011#issuecomment-315269826)

This should resolve https://github.com/twitter/heron/pull/1567 (cc @ajorgensen )

### DIFF
--- a/heron/common/src/cpp/basics/mempool.cpp
+++ b/heron/common/src/cpp/basics/mempool.cpp
@@ -18,8 +18,15 @@
 // Just defines the static member of Mempool class
 ////////////////////////////////////////////////////////////////////////////////
 #include "basics/basics.h"
+#include "glog/logging.h"
 
 // TODO(nlu): get the pool size limit from config
 MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
                                    new MemPool<google::protobuf::Message>(512);
 std::mutex __global_protobuf_pool_mutex__;
+
+void __global_protobuf_pool_set_pool_max_size(sp_int32 _pool_limit) {
+  std::lock_guard<std::mutex> guard(__global_protobuf_pool_mutex__);
+  LOG(INFO) << "Set max size of each memory pool to " << _pool_limit;
+  __global_protobuf_pool__->set_pool_max_size(_pool_limit);
+}
--- a/heron/common/src/cpp/basics/mempool.cpp
+++ b/heron/common/src/cpp/basics/mempool.cpp
@@ -29,8 +29,8 @@ MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
                                    new MemPool<google::protobuf::Message>(512);
 std::mutex __global_protobuf_pool_mutex__;
 
-void __global_protobuf_pool_set_pool_max_size(sp_int32 _pool_limit) {
+void __global_protobuf_pool_set_pool_max_number_of_messages__(sp_int32 _pool_limit) {
   std::lock_guard<std::mutex> guard(__global_protobuf_pool_mutex__);
   LOG(INFO) << "Set max size of each memory pool to " << _pool_limit;
-  __global_protobuf_pool__->set_pool_max_size(_pool_limit);
+  __global_protobuf_pool__->set_pool_max_number_of_messages(_pool_limit);
 }
--- a/heron/common/src/cpp/basics/mempool.cpp
+++ b/heron/common/src/cpp/basics/mempool.cpp
@@ -20,7 +20,11 @@
 #include "basics/basics.h"
 #include "glog/logging.h"
 
-// TODO(nlu): get the pool size limit from config
+/* The initial value of pool size limit is 512. This value will
+   be overridden by value read from config during initialization
+   of stream manager. The reason why we do this is to avoid
+   circular dependency problem. See https://github.com/twitter/heron/pull/2073
+   for details */
 MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
                                    new MemPool<google::protobuf::Message>(512);
 std::mutex __global_protobuf_pool_mutex__;
--- a/heron/common/src/cpp/basics/mempool.cpp
+++ b/heron/common/src/cpp/basics/mempool.cpp
@@ -21,5 +21,5 @@
 
 // TODO(nlu): get the pool size limit from config
 MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
-                                   new MemPool<google::protobuf::Message>(50 * 1024 * 1024);
+                                   new MemPool<google::protobuf::Message>(512);
 std::mutex __global_protobuf_pool_mutex__;
--- a/heron/config/src/yaml/conf/yarn/heron_internals.yaml
+++ b/heron/config/src/yaml/conf/yarn/heron_internals.yaml
@@ -65,8 +65,8 @@ heron.streammgr.cache.drain.size.mb: 100
 # For efficient acknowledgements
 heron.streammgr.xormgr.rotatingmap.nbuckets: 3
 
-# The max size of the memory pool for any type of message
-heron.streammgr.mempool.size: 512
+# The max number of messages in the memory pool for each message type
+heron.streammgr.mempool.max.message.number: 512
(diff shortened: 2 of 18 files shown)


srkukarni:
In principle I like this change, however I think you should add a heron internals config variable to control this parameter?

objmagic:
@srkukarni Thanks. Add configuration variable takes time because there is circular dependency problems.

heron/common/src/cpp/basics/mempool.cpp
```diff
@@ -21,5 +21,5 @@
 
 // TODO(nlu): get the pool size limit from config
 MemPool<google::protobuf::Message>* __global_protobuf_pool__ =
-                                   new MemPool<google::protobuf::Message>(50 * 1024 * 1024);
+                                   new MemPool<google::protobuf::Message>(512);
```
billonahill: this review looks like you're just calculating size of the pool differently, but what causes the pool size logic to change from being based on number of bytes to number of message in the pool impl?

objmagic: This is explained above: https://github.com/twitter/heron/pull/2073#issue-243279094

srkukarni:
@objmagic could you please elaborate what these cir dep are?

objmagic:
@srkukarni `config-cxx` -> `network-cxx` -> `basics-cxx`. Inside `basics-cxx`, `mempool.h` needs to use `heron-internals-config-reader.`, which is in `config-cxx`. Thus circular dependency.

objmagic:
@srkukarni I have added a method to set limit of memory pool externally. We then set limit of memory pool in `Init`. So we can avoid the circular dependency problem here.

srkukarni:
There might be ways to avoid this. One way is to actually send this parameter as part of the init method. Or have an explicit InitBuffer pool method inside main of stmgr/tmaster.

srkukarni:
@objmagic Sorry, looks like our comments crossed each other. I see that you have done it. Let me take a look at this. Thanks!

heron/common/src/cpp/basics/mempool.cpp
billonahill: Got it. Looks like your changing the 512 to be configurable. Shouldn't the config-driven value be used here now?

objmagic: it is config driven. The initial value is 512, and will be overridden here: https://github.com/twitter/heron/pull/2073/files#diff-5ff5ed00c50afd2302f522a9181e50a1R90

As for why initial value cannot be read from config, see the discussion between Sanjeev and I about circular dependency.

billonahill: Got it. Could you update the TODO nlu comment above to explain why this is config-driven, but hard coded here.

objmagic: @billonahill Updated here: https://github.com/twitter/heron/pull/2073/commits/01acca884a923b0fbe556e266586b8bd54019c76

heron/config/src/yaml/conf/yarn/heron_internals.yaml
```diff
@@ -65,6 +65,9 @@ heron.streammgr.cache.drain.size.mb: 100
 # For efficient acknowledgements
 heron.streammgr.xormgr.rotatingmap.nbuckets: 3
 
+# The max size of the memory pool for any type of message
+heron.streammgr.mempool.size.mb: 512
```
nlu90: Is this config for limiting the # of messages in the mempool or the total byte size in this mempool? This config is different than the other above 9 config.

Could you make them consistent and update the comment and even the config name to avoid confusion?

objmagic: oh you are right. this is typo.

congwang:
This is still ugly. The memory leaks are leaks, adding a limit is just adding a cap to these leaks, doesn't resolve them at all. How would you expect users to figure out how many messages they actually need?

The real solution here is probably to explore the new protobuf arena API to see how we can integrate it with mempool. But that requires protobuf 3.0 at least, fortunately I already have this here: https://github.com/twitter/heron/tree/cwang/protobuf30

srkukarni:
@congwang unless I'm mistaken, afaik, we are not talking about memory leak here. Its just about restricting the size of mempool from growing and remaining high for a long time.

objmagic:
@congwang I agree this solution does not look nice to me.

@srkukarni @congwang I have a theory about the cause of "memory leak". If one does not call 'Clear' on 'Message', used space in 'Message' will not be zero'ed out. So next time we add tuple in it, `Message` thinks it does not have enough space and will enlarge the strings in itself. C++ STL will reserve extra space in string when you append (just like how it uses some heuristics to reserve extra space for vector). Thus we are always allocating extra memory here. 

If my theory is correct, https://github.com/twitter/heron/pull/2057 should solve the issues, along with this PR.

See our heap leak check graph below and [backtrace we get from core dump](https://gist.github.com/objmagic/13a59d0efcc7dcacd096dec59fd9d720) for details.
fake85

srkukarni:
@objmagic Thats not true. If you notice, acquire does a Clear before giving the proto structure to the client, so Clear is called.
You can do one expt. Remove the Clear from acquire to release.

ttim:
@srkukarni protobuf documentation says about this (https://developers.google.com/protocol-buffers/docs/cpptutorial, Optimization Tips):
> Reuse message objects when possible. Messages try to keep around any memory they allocate for reuse, even when they are cleared. Thus, if you are handling many messages with the same type and similar structure in succession, it is a good idea to reuse the same message object each time to take load off the memory allocator. However, objects can become bloated over time, especially if your messages vary in "shape" or if you occasionally construct a message that is much larger than usual. You should monitor the sizes of your message objects by calling the SpaceUsed method and delete them once they get too big.

So even if you do `Clear` you still have some *kind of* memory leak.

Having limited number of objects in pool can fix this issue thou: they will be rotated over the time so most likely each message eventually will be deleted. Having big enough pool at the same time almost guarantee all messages will only become bigger and memory will keep increasing.

objmagic:
@srkukarni You are right. Before your PR, `Clear` is already called properly for all messages.

objmagic:
@ttim Yes. However, current mempool is a `vector`. We push and pop. There is no `rotation` going on here. Maybe we should use `deque`?

PR #2075, opened 2017-07-17
Added Stateful processing related message handling in TMasterClient

srkukarni:
While running in stateful mode, TMaster sends the following messages to the stmgr
1. Periodic Checkpoint messages(StartStatefulCheckpoint) so that stmgr can send them to local instances for checkpointing their state
2. Restore Topology messages(RestoreTopologyStateRequest) so that stmgr can start restoring the state of the local instances to a specified checkpoint_id
3. Start Processing message(StartStmgrStatefulProcessing) to start the processing when all stmgrs have restored their local instances' state.
This change adds the ability for TMasterClient in stmgr to handle those messages and call the appropriate stmgr callback. Note that the stmgr callbacks themselves are not yet implemented, that will be coming in the next set of prs.

### DIFF
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -112,8 +112,15 @@ message RestoreTopologyStateResponse {
 // to some valid checkpoint. This is sent either if stmgr dies
 // and comes back up or if an instance dies.
 message ResetTopologyState {
+  // Was there a dead/recovered stmgr connection that was the reason
+  // for this request?
   optional string dead_stmgr = 1;
+  // Was there a dead/recovered local instance connection that was the reason
+  // for this request?
   optional int32 dead_taskid = 2;
+  // This one describes the reason for sending the message.
+  // Used primarily for debugging
+  required string reason = 3;
 }
 
 // Message sent by stmgr to tmaster informing it about
--- a/heron/stmgr/src/cpp/manager/stmgr.h
+++ b/heron/stmgr/src/cpp/manager/stmgr.h
@@ -138,6 +138,20 @@ class StMgr {
   void BroadcastTmasterLocation(proto::tmaster::TMasterLocation* tmasterLocation);
   void BroadcastMetricsCacheLocation(proto::tmaster::MetricsCacheLocation* tmasterLocation);
 
+  // Called when TMaster sends a InitiateStatefulCheckpoint message with a checkpoint_id
+  // This will send intiate checkpoint messages to local instances to capture their state.
+  void InitiateStatefulCheckpoint(sp_string checkpoint_id);
+
+  // Invoked when TMaster asks us to restore all our local instances state to
+  // the checkpoint represented by _checkpoint_id. This starts the
+  // Restore state machine
+  void RestoreTopologyState(sp_string _checkpoint_id, sp_int64 _restore_txid);
+
+  // Invoked when TMaster sends the StartStatefulProcessing request to kick
+  // start the computation. We send the StartStatefulProcessing to all our
+  // local instances so that they can start the processing.
+  void StartStatefulProcessing(sp_string _checkpoint_id);
+
   heron::common::HeronStateMgr* state_mgr_;
   proto::system::PhysicalPlan* pplan_;
   sp_string topology_name_;
--- a/heron/stmgr/src/cpp/manager/tmaster-client.cpp
+++ b/heron/stmgr/src/cpp/manager/tmaster-client.cpp
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
+#include <string>
 #include <vector>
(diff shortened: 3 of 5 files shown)


heron/stmgr/src/cpp/manager/tmaster-client.cpp
```diff
@@ -217,5 +235,48 @@ void TMasterClient::SendHeartbeatRequest() {
   return;
 }
 
+void TMasterClient::SavedInstanceState(const proto::system::Instance& _instance,
+                                       const std::string& _checkpoint_id) {
+  proto::ckptmgr::InstanceStateStored message;
```
nlu90: should we use the global mempool here and any following places?

heron/stmgr/src/cpp/manager/tmaster-client.cpp
```diff
@@ -217,5 +235,48 @@ void TMasterClient::SendHeartbeatRequest() {
...
+  SendMessage(message);
+}
+
+void TMasterClient::HandleRestoreTopologyStateRequest(
+              proto::ckptmgr::RestoreTopologyStateRequest* _message) {
+  restore_topology_watch_(_message->checkpoint_id(), _message->restore_txid());
+  __global_protobuf_pool_release__(_message);
+}
+
+void TMasterClient::HandleStartStmgrStatefulProcessing(
```
nlu90: Why do we need this start_stateful_processing callback? Whether a topology is stateful or not is configured in user's topology code. Once a stmgr is started, it should already know if it's stateful or not by reading the topology config. And thus can behaves accordingly.

heron/proto/ckptmgr.proto
```diff
@@ -112,8 +112,15 @@ message RestoreTopologyStateResponse {
 // to some valid checkpoint. This is sent either if stmgr dies
 // and comes back up or if an instance dies.
 message ResetTopologyState {
+  // Was there a dead/recovered stmgr connection that was the reason
+  // for this request?
   optional string dead_stmgr = 1;
+  // Was there a dead/recovered local instance connection that was the reason
+  // for this request?
   optional int32 dead_taskid = 2;
```
nlu90: should we allow multiple tasks death?

heron/stmgr/src/cpp/manager/tmaster-client.cpp
srkukarni: My feeling is that its not necessary since its on control path

heron/stmgr/src/cpp/manager/tmaster-client.cpp
srkukarni: So topologies not doing exactly once can start processing right away as soon as assignment are propagated to all stmgrs/instances. However for exactly once semantics, we have to wait until all of the stmgrs/instances are restored to a certain globally consistent checkpoint before proceeding. Once that determination is done by tmaster, this is the message that it sends to the stmgr to start the actual processing

heron/proto/ckptmgr.proto
srkukarni: Not necessary. We issue these messages to the tmaster as soon as we notice any instance failure. The same applies for any stmgr client connection failures

nlu90:
👍

heron/stmgr/src/cpp/manager/stmgr.h
```diff
@@ -138,6 +138,20 @@ class StMgr {
   void BroadcastTmasterLocation(proto::tmaster::TMasterLocation* tmasterLocation);
   void BroadcastMetricsCacheLocation(proto::tmaster::MetricsCacheLocation* tmasterLocation);
 
+  // Called when TMaster sends a InitiateStatefulCheckpoint message with a checkpoint_id
+  // This will send intiate checkpoint messages to local instances to capture their state.
```
billonahill: typo: intiate

heron/stmgr/src/cpp/manager/stmgr.h
```diff
@@ -138,6 +138,20 @@ class StMgr {
...
+
+  // Invoked when TMaster asks us to restore all our local instances state to
+  // the checkpoint represented by _checkpoint_id. This starts the
+  // Restore state machine
+  void RestoreTopologyState(sp_string _checkpoint_id, sp_int64 _restore_txid);
+
+  // Invoked when TMaster sends the StartStatefulProcessing request to kick
+  // start the computation. We send the StartStatefulProcessing to all our
+  // local instances so that they can start the processing.
+  void StartStatefulProcessing(sp_string _checkpoint_id);
```
billonahill: why is the checkpoint id necessary to start processing?

srkukarni: Mostly for correctness checking purposes. Instances will check that they have recovered to this checkpoint id before starting. They should die if this checkpoint id does not match theirs.

PR #2076, opened 2017-07-18
Instantiate checkpoint manager client in stmgr

srkukarni:
The stmgrs in Stateful Processing Heron topologies interact with locally running checkpoint managers to get/save the instance state. This change makes executor pass checkpoint manager port/id to stmgr which allows it to establish a connection.


billonahill:
👍

nlu90:
👍

PR #2078, opened 2017-07-18
Instantiate ClientMgr in constructor

srkukarni:
There is no reason not to construct all relevant datastructures(like clientmgr) in the constructor itself. This does not change any logic because clientmgr does not start opening connections until the StartConnections is called.
This removes a long pending TODO item and also paves way for some pending unittest refactor


huijunw:
One note for the pending unit test refactor #1852,
There is dependency cycle to be solved if we keep the paradigm 'putting all initialization in constructor':
- StartStmgrServer() depends on metrics_manager_client_
- metrics_manager_client_ needs stmgr_port_ after StartStmgrServer()

PR #2080, opened 2017-07-18
Do not log exception when OOM.

objmagic:
In one production topology at Twitter, an instance OOM'ed but did not halt itself. @ttim investigated the issue and believed **another OOM** happened when we were trying to convert exception to string during logging. 

I suggest we remove the logging at all, but @maosongfu thinks we can at least log thread id and names.

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
@@ -235,10 +235,13 @@ public class HeronInstance {
     private void handleException(Thread thread, Throwable exception) {
       // We would fail fast when errors occur to avoid unexpected bad situations
       if (exception instanceof Error) {
-        LOG.log(Level.SEVERE,
-            "Error caught in thread: " + thread.getName()
-                + " with thread id: " + thread.getId() + ". Process halting...");
-        Runtime.getRuntime().halt(1);
+        if (exception instanceof OutOfMemoryError) {
+          // Fail fast without logging, otherwise another OOM may happen
+          Runtime.getRuntime().halt(1);
+        } else {
+          LOG.log(Level.SEVERE, "Error caught in thread: " + thread.getName()
+                + " with thread id: " + thread.getId() + ". Process halting...", exception);
+        }
       }
 
       LOG.log(Level.SEVERE,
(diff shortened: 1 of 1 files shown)


heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
```diff
@@ -237,8 +237,7 @@ private void handleException(Thread thread, Throwable exception) {
       if (exception instanceof Error) {
         LOG.log(Level.SEVERE,
             "Error caught in thread: " + thread.getName()
-                + " with thread id: " + thread.getId() + ". Process halting...",
```
nlu90: Just a clarification, this change basically doesn't log actual message whenever any error happens. Not only the OOM error.

nlu90: I'm fine with this change. Please check with others.

maosongfu: we should either add a "if" or update the description of this pull request.

billonahill: If the concern is only OOM errors I say we check for that and log a message that it's an OOM without logging the error. If it's anything else, we should log it.

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
```diff
@@ -235,10 +235,13 @@ public void uncaughtException(Thread thread, Throwable exception) {
     private void handleException(Thread thread, Throwable exception) {
       // We would fail fast when errors occur to avoid unexpected bad situations
       if (exception instanceof Error) {
-        LOG.log(Level.SEVERE,
-            "Error caught in thread: " + thread.getName()
-                + " with thread id: " + thread.getId() + ". Process halting...");
-        Runtime.getRuntime().halt(1);
+        if (exception instanceof OutOfMemoryError) {
+          // Fail fast without logging, otherwise another OOM may happen
+          Runtime.getRuntime().halt(1);
```
ttim: I guess you can do something like 
```
try {
  LOG.log("...")
} finally {
  Runtime.getRuntime().halt(1);
}
```

This both will log and most likely will halt process.

PR #2084, opened 2017-07-19
travis ci fix

huijunw:
fix https://github.com/twitter/heron/issues/2082
test passed https://travis-ci.org/twitter/heron/builds/255378022

PR #2085, opened 2017-07-19
comment instead of removing 

huijunw:
after #2084

update with comment instead of removing that 'npm install' line.
the root cause is not clear. it may relate to npm install.


billonahill:
I commented in #2084 but we should find and fix the root cause, since this is necessary to validate that the site builds and the links are valid.

PR #2087, opened 2017-07-19
Fix CI build errors (work in progress)

lucperkins:
The CI build is throwing a variety of errors related to the website build. In this PR, I delegate the lion's share of the dependency installation for the website to the Travis CI config rather than the scripts in the `website` folder.


lucperkins:
I'm going to close the PR and propose a new one that simply removes *all* website publishing logic from the Travis CI jobs.

As it stands, the website isn't being published by the CI job anyway, so any website-related failures will cause builds to fail and any successful website builds won't go public anyway.

PR #2088, opened 2017-07-19
Remove website build from Travis CI build

lucperkins:
At the moment, the CI job for Heron builds the website but doesn't actually *publish* the website (which requires a manual process described [here](https://github.com/twitter/heron/blob/master/website/README.md)). This means that the CI job is essentially building the website for no good reason, and when website build failures emerge they make the whole CI job fail, and needlessly so.

This PR completely removes the website build process from CI (just a few stray lines in the config...most of it has already been removed). We should have some kind of collective decision process soon in which we determine how we'd like to proceed in the future. A few options:

* Keep the status quo and build the site manually
* Add the current website publishing process (not just the build process) to the CI job
* Publish the website via a different method, for example:
  * Keep using GitHub pages but don't use the separate `gh-pages` branch
  * Using a different hosting provider altogether

I, personally, am agnostic on this question. Until we get it sorted out, though, I think that these changes are necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,6 @@ before_install:
   - ./bazel-0.3.1-installer-linux-x86_64.sh --user
 
 before_script:
-  # required for building the website docs
-  # - (cd website && npm install && npm install gulp && which gulp)
-  - wget -q https://github.com/gohugoio/hugo/releases/download/v0.24.1/hugo_0.24.1_Linux-64bit.deb
-  - sudo dpkg -i hugo_0.24.1_Linux-64bit.deb
-  - hugo version
-
   # install python module for wheel files
   # required for python packaging
   - sudo pip install wheel
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ env:
   - CC=gcc-4.8 CXX=g++-4.8 CPP=cpp-4.8 CXXCPP=cpp-4.8
 
 before_install:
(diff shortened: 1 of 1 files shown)


lucperkins:
@objmagic Cool. Feel free to merge pending the CI run.

billonahill:
We build the site in CI so we can verify that a.) the site build still works, and b.) all the site links are valid (via the link checker). We could try to do something smart and only do this if there is a change under `website/`, but we should have a site check as part of the build. That way someone can't push bad docs.

We should only publish the site when we cut a release.

lucperkins:
@billonahill The issue, though, is that the website-related stuff is adding a lot of extra time (several minutes to each build) and a lot of extra brittleness that is imposing frustration and other costs on engineers. I do agree that periodic checks are a good thing, but performing these actions on *every* build, across many thousands of builds a year, seems excessive. I'll see if I can come up with a better solution that both uses automation to keep the documentation up to our standards but at a more reasonable cost.

kramasamy:
Perhaps another CI that runs weekly or every other day might help.

lucperkins:
@kramasamy @billonahill That could work, too, Karthik. I'm doing some exploring today. I think I may have a solution that only does a basic site build (without static assets) and checks links. I'll submit a PR hopefully by EOD.

billonahill:
Thanks @lucperkins. Yeah, two ideas come to mine: a second CI or maybe only selectively doing this check if changes under `website/` are detected.

Issue #1904, opened 2017-05-26
Users should be able to specify their desired topology semantics


billonahill:
Instead of having to configure `setAckingEnabled(boolean)` and/or `setTopologyExactlyOnceEnabled(boolean)` of the Heron config objects, users should be able to set their tuple semantics as one of a mutually exclusive set of supported options. We could do this with something like this:

```
enum TupleDeliverySemantics {
  AT_LEAST_ONCE,
  AT_MOST_ONCE,
  EXACTLY_ONCE
}

Config {
  setTupleDeliverySemantics(TupleDeliverySemantics semantics);
}
```

This will allow users to set configs in the way that we discuss semantics. The `setTopologyExactlyOnceEnabled()` can be removed since we haven't shipped that yet and the  `setEnableAcking(..)` method should be deprecated from user-facing config APIs once we support setting the processing semantics. See additional discussion in #1898.

We should do this before shipping support for `setTopologyExactlyOnceEnabled()` so we don't need to break that API later.

maosongfu:
👍

srkukarni:
Solved by #2089

PR #2089, opened 2017-07-19
New API for topology reliability modes

srkukarni:
Currently Heron topologies can run in either atmost_once, atleast_once(and in the upcoming change exactly_once or stateful) modes. The way you specify the mode is not very obvious. For instance, in order to run in atmost once, you need to setEnableAcking(false), and for atleast once, setEnableAcking(true). This way of specifying actually attaches acking(which is the Heron;'s way of tracking tuples) with the semantics of atleast/atmost once. Further more, newer mode like exactly once cause more confusion with more flags that are added and one now needs to worry about which flags are set and which are unset to be really sure of whats going on.
This change introduces TopologyReliabilityMode which is an enum of the various different modes that Heron supports. This is a one stop and intuitive way of specifying the reliability mode that you want to run the topology in. 
Legacy setEnableAcking is now deprecated and will be removed in future releases.


srkukarni:
Fixes #1904

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -100,13 +100,65 @@
...
+     * all of the components of the topology every
+     * TOPOLOGY_STATEFUL_CHECKPOINT_INTERVAL_SECONDS seconds. Upon failure of
+     * any component, its state is recovered to the last checkpoint saved. The
+     * spouts/bolts of the topology have to implement IStatefulComponent for state
+     * to be actually captured.
+     * While this mode provides a way for topology components to recover their state
+     * upon their death, it does not guard against tuple losses that can happen because
+     * of network partitions and other reasons.
+     */
+    STATEFUL,
```
billonahill: This seems orthogonal to the reliability mode IMO. If exactly once leverages the stateful impl that's an implementation detail. From the users perspective these are different things. The reliability mode should specify guarantees with tuple reliability, while the stateful mode should define whether herons state management is being leveraged and enabled by the user logic in the topology.

IIRC the reason for requiring the stateful enabled flag was so that users could turn stateful processing of their topology on or off. If that's the case they might choose to do either while still keeping exactly once enabled.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -302,8 +319,13 @@ public static void setSerializationClassName(Map<String, Object> conf, String cl
     conf.put(Config.TOPOLOGY_SERIALIZER_CLASSNAME, className);
   }
 
+  // This is deprecated. Please use setReliabilityMode instead
```
billonahill: For this and others we should annotate as deprecated:
```
/**
 * Acking is now internally set based on the reliability mode being used.
 *
 * @deprecated use {@link #setReliabilityMode()} instead.  
 */
@Deprecated
public static void setEnableAcking...
```

srkukarni: Fixed

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: Thats an interesting thought! It seems in general there might be lack of consensus/clarity on this one. So I've removed STATEFUL from the list of the modes. We'll add it back later.

heron/instance/src/java/com/twitter/heron/instance/AbstractOutputCollector.java
```diff
@@ -37,6 +37,7 @@
   private long totalTuplesEmitted;
   private PhysicalPlanHelper helper;
 
+  @SuppressWarnings("deprecation")
```
billonahill: Where we add this, we should add a comment that it's only need until TOPOLOGY_ENABLE_ACKING is removed. That way it doesn't linger after TOPOLOGY_ENABLE_ACKING is removed. Same for other occurrences in this pr.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -100,13 +100,53 @@
...
-   * How many executors to spawn for ackers.
+   * Is the topology running in atleast-once mode?
    * <p>
-   * <p>If this is set to 0, then Heron will immediately ack tuples as soon
+   * <p>If this is set to false, then Heron will immediately ack tuples as soon
    * as they come off the spout, effectively disabling reliability.</p>
+   * @deprecated use {@link #TOPOLOGY_RELIABILITY_MODE} instead.
    */
+  @Deprecated
   public static final String TOPOLOGY_ENABLE_ACKING = "topology.acking";
```
billonahill: FWIW, I found no occurrences of `TOPOLOGY_ENABLE_ACKING` or `setEnableAcking` in twitters codebase. @nlu90 @maosongfu do you know how topologies set acking without using those?

heron/instance/src/java/com/twitter/heron/instance/AbstractOutputCollector.java
srkukarni: Added comments

heron/api/src/java/com/twitter/heron/api/Config.java
srkukarni: I remember @maosongfu @kramasamy mentioning that there are some native heron topologies inside Twitter that might be using these features.

billonahill: I thought there were too, but I don't see it being enabled by using either of these. Is there any other way?

billonahill:
👍

PR #2090, opened 2017-07-20
New website checker for Travis CI

lucperkins:
Included in this PR:

* A new link checker for the website ([`html-proofer`](https://github.com/gjtorikian/html-proofer)), which is faster and more configurable than the original link checker
* The link checker almost always fails. Here, the link checker output is shown in CI but link checker failures don't bring down the build, as was the case previously.
* All broken links have been fixed.
* Changes to the Travis CI setup that eliminate the need for a separate setup script (the previous setup script installed Node.js and npm, which can be done via Travis config instead).
* The new website checker doesn't build the static assets (another source of brittleness in the past).

I think that this is a reasonable middle ground between performing only manual website checks when publishing the site and doing a potentially build-breaking website check on every CI run.


billonahill:
Can we fix the broken links to the checker doesn't fail, and then enable it in the build? The idea is to catch bad edits before they're merged, so the person publishing the site doesn't have to deal it.

kramasamy:
@lucperkins - some error in javadocs. Can you please fix it so that we can check it in?

lucperkins:
I'm working on it. The errors are pretty obscure to me.

billonahill:
The errors are because these method references are not valid because their is no no-arg method signature:
```
/home/travis/build/twitter/heron/heron/api/src/java/com/twitter/heron/api/Config.java:312: error: reference not found

   * @deprecated use {@link #setTopologyReliabilityMode()} instead.

                            ^

/home/travis/build/twitter/heron/heron/api/src/java/com/twitter/heron/api/Config.java:497: error: reference not found

   * @deprecated use {@link #setTopologyReliabilityMode()} instead
```

Instead one of these must be used depending on which one we want to sent people to:
```
   * @deprecated use {@link #setTopologyReliabilityMode(TopologyReliabilityMode)} instead.
   * @deprecated use {@link #setTopologyReliabilityMode(Map, TopologyReliabilityMode)} instead.


```

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -309,7 +309,8 @@ public static void setSerializationClassName(Map<String, Object> conf, String cl
 
   /**
    * Is topology running with acking enabled?
-   * @deprecated use {@link #setTopologyReliabilityMode()} instead.
+   * @deprecated use {{@link #setTopologyReliabilityMode(Map, TopologyReliabilityMode)}}
```
billonahill: This occurrence should redirect people to the one with the similar constructor, so the first one with the map. The other javadoc error in this class should send them to the second one without it.

lucperkins: Okay, will change.

PR #2094, opened 2017-07-22
Add base framework for api server.

cckellogg:
What?
Initial setup for heron api server.

initial paths (currently not implemented):
list topologies GET /api/v1/topologies
submit topology POST /api/v1/topologies
kill topology DELETE /api/v1/topologies/topology-name

This will be built in stages and this first path is to setup the framework.


huijunw:
what is the motivation for api server ?

kramasamy:
@huijunw - this is needed for multiple different deployment modes.

WORKSPACE
```diff
@@ -371,3 +371,146 @@ maven_jar(
   artifact = "com.yahoo.pulsar:pulsar-client:jar:shaded:1.18"
 )
 # end Pulsar Client
+
+
+# heron api server
+# jetty
+jetty_version = "9.4.6.v20170531"
```
kramasamy: can you move it to the top - all the version so that it is all in one place.

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/Constants.java
```diff
@@ -0,0 +1,22 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.apiserver;
+
+public class Constants {
+
+  static final int DEFAULT_PORT = 9000;
```
kramasamy: do you want to keep at 8888 similar to tracker?

WORKSPACE
cckellogg: ok

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/Constants.java
cckellogg: I wanted something different until we collapse it so it doesn't interfere with the tracker and ui default ports.

cckellogg:
@huijunw we will add a doc soon. The purpose of this is to add another way to submit topologies and to provide an option that makes it easier to control (submit/kill/etc..) topologies within a clusters (DCOS and kubernetes). For example, users must have their ssh keys on a node in the cluster so they can access zookeeper when submitting or killing. The api server would fix this issue.

huijunwu:
notes from slack channnel
> existing - library mode:
> currently the mode of deployment is called library mode, where we don't have any services running - but a lot of configuration is in the client side. this mode is advantageous since there is no need to run any services.
> 
> this pr - services mode:
> the api server adds another mode of deployment. in this mode, there will not be any configuration at the client side - api server manages everything, include submission/kill/activate/deactivate/ etc. heron cli will use the rest api to even submit jobs. note that we will support both the modes
> 
> future - tracker mode:
> heron tracker in python duplicates state manager code in python as well. if we can make api-server support the heron tracker api in addition to other api, we can run the api server - both in tracker mode and api server mode. 
> initially current tracker and api server will be in separate directory until the good amount of testing is done. once it is done, we can think about cutting over to use api server just as tracker in library mode.
> 
> ps:
> none of the existing deployment will change
>

third_party/java/BUILD
```diff
@@ -230,3 +230,70 @@ java_library(
...
+        "@javax_ws_rs_api//jar",
+        "@hk2_api//jar",
+        "@hk2_utils//jar",
+        "@hk2_locator//jar",
+        "@hk2_aopalliance_repackaged//jar",
+        "@hk2_osgi_resource_locator//jar",
+        "@org_javassit//jar"
+    ],   
+)
+
```
objmagic: lots of unnecessary newlines here

cckellogg: will remove

billonahill:
Before submitting/reviewing code for this could we discussion the motivation and proposed approach for this on the dev apache list, and share any design docs towards this effort. We should have better communication/collaboration before deciding to create a new core architectural component.

PR #2106, opened 2017-07-26
add stateful support into Slave

nlu90:
This PR add stateful support into the SlaveThread. Now the slave thread need to handle two new type of control message:
1. StartStatefulProcessing
2. RestoreInstanceState

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/IInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/IInstance.java
@@ -16,6 +16,7 @@ package com.twitter.heron.instance;
 
 import com.google.protobuf.Message;
 
+import com.twitter.heron.api.state.State;
 import com.twitter.heron.classification.InterfaceAudience;
 import com.twitter.heron.classification.InterfaceStability;
 import com.twitter.heron.common.basics.Communicator;
@@ -30,7 +31,7 @@ public interface IInstance {
   /**
    * Do the basic setup for HeronInstance
    */
-  void start();
+  void start(State<byte[], byte[]> state);
 
   /**
    * Do the basic clean for HeronInstance
--- a/heron/instance/src/java/com/twitter/heron/instance/IInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/IInstance.java
@@ -14,6 +14,8 @@
 
 package com.twitter.heron.instance;
 
+import java.io.Serializable;
+
 import com.google.protobuf.Message;
 
 import com.twitter.heron.api.state.State;
@@ -31,7 +33,7 @@ public interface IInstance {
   /**
    * Do the basic setup for HeronInstance
    */
-  void start(State<byte[], byte[]> state);
+  void start(State<? extends Serializable, ? extends Serializable> state);
 
   /**
    * Do the basic clean for HeronInstance
--- a/heron/instance/src/java/com/twitter/heron/instance/Slave.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/Slave.java
@@ -55,7 +55,7 @@ public class Slave implements Runnable, AutoCloseable {
   private final Communicator<Message> streamOutCommunicator;
   private final Communicator<InstanceControlMsg> inControlQueue;
   private final Communicator<Metrics.MetricPublisherPublishMessage> metricsOutCommunicator;
-  private final IPluggableSerializer serializer;
+  private IPluggableSerializer serializer;
   private IInstance instance;
   private PhysicalPlanHelper helper;
   private SystemConfig systemConfig;
@@ -75,8 +75,6 @@ public class Slave implements Runnable, AutoCloseable {
     this.streamOutCommunicator = streamOutCommunicator;
     this.inControlQueue = inControlQueue;
     this.metricsOutCommunicator = metricsOutCommunicator;
-    this.serializer =
-        SerializeDeSerializeHelper.getSerializer(helper.getTopologyContext().getTopologyConfig());
 
     // TODO (nlu): Create the state with corresponding type according to config
     instanceState = new HashMapState<>();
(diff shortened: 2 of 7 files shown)


heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -183,5 +197,115 @@ public void close() {
...
+    if (instance != null) {
+      instance.stop();
+    }
+
+    // Restore the state
+    instanceState.clear();
+    if (request.getState().hasState() && !request.getState().getState().isEmpty()) {
+      // TODO(nlu): Add interface to allow customized serialization
+      @SuppressWarnings("unchecked")
+      Map<byte[], byte[]> stateToRestore = (Map<byte[], byte[]>) Utils.deserialize(
```
maosongfu: It implies that we would only use java default serialization/deserialization. It is good to have an interface for it, similar to tuple serialization/deserialization. Users could provide an implemented one, or a default one will be used (java default or kryo).

srkukarni: For what its worth, its better to just piggy back on the regular serialization that ppl use for tuples instead of having multiple serialization interfaces.

heron/instance/src/java/com/twitter/heron/instance/IInstance.java
```diff
@@ -30,7 +31,7 @@
   /**
    * Do the basic setup for HeronInstance
    */
-  void start();
+  void start(State<byte[], byte[]> state);
```
billonahill: Why not use generics in the interface here so instances can handle any types of state?

heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -62,6 +74,13 @@ public Slave(SlaveLooper slaveLooper,
     this.streamInCommunicator = streamInCommunicator;
     this.streamOutCommunicator = streamOutCommunicator;
     this.inControlQueue = inControlQueue;
+    this.metricsOutCommunicator = metricsOutCommunicator;
+    this.serializer =
+        SerializeDeSerializeHelper.getSerializer(helper.getTopologyContext().getTopologyConfig());
+
+    // TODO (nlu): Create the state with corresponding type according to config
+    instanceState = new HashMapState<>();
```
billonahill: why have the Slave create empty state? shouldn't this be injected? there could be previous state to be loaded.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -57,6 +58,8 @@
   // The bolt will read Data tuples from streamInQueue
   private final Communicator<Message> streamInQueue;
 
+  private State<byte[], byte[]> instanceState;
```
billonahill: Why are State types being limited to byte[] here? A bolt implementor should be able to define state as anything that satisfies `State<? extends Serializable, ? extends Serializable>` no?

nlu90: good suggestion. Updating it.

nlu90: updated

heron/instance/src/java/com/twitter/heron/instance/Slave.java
nlu90: This empty state is just for initialization. If the instance is instructed to restore from a state, they will receive a `RestoreInstanceStateRequest` request, and thus `handleRestoreInstanceStateRequest ` will take care of loading the state

billonahill: But the state should always be injected from the state server right? The slave should not assume that the state is empty state until it hears from the state server. It should be told this state by someone else, even if it's empty.

nlu90: When `HeronInstance` starts the slave thread, it also doesn't know what's the state for it. So it's not the guy who can inject the state into an instance. Actually, stream manager is the one who sends the state to a local instance via the network. Instead of an empty state, a `null` might be more proper here.

srkukarni: Agree with Neng, null might be more appropriate

billonahill: +1 for null

heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -183,5 +204,116 @@ public void close() {
...
+    // Create a new MetricsCollector with new clean slave looper
+    metricsCollector = new MetricsCollector(slaveLooper, metricsOutCommunicator);
+
+    // Stop current working instance if there is one
+    if (instance != null) {
+      instance.stop();
+    }
+
+    // Restore the state
+    instanceState.clear();
```
billonahill: null check?

heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -170,10 +178,23 @@ public void run() {
...
+    //  2. The TopologyState == RUNNING
+    //  3. - If the topology is stateful and we got the stateful processing start signal
+    //     - If the topology is not stateful
+    if (helper != null && helper.getTopologyState().equals(TopologyAPI.TopologyState.RUNNING)) {
+      Map<String, Object> config = helper.getTopologyContext().getTopologyConfig();
+      boolean isTopologyStateful = String.valueOf(Config.TopologyReliabilityMode.EXACTLY_ONCE)
+          .equals(config.get(Config.TOPOLOGY_RELIABILITY_MODE));
+
+      if (!isTopologyStateful || isStatefulProcessingStarted) {
+        instance.start(instanceState);
```
billonahill: is it possible that instanceState hasn't yet been set here?

heron/instance/src/java/com/twitter/heron/instance/Slave.java
nlu90: good catch

heron/instance/src/java/com/twitter/heron/instance/Slave.java
nlu90: If the topology is not a stateful one, then a `null` state won't affect anything. It's not used by user code at all.

If the topology is a stateful one, and we haven't received the RestoreInstanceStateRequest. Then the components will start with a null value at first. But since the stateful topology wont start processing tuples until receives the `StartStatefulProcessingRequest`, so a null state won't break anything for the first. 

We just need to make sure that, there's always a `RestoreInstanceStateRequest` sent before the `StartStatefulProcessingRequest`. We can guarantee that the `instanceState` is properly initialized when the topology starts to processing

PR #2118, opened 2017-07-28
Added a new socket option to take the maximum packet size accepted

srkukarni:


heron/config/src/yaml/conf/aurora/heron_internals.yaml
```diff
@@ -160,10 +160,16 @@ heron.metricsmgr.network.options.socket.send.buffer.size.bytes: 6553600
...
 
+# The maximum packet size that metrics mgr can read
+heron.metricsmgr.network.options.maximum.packetsize.bytes: 1048576
+
 ################################################################################
 # Configs related to Heron Instance, starts with heron.instance.*
 ################################################################################
 
+# The maximum size of packets that instance can read
+heron.instance.network.options.maximum.packet.bytes: 10485760
```
huijunw: Does `heron.instance.set.data.tuple.size.bytes` have to be less than `heron.instance.network.options.maximum.packet.bytes`? If yes, shall we validate them before use them?

srkukarni: We could. Do you know where is validation done?

PR #2121, opened 2017-07-28
Move away from HeronTupleSet

srkukarni:
Currently the communication between stmgr <-> instance is asymmetric
1. stmgr sends HeronTupleSet2 messages to Instance
2. Instance sends HeronTupleSet messages to stmgr

HeronTupleSet2 was added as a performant variant to decrease the deserialization cost. But the instance deals entirely in HeronTupleSet(to the point of converting the incoming HeronTupleSet2 messages to HeronTupleSet).
This change moves everything in instance deal with HeronTupleSet2.


srkukarni:
On testing on local mac and a linux based local mode in gcloud, i found 7-8% perf regression due to this change. Hence closing this pr.

PR #2122, opened 2017-07-28
Organize the docker directory

kramasamy:


docker/scripts/compile-platform.sh
```diff
@@ -35,8 +35,8 @@ fi
 bazel clean
 
 echo "Creating packages"
-bazel build -c opt --config=$CONFIG_PLATFORM scripts/packages:tarpkgs
-bazel build -c opt --config=$CONFIG_PLATFORM scripts/packages:binpkgs
+bazel build -c opt --jobs 25 --config=$CONFIG_PLATFORM scripts/packages:tarpkgs
```
aahmed-se: What is jobs 25 parameter doing , is it controlling parallelism

kramasamy: It is controlling the number of jobs that can run in parallel. Docker containers in Mac by default comes with 4 cores and 2 GB of memory. Within that constraint, we are able to run 25 jobs without bazel terminating with the error 'EOF'.

cckellogg:
👍

PR #2125, opened 2017-07-29
limit stmgr client reconnect attempts

nlu90:
We found that when network partition happens, stmgr client keeps reconnecting to the other stmgr without any limit. If the network partition problem keeps going on, the whole topology is stuck. 

This PR is to mitigate this problem at heron level. Now stmgr client will kill the whole stmgr process after a configured amount of reconnecting. And if a stmgr process is killed multiple times, heron-executor will die and cause the scheduler to reschedule it. There's a great chance that after rescheduling, the whole container is scheduled to another host and thus avoid the network partition problem.

The PR is tested with local cluster.

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-client.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-client.cpp
@@ -94,6 +94,10 @@ void StMgrClient::HandleConnect(NetworkErrorCode _status) {
     LOG(INFO) << "Connected to stmgr " << other_stmgr_id_ << " running at "
               << get_clientoptions().get_host() << ":" << get_clientoptions().get_port()
               << std::endl;
+
+    // reset the reconnect attempt once connection established
+    reconnect_attempts_ = 0;
+
     if (quit_) {
       Stop();
     } else {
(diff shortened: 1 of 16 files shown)


heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -160,7 +163,16 @@ void StMgrClient::HandleHelloResponse(void*, proto::stmgr::StrMgrHelloResponse*
   client_manager_->HandleStMgrClientRegistered();
 }
 
-void StMgrClient::OnReConnectTimer() { Start(); }
+void StMgrClient::OnReConnectTimer() {
+  reconnect_attempts_ += 1;
```
huijunwu: If stmgr A lost connection to stmgr B and tried 28 times on Monday, and re-connected successfully at the 29th attempt. Four days later on Friday, stmgr A lost connection to stmgr B again and tried to re-connect 2 times and quit.
I suggest resetting `reconnect_attempts_ `to 0 after A resumed connection to B on Monday.

nlu90: Good idea. Updated

srkukarni:
@nlu90 Just looked a little closely at the reconnect interval. Its currently set at 1 second. Which basically means that if we fail for 30 seconds we kill the stmgr. I believe this is too aggressive. Maybe increase the max reconnect attempts from 30 to 300?

nlu90:
@srkukarni Updated.

srkukarni:
looks good!

PR #2126, opened 2017-07-29
Print Command details only in verbose mode

srkukarni:
Currently heron submit spews tons of information even under non verbose mode. This is actually very confusing and non expected behavior. Heron already has the verbose mode to print out details of the submission process. This change ensures that the verbose mode is adhered to.

### DIFF
--- a/heron/tools/cli/src/python/main.py
+++ b/heron/tools/cli/src/python/main.py
@@ -240,7 +240,7 @@ def main():
 
   if command not in ('help', 'version'):
     sys.stdout.flush()
-    Log.info('Elapsed time: %.3fs.', (end - start))
+    Log.debug('Elapsed time: %.3fs.', (end - start))
 
   return 0 if result.is_successful(results) else 1
 
--- a/heron/tools/common/src/python/utils/config.py
+++ b/heron/tools/common/src/python/utils/config.py
@@ -245,7 +245,7 @@ def get_heron_cluster(cluster_role_env):
 def parse_cluster_role_env(cluster_role_env, config_path):
   """Parse cluster/[role]/[environ], supply default, if not provided, not required"""
   parts = cluster_role_env.split('/')[:3]
-  Log.info("Using config file under %s" % config_path)
+  Log.debug("Using config file under %s" % config_path)
   if not os.path.isdir(config_path):
     Log.error("Config path cluster directory does not exist: %s" % config_path)
     raise Exception("Invalid config path")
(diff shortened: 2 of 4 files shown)


heron/tools/common/src/python/utils/config.py
```diff
@@ -245,7 +245,7 @@ def get_heron_cluster(cluster_role_env):
 def parse_cluster_role_env(cluster_role_env, config_path):
   """Parse cluster/[role]/[environ], supply default, if not provided, not required"""
   parts = cluster_role_env.split('/')[:3]
-  Log.info("Using config file under %s" % config_path)
+  Log.debug("Using config file under %s" % config_path)
```
srkukarni: Why do we need it in non verbose mode?

heron/tools/cli/src/python/main.py
```diff
@@ -240,7 +240,7 @@ def main():
 
   if command not in ('help', 'version'):
     sys.stdout.flush()
-    Log.info('Elapsed time: %.3fs.', (end - start))
+    Log.debug('Elapsed time: %.3fs.', (end - start))
```
srkukarni: Again the same reason as above. Why do we need it in non verbose mode?

kramasamy:
Basically, we need the following when we submit a topology 

[INFO]: Using config file under /Users/karthikz/.heron/conf/local
[INFO]: Launching topology: 'etc'
[INFO]: Uploading topology jar to <whatever the uploader is>
[INFO]: Scheduling topology in <whatever the scheduler is>
[INFO]: Successfully launched topology 'etc'
[INFO]: Elapsed time: 2.405s.

Providing minimal information as each step is complete - so that user knows at a high level what is the current step, if that step is successful and the total time it took to complete. We print Elapsed time for all the commands just - to give an idea for the user - similar to what bazel does.

srkukarni:
@kramasamy 
The config file looks like stuff in verbose mode. When submitting regularly, I really don't look at the config file.
I agree that we need to add some more comments in Uploading to and scheduling in. I can add those in later prs.
The Elapsed time comparison to bazel is slightly is not appropriate imo. bazel is a compilation tool which values how much time it spent while compiling. We are just submitting the job, submission time is something that's not really a highlight.

kramasamy:
Elapsed time is need since the jobs are varying size, operating over different schedulers and different networks. I would feel comfortable keeping it. In fact it was added based on feedback from teams. On the config file, if you have multiple clusters and different config paths being provided, it is good to display so that you have an idea which config file is being used - since there are typically multiple of them.

PR #2129, opened 2017-07-31
While unzipping pex files, don't overwrite existing files.

srkukarni:
This is useful while running multi container python topologies in local mode
where multiple executors are unzipping the same file

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -677,7 +677,7 @@ class HeronExecutor(object):
     if self.pkg_type == "tar":
       os.system("tar -xvf %s" % self.topology_bin_file)
     elif self.pkg_type == "pex":
-      os.system("unzip %s" % self.topology_bin_file)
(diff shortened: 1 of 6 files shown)


heron/executor/src/python/heron_executor.py
```diff
@@ -677,7 +677,7 @@ def _untar_if_needed(self):
     if self.pkg_type == "tar":
       os.system("tar -xvf %s" % self.topology_bin_file)
     elif self.pkg_type == "pex":
-      os.system("unzip %s" % self.topology_bin_file)
+      os.system("unzip -qq -n %s" % self.topology_bin_file)
```
kramasamy: What happens when you relaunch the topology after changing code? Then the new code will not take effect?

srkukarni: The thing is that when launching we need to clean the directory. I've updated the pr to ensure that we clean the working directory while setting it up. This way, new topology launches will be ok.

PR #2134, opened 2017-08-02
Initialize the instance state before open/prepare instance

nlu90:
This PR made the following two changes:
1. call `initState()` if the component is a statefulComponent during `start()`
2. Replace the `? extends Serializable`  with `Serializable` for `State` related variables and methods in Instance code

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/Slave.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/Slave.java
@@ -62,7 +62,7 @@ public class Slave implements Runnable, AutoCloseable {
 
   private boolean isInstanceStarted = false;
 
-  private State<? extends Serializable, ? extends Serializable> instanceState;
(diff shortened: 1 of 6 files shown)


heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -249,8 +249,8 @@ private void handleRestoreInstanceStateRequest(InstanceControlMsg instanceContro
     }
     if (request.getState().hasState() && !request.getState().getState().isEmpty()) {
       @SuppressWarnings("unchecked")
```
srkukarni: why is the SuppressWarnings needed?

nlu90: yes, we need to suppress the cast of `State<Serializable, Serializable>`, java's type erasure the two parameter type of the `State` is lost.

PR #2145, opened 2017-08-03
Fix the paths for local stateful storage

srkukarni:


heron/config/src/yaml/conf/local/stateful.yaml
```diff
@@ -3,7 +3,7 @@ heron.statefulstorage.classname:            "com.twitter.heron.statefulstorage.l
 
 heron.statefulstorage.config:
   heron.statefulstorage.classpath:           ""
-  heron.statefulstorage.localfs.root.path:   "./checkpoints"
+  heron.statefulstorage.localfs.root.path:   ~/.herondata/checkpoints
```
kramasamy: you should use ${HOME}/.herondata/checkpoints

heron/config/src/yaml/conf/localzk/stateful.yaml
```diff
@@ -3,7 +3,7 @@ heron.statefulstorage.classname:            "com.twitter.heron.statefulstorage.l
 
 heron.statefulstorage.config:
   heron.statefulstorage.classpath:           ""
-  heron.statefulstorage.localfs.root.path:   "./checkpoints"
+  heron.statefulstorage.localfs.root.path:   ~/.herondata/checkpoints
```
kramasamy: you should use ${HOME}/.herondata/checkpoints

heron/config/src/yaml/conf/local/stateful.yaml
srkukarni: It's much easier to expand ~ than using ${HOME}.
BTW the other uses of ${HOME} used in files like statemgr.yaml is really not used by the code. Its being overriden manually. See https://github.com/twitter/heron/blob/master/heron/statemgrs/src/java/com/twitter/heron/statemgr/localfs/LocalFileSystemStateManager.java#L245
I will fix this as well in another pr

srkukarni:
Related to #2146

srkukarni:
https://github.com/twitter/heron/issues/2146 will track moving away from CheckpointManagerConfig to the generic Config class.

PR #2153, opened 2017-08-04
Add pom generation to heron-api

aahmed-se:
Make version tag externally configurable in future PR

### DIFF
--- a/heron/api/src/java/BUILD
+++ b/heron/api/src/java/BUILD
@@ -45,3 +45,16 @@ java_library(
     srcs = glob(["com/twitter/heron/classification/**/*.java"]),
     resources = ["//heron/api/src:processor-service-conf"],
 )
+
+genrule(
+    name = "pom",
+    outs = ["pom.xml"],
+    cmd = "$(location generate_pom) >$@",
+    output_to_bindir = 1,
+    tools = [":generate_pom"],
+)
+
+cc_binary(
+    name = "generate_pom",
+    srcs = ["generate_pom.cc"],
+)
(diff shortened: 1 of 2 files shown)


heron/api/src/java/BUILD
```diff
@@ -45,3 +45,16 @@ java_library(
     srcs = glob(["com/twitter/heron/classification/**/*.java"]),
     resources = ["//heron/api/src:processor-service-conf"],
 )
+
```
kramasamy: generate_pom.cc should go into tools - this looks a generation tool that can be shared by others.

aahmed-se: Will have to parameterize it to make it shareable , will do that.

heron/api/src/java/generate_pom.cc
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+
+#define HERON_VERSION "0.14.6"
+
+int main(int argc, char** argv) {
+  std::string tmpl(R"EOF(<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.twitter.heron</groupId>
+  <artifactId>libtensorflow</artifactId>
```
objmagic: libtensorflow?

aahmed-se: Will change that and take a look at the existing pom scripts.

objmagic: Yes, we **must** follow existing POM files. The generated POM files **must** be the same one generated now.

Please read https://github.com/twitter/heron/pull/1811 if you want to know why.

And why a C++ program to do this job? A Python program should do well.

Issue #2154, opened 2017-08-04
Calling SLF4J Logger.error(String, Throwable) can crash HI in catch block


yohan233:
*Note*: I've tracked this issue down and fixed it, but I can't figure out how to test the fix, so I am making an issue for now

In bolt code, I often wrap the entire execute() method in a try-catch block, and then handle different exceptions that may arise (some cause the tuple to ack, some to fail, etc).  I nearly always call Logger.error(String, Throwable) in these cases. 

Doing so inside of a catch block can crash the HI if null is passed into the logger function:

```
[2017-08-03 20:53:20 +0000] [SEVERE] com.twitter.heron.instance.HeronInstance: Exception caught in thread: SlaveThread with id: 12 
java.lang.NullPointerException
    at com.twitter.heron.proto.system.Metrics$ExceptionData$Builder.setLogging(Metrics.java:1889)
    at com.twitter.heron.common.utils.logging.ErrorReportLoggingHandler.publish(ErrorReportLoggingHandler.java:107)
    at java.util.logging.Logger.log(Logger.java:738)
    at org.slf4j.impl.JDK14LoggerAdapter.log(JDK14LoggerAdapter.java:582)
    at org.slf4j.impl.JDK14LoggerAdapter.error(JDK14LoggerAdapter.java:559)
    at com.sis.sis_storm.feeds.processor.AggregateStageBolt.execute(AggregateStageBolt.java:312)
    at com.twitter.heron.instance.bolt.BoltInstance.readTuplesAndExecute(BoltInstance.java:209)
    at com.twitter.heron.instance.bolt.BoltInstance$1.run(BoltInstance.java:154)
    at com.twitter.heron.common.basics.WakeableLooper.executeTasksOnWakeup(WakeableLooper.java:136)
    at com.twitter.heron.common.basics.WakeableLooper.runOnce(WakeableLooper.java:75)
    at com.twitter.heron.common.basics.WakeableLooper.loop(WakeableLooper.java:65)
    at com.twitter.heron.instance.Slave.run(Slave.java:169)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
```

The problem is in ErrorReportLoggingHandler.java, line 107:

`        exceptionDataBuilder.setLogging(record.getMessage());`

[...]

PR #2161, opened 2017-08-07
Fix NPE when logging exceptions that have no message using SLF4J

yohan233:
What?

Fix NullPointerException that can occur when logging certain types of exceptions

Why?
If logged in a catch block, whole HI can crash. 

Issue with more explanation: 

https://github.com/twitter/heron/issues/2154


yohan233:
Seems to have failed due to not being able to reach maven.. is there a way to retry?

srkukarni:
@yohan233 restarted build

PR #2164, opened 2017-08-08
Add kubernetes stateful config.

cckellogg:


heron/config/src/yaml/conf/kubernetes/stateful.yaml
```diff
@@ -0,0 +1,21 @@
+# indicate the type of storage for checkpoint is local file system
+heron.statefulstorage.classname: "com.twitter.heron.statefulstorage.localfs.LocalFileSystemStorage"
+
+heron.statefulstorage.config:
+heron.statefulstorage.classpath: ""
```
kramasamy: It should be a sub block

PR #2176, opened 2017-08-10
Create a connectors directory for spouts and bolts (sources/sinks)

cckellogg:
Create a common place to put spouts and bolts.


srkukarni:
One broad comment. This change seems to be making a single connectors and putting spouts/sinks in there. Instead would a second level differentiation help? I.e. connectors/spouts, connections/sinks?

cckellogg:
We could have a second level that groups things by type connectors/pulsar, connectors/jdbc, etc... The source (spout) and sink (bolt) would be together in the second level in case they need to share common code. This is what other projects do as well. Thoughts?

srkukarni:
@cckellogg makes sense

heron/examples/src/python/BUILD
```diff
@@ -55,7 +55,7 @@ pex_binary(
     deps = [
       '//heron/api/src/python:heron-python-py',
       '//heron/dsl/src/python:heron-pythondsl-py',
-      '//heron/spouts/src/python:heron-pythonspouts-py',
+      '//heron/connectors/src/python:heron-pythonspouts-py',
```
srkukarni: pythonconnectors-py. Here and other places

PR #2181, opened 2017-08-11
Add health manager operation mode

ashvina:
This PR adds cluster operation mode to the health manager. The cluster mode would be needed to execute the HM on container 0.

### DIFF
--- a/heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
+++ b/heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
@@ -159,7 +159,7 @@ public class HealthManager {
       throw new RuntimeException("Error parsing command line options: ", e);
     }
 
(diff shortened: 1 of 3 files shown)


heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -173,6 +224,18 @@ public static void main(String[] args) throws Exception {
     }
   }
 
+  private static boolean hasOption(CommandLine cmd, CliArgs argName) {
```
huijunw: If these functions include a single line simple body, I would suggest calling its body directly. I do not see the necessity of wrapping it in function.

ashvina: The methods are utility methods related to use of enum to replace hard-coded argument names. Instead of adding `.text` everywhere, which seems annoying, I added these utility methods

huijunw: got it

PR #2194, opened 2017-08-14
Move over from using iteritems to items since its more python3 friendly

srkukarni:
Python3 dropped iteritems in favor of items. This pr removes all iteritems usages in heron repository replacing them with items.
One difference between items and iteritems while running in python2 is that items actually materializes the list. However apart from the usage in python instance, all other usages of iteritems actually have lists that are typically very small(usually in 10s of items and never more than 1000). Thus the performance implications are minimal.
For python instance usage, we have refactored the code to using next(iter()) which is equally fast in both python2 and 3.

### DIFF
--- a/heron/api/src/python/component/component_spec.py
+++ b/heron/api/src/python/component/component_spec.py
@@ -108,7 +108,7 @@ class HeronComponentSpec(object):
     # iterate through self.custom_config
     if self.custom_config is not None:
       sanitized = self._sanitize_config(self.custom_config)
-      for key, value in sanitized.iteritems():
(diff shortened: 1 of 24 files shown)


objmagic:
pretty sure we need some time to make Python3 default inside twtr

srkukarni:
@objmagic Note that this change does not force python3. Rather it moves away from using apis that are python2 only to apis that are existing in both python2 and python3. This way existing stuff we can run Heron on both python2(which is what Twitter's environment is) as well as python3.

objmagic:
ok sounds good

PR #2200, opened 2017-08-15
Upgrade gflags to 2.2.1

srkukarni:
The current gflags version (2.0) was released in 2013 and is thus 4 years old. This one updates it to the latest version


kramasamy:
Ship it!

PR #2203, opened 2017-08-16
Ability to launch health manager on container-0

ashvina:
If configured to run in cluster mode, the executor will start the health-manager on container-0.

Discussion: https://lists.apache.org/thread.html/438381a6efef28b6e8f33341b87e766734e72e2c9a0eab180ebf8008@%3Cdev.heron.apache.org%3E

I have tested the changes on local scheduler and executing some tests on yarn. This PR is for early review and feedback.


heron/config/src/yaml/conf/aurora/heron.aurora
```diff
@@ -20,7 +20,7 @@ fetch_user_package = Process(
   cmdline = 'curl %s -o %s && tar zxf %s' % (heron_topology_jar_uri, topology_package_file, topology_package_file)
 )
 
-command_to_start_executor = '{{EXECUTOR_BINARY}} {{mesos.instance}} {{TOPOLOGY_NAME}} {{TOPOLOGY_ID}} {{TOPOLOGY_DEFINITION_FILE}} {{STATEMGR_CONNECTION_STRING}} {{STATEMGR_ROOT_PATH}} {{TMASTER_BINARY}} {{STMGR_BINARY}} "{{METRICSMGR_CLASSPATH}}" {{INSTANCE_JVM_OPTS_IN_BASE64}} "{{TOPOLOGY_CLASSPATH}}" {{thermos.ports[port1]}} {{thermos.ports[port2]}} {{thermos.ports[port3]}} {{SYSTEM_YAML}} {{COMPONENT_RAMMAP}} {{COMPONENT_JVM_OPTS_IN_BASE64}} {{TOPOLOGY_PACKAGE_TYPE}} {{TOPOLOGY_BINARY_FILE}} {{JAVA_HOME}} {{thermos.ports[http]}} {{SHELL_BINARY}} {{thermos.ports[port4]}} {{CLUSTER}} {{ROLE}} {{ENVIRON}} "{{INSTANCE_CLASSPATH}}" {{METRICS_YAML}} "{{SCHEDULER_CLASSPATH}}" "{{thermos.ports[scheduler]}}" {{PYTHON_INSTANCE_BINARY}} {{METRICSCACHEMGR_CLASSPATH}} {{thermos.ports[metricscachemgr_masterport]}} {{thermos.ports[metricscachemgr_statsport]}} {{IS_STATEFUL_ENABLED}} "{{CKPTMGR_CLASSPATH}}" {{thermos.ports[ckptmgr_port]}} {{STATEFUL_CONFIG_YAML}}'
```
nlu90: Since this change changes the aurora conf file and also adds new yaml conf file, it will break the current topology deployment process for Twitter's environment. Please coordinate with @huijunw @objmagic or me to make sure that the necessary changes inside twitter are also made before merging it

huijunw: thanks for reminder

ashvina: Ok, please let me know once the private env is ready.

Also, I think it is time to make the `executor` smart. As per the current pattern, the scheduler loads the configs and passes to the `executor`. This could be avoided as the executor has access to the config files. Its a separate PR/discussion though. WDYT?

huijunw: The original idea of heron-executor is the simple entry point in the container to start and monitor the other processes in the same container. Loading configs is not heron-executor's job. However, heron-executor evolves complex. Correct me if i am wrong.

We may need better design which config should passed from scheduler and which should be loaded.

heron/executor/src/python/heron_executor.py
```diff
@@ -214,6 +215,9 @@ def init_parsed_args(self, args):
     self.ckptmgr_classpath = parsed_args.ckptmgr_classpath
     self.ckptmgr_port = parsed_args.ckptmgr_port
     self.stateful_config_file = parsed_args.stateful_config_file
+    self.healthmgr_mode = parsed_args.healthmgr_mode
+    self.healthmgr_classpath = '%s:%s' % (self.scheduler_classpath, parsed_args.healthmgr_classpath)
```
huijunw: why is scheduler class path in healmgr class path

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -159,22 +161,12 @@ public static void main(String[] args) throws Exception {
...
-      loggingLevel = Level.FINE;
-    }
-    LoggingHelper.loggerInit(loggingLevel, false);
-
     HealthManagerMode mode = HealthManagerMode.cluster;
     if (hasOption(cmd, CliArgs.MODE)) {
       mode = HealthManagerMode.valueOf(getOptionValue(cmd, CliArgs.MODE));
     }
 
+    LOG.info("Health manager mode: " + mode);
```
huijunw: This LOG is before `setupLogging(cmd, config);`?

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
```diff
@@ -200,12 +192,20 @@ public static void main(String[] args) throws Exception {
...
+    setupLogging(cmd, config);
 
     LOG.info("Static Heron config loaded successfully ");
     LOG.fine(config.toString());
 
+    // load the default config value and override with any command line values
+    String metricSourceClassName = config.getStringValue(PolicyConfigKey.METRIC_SOURCE_TYPE.key());
+    metricSourceClassName = getOptionValue(cmd, CliArgs.METRIC_SOURCE_TYPE, metricSourceClassName);
+
+    String metricsUrl = config.getStringValue(PolicyConfigKey.METRIC_SOURCE_URL.key());
```
huijunw: If MetricsCacheMgr is the source, MetricsCacheMgr port is not determined in the config yaml. The port in the url has to be passed by command line. I suggest pass the url in command line for MetricsCacheMgr. The heron-executor knows MetricsCacheMgr port.

heron/executor/src/python/heron_executor.py
ashvina: The health manager needs repacking algorithm, state manager and scheduler client. All these components are part of scheduler classpath.

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
ashvina: I realized that the cache manager port is dynamic. As the executor does not know the type of the provider and the metrics cache port does not apply to other metrics providers. So that is not feasible. The metrics cache port is saved in state manager. Health manager has the state manager connection. So the metrics cache provider can read the port from state manager. It will ignore the url provided in the config. If you agree I will create a PR for this?

huijunwu: agree.

huijunw: @ashvina any update?

ashvina:
Marking the PR as ready to merge after discussing with @huijunw

heron/healthmgr/src/java/com/twitter/heron/healthmgr/HealthManager.java
ashvina: Fixed by removing the log message.

PR #2207, opened 2017-08-17
changes to make heron compile with bazel 0.5.4

kramasamy:


nlu90:
Verified with twitter's environment, the bazel 0.5.4 can compile the project successfully.

Please fix the falied check for travis here so that we can approve it.

srkukarni:
@nlu90 It passes now

nlu90:
👍

PR #2211, opened 2017-08-18
upgrade gflags and glog to newer versions

kramasamy:


third_party/gflags/BUILD
```diff
@@ -33,12 +32,13 @@ genrule(
         'cd $$TMP_DIR',
         'tar xfz ' + package_file,
         'cd ' + package_dir,
-        './configure --prefix=$$INSTALL_DIR --enable-shared=no',
+        'cmake -DCMAKE_INSTALL_PREFIX:PATH=$$INSTALL_DIR .',
```
aahmed-se: why are we switching to cmake ?

kramasamy: new gflags use cmake now - instead of configure/make in the older versions.

srkukarni:
I didnt see the old gflags package being removed?

PR #2214, opened 2017-08-18
remove inputs from streamlet since the core streamlet class does not use it

srkukarni:


heron/dsl/src/python/filterbolt.py
```diff
@@ -79,5 +79,5 @@ def _build_this(self, builder):
     if not callable(self._filter_function):
       raise RuntimeError("filter function must be callable")
     builder.add_bolt(self._stage_name, FilterBolt, par=self._parallelism,
-                     inputs=self._inputs,
+                     inputs=self._calculate_inputs(),
```
aahmed-se: You are deleting this function

srkukarni: I'm deleting the interface function from streamlet, but letting all other derivates implement calculating inputs in their own way. Thus the _calculate_inputs method here is the local one defined above

PR #2216, opened 2017-08-19
Add a heron downloader to download and extract topology packages.

cckellogg:
This is needed to offer support for downloading topology packages from storage systems that do not support a http interface.


kramasamy:
Ship it - once the CI passes.

PR #2223, opened 2017-08-21
Convert kubernetes scheduler to use heron downloader.

cckellogg:


cckellogg:
@jrcrawfo what do you think of this change?

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -400,9 +404,9 @@ protected ArrayNode getPorts(ObjectMapper mapper) {
   /**
    * Get the command that will be used to retrieve the topology JAR
    */
-  protected String getFetchCommand() {
-    return "cd /opt/heron/ && curl " + Runtime.topologyPackageUri(runtime).toString()
-            + " | tar xvz";
```
jrcrawfo: No need to extract the TAR file anymore? Assuming you've tested this.

cckellogg: the downloader extracts the tar.gz on the file. There is a test for this (ExtractorTests under heron/downloaders/tests)

jrcrawfo:
👍  looks good to me

PR #2228, opened 2017-08-22
Make bazel configure python3 friendly

srkukarni:
As part of general cleanup and making things Python3 friendly, this change sanitizes bazel_configure to use functions that work both in python2 and python3.

### DIFF
--- a/bazel_configure.py
+++ b/bazel_configure.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7
+#!/usr/bin/env python
 #
 # Verifies required libraries and tools exist and are valid versions.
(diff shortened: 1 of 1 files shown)


huijunw:
do we plan to refactor the present src code to be compatible with both python 2 and python 3?

srkukarni:
Yes. One step at a time :)
First we need to make sure that heron compiles both in python2 and python3. Currently the sticking points for this are a) bazel needs to be upgraded to 0.5.4 and b) pex needs to be upgraded in place

huijunw:
Cool. Looks like a long journey. :)

PR #2229, opened 2017-08-22
Move api, dsl, tools, proto and connectors

aahmed-se:
This commit is to reorg the code to prepare a new top
level project heronpy


huijunw:
what is the motivation to separate python from the current directory structure?

aahmed-se:
We are trying to segregate the external client python package from the internal python code and adopting a flat directory structure so import statements become more straightforward.

srkukarni:
One issue with this pr is that it is making system components(like executor) depend on heronpy which is not what we want to do. I don't think thats a good idea. One issue that we need to address in this pr is to make sure that heron/common does not depend on heronpy.

PR #2232, opened 2017-08-23
Make metrics cache client in healthMgr fetch location from statemgr

ashvina:
The metrics cache port is selected at runtime. So a user cannot provide
it as a configuration. The cache client needs to fetch it.

### DIFF
--- a/heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/MetricsCacheMetricsProvider.java
+++ b/heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/MetricsCacheMetricsProvider.java
@@ -27,6 +27,7 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.InvalidProtocolBufferException;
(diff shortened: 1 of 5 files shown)


heron/healthmgr/src/java/com/twitter/heron/healthmgr/sensors/MetricsCacheMetricsProvider.java
```diff
@@ -37,23 +38,28 @@
...
+  private SchedulerStateManagerAdaptor stateManagerAdaptor;
+  private String topologyName;
 
   private Clock clock = new Clock();
+  private String metricsCacheLocation;
 
   @Inject
-  public MetricsCacheMetricsProvider(@Named(CONF_METRICS_SOURCE_URL) String metricsCacheURL) {
-    LOG.info("Metrics will be provided by MetricsCache at :" + metricsCacheURL);
+  public MetricsCacheMetricsProvider(SchedulerStateManagerAdaptor stateManagerAdaptor,
```
huijunw: The buildCommonConfigModule() is associated with the injector inside HealthMgr. How does Guice link that inject to here.

PR #2243, opened 2017-08-25
Add prometheus metrics web sink.

cckellogg:
Provides a simple web server that exposes an endpoint for prometheus to scrape metrics. Prometheus metrics need to be exported in a certain format.

https://prometheus.io/docs/instrumenting/exposition_formats/


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/Metrics.java
```diff
@@ -0,0 +1,41 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.metricsmgr;
+
+import com.twitter.heron.spi.metricsmgr.metrics.MetricsRecord;
+
+public final class Metrics {
```
huijunw: This class seems like a Util class. The class name may be Util rather than Metrics

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/AbstractWebSink.java
```diff
@@ -0,0 +1,165 @@
...
+              + " Make sure the file is readable,"
+              + " only contains the port on which the service should run"
+              + " and is UTF8 encoded", e);
+        }
+      } else {
+        throw new IllegalArgumentException("Neither 'port' nor 'port_file' "
+            + "were specified in config for metrics sink " + context.getSinkId());
+      }
+    }
+    startHttpServer(path, port);
```
huijunw: what is this httpserver for ? who will send request to this server?

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/Metrics.java
cckellogg: I will update the name.

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/AbstractWebSink.java
cckellogg: It is there from the old WebSink sink class. This provides an endpoint for the prometheus service to scrape metrics from.  I added this class so both the WebSink and PrometheusSink could share the http server code. Thoughts?

huijunw: I though MetricsSink is a client which send metrics to Prometheus server.
If we need to make MetricsSink as server, why not use tmaster/metricscache httpserver https://github.com/twitter/heron/blob/master/heron/tmaster/src/cpp/manager/stats-interface.h#L43 instead?

cckellogg: The metrics for prometheus need to be in a certain format for exporting (https://prometheus.io/docs/instrumenting/exposition_formats/). This why the http server is used. I think keeping it here makes it less invasive with the core code and less stress on the tmaster service. Thoughts?

srkukarni: It might also be a more scalable way of doing things because now you have prometheus server connecting to each metrics manager to get the stats so the load is more evenly distributed.

cckellogg: @huijunw do you have any more concerns? we would like to get this in. thanks.

huijunwu: @cckellogg no more concerns, LGTM

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/tmaster/TMasterSink.java
```diff
@@ -191,7 +192,7 @@ public void processRecord(MetricsRecord record) {
     // The format of source is "host:port/componentName/instanceId"
     // So source.split("/") would be an array with 3 elements:
```
nlu90: this line of comment is no longer valid. Please remove it or update it

heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/metricscache/MetricsCacheSink.java
```diff
@@ -189,7 +190,7 @@ public void processRecord(MetricsRecord record) {
     // The format of source is "host:port/componentName/instanceId"
     // So source.split("/") would be an array with 3 elements:
```
nlu90: this line of comment is no longer valid. Please remove it or update it

PR #2267, opened 2017-08-31
add error log for physical plan check in tmaster

nlu90:
Encountered this `Check failed: ValidateStMgrsWithPhysicalPlan(*current_pplan_) ` problem, add more log for better tracking the root cause.

### DIFF
--- a/heron/tmaster/src/cpp/manager/tmaster.cpp
+++ b/heron/tmaster/src/cpp/manager/tmaster.cpp
@@ -922,7 +922,8 @@ bool TMaster::ValidateStMgrsWithPhysicalPlan(proto::system::PhysicalPlan _pplan)
       return false;
     }
     if (!iter->second->VerifyInstances(stmgr_to_instance_map[iter->first])) {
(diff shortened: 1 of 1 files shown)


heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -917,9 +917,12 @@ bool TMaster::ValidateStMgrsWithPhysicalPlan(proto::system::PhysicalPlan _pplan)
   }
   for (StMgrMapIter iter = stmgrs_.begin(); iter != stmgrs_.end(); ++iter) {
     if (stmgr_to_instance_map.find(iter->first) == stmgr_to_instance_map.end()) {
+      LOG(ERROR) << "Instances info from " << iter->first
+                 << " is missing. Bailing out..." << std::endl;
       return false;
     }
     if (!iter->second->VerifyInstances(stmgr_to_instance_map[iter->first])) {
+      LOG(ERROR) << "Instances verification failed. Bailing out..." << std::endl;
```
srkukarni: can we print iter->first here as well

PR #2269, opened 2017-09-01
log large packet without interrupting the heron-instance exectuion

huijunwu:
Sometimes the large packet was received. The heron-instance closes the connection to stmgr which disconnects the stmgr and heron-instance and the whole topology is down.

This PR lets the heron-instance continue execution. The unexpected large packet is written in log but the execution is not interrupted.


srkukarni:
This is a really bad change. Essentially the system is throwing things out without informing the upper layers. Imagine effectively once topologies encountering this.

huijunw:
this temp pr is deprecated. close it.

objmagic:
@srkukarni @huijunw wanted to add this because when a topology just started, it seems physical plan message was so large that instance rejected it because of this limit.

This limit was defined here: https://github.com/twitter/heron/pull/2118/files

A solution we made internally at Twitter is to increase this value to 64Mbytes

PR #2274, opened 2017-09-01
Topology package fetcher.

objmagic:
Add a topology fetcher in heron-client so that we can support fetching package from a remote location (for twitter internal package manager). This allows easier rollback, etc.

### DIFF
--- a/heron/tools/cli/src/python/fetcher/__init__.py
+++ b/heron/tools/cli/src/python/fetcher/__init__.py
@@ -1,0 +1,3 @@
+''' __init__ '''
+from base_fetcher import BaseFetcher
+from local_fetcher import LocalFetcher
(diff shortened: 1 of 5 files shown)


nlu90:
@srkukarni The fetcher abstraction is to allow heron client to download topology files from remote storage like Packer within twitter and submit it. We need to update the heron client is because the topology's main method must be executed locally to generate information for topology submission.

srkukarni:
@objmagic I understand that. However why are we duplicating functionlity that already exists in Java. Shouldn't the pathway be
1. If jar is regular file, continue as before
2. If jar is uri of some sort, invoke a downloader(analogous to the uploader). I believe the downloader functionality already exists in the java spi side of things.

huijunw:
agree with @srkukarni 
re-use heron/heron/downloaders/ code

huijunwu:
replaced by #2886. close this pr

PR #2278, opened 2017-09-02
Fixing bug involving race condition with initializing in and out queues

jerrypeng:
Fixing bug involving race condition with initializing in and out queues and downloading the physical plan.  In and out queues are initialized after the physical plan was download by the instance.  However, physical plans are not going to be download all at the same time causing some components to start sending tuples earlier than others. By initializing the queues earlier allows tuples to be start processing earlier

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/Slave.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/Slave.java
@@ -93,9 +93,9 @@ public class Slave implements Runnable, AutoCloseable {
     this.metricsCollector = new MetricsCollector(slaveLooper, metricsOutCommunicator);
 
     /**
-     * initializing queues with default values
-     * this is done after the physical plan is download and here since
-     * components might download physical plans at different rates causing
+     * Initializing queues with default values.
+     * This is done after the physical plan is downloaded and here since
+     * components might download physical plans at different rates
      * and components upstream might already start sending tuples downstream
      */
     this.streamInCommunicator.init(INSTANCE_INTERNAL_BOLT_READ_QUEUE_CAPACITY,
--- a/heron/instance/src/java/com/twitter/heron/instance/Slave.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/Slave.java
@@ -66,6 +66,11 @@ public class Slave implements Runnable, AutoCloseable {
   private State<Serializable, Serializable> instanceState;
   private boolean isStatefulProcessingStarted;
 
+  /**
+   * Default values used to initialize in and out queues prior to downloading the physical plan
+   * These values are transient and will be overwritten when the queue gets reinitialized after
+   * the physical plan is downloaded to values set in heron_interals.yaml
+   */
   private static final int INSTANCE_INTERNAL_BOLT_READ_QUEUE_CAPACITY = 128;
   private static final int INSTANCE_TUNING_EXPECTED_BOLT_READ_QUEUE_SIZE = 8;
   private static final double INSTANCE_TUNING_CURRENT_SAMPLE_WEIGHT = 0.8;
--- a/heron/instance/src/java/com/twitter/heron/instance/Slave.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/Slave.java
@@ -66,6 +66,12 @@ public class Slave implements Runnable, AutoCloseable {
   private State<Serializable, Serializable> instanceState;
   private boolean isStatefulProcessingStarted;
 
+  private static final int INSTANCE_INTERNAL_BOLT_READ_QUEUE_CAPACITY = 128;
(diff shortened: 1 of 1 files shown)


jerrypeng:
@srkukarni and @kramasamy can you guys take a look?  Thanks!

heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -66,6 +66,12 @@
   private State<Serializable, Serializable> instanceState;
   private boolean isStatefulProcessingStarted;
 
+  private static final int INSTANCE_INTERNAL_BOLT_READ_QUEUE_CAPACITY = 128;
```
kramasamy: @jerrypeng - comment about how these values are being chosen will be helpful. Are these values configurable by the user?

heron/instance/src/java/com/twitter/heron/instance/Slave.java
```diff
@@ -86,6 +92,19 @@ public Slave(SlaveLooper slaveLooper,
 
     this.metricsCollector = new MetricsCollector(slaveLooper, metricsOutCommunicator);
 
+    /**
+     * initializing queues with default values
+     * this is done after the physical plan is download and here since
```
kramasamy: downloaded?

jerrypeng: will fix

heron/instance/src/java/com/twitter/heron/instance/Slave.java
jerrypeng: @kramasamy I discussed with @srkukarni and he recommended we just use some hard coded values since these values for the queue are going to be transient anyways

kramasamy: @jerrypeng - that is fine. please just add a comment saying that we are using arbitrary values since it is going to transient. Later when somebody reads the code, they might not be able to know the context.

jerrypeng: The values used are default values chosen from the heron_internals.yaml file

jerrypeng: @kramasamy sure I can add some comments

jerrypeng:
@kramasamy thanks for the review! I think I have addressed your comments

maosongfu:
Sorry but I don't understand the description: 

>  physical plans are not going to be download all at the same time causing some components to start sending tuples earlier than others. By initializing the queues earlier allows tuples to be start processing earlier

If one instance hasn't got the physical plan, it does not know who it is and let alone to process incoming tuples. I don't see the point for an instance to accept data tuples before getting the physical plan.

jerrypeng:
@maosongfu Sorry my initial explanation of this bug might not be correct.  Let me first talk about what I have observed.  I was testing the windowing function that was recently merged into master.  I have a topology that has one instance of a spout and one instance of a bolt.  The spout emits a tuple every second.  However, for the first several tuples the bolt would receive them all at once and not in the time interval that the spout has sent them.  And this only happens at the beginning of a topology. 

I think I have found the reason.  The doWait function for the NIOLooper:

https://github.com/twitter/heron/blob/master/heron/common/src/java/com/twitter/heron/common/basics/NIOLooper.java#L53

is blocking for around 4 seconds before it timeout causing readStreamMessageIfNeeded:

https://github.com/twitter/heron/blob/master/heron/instance/src/java/com/twitter/heron/network/StreamManagerClient.java#L104

not to be called for that duration thus tuples are not read during that time.  Not sure why the selector there is blocking until it times out.

maosongfu:
@jerrypeng 
Thanks for sharing the context. I think we need to investigate the root cause, rather than commit something tentatively, since we don't know whether it does fix the issues, or potentially it can introduce other bugs.

If you need to skip this for local development, you could have it on a seprate branch or patch.

srkukarni:
@maosongfu Can you tell me who wakes up the gateway thread once the slave processes the assignment message? Currently the logic is as follows
1. We start reading from stmgr if either we have capacity to absorb messages or if helper is null(that is assignment is not done). This ensures that we read the assignment message
2. Once we read the assignment message we promptly stopReading from stmgr(which is fine)
3. But after assignment message is processed by the slave, I would have expected us to start reading stmgr immediately. Rather what happens is that nobody informs the gateway thread about this. After around 5 seconds the slave thread wakes up(not sure who is doing that) and then it realizes that it now has capacity to absorb and thus starts reading from stmgr.
My feeling is that this change wont be necessary if we figure out why slave thread is not communicating back to the gateway thread after processing the assignment message.

maosongfu:
@jerrypeng @srkukarni 
I went through the code and I found the issue. I filed a pull request to fix it: #2279. With that fix, on the one hand, it makes more sense; on the other hand, we don't need to specify magic default values outside the Communicator.

PR #2279, opened 2017-09-02
Notify both producer and consumer when init the Communicator

maosongfu:
Currently, when we init the Communicator, the changes will be picked up only when the producer/consumer
are notified by other events. It can lead to a race condition.

This pull request fixes the issue by notifying both producer/consumer immediately to run with new configs/values.

It is fixing the same issue mentioned in #2278. Reproduced the issue and have tested the fixing with this pull request with Local Scheduler.


maosongfu:
I don't know why but I can not add @jerrypeng as a reviewer. So at you @jerrypeng here for review this pull request.

jerrypeng:
@maosongfu Thanks for looking into this.  Your patch does fix the issue I was seeing.  Your solution is definitely more elegant so I am +1 for this patch

jerrypeng:
I will also close the PR I submitted to solve this issue

PR #2230, opened 2017-08-22
add `restartcontainer` resolver to healthmgr

huijunwu:
This PR add a resolver on top of Dhalion.
The proposed resolver restarts a container, which is part of `auto restart backpressure container` feature.

Replace: https://github.com/twitter/heron/pull/1770

PR #2283, opened 2017-09-05
healthmgr: aurora heron shell controller

huijunwu:
follow #2230:
The healthMgr/Dhalion resolver (RestartContainerResolver) calls Aurora scheduler to restart container. This PR adds a aurora scheduler which does not depend on Aurora CLI and which can be called inside Aurora container.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
@@ -30,44 +30,29 @@ import com.twitter.heron.spi.utils.NetworkUtils;
 import com.twitter.heron.spi.utils.ReflectionUtils;
 
 /**
- * Implementation of AuroraController that is a wrapper of AuroraCLIController
- * The difference is: the `restart` method implementation is changed to heron-shell
+ * Implementation of AuroraController that is a wrapper of AuroraCLIController The difference is:
+ * the `restart` method implementation is changed to heron-shell
  */
 class AuroraHeronShellController implements AuroraController {
   private static final Logger LOG = Logger.getLogger(AuroraHeronShellController.class.getName());
 
   private final String topologyName;
   private final AuroraCLIController cliController;
-  private SchedulerStateManagerAdaptor stateMgrAdaptor;
+  private final SchedulerStateManagerAdaptor stateMgrAdaptor;
 
-  AuroraHeronShellController(
-      String jobName,
-      String cluster,
-      String role,
-      String env,
-      String auroraFilename,
-      boolean isVerbose) {
+  AuroraHeronShellController(String jobName, String cluster, String role, String env,
+      String auroraFilename, boolean isVerbose)
+      throws ClassNotFoundException, InstantiationException, IllegalAccessException {
     this.topologyName = jobName;
-    this.cliController = new AuroraCLIController(
-        jobName, cluster, role, env, auroraFilename, isVerbose);
+    this.cliController =
+        new AuroraCLIController(jobName, cluster, role, env, auroraFilename, isVerbose);
 
-    stateMgrAdaptor = null;
     Config config =
         Config.toClusterMode(Config.newBuilder().putAll(ConfigLoader.loadClusterConfig()).build());
     String stateMgrClass = Context.stateManagerClass(config);
-    IStateManager stateMgr = null;
-    try {
-      stateMgr = ReflectionUtils.newInstance(stateMgrClass);
-
-      stateMgr.initialize(config);
-      stateMgrAdaptor = new SchedulerStateManagerAdaptor(stateMgr, 5000);
-    } catch (ClassNotFoundException e) {
-      LOG.severe(e.getMessage());
-    } catch (InstantiationException e) {
-      LOG.severe(e.getMessage());
-    } catch (IllegalAccessException e) {
-      LOG.severe(e.getMessage());
-    }
+    IStateManager stateMgr = ReflectionUtils.newInstance(stateMgrClass);
+    stateMgr.initialize(config);
+    stateMgrAdaptor = new SchedulerStateManagerAdaptor(stateMgr, 5000);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
```diff
@@ -0,0 +1,119 @@
...
+
+      stateMgr.initialize(config);
+      stateMgrAdaptor = new SchedulerStateManagerAdaptor(stateMgr, 5000);
+    } catch (ClassNotFoundException e) {
+      LOG.severe(e.getMessage());
+    } catch (InstantiationException e) {
+      LOG.severe(e.getMessage());
+    } catch (IllegalAccessException e) {
+      LOG.severe(e.getMessage());
+    }
```
maosongfu: Need we throw an exception if we could not initialize `stateMgr` and `stateMgrAdaptor`? Otherwise, the `AuroraHeronShellController` is in a wrong state.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
```diff
@@ -0,0 +1,119 @@
...
+      throw new UnsupportedOperationException("Not implemented");
+    }
+
+    if (stateMgrAdaptor == null) {
+      LOG.warning("SchedulerStateManagerAdaptor not initialized");
+      return false;
+    }
+
+    StMgr contaienrInfo = stateMgrAdaptor.getPhysicalPlan(topologyName).getStmgrs(containerId);
+    String ip = contaienrInfo.getHostName();
```
maosongfu: call the variable `host` or `hostname`. `ip` is confusing here.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
```diff
@@ -0,0 +1,119 @@
...
+    int port = contaienrInfo.getShellPort();
+    String url = "http://" + ip + ":" + port + "/killexecutor";
+
+    String payload = "secret=" + topologyName;
+    LOG.info("sending `kill container` to " + url + "; payload: " + payload);
+
+    HttpURLConnection con = NetworkUtils.getHttpConnection(url);
+    NetworkUtils.sendHttpPostRequest(con, "X", payload.getBytes());
+    boolean ret = NetworkUtils.checkHttpResponseCode(con, 200);
```
maosongfu: Have a try-catch-finally block and put `con.disconnect()` in finally

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -66,14 +67,28 @@ public void initialize(Config mConfig, Config mRuntime) {
...
+    String klass = config.getStringValue(Key.AURORA_CONTROLLER_CLASS);
     Config localConfig = Config.toLocalMode(this.config);
-    return new AuroraCLIController(
-        Runtime.topologyName(runtime),
-        Context.cluster(localConfig),
-        Context.role(localConfig),
-        Context.environ(localConfig),
-        AuroraContext.getHeronAuroraPath(localConfig),
-        Context.verbose(localConfig));
+    switch (klass) {
```
maosongfu: This implementation here is really ugly and bad.
One alternative is to use constant enum;
another alternative is to use a boolean to decide which one to use since you have only two cases and it will fail otherwise.
Polymorphism can be another one.

heron/spi/src/java/com/twitter/heron/spi/common/Key.java
```diff
@@ -71,6 +71,8 @@
   PACKING_CLASS            ("heron.class.packing.algorithm",   Type.STRING),
   REPACKING_CLASS          ("heron.class.repacking.algorithm", Type.STRING),
   STATE_MANAGER_CLASS      ("heron.class.state.manager",       Type.STRING),
+  AURORA_CONTROLLER_CLASS  ("heron.class.scheduler.aurora.controller",
```
maosongfu: given you have only 2 options, you can consider this is a boolean. You could evolve from it in future if needed, but I seriously doubt that.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
huijunw: Eventually any exception has to be caught before AuroraScheduler.initialize(). If the exception is not caught here, where shall we catch them?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
huijunw: done

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: changed to boolean

heron/spi/src/java/com/twitter/heron/spi/common/Key.java
huijunw: changed to boolean

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
huijunw: done

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
maosongfu: Bubble up a RuntimeException to external logic. Otherwise, the program will execute with a bad state but it is no way to execute with a bad state.

huijunw: done: throw to external logic

maosongfu:
👍 LGTM

huijunw:
thanks @maosongfu

PR #2291, opened 2017-09-07
fix tmaster client sends incomplete register request to tmaster

nlu90:
Previously, both tmaster client and stmgr have pointers to instances' information. This is problematic that between the time tmaster client gets the instance pointer and sends the register request using the pointed instances' information, it's possible that stmgr changes some instance info. Then tmaster client sends the changed instance info to tmaster without any knowledge. This causes severe problem for tmaster to compose a complete physical plan.

In this PR, we fix this problem by composing the `StmgrRegisterRequest` instead of maintaining the pointers to instance in tmaster client when stmgr is ready to connect to tmaster. In this way, the request sent by tmaster client is always complete and won't be updated unexpectedly. If the instance is updated between the request is composed and sent, the old complete request is still sent to tmaster and then a new complete request is sent. This won't causing any incomplete physical plan problem.

### DIFF
--- a/heron/stmgr/src/cpp/manager/tmaster-client.cpp
+++ b/heron/stmgr/src/cpp/manager/tmaster-client.cpp
@@ -49,6 +49,7 @@ TMasterClient::TMasterClient(EventLoop* eventLoop, const NetworkOptions& _option
       stateful_checkpoint_watch_(std::move(_stateful_checkpoint_watch)),
       restore_topology_watch_(std::move(_restore_topology_watch)),
       start_stateful_watch_(std::move(_start_stateful_watch)),
+      register_request_(nullptr),
       reconnect_timer_id(0),
       heartbeat_timer_id(0) {
   reconnect_tmaster_interval_sec_ = config::HeronInternalsConfigReader::Instance()
@@ -69,7 +70,9 @@ TMasterClient::TMasterClient(EventLoop* eventLoop, const NetworkOptions& _option
   InstallMessageHandler(&TMasterClient::HandleStartStmgrStatefulProcessing);
 }
 
-TMasterClient::~TMasterClient() {}
+TMasterClient::~TMasterClient() {
+  delete register_request_;
+}
 
 void TMasterClient::Die() {
   LOG(INFO) << "Tmaster client is being destroyed " << std::endl;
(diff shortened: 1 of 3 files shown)


heron/stmgr/src/cpp/manager/tmaster-client.h
```diff
@@ -86,7 +84,8 @@ class TMasterClient : public Client {
   sp_string stmgr_host_;
   sp_int32 stmgr_port_;
   sp_int32 shell_port_;
-  std::vector<proto::system::Instance*> instances_;
+  proto::tmaster::StMgrRegisterRequest* register_request_;
```
srkukarni: why not make it a an actual instance of register request rather than a pointer? That way the cleaning logic is greatly simplified

nlu90: Good suggestion. The original reason I want to use pointer here is I can do safety check before sending out the request. Now I updated the code: 1. use an instance instead of pointer; 2. introduce a boolean value indicating if the request is initialized properly  for safety check

heron/stmgr/src/cpp/manager/tmaster-client.cpp
```diff
@@ -207,31 +208,38 @@ void TMasterClient::OnHeartbeatTimer() {
...
+    stmgr->set_host_name(stmgr_host_);
+    stmgr->set_data_port(stmgr_port_);
+    stmgr->set_local_endpoint("/unused");
+    stmgr->set_cwd(cwd);
+    stmgr->set_pid((sp_int32)ProcessUtils::getPid());
+    stmgr->set_shell_port(shell_port_);
+    for (auto iter = _instances.begin(); iter != _instances.end(); ++iter) {
+      register_request_.add_instances()->CopyFrom(*(*iter));
+    }
+}
```
srkukarni: I belive we could do a simplification here.
Can we rename SetStmgrRegisterRequest to SetInstanceInfo, where the tmaster client copies the instances into local variable. And then during send register, just copies them back into the register request and sends it.

nlu90: The proposed solution is a cleaner way to do the thing. But why is it a simplification? There're two sets of copy operations instead of one. Plus, it requires the management of some objects in the memory.

srkukarni:
We still need to do memory management of the register request anyways. That way the memory overhead is the same.
Maybe I worded it incorrectly when I said simplification. Its probably clearer, that when you say getInstanceInfo, you are just collecting instance info and storing them over at the tmaster client to be used later.

nlu90:
Sure, I agree we should move to the proposed way to manage them. Let's make the change.

PR #2295, opened 2017-09-08
Make submitting examples easier from the command line.

cckellogg:
New commands:
  heron examples list
  heron examples run <cluster> <example-id>

This patch is to provide easier commands to get the heron examples up and running.

### DIFF
--- a/heron/tools/common/src/python/utils/config.py
+++ b/heron/tools/common/src/python/utils/config.py
@@ -34,6 +34,8 @@ ENVIRON = "default"
 BIN_DIR = "bin"
 CONF_DIR = "conf"
 ETC_DIR = "etc"
+EXAMPLES_DIR = "examples"
+EXAMPLES_JAR = "heron-examples.jar"
 LIB_DIR = "lib"
 CLI_DIR = ".heron"
 RELEASE_YAML = "release.yaml"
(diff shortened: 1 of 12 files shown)


heron/tools/common/src/python/utils/config.py
```diff
@@ -95,6 +97,7 @@ def get_subparser(parser, command):
   for subparsers_action in subparsers_actions:
     # get all subparsers
     for choice, subparser in subparsers_action.choices.items():
+      print choice, subparser
```
nlu90: is this `print` expected behavior?

srkukarni:
This is a very interesting link between codes(java examples to python cli). Are we sure about this dep?

heron/tools/common/src/python/utils/config.py
cckellogg: no left over from debugging. will remove

cckellogg:
It is a little different and strange but I think we need an easier way to get started with examples. Currently, the user needs to know where the examples are installed and the fully qualified class name. This is a lot to ask to get started. I'm open to suggestions on how make this better.

huijunw:
With #2300, we have two sets of examples (heron-examples.jar and storm-compatibility-examples). Does this PR run the examples for both heron and storm-compatibility?

srkukarni:
Closing this one

PR #2297, opened 2017-09-08
Use bufferevent to do Network I/O

srkukarni:
libevent has buffer event library that takes care of actually reading/writing from socket descriptors. Managing write readiness, reading/writing using vectors efficiently, zero copy buffer management are all done internally by bufferevent.
In this pr, we start using bufferevent to do the io rather than doing native i/o. This change greatly simplifies the networking code. There should be no performance penalty.
In addition to simplification of the code, this change has several benefits
1. bufferevents allow sophisticated rate limiting techniques(like leaky buckets) to control the rate of io.
2. bufferevents have priority groups which would in the future allow us to do differentiated io for spouts/bolts
3. adding ssl support for end to end encryption becomes trivial since bufferevents supports it natively.
4. Makes running stmgr on non posix api machines straightforward(windows)

PR #2300, opened 2017-09-11
Refactoring storm api

jerrypeng:
The following was done:

- Moved all the Storm API to storm-compatibility/
- Modified heron/examples to use storm api only and move to storm-compatibility-examples/. Also removed all dependencies of the twitter API from these examples
- Modified heron/examples to use the heron api only and moved to examples/
- Made sure the correct headers are on the storm api related files


srkukarni:
This mostly looks good to me. Adding @maosongfu and @nlu90 for their thoughts

jerrypeng:
some unrelated tests failed. Can someone kick travis again?

srkukarni:
@jerrypeng It did it two times. I don't see the same failures in other prs. Are all the unittests working in your working env?

srkukarni:
@jerrypeng I figured out the issue wrt failures. This has been fixed in the move to the buffer events pr. https://github.com/twitter/heron/pull/2297

examples/src/java/com/twitter/heron/examples/ExclamationTopology.java
```diff
@@ -68,17 +65,9 @@ public static void main(String[] args) throws Exception {
         ExampleResources.getContainerRam(spouts + bolts, parallelism));
     com.twitter.heron.api.Config.setContainerCpuRequested(conf, 1);
 
-    if (args != null && args.length > 0) {
-      conf.setNumWorkers(parallelism);
-      StormSubmitter.submitTopology(args[0], conf, builder.createTopology());
-    } else {
-      System.out.println("Topology name not provided as an argument, running in simulator mode.");
```
huijunw: do we keep the simulator in the heron example?
same comment to the other LocalCluster example

jerrypeng: Do we really need a simulator in Heron?  A LocalCluster or simulator is useful in Storm because running a single topology requires you start up a cluster on a single node (Nimbus, zookeeper, Supervisor, etc.) which can be a lot of work.  This is not the case in Heron.  It is super easy to run a topology in local mode.  Thus the usefulness for a simulator is diminished.

huijunw: as far as i know, the heron-simulator runs all heron-components in single process for topology writers to easily debug, which is different from multi-process local mode.

jerrypeng:
Weird travis error:

ERROR: (09-11 21:16:36.307) /home/travis/build/twitter/heron/third_party/java/BUILD:200:1: no such package '@google_http_client//jar': Failed to fetch Maven dependency: Could not transfer artifact com.google.http-client:google-http-client:jar:1.22.0 from/to google_http_client (http://central.maven.org/maven2/): Connection reset and referenced by '//third_party/java:google-api-services-storage'.

examples/src/java/com/twitter/heron/examples/ExclamationTopology.java
jerrypeng: gotcha, but is it easier to debug if everything just threads in a single process? A jstack might be confusing as there are many threads executing at the same time.  People often say it was hard to debug storm workers because there are many instances/threads of different components running in each

jerrypeng: but perhaps this discussion about the usefulness of the simulator is not in the scope of this pull request.  I can just add back the code for the simulator and we can discuss the usefulness of having a simulator in heron somewhere else

jerrypeng:
@huijunw thanks for the reviewing my PR!  I have added back to simulator code in the examples that originally contain it

nlu90:
This change in general is a really good one.  My only question is once this change merged, does the heron-client.tar.gz still has `heron-examples.jar` with it? The heron-examples.jar is really useful for verification and debugging purpose. Hope we can keep this unchanged.

jerrypeng:
@nlu90 yes that will remain the same.  In heron-client.tar.gz, heron examples will be at:
./examples/heron-examples.jar

PR #2303, opened 2017-09-13
Split executor args in aurora file so easy to read diff.

objmagic:


objmagic:
still needs some work... whitespace separator is gone.

objmagic:
tmaster unittest being flaky. Twitter internal test passed.

PR #2304, opened 2017-09-13
Bug fixes and integration test for Python DSL

jerrypeng:
- Added integration test for python dsl
- fixed bug in reducebykeyandwindow involving not correctly reducing values
- fixed bug in integration_test_bolt involving not invoking the underline process_tick tuple

edit

As @srkukarni described below this PR will also fix the python packaging issue with heron python instance


srkukarni:
After examining the fix to the build file of the heron instance, it seems to be a valid way of fixing the inclusion issue. I'll try to summarize below
HeronPython Instance depends on heronpy because it implements the api functionality. In the current version, we include heronpy.api inside the instance itself. This means that when the instance starts up, it can instantiate its heronpy references. But, all subsequent heronpy references will now be directed at this instantiation. Which means that if there is a dsl topology that refers heronpy.dsl or heronpy.connectors, they won't be found because the instantiated heronpy embedded inside the instance, only contains api.
This change includes the heronpy.api/dsl/connectors inside the instance thus fixing that issue.

heron/instance/src/python/BUILD
```diff
@@ -7,8 +7,10 @@ pex_library(
     srcs = glob(['**/*.py']),
     deps = [
         '//heronpy/proto:proto-py',
-        "//heronpy/api:heron-python-py",
-        '//heron/common/src/python:common-py'
+        '//heron/common/src/python:common-py',
```
srkukarni: can heronpy dep come first before the common?

integration_test/src/python/integration_test/core/integration_test_bolt.py
```diff
@@ -98,6 +98,7 @@ def process(self, tup):
 
   def emit(self, tup, stream=Stream.DEFAULT_STREAM_ID, anchors=None,
            direct_task=None, need_task_ids=False):
+    Log.info("emitting tuple: %s", tup)
```
srkukarni: pl remove

heron/instance/src/python/BUILD
jerrypeng: sure

integration_test/src/python/integration_test/core/integration_test_bolt.py
jerrypeng: I intentionally left this comment in here so it is easier to debug the integration tests.  Tuples received is already printed but tuples emitted was missing

PR #2305, opened 2017-09-13
Fixed the path of the python examples

srkukarni:


jerrypeng:
+1 @srkukarni thanks for the fix!

Issue #2272, opened 2017-09-01
--config-property failed to overwrite the maxpacketsize config for heron-instance


objmagic:
A walkaround for an issue internal at Twitter is to set the `heron.instance.network.options.maximum.packetsize.bytes` to larger value, to mitigate the heron-instance packet size checking, so that the heron-instance can run with unexpected large packet.

When user tried `heron submit --config-property heron.instance.network.options.maximum.packetsize.bytes=104857600 s/xxx/yyy heron-native.tar jobname`, the topology did not pick up the overridden value.

The heron-instance log showed the old value:
```
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.common.network.HeronClient: Connecting to endpoint: /127.0.0.1:31053  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.network.StreamManagerClient: Connected to Stream Manager. Ready to send register request  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.network.StreamManagerClient: Stop writing due to not yet connected to Stream Manager.  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.network.StreamManagerClient: Stop writing due to not yet connected to Stream Manager.  
[2017-08-31 23:22:36 +0000] [SEVERE] com.twitter.heron.common.network.IncomingPacket: packet size 15412455 exceeds limit 10485760  
[2017-08-31 23:22:36 +0000] [SEVERE] com.twitter.heron.common.network.SocketChannelHelper: Something bad happened while reading from channel: /127.0.0.1:31053  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.common.network.HeronClient: Handling Error. Cleaning states in HeronClient.  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.common.network.HeronClient: Successfully closed the channel: java.nio.channels.SocketChannel[closed]  
[2017-08-31 23:22:36 +0000] [SEVERE] com.twitter.heron.network.StreamManagerClient: Disconnected from Stream Manager.  
[2017-08-31 23:22:36 +0000] [INFO] com.twitter.heron.network.StreamManagerClient: Clean the old PhysicalPlanHelper in StreamManagerClient.  
[2017-08-31 23:22:36 +0000] [WARNING] com.twitter.heron.network.StreamManagerClient: Error connecting to Stream Manager with status: CONNECT_ERROR, Retrying...  
[...]

PR #2311, opened 2017-09-15
Take override.yaml in stmgr, metricsmgr, metricsmgr-cache, tmaster, instance

objmagic:
resolves https://github.com/twitter/heron/issues/2272


objmagic:
@srkukarni others directly or indirectly use [`SchedulerUtils.executorCommandArgs`](https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java#L189), which I have modified.

huijunw:
looks good to me. fix ci pls

PR #2313, opened 2017-09-16
Seperate Stmgr and Instance Serving in Stmgr

srkukarni:
In the current stmgr, all incoming traffic(from peer stmgr and local instances) are multiplexed in the same server. This makes the class complex. Secondly when connection setup/disconnect happen it is not completely clear whether it was stmgr or an instance related. Plus it doesn't allow us to do differentiated handling between the two traffic. One is a inter-container traffic and the other is a intra-container traffic. We could perhaps encrypt the inter-container traffic, while intra-container can be sent via shared memory.
This pr seperates out this functionality. StMgrServer handles traffic from other stmgrs while InstanceServer handles traffic from local containers.


heron/executor/src/python/heron_executor.py
```diff
@@ -621,6 +621,7 @@ def _get_streaming_processes(self):
         ','.join(map(lambda x: x[0], instance_info)),
         self.master_host,
         self.master_port,
+        self.tmaster_controller_port,
```
huijunwu: is it `local_data_port`?

srkukarni: in the executor, the ports are called master_port and tmaster_controller_port. They are labelled that way for both the tmaster container, as well as the regular container.
What we are doing is that using the tmaster_controller_port to be the local data port.

objmagic:
LGTM. Also looking forward to sending intra message via shared memory! Always want to see its implementation in real world code.

heron/proto/physical_plan.proto
```diff
@@ -11,7 +11,12 @@ import "topology.proto";
 message StMgr {
   required string id = 1;
   required string host_name = 2;
+  // This is the data port for communication
+  // accross the containers(i.e. peer stmgrs)
   required int32  data_port = 3;
```
nlu90: stmgr_port and instance_port should be better. There are both data and control infos passed to and from these two ports.

srkukarni: I prefer not to change any 'names' in this pr because that will pollute this pr. I will send out a follow up pr to fix the names

nlu90: 👍

PR #2314, opened 2017-09-16
Upgrade pex

kramasamy:
- Upgrade PEX and its dependencies to 1.2.11
- It uses bazel http file to download the repo and compile
- It loads the pex rules and java_tests apriori using tools/build_rules/prelude_bazel (this is convenient so that you don't need to include them in every files)


nlu90:
This requires some changes inside Twitter's related repo correspondingly to avoid issues. Please don't merge it until we verifies it works fine with twitter environment.

nlu90:
Verified. The pex change doesn't cause any building problems.

PR #2316, opened 2017-09-18
upgrade to protobuf3.4

kramasamy:
With proto2 syntax. This is completely backward compatible with current proto 2.5


nlu90:
Please don't merge it until we verify it doesn't cause any problem.

third_party/protobuf/protobuf-3.4.0.patch
```diff
@@ -0,0 +1,295 @@
+diff -rup protobuf-3.4.0.orig/configure.ac protobuf-3.4.0/configure.ac
```
huijunwu: what is this patch file for?

huijunwu:
does this pr replace #2301?

kramasamy:
Yes. I have closed #2301

third_party/protobuf/protobuf-3.4.0.patch
kramasamy: this patch is to ensure we use a fast path for C++11x so that performance is not affected. The patch was originally done by @congwang - just redid it for protobuf 3.4.0

PR #2319, opened 2017-09-19
fix python state manager

nlu90:
Recent internal zookeeper server unstable issue caused our topologies fail to be started. The log shows that heron executor  continues execution even after an exception is logged.

```
[2017-09-16 00:21:40 +0000] [ERROR]: Exception while connecting to state_manager.
[2017-09-16 00:21:40 +0000] [INFO]: Adding data watch for path: ${SOME_PATH}
[2017-09-16 00:21:40 +0000] [INFO]: Registered state watch for packing plan changes with state manager <heron.statemgrs.src.python.zkstatemanager.ZkStateManager object at ${SOME_ADDR}>.
[2017-09-16 00:43:27 +0000] [INFO]: signal_handler invoked with signal 15
```

With further investigation, we found that methods in `statemanagerfactory` only logs an exception and then swallows it. Ideally, the state manager should either fail fast or propagate exceptions to upper layer.

This PR is to fix the problem. `statemanagerfactory` propagate exception after logging it and `heron_executor` exit immediately if it catches any exception from state manager.

I tested the logic locally by throwing an exception on purpose. And the heron-executor exits once caught the exception

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -898,7 +898,11 @@ class HeronExecutor(object):
     """
     statemgr_config = StateMgrConfig()
     statemgr_config.set_state_locations(configloader.load_state_manager_locations(self.cluster))
-    self.state_managers = statemanagerfactory.get_all_state_managers(statemgr_config)
+    try:
+      self.state_managers = statemanagerfactory.get_all_state_managers(statemgr_config)
+    except Exception as ex:
+      Log.error("Found exception: %s. Bailing out..." % ex)
+      sys.exit(1)
 
     # pylint: disable=unused-argument
     def on_packing_plan_watch(state_manager, new_packing_plan):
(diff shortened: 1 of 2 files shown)


heron/executor/src/python/heron_executor.py
```diff
@@ -898,7 +898,11 @@ def start_state_manager_watches(self):
     """
     statemgr_config = StateMgrConfig()
     statemgr_config.set_state_locations(configloader.load_state_manager_locations(self.cluster))
-    self.state_managers = statemanagerfactory.get_all_state_managers(statemgr_config)
+    try:
+      self.state_managers = statemanagerfactory.get_all_state_managers(statemgr_config)
+    except Exception as ex:
+      Log.error("Found exception: %s. Bailing out..." % ex)
```
objmagic: "Found exception when initializing state managers"

PR #2324, opened 2017-09-19
Delivery semantics documentation (WIP)

lucperkins:
This is a document that we'll need to improve over time, but for now I think it's important to include it here and iterate. I also included some housekeeping items related to the site generation process in this PR.


website/content/docs/concepts/delivery-semantics.md
```diff
@@ -0,0 +1,62 @@
...
+At least once | Tuples injected into the Heron topology are guaranteed to be processed at least once; no tuple will fail to be processed. It's possible, however, that any given tuple is processed more than once in the presence of various failures, retries, or other contingencies. | When you need to guarantee no data loss
+Effectively once | Heron ensures that the data it receives is processed effectively once---even in the presence of various failures---leading to accurate results. **This applies only to [stateful topologies](#stateful-topologies)**. "Effectively" in this case means that there's a guarantee that tuples that cause [state changes](#stateful-topologies) will be processed once (that is, they will have *an effect* on state once). | When you're using [stateful topologies](#stateful-processing) and need strong
```
huijunw: `and need strong ` .. 
this sentence seems not complete.

website/content/docs/concepts/delivery-semantics.md
```diff
@@ -0,0 +1,62 @@
...
+* In **non-stateful topologies**, there is no requirement that any processing components store a state snapshot. Non-stateful topologies can provide at-most-once or at-least-once semantics, but never effectively-once semantics.
+
+> Heron currently supports two state managers: [ZooKeeper](../../operators/deployment/statemanagers/zookeeper) and the [local filesystem](../../operators/deployment/statemanagers/localfs), although others are currently under development.
+
+Stateful topologies, in turn, are of two types:
+
+* **Idempotent** stateful topologies are stateful topologies in which applying the processing graph to an input more than once, it will continue to return the same result. A basic example is multiplying a number by 0. The first time you do so, the number will change (always to 0), but if you apply that transformation again and again, it will not change.
+
+    For topologies to provide effectively-once semantics, they need to transform tuple inputs idempotently as well. If they don't, and applying the topology's processing graph multiple times yields different results, then effectively-once semantics *cannot* be achieved.
```
huijunw: if topology is idempotent, at-least once can achieve the same result ? no need to enable `effectively once`?

website/content/docs/concepts/delivery-semantics.md
```diff
@@ -0,0 +1,62 @@
...
+
+    If you'd like to create idempotent stateful topologies, make sure to write tests to ensure that idempotency requirements are being met.
+* **Non-idempotent** stateful topologies are stateful topologies that do not apply processing logic along the model of "multiply by zero" and thus cannot provide effectively-once semantics. An example of a non-idempotent
```
huijunw: `An example of a non-idempotent` ..
this sentence is not complete

website/content/docs/concepts/delivery-semantics.md
lucperkins: Sorry, forgot to add that the PR is a work in progress

website/content/docs/concepts/delivery-semantics.md
srkukarni: I would reword this as
they need to ensure that any state changes caused by execution of the tuple is captured in the heron state variable thats passed at startup time. This way Heron can keep track of state of the processing logic and rewind state when failures occur.

lucperkins: @huijunw As far as I know, idempotency and effectively once correspond to different concerns. A topology can be idempotent and process tuples more than once, which is a problem if you have effectively once requirements.

lucperkins:
I'm going to merge this for now and iterate this upcoming week, because there continue to be gaps.

PR #2335, opened 2017-09-21
Python3 compatibility

srkukarni:
A set of changes to make heron compatible with both python2.7 and python3


huijunwu:
the motivation looks good to me.
another curious question, what version does heronpy use?

srkukarni:
We might need some changes in heronpy as well to make it python3 compatible. Currently it works in python2.7.
But want to do this in smaller prs.

nlu90:
👍

PR #2353, opened 2017-09-26
Move java dsl to be part of heron api

aahmed-se:


huijunw:
This PR reminds me of heronpy. What is the relation between heronpy and heron-api/python?

aahmed-se:
there is no real relation, heronpy was moved out to separate client api from internal code, we are doing somewhat the same here, dsl is now ready to be exposed as part of the  api jar that is published on maven.

srkukarni:
@huijunw The aim is to provide the dsl func in the same jar as regular api.

PR #2357, opened 2017-09-28
Changing TimeTriggerPolicy to use tick tuples as a timer

jerrypeng:
TimeTriggerPolicy used an additional thread to run periodic tasks.  Tuple acking was conducted as part of these periodic tasks.  However, OutgoingTupleCollection isn't thread safe thus causing NPEs.  Thus I removed the separate thread to run periodic tasks and instead used tick tuples to trigger the tasks. 

I also changed Tick Tuple to frequency to milliseconds since windowing needs to support windows in milliseconds.  I left a backwards compatible api for seconds:

public static void setTickTupleFrequency(Map<String, Object> conf, int seconds) 

Which will translate to milliseconds

I also add logic in bolt instance to persist config changes done in topology logic when updates to the physical plan occurs

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -50,10 +50,10 @@ public class Config extends HashMap<String, Object> {
    */
   public static final String TOPOLOGY_COMPONENT_JVMOPTS = "topology.component.jvmopts";
   /**
-   * How often a tick tuple from the "__system" component and "__tick" stream should be sent
+   * How often (in milliseconds) a tick tuple from the "__system" component and "__tick" stream should be sent
    * to tasks. Meant to be used as a component-specific configuration.
    */
-  public static final String TOPOLOGY_TICK_TUPLE_FREQ_SECS = "topology.tick.tuple.freq.secs";
+  public static final String TOPOLOGY_TICK_TUPLE_FREQ_MS = "topology.tick.tuple.freq.ms";
   /**
    * True if Heron should timeout messages or not. Defaults to true. This is meant to be used
    * in unit tests to prevent tuples from being accidentally timed out during the test.
@@ -248,7 +248,7 @@ public class Config extends HashMap<String, Object> {
     apiVars.add(TOPOLOGY_WORKER_CHILDOPTS);
     apiVars.add(TOPOLOGY_COMPONENT_JVMOPTS);
     apiVars.add(TOPOLOGY_SERIALIZER_CLASSNAME);
-    apiVars.add(TOPOLOGY_TICK_TUPLE_FREQ_SECS);
+    apiVars.add(TOPOLOGY_TICK_TUPLE_FREQ_MS);
     apiVars.add(TOPOLOGY_ENABLE_MESSAGE_TIMEOUTS);
     apiVars.add(TOPOLOGY_CONTAINER_CPU_REQUESTED);
     apiVars.add(TOPOLOGY_CONTAINER_DISK_REQUESTED);
@@ -333,7 +333,11 @@ public class Config extends HashMap<String, Object> {
   }
 
   public static void setTickTupleFrequency(Map<String, Object> conf, int seconds) {
-    conf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_SECS, Integer.toString(seconds));
+    setTickTupleFrequencyMs(conf, (long) (seconds * 1000));
+  }
+
+  public static void setTickTupleFrequencyMs(Map<String, Object> conf, long millis) {
+    conf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_MS, millis);
   }
 
   public static void setTopologyReliabilityMode(Map<String, Object> conf,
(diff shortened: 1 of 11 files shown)


jerrypeng:
@srkukarni @maosongfu @nlu90 Can you review this?  Thanks!

heron/api/src/java/com/twitter/heron/api/windowing/WindowManager.java
```diff
@@ -121,7 +121,7 @@ public void add(T event, long ts) {
    */
   public void add(Event<T> windowEvent) {
     // watermark events are not added to the queue.
-    if (!windowEvent.isWatermark()) {
+    if (!windowEvent.isWatermark() && !windowEvent.isTimer()) {
       queue.add(windowEvent);
     } else {
       LOG.fine(String.format("Got watermark event with ts %d", windowEvent.getTimestamp()));
```
nlu90: log also needs to handle timer event

heron/api/src/java/com/twitter/heron/api/windowing/triggers/TimeTriggerPolicy.java
```diff
@@ -47,88 +37,45 @@ public TimeTriggerPolicy(long millis, TriggerHandler handler, EvictionPolicy<T>
...
-    checkFailures();
+    if (started && event.isTimer()) {
+      triggerTask();
+    }
   }
 
   @Override
   public void reset() {
-    checkFailures();
+
```
nlu90: why the `reset()` doesn't triggerTask here?

heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
```diff
@@ -258,6 +260,8 @@ private boolean isTupleTs() {
         return new WatermarkTimeTriggerPolicy<>(slidingIntervalDurationMs, manager,
             evictionPolicy, manager);
       } else {
+        // set tick tuple frequency for timer in TimeTriggerPolicy
+        Config.setTickTupleFrequency(topoConf, slidingIntervalDurationMs);
```
nlu90: Instead of updating the config here within a bolt, can we push this task to topology creation time and provide some default value if user doesn't set it? In this way, you don't need to change the `BoltInstance` code and pass around the conf object

jerrypeng:
@nlu90 we could definitely discuss about making acking thread safe, but i think that might be part of a larger discussion and research as it may be performance impacting.  That discussion may involve architectural changes to heron and could be time consuming.  While I am definitely to take part in that discussion, this PR addresses the bug immediately.

heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
jerrypeng: @nlu90 not all types of windows need the timer.  Only processing time based windows that use the TimerTriggerPolicy need a timer thus tick tuples other window types do not need it.   
 Its not eloquent to determine what kind of window it is on the client side. The existing python windowing implementation also sets the tick tuple frequency in the initialize/prepare method.

heron/api/src/java/com/twitter/heron/api/windowing/triggers/TimeTriggerPolicy.java
jerrypeng: The point of reset is to clear some internal state of the trigger.  Please look at the CountTriggerPolicy.  The TimeTriggerPolicy does not have any internally held state to clear since its just a timer

heron/api/src/java/com/twitter/heron/api/windowing/WindowManager.java
jerrypeng: +1 thanks for pointing that out

jerrypeng:
@nlu90 thanks for taking the time to review my PR!  I have addressed your comments.

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -333,7 +333,11 @@ public static void setMaxSpoutPending(Map<String, Object> conf, int max) {
   }
 
   public static void setTickTupleFrequency(Map<String, Object> conf, int seconds) {
-    conf.put(Config.TOPOLOGY_TICK_TUPLE_FREQ_SECS, Integer.toString(seconds));
+    setTickTupleFrequency(conf, (long) (seconds * 1000));
```
srkukarni: can we name them explicitly. like setTickTupleFreqencySec and Msec

jerrypeng: but that might break backwards compatibility right?

srkukarni: Ok. How about the new api then. Make it explicitly say Ms?

jerrypeng:
@srkukarni thanks for your review.  I have addressed your comments

srkukarni:
looks good

jerrypeng:
just squashed commits into one.  Are we ready to merge?

PR #2364, opened 2017-09-29
Add component type label for prometheus metrics.

cckellogg:


heron/metricsmgr/src/java/com/twitter/heron/metricsmgr/sink/PrometheusSink.java
```diff
@@ -43,6 +43,12 @@
 
   private static final String DELIMITER = "\n";
 
+  // bolt metric
+  private static final String METRIC_EXECUTE_TIME_NS = "__execute-time-ns/default";
```
srkukarni: Just __execute-time-ns should do the trick. Since the stream name could be different

cckellogg: I tried without the default on some of the examples and seems to not work. It seems like the default is needed.

msa/exclaim1/container_1_exclaim1_3:
{__execute-time-ns/word/default=4.36168882E8, __fail-count/word/default=0.0, __execute-latency/default=10412.989280683745, __process-latency/default=9611.879413660563, __fail-count/default=0.0, __process-latency/word/default=9611.879413660563, __execute-count/word/default=41887.0, __execute-latency/word/default=10412.989280683745, __ack-count/word/default=41887.0, __tuple-serialization-time-ns/default=9.5455816E7, __ack-count/default=41887.0, __execute-count/default=41887.0, __auto__/selected_items=4.0, __out-queue-full-count=103.0, __emit-count/default=41887.0, __execute-time-ns/default=4.36168882E8}

In what cases would I not see default at the end?

srkukarni:
if users use collector.emit(new Values(..)), then you would see default.
In cases when they use collector.emit(streamName, new Values(...)) like the underlying spouts/bolts in dsl, you would see the actual stream name.

PR #2372, opened 2017-10-02
Make bazel updates for 6.0

aahmed-se:
This maintains backward compatibility with 0.54

### DIFF
--- a/tools/rules/genproto.bzl
+++ b/tools/rules/genproto.bzl
@@ -42,7 +42,7 @@ def genproto_java_impl(ctx):
       command = java_cmd,
       use_default_shell_env = True)
 
-  return struct(files = set([srcjar]))
+  return struct(files = depset([srcjar]))
 
 genproto_java = rule(
     genproto_java_impl,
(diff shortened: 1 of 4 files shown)


huijunw:
I have a question, what is our policy to update bazel version? Do we follow bazel version closely for each version?

srkukarni:
@huijunw Its ok to introduce non breaking changes. i.e. there is a officially supported version(currently 0.5.4). As long as things work in that officially supported version, making any new changes to support newer version should be ok.
What do you think?

huijunw:
@srkukarni thank you for the clarification.
needs @nlu90 comment

nlu90:
@huijunw I suggest 1). we make sure things are not broken with any such dev tool upgrades; 2).  with the first point guaranteed, we move as fast as we can to help the oss community.

kramasamy:
@nlu90 @huijunw - these changes are backward compatible withh bazel 0.5.4 and also works with bazel 0.6.0 as well.

PR #2373, opened 2017-10-02
introduce sandbox config

kramasamy:
if we have sandbox config - we can further reduce docker image size.

### DIFF
--- a/heron/config/src/yaml/BUILD
+++ b/heron/config/src/yaml/BUILD
@@ -23,6 +23,11 @@ filegroup(
 )
 
 filegroup(
+    name = "conf-sandbox-yaml",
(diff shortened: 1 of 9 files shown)


heron/config/src/yaml/BUILD
```diff
@@ -23,6 +23,11 @@ filegroup(
 )
 
 filegroup(
+    name = "conf-sandbox-yaml",
```
huijunw: who will use these files `conf-sandbox-yaml`

kramasamy: @huijunw - this is for docker sandbox image.

huijunw: thanks @kramasamy

nlu90:
👍

PR #2375, opened 2017-10-02
Adding new metrics from Apache Storm

jerrypeng:
adding CountStatAndMetric and LatencyStatAndMetric metrics. To be more compatible with the Storm API

Most of the code is take from Apache Storm


jerrypeng:
@kramasamy @srkukarni please review

PR #2376, opened 2017-10-02
Modify local scheduler to work without a core uri.

cckellogg:
This adds a config option to use an installed version of
heron so we don't need to package the heron client within
a docker image to run the sandbox.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
+++ b/heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
@@ -121,6 +121,15 @@ public final class FileUtils {
   }
 
   public static boolean deleteDir(File dir, boolean deleteSelf) {
+    if (Files.isSymbolicLink(dir.toPath())) {
+      try {
+        Files.delete(dir.toPath());
+        return true;
+      } catch (IOException e) {
+        return false;
+      }
+    }
+
     if (dir.isDirectory()) {
       String[] children = dir.list();
 
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
@@ -369,26 +369,13 @@ public final class SchedulerUtils {
   }
 
   /**
-   * Setup the working directory:
-   * <br> 1. Download heron core and the topology packages into topology working directory,
-   * <br> 2. Extract heron core and the topology packages
-   * <br> 3. Remove the downloaded heron core and the topology packages
+   * Setup the working directory: Create the directory if it does not exist
+   * otherwise clean the directory.
    *
    * @param workingDirectory the working directory to setup
-   * @param coreReleasePackageURL the URL of core release package
-   * @param coreReleaseDestination the destination of the core release package fetched
-   * @param topologyPackageURL the URL of heron topology release package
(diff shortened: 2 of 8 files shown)


heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
```diff
@@ -121,6 +121,15 @@ public static boolean deleteDir(String dir) {
   }
 
   public static boolean deleteDir(File dir, boolean deleteSelf) {
+    if (Files.isSymbolicLink(dir.toPath())) {
```
nlu90: will this symbolic file deletion work with directories?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -369,26 +369,13 @@ public static String decodeJavaOpts(String encodedJavaOpts) {
...
    *
    * @param workingDirectory the working directory to setup
-   * @param coreReleasePackageURL the URL of core release package
-   * @param coreReleaseDestination the destination of the core release package fetched
-   * @param topologyPackageURL the URL of heron topology release package
-   * @param topologyPackageDestination the destination of heron topology release package fetched
-   * @param isVerbose display verbose output or not
    * @return true if successful
    */
-  public static boolean setupWorkingDirectory(
```
nlu90: this method no longer sets up directory, should rename it peoperly

heron/schedulers/src/java/com/twitter/heron/scheduler/local/LocalLauncher.java
```diff
@@ -114,13 +120,36 @@ protected boolean setupWorkingDirectory() {
...
         topologyWorkingDirectory, "topology.tar.gz").toString();
 
-    return SchedulerUtils.setupWorkingDirectory(
-        topologyWorkingDirectory,
-        coreReleasePackageURI,
-        coreReleaseFileDestination,
-        topologyPackageURI,
-        topologyPackageDestination,
-        Context.verbose(config));
+    if (!SchedulerUtils.setupWorkingDirectory(topologyWorkingDirectory)) {
```
nlu90: code on line 123 to line 148 is doing the actual setup work. Can we move them into the `SchedulerUtils. setupWorkingDirectory`? 

If there's a reason we couldn't put them there, we can create a local method here to make the code here clear.

heron/common/src/java/com/twitter/heron/common/basics/FileUtils.java
cckellogg: This will only delete the symlink and not any of the potential files/directories under it.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
cckellogg: I will rename it to something better.

heron/schedulers/src/java/com/twitter/heron/scheduler/local/LocalLauncher.java
cckellogg: I wanted to separate the create/clean directory and the tar.gz file extraction. This decouples things for when a scheduler would like to use an installed version of the code. The extraction of the core.tar.gz is not needed but you had to pass those values in previous. Thoughts?

PR #2378, opened 2017-10-03
Log exception when scheduler fails to deploy config.

cckellogg:


kramasamy:
@cckellogg - can we refine the error messages to 

* Unable to connect the kubernetes schheduler or api server
* Invalid config

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesController.java
```diff
@@ -142,11 +135,13 @@ protected boolean submitTopology(String[] appConfs) {
       try {
         deployContainer(appConf);
       } catch (IOException ioe) {
-        LOG.log(Level.SEVERE, "Problem deploying container with config: " + appConf);
-        return false;
+        String message = (ioe instanceof java.net.ConnectException)
```
nlu90: Instead of using `instanceof`, we should catch the `ConnectException` directly and handle it separately. Something like:
```
try {
...
} catch (ConnectException) {
...
} catch (IOException) {
...
}
```

nlu90:
Just a reminder, if we will do similar repeated exception handling in the future, we need to find out a cleaner way.

PR #2380, opened 2017-10-03
Be explicit about subprocess pipe's EOF.

objmagic:
We should not check if a subprocess has terminated to determine if we have read all bytes from subprocess's pipe. This PR makes EOF explicit when reading from pipe.

Also use list and join instead of concatenation to collect output faster.

### DIFF
--- a/heron/common/src/python/utils/proc.py
+++ b/heron/common/src/python/utils/proc.py
@@ -22,9 +22,9 @@ def _stream_process_fileno(fileno, handler):
   """
   while 1:
     line = fileno.readline()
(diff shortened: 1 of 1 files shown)


maosongfu:
@objmagic Did we verify the problematic topology mentioned by Pankaj?

objmagic:
@maosongfu no. that's user's topology and we cannot experiment this on their topology. If I have time I can probably ask someone to help to test.

maosongfu:
@objmagic Thanks! That would be awesome! Good job!

PR #2386, opened 2017-10-04
Documentation for self-regulating topologies (WIP)

lucperkins:


ashvina:
@lucperkins Thanks for this PR ! 
The doc looks generally looks good to me. We are still developing and improving the feature. I think it would be better to mark this feature as a "beta" feature till we have tested and hardened it enough. The document will also need a section on "How to create and manage Self-regulating topologies".

lucperkins:
@ashvina Yep, just a work in progress 😄 This PR will eventually include practical docs for both Python and Java as well as a more full-fledged theoretical discussion of Dhalion. I'll be in touch with Karthik and Sanjeev and we can push forward on this when the time comes 👍

PR #2388, opened 2017-10-04
Update Functional API docs (WIP)

lucperkins:
Previously, I created a preliminary version of the documentation for the Functional API that was missing some information and provided just an initial glimpse of the API. In this PR, I seek to create a more complete picture from both a theoretical perspective and for the Java API.


lucperkins:
I'm going to close this, as we're changing all of the terminology surrounding the new API from "DSL" to "Streamlets"

PR #2390, opened 2017-10-05
Add bookkeeper kubernetes yaml script.

cckellogg:


cckellogg:
Going to move this somewhere else.

PR #2396, opened 2017-10-06
Add bazel dev command to bypass checkstyle

aahmed-se:


aahmed-se:
Will cancel this after consultation with @kramasamy

PR #2398, opened 2017-10-07
Stateful windowing

jerrypeng:
This PR allows window bolts to have state.  This PR also implements internal state of windows to keep track of which tuples are queued but still haven't been triggered yet.  This allows effectively once guarantees to be maintained for windowing operations.


jerrypeng:
@srkukarni @nlu90 @kramasamy please review this PR.  Thanks!

examples/src/java/com/twitter/heron/examples/api/StatefulSlidingWindowTopology.java
```diff
@@ -0,0 +1,171 @@
...
+      this.state = state;
+      sum = state.getOrDefault("sum", 0L);
+    }
+
+    @Override
+    public void execute(TupleWindow inputWindow) {
+      for (Tuple tuple : inputWindow.get()) {
+        sum += tuple.getLongByField("value");
+      }
+      state.put("sum", sum);
```
srkukarni: You could move this line to preSave. That way the topoliogy can be used both in effectively once as well as atleast/atmost once mode

jerrypeng: +1

heron/api/src/java/com/twitter/heron/api/tuple/Tuple.java
```diff
@@ -29,7 +30,7 @@
  * use another type, you'll need to implement and register a serializer for that type.
  * @see <a href="https://storm.apache.org/documentation/Serialization.html">Storm serialization</a>
  */
-public interface Tuple {
+public interface Tuple extends Serializable {
```
srkukarni: Is this a breaking change? @maosongfu?

heron/api/src/java/com/twitter/heron/api/bolt/StatefulWindowedBoltExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+  private State state;
+
+  private static final String USER_STATE = "user.state";
+
+  public StatefulWindowedBoltExecutor(IStatefulWindowedBolt bolt) {
+    super(bolt);
+    this.statefulWindowedBolt = bolt;
+  }
+
+  @Override
```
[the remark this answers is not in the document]
jerrypeng: you are rights it redundant. I can remove

heron/api/src/java/com/twitter/heron/api/tuple/Tuple.java
maosongfu: it does not break current code. But I am wondering why we need this change?

heron/api/src/java/com/twitter/heron/api/topology/TopologyBuilder.java
```diff
@@ -185,6 +189,50 @@ public BoltDeclarer setBolt(String id, IWindowedBolt bolt, Number parallelismHin
...
+   * @param id the id of this component.
+   * This id is referenced by other components that want to consume this bolt's outputs.
+   * @param bolt the stateful windowed bolt
+   * @param <K> Type of key for {@link com.twitter.heron.api.state.HashMapState}
+   * @param <V> Type of value for {@link com.twitter.heron.api.state.HashMapState}
+   * @return use the returned object to declare the inputs to this component
+   * @throws IllegalArgumentException {@code parallelism_hint} is not positive
+   */
+  @SuppressWarnings("rawtypes")
+  public <K extends Serializable, V extends Serializable> BoltDeclarer setBolt(
```
maosongfu: I would argue let `IWindowedBolt extends IBolt`, then 
- we no longer need `setBolt(..., IStatefulWindowBolt, ...)` method.
- it follows the convention we used for low-level heron stateful API as: https://github.com/twitter/heron/blob/master/examples/src/java/com/twitter/heron/examples/api/StatefulWordCountTopology.java
- `IStatefulWindowedBolt` should be a type of `IBolt`. This method signature could potentially lead to ambiguous signature matching.

heron/api/src/java/com/twitter/heron/api/topology/TopologyBuilder.java
```diff
@@ -185,6 +189,50 @@ public BoltDeclarer setBolt(String id, IWindowedBolt bolt, Number parallelismHin
...
+   * @param bolt the stateful windowed bolt
+   * @param parallelismHint the number of tasks that should be assigned to execute this bolt.
+   * Each task will run on a thread in a process somwehere around the cluster.
+   * @param <K> Type of key for {@link com.twitter.heron.api.state.HashMapState}
+   * @param <V> Type of value for {@link com.twitter.heron.api.state.HashMapState}
+   * @return use the returned object to declare the inputs to this component
+   * @throws IllegalArgumentException {@code parallelism_hint} is not positive
+   */
+  @SuppressWarnings("rawtypes")
+  public <K extends Serializable, V extends Serializable> BoltDeclarer setBolt(
```
maosongfu: Same as above.

heron/api/src/java/com/twitter/heron/api/windowing/WindowManager.java
```diff
@@ -31,15 +34,20 @@
...
 /**
  * Tracks a window of events and fires {@link WindowLifecycleListener} callbacks
  * on expiry of events or activation of the window due to {@link TriggerPolicy}.
  *
  * @param <T> the type of event in the window.
  */
-public class WindowManager<T> implements TriggerHandler {
+public class WindowManager<T extends Serializable> implements TriggerHandler {
   private static final Logger LOG = Logger.getLogger(WindowManager.class.getName());
+  private static final String EVICTION_STATE_KEY = "es";
```
maosongfu: Will users mistakenly use the same key for their own states leading to a conflict? Same for below.

Also, should we have another separate `xxxState` class wrapping up these fields, and provide `restore(..)` and `getState(...)`?

heron/api/src/java/com/twitter/heron/api/tuple/Tuple.java
jerrypeng: @maosongfu The change is need because we need a way to store buffered tuples in windows that haven't been triggered yet.  So that when a checkpoint a happens, the state of all untriggered windows, which is essentially just buffered tuples, will be persisted. This is to maintain the effectively once guarantee which windowing components.

heron/api/src/java/com/twitter/heron/api/windowing/WindowManager.java
jerrypeng: @maosongfu the hashmap state is partitioned into USER_STATE and WINDOWING_INTERNAL_STATE and the user will only get state.get(USER_STATE). Thus even if the user uses the same key it won't matter since its part of a different namespace as you might think of it.

The getState and restore function called in this class will in turn call the same functions for different trigger policy and eviction policy classes

jerrypeng: We could have another class e.g. WindowState that contain these objects but what are the benefits of that?  It just seems like a wrapper class that will only be used in the WindowManager

heron/api/src/java/com/twitter/heron/api/topology/TopologyBuilder.java
jerrypeng: IWindowedBolt cannot extend IBolt since the execute API for IBolt is execute(Tuple tuple) but for windowed bolts in needs to be execute(TupleWindow window)

jerrypeng:
@maosongfu thanks for your review.  I have addressed your comments

maosongfu:
LGTM.

PR #2400, opened 2017-10-09
Add flags to the heron executor.

cckellogg:
Add executor flags to variables so they no longer have to have an order. Also, this makes it easier to debug and reason about the values.


srkukarni:
You need to modify aurora file as well

nlu90:
any update on this PR? it failed the ci test.

srkukarni:
@huijunw comments?

PR #2407, opened 2017-10-10
fixing flapMap signature

jerrypeng:


heron/api/src/java/com/twitter/heron/dsl/Streamlet.java
```diff
@@ -85,7 +85,7 @@
    * flattening the result
    * @param flatMapFn The FlatMap Function that should be applied to each element
   */
-  <T> Streamlet<T> flatMap(SerializableFunction<? super R, Iterable<? extends T>> flatMapFn);
+  <T> Streamlet<T> flatMap(SerializableFunction<? super R, ? extends Iterable<T>> flatMapFn);
```
srkukarni: Would we rather have SerializableFunction<? super R, ? extends Iterable<? extends T>> flatMapFn

jerrypeng: What does that really buy us though?

jerrypeng: Ya we can change it to that

jerrypeng:
@srkukarni thanks for the review! I have addressed your comments

PR #2416, opened 2017-10-11
allow users to set topology environment properties

jerrypeng:
I added TOPOLOGY_ENVIRONMENT to make heron more compatible with Storm since Storm has this feature.

We can use TOPOLOGY_WORKER_CHILDOPTS to set these environment properties but it might not be an elegant way of doing it. We would essentially have to put all the user defined environment properties on the java command line that starts the heron instance in a format: -Denv.variable=foo . If the user has many environment properties, this could cause the java command line to start the heron instance to be very large. I don't know if that is a good idea. Doing a System.setProperty in the start of the heron instance seems like a more eloquent approach

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -473,7 +473,7 @@ public class Config extends HashMap<String, Object> {
   public static void setTopologyStatefulStartClean(Map<String, Object> conf, boolean clean) {
     conf.put(Config.TOPOLOGY_STATEFUL_START_CLEAN, String.valueOf(clean));
   }
-  public static void setEnvironment(Map<String, Object> conf, Map<String, String> env) {
+  public static void setEnvironment(Map<String, Object> conf, Map env) {
     conf.put(Config.TOPOLOGY_ENVIRONMENT, env);
   }
 
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -235,6 +235,13 @@ public class Config extends HashMap<String, Object> {
   public static final String TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS =
       "topology.update.reactivate.wait.secs";
 
+  /**
+   * Topology-specific environment variables for the worker child process.
+   * This is added to the existing environment (that of the supervisor).
+   * This variable contains Map<String, String>
+   */
+  public static final String TOPOLOGY_ENVIRONMENT = "topology.environment";
+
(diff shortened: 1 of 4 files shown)


heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -235,6 +235,13 @@
   public static final String TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS =
       "topology.update.reactivate.wait.secs";
 
+  /**
+   * Topology-specific environment variables for the worker child process.
```
kramasamy: There is no notion of worker child process in heron - can you rephrase it?

storm-compatibility/src/java/org/apache/storm/Config.java
```diff
@@ -357,6 +357,12 @@
       = "topology.bolts.watermark.event.interval.ms";
 
   /**
+   * Topology-specific environment variables for the worker child process.
```
kramasamy: the same as before

maosongfu:
Quick questions: 
Is it feasible to pass env variables by using `TOPOLOGY_WORKER_CHILDOPTS ` and `TOPOLOGY_COMPONENT_JVMOPTS` in Config: https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/Config.java#L51?

What are the major benefits to have a specific Config to allow users to set topology environment properties? (Can you also put this in the description of this pull request)

storm-compatibility/src/java/org/apache/storm/Config.java
jerrypeng: Lets keep this same as it is for Storm

heron/api/src/java/com/twitter/heron/api/Config.java
jerrypeng: will fix

jerrypeng:
@maosongfu I added TOPOLOGY_ENVIRONMENT to make heron more compatible with Storm since Storm has this feature. 

We can use TOPOLOGY_WORKER_CHILDOPTS to set these environment properties but it might not be an elegant way of doing it.  We would essentially have to put all the user defined environment properties on the java command line that starts the heron instance in a format:  -Denv.variable=foo .  If the user has many environment properties, this could cause the java command line to start the heron instance to be very large.  I don't know if that is a good idea.

maosongfu:
@jerrypeng sounds fair enough. Looks good to me. Can you also add the justification you mentioned in the description of this pull request?

jerrypeng:
@maosongfu sounds good!  I have added the description.  Thanks for reviewing!

PR #2427, opened 2017-10-12
Make "delivery semantics" the default terminology for the older topology API

lucperkins:
This may be pointless but it just took a minute. Feel free to reject if there's no need. I just feel like having multiple terminologies in play produces unnecessary confusion, even when it involves older APIs.


lucperkins:
@srkukarni I do see your point, but I also feel like this is very confusing as well: https://twitter.github.io/heron/docs/developers/java/effectively-once/#specifying-delivery-semantics

lucperkins:
I'll close this for now but in the future we should try to be more aggressive about standardizing on terminology. The conceptual surface area of Heron is already quite intimidating for most users and we should do whatever we can to eliminate potential sources of confusion.

PR #2435, opened 2017-10-14
Add dual license to api files

aahmed-se:


HeartSaVioR:
Some questions on this change:

- Could you explain how dual license works and why it is needed? I think the source files Twitter created don't need to have ASF license header, and adding Twitter copyright to the source files which are copied from Apache Storm doesn't feel right. Yes I'm not a lawyer, but I'm wondering it as a PMC member of Apache Storm.
- Why Heron would want to keep Twitter copyright even after the project went to the ASF incubator? Moving the codebase should be followed up also.

aahmed-se:
Twitter copyright  will be removed once move to asf is complete, this is an interim solution , as files are intermixed with copied portions and modifications.

kramasamy:
Waiting for @maosongfu at Twitter to approve

PR #2436, opened 2017-10-16
fixing join operator

jerrypeng:
The join operator didn't have the correct processing semantics.  This PR fixes them.


srkukarni:
Looks good. Maybe make analogous change in python as well?

PR #2450, opened 2017-10-18
[DO NOT MERGE] - K8S Conversion to using Replica Sets

jrcrawfo:
What?
Changes to the k8s scheduler to use replica sets as the mechanism for deployment of topology containers

Why?
Using pods (previous version) wasn't fault tolerant when k8s nodes would fail.

### DIFF
--- a/heron/config/src/yaml/conf/kubernetes/packing.yaml
+++ b/heron/config/src/yaml/conf/kubernetes/packing.yaml
@@ -1,2 +1,3 @@
 # packing algorithm for packing instances into containers
 heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.RoundRobinPacking
+heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking
--- a/heron/config/src/yaml/conf/kubernetes/packing.yaml
+++ b/heron/config/src/yaml/conf/kubernetes/packing.yaml
@@ -1,3 +1,2 @@
 # packing algorithm for packing instances into containers
 heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.RoundRobinPacking
-heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking
--- a/heron/config/src/yaml/conf/kubernetes/scheduler.yaml
+++ b/heron/config/src/yaml/conf/kubernetes/scheduler.yaml
@@ -37,4 +37,4 @@ heron.scheduler.tunnel.retry.interval.ms:        1000
 heron.scheduler.tunnel.verify.count:             10
 
 # SSH tunnel host
-heron.scheduler.tunnel.host:              john@35.202.157.213
+heron.scheduler.tunnel.host:              <user>@<k8s_host>
--- a/heron/config/src/yaml/conf/kubernetes/scheduler.yaml
+++ b/heron/config/src/yaml/conf/kubernetes/scheduler.yaml
@@ -14,10 +14,27 @@ heron.directory.conf:                       "./heron-conf/"
 heron.directory.home:                       $HERON_HOME
 
 # The URI of kubernetes API
-heron.kubernetes.scheduler.uri:             <kubernetes_api_addr>
+heron.kubernetes.scheduler.uri:             http://localhost:8001
 
 # Invoke the IScheduler as a library directly
 heron.scheduler.is.service:                  False
 
 # docker repo for executor
-heron.executor.docker.image:                'ndustrialio/heron-core:latest'
+heron.executor.docker.image:                'streamlio/heron:latest'
+
+heron.scheduler.is.tunnel.needed: True
+
+# The connection timeout in ms when trying to connect to zk server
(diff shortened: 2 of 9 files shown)


jrcrawfo:
Still a work in progress as there is a bug in the update function -- not quite sure of the source of the bug, as it's not related to the scheduler. This bug is preventing me from testing the update functionality, however.

srkukarni:
@jrcrawfo What is the symptoms of the bug? Does it exist in master as well?

jrcrawfo:
@srkukarni i can look against master later, but here's the exception i was getting:

```
com.twitter.heron.spi.packing.PackingException: Could not initialize containers using existing packing plan
        at com.twitter.heron.packing.builder.PackingPlanBuilder.initContainers(PackingPlanBuilder.java:259)
        at com.twitter.heron.packing.builder.PackingPlanBuilder.addInstance(PackingPlanBuilder.java:153)
        at com.twitter.heron.packing.builder.PackingPlanBuilder.addInstance(PackingPlanBuilder.java:141)
        at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.placeFFDInstance(FirstFitDecreasingPacking.java:312)
        at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.assignInstancesToContainers(FirstFitDecreasingPacking.java:265)
        at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.getFFDAllocation(FirstFitDecreasingPacking.java:246)
        at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.repack(FirstFitDecreasingPacking.java:180)
        at com.twitter.heron.scheduler.RuntimeManagerRunner.buildNewPackingPlan(RuntimeManagerRunner.java:303)
        at com.twitter.heron.scheduler.RuntimeManagerRunner.updateTopologyHandler(RuntimeManagerRunner.java:182)
        at com.twitter.heron.scheduler.RuntimeManagerRunner.call(RuntimeManagerRunner.java:81)
        at com.twitter.heron.scheduler.RuntimeManagerMain.callRuntimeManagerRunner(RuntimeManagerMain.java:448)
        at com.twitter.heron.scheduler.RuntimeManagerMain.manageTopology(RuntimeManagerMain.java:396)
        at com.twitter.heron.scheduler.RuntimeManagerMain.main(RuntimeManagerMain.java:317)
Caused by: com.twitter.heron.packing.ResourceExceededException: Insufficient container resources to add instancePlan {component-name: exclaim1, task-id: 4, component-index: 3, instance-resource: {cpu: 1.000000, ram: ByteAmount{256 MB (268435456 bytes)}, disk: ByteAmount{1 GB (1073741824 bytes)}}} to container {containerId=2, instances=[{component-name: exclaim1, task-id: 2, component-index: 1, instance-resource: {cpu: 1.000000, ram: ByteAmount{256 MB (268435456 bytes)}, disk: ByteAmount{1 GB (1073741824 bytes)}}}], capacity={cpu: 1.000000, ram: ByteAmount{3 GB (2952790016 bytes)}, disk: ByteAmount{3 GB (3221225472 bytes)}}, paddingPercentage=10}[2017-10-18 09:49:32 -0400] [ERROR]: Could not initialize containers using existing packing plan

        at com.twitter.heron.packing.builder.PackingPlanBuilder.getContainers(PackingPlanBuilder.java:392)
[2017-10-18 09:49:32 -0400] [ERROR]: Failed to update topology: exclamationtopology
[2017-10-18 09:49:32 -0400] [DEBUG]: Elapsed time: 8.798s.
```

Was using the `heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking` repacking class

srkukarni:
@cckellogg Might be related to the same issue that you are looking into

cckellogg:
This is different. The issue I was seeing earlier was with the RoundRobinPackingPlan. It would be helpful to have a better to have more meaningful error message.

cckellogg:
What about using stateful sets instead? We would only have to make one rest call to start the topology then. Also, pods will have a predictable name too. The trick is to find the ordinal for the pod and then we can use that as the shard?

jrcrawfo:
@cckellogg i personally wouldn’t recommend using a stateful set, as it’s beta in the newest version (1.8 not available on google compute engine yet either). I'm also not sure you'd be able to get away with making one rest call, as every heron instance (pod) has a different command line argument. Keep in mind the specs that work for `kubectl` commands don't necessarily translate to API commands as well. `kubectl` does a lot of work itself making multiple API calls to get things deployed even it is just one command using `kubectl`

cckellogg:
We have been using stateful sets for zookeeper for a while so I'm not concerned with the beta part. Also, replicasets are in beta too. What i like about the stateful sets is that it groups everything together (with replica sets we would have to create one for each container/pod) and they provides a predictable name for the pod. The only thing that changes in the heron command currently is the shard-id right?
Stateful sets will give each pod an ordinal and if one of those pods does down it will restart with the same ordinal. 
Thoughts?

cckellogg:
@jrcrawfo is this patch ready with the replica sets?

nicknezis:
Closing this merge request because Heron is now using Kubernetes StatefulSets

PR #2456, opened 2017-10-18
Update python requests version

aahmed-se:


objmagic:
You forgot to update WORKSPACE file. cc @srkukarni

PR #2464, opened 2017-10-19
Make container ram padding configurable for round robin packing.

cckellogg:


heronpy/api/api_constants.py
```diff
@@ -65,6 +65,8 @@ class TopologyReliabilityMode(object):
 TOPOLOGY_CONTAINER_MAX_DISK_HINT = "topology.container.max.disk.hint"
 # Padding percentage for this container.
 TOPOLOGY_CONTAINER_PADDING_PERCENTAGE = "topology.container.padding.percentage"
+# Amount of ram padding per container.
+TOPOLOGY_CONTAINER_RAM_PADDING = "topology.container.ram.padding"
```
srkukarni: Can you add one for cpp as well?

PR #2466, opened 2017-10-19
Clear bookkeeper data, when redeploying minikube

jerrypeng:


jerrypeng:
@cckellogg can you take a look?

PR #2468, opened 2017-10-20
adding remote debugging functionality

jerrypeng:
- adding remote debugging functionality for LocalScheduler and KubernetesScheduler
- allow override config in heron_executor.py to actually override heron internal configs in HeronInstance
- Changed arguments passed to HeronInstance to -argument <argument_value>


kramasamy:
Nice functionality!

PR #2471, opened 2017-10-20
changing commandline arguments to be passed in a key value fashion

jerrypeng:
- changing commandline arguments to be passed in a key value fashion
- allow override config in heron_executor.py to actually override heron internal configs in HeronInstance

Splitting up PR: https://github.com/twitter/heron/pull/2468

### DIFF
--- a/heron/instance/src/java/BUILD
+++ b/heron/instance/src/java/BUILD
@@ -13,6 +13,7 @@ instance_deps_files =  \
         "//heron/common/src/java:config-java",
         "//heron/common/src/java:network-java",
         "//heron/common/src/java:utils-java",
+        "@commons_cli_commons_cli//jar"
     ]
 
 java_library(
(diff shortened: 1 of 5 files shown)


jerrypeng:
@srkukarni @maosongfu can you take a look at this. Thanks!

heron/instance/src/java/BUILD
```diff
@@ -13,6 +13,7 @@ instance_deps_files =  \
         "//heron/common/src/java:config-java",
         "//heron/common/src/java:network-java",
         "//heron/common/src/java:utils-java",
+        "@commons_cli_commons_cli//jar"
```
srkukarni: If we are introducing this dep, we need to add a shading rule for this.

jerrypeng: I have added a rule to shade this dep

objmagic:
have you changed Python instance as well?

jerrypeng:
@objmagic I have not, but this change should not break it.  Feel free to work on changing Python Heron instance to accept arguments in this fashion if you want to.

maosongfu:
@huijunw @objmagic We construct the arguments for heron-executor on aurora file, i.e. on the heron-cli side. We need to make sure a heron-cli will always deploy with the corresponding version of heron-core; otherwise, it will fail.

PR #2473, opened 2017-10-20
Add formattedLog method to Streamlet class

lucperkins:
I'm not sure if this is the best way to accomplish this, but I'm finding the default `log()` function to be a bit difficult to work with. It'd be nice to provide topology developers with the ability to format their final log output in accordance with a formatting function of their choosing rather than relying on the default `String.valueOf`, which may produce opaque output for more complex data types.


lucperkins:
Something like this:

```java


builder.newKVSource(() -> new KeyValue<>("player1", 1234))
        .formattedLog((k, v) -> String.format("Player: %s. Score: %d", k, v));
```

lucperkins:
@srkukarni Now that I think about it, it seems like it'd be better to just use the `consume` operator for this. Like so:

```java
builder.newKVSource(() -> new KeyValue<>("player1", 1234))
        .consume(s -> LOG.info(/* formatted string */));
```

I don't think that providing a new operator is much of a win. Happy to close this if you agree.

srkukarni:
+1

PR #2474, opened 2017-10-21
fixing bugs in python streamlet join

jerrypeng:


jerrypeng:
@srkukarni can you look at this

PR #2477, opened 2017-10-22
Add gcp bookkeeper deploy file.

cckellogg:


deploy/kubernetes/gcp/bk-apiserver.yaml
```diff
@@ -0,0 +1,47 @@
+##
```
kramasamy: can you rename the file to bookkeeper-apiserver.yaml? just to make it easy.

Issue #2079, opened 2017-07-18
duplicate active stmgrs due to zombie container


huijunw:
Zombie container:
- The Aurora thought the container X was dead.
- The connections between the container X and some other containers are lost, but the connection between X and Tmaster exists.
- The processes are still running inside the container X.
- Aurora starts a new container Y for the zombie container X.
- Two stmgrs X and Y are now sending heartbeat to the Tmaster.

The present behavior of Tmaster:
Tmaster rejects the new stmgr Y registration and waits for the old stmgr X to timeout. However, we find that the zombie container X does not stop for long time.

Aurora is supposed to assure no duplicate containers, however this assumption is not always truth. A proposal on Heron side:
When Tmaster finds two duplicate stmgrs X and Y, Tmaster sends _killcontainer_ to both X and Y. The container X exits and Aurora does not restart it. The container Y exits and Aurora restarts it. In the end, only one container Y survives.
@billonahill @srkukarni @kramasamy comments are welcome.

srkukarni:
Could you elaborate more on how/who the Tmaster would send the 'killcontainer' message? Is it to the stmgrs in X and Y. Or to the scheduler somehow?

huijunw:
There is a handler in heron-shell, which can make heron-executor suicide. 
https://github.com/twitter/heron/blob/master/heron/shell/src/python/main.py#L37

Tmaster may send request to this heron-shell handler

nlu90:
The assumption here is once the executor in zombie host's container X is killed, it will not be restarted by Aurora again. But the executor in normal host's container Y is killed, Aurora restarts it again. Thus, only one executor for a particular container number is running eventually.

srkukarni:
Are we sure that if executor dies in a zombie container, internal aurora handlers within that container won't restart it?

huijunw:
Zombie container scenario is hard to re-produce. I would suggest enable the proposed implementation and wait for the zombie container to appear to verify.

billonahill:
@huijunw before going to far down the path of working around this from the Heron side, could you check in with the Aurora team to see where they are towards fixing the root cause. They had previously planned to work on this is Q2.

huijunw:
All right @billonahill

huijunw:
we confirmed with aurora team that aurora does not restart the zombie container. aurora team told us, if a new container is scheduled, the old zombie is already marked as `task lost` in aurora state, thus aurora does restart the `task lost` old zombie if it dies. @srkukarni

[...]

PR #2480, opened 2017-10-23
Kill both containers when zombie host happens

objmagic:
Kill both containers when the zombie one still attempts to connect TMaster so that zombie one will not get rescheduled by scheduler (idea by @huijunw). Test done locally via a simulation of zombie host scenario by local scheduler.

resolves https://github.com/twitter/heron/issues/2079


heron/tmaster/src/cpp/manager/tmaster.cpp
```diff
@@ -595,22 +600,51 @@ void TMaster::HandleCleanStatefulCheckpointResponse(proto::system::StatusCode _s
   tmaster_controller_->HandleCleanStatefulCheckpointResponse(_status);
 }
 
+void TMaster::KillContainer(const std::string& host_name,
+    sp_int32 shell_port, sp_string stmgr_id) {
+  LOG(ERROR) << "Start killing " << stmgr_id << " on " <<
```
srkukarni: This should be INFO

PR #2481, opened 2017-10-23
Add Simulator test

aahmed-se:


kramasamy:
@aahmed-se - a bunch of checkstyle errors that you need to fix.

PR #2483, opened 2017-10-23
validate topology before submitting

jerrypeng:
- Validate topology before submitting.  

 - Validate code will throw InvalidTopologyException.  InvalidTopologyException was caught in places before but never thrown so lets actually use it

- moved TopologyUtils to heron api because heron-api cannot depend on common-utils since common-utils depends on heron-api


heron/api/tests/java/BUILD
```diff
@@ -3,6 +3,7 @@ load("/tools/rules/heron_deps", "heron_java_api_proto_files")
 api_deps_files = [
     "//heron/api/src/java:api-java",
     "//heron/common/src/java:utils-java",
+    "//heron/common/src/java:basics-java",
```
srkukarni: why are we adding this dep?

jerrypeng: com.twitter.heron.common.basics.ByteAmount is from there

PR #2484, opened 2017-10-23
adding remote debugging functionality

jerrypeng:


heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -280,6 +282,11 @@ private SchedulerUtils() {
     commands.add(createCommandArg(ExecutorFlag.HealthManagerMode, healthMgrMode));
     commands.add(createCommandArg(ExecutorFlag.HealthManagerClasspath,
         Context.healthMgrClassPath(config)));
+    if (freePorts.size() > 9) {
```
nwangtw: Why 9? Suggest to make a const for it with a name.

Also I am wondering if it could be useful to have an else {} block with a log?

jerrypeng: This is because ports are pass in as positional arguments.  There are 9 ports prior to the debugging ports which could be more than one port.  

This PR should address the positional arguments issue:

https://github.com/twitter/heron/pull/2516

nwangtw: I see. Feels a little scary for future maintenance, especially for adding or removing ports. I looked at the other PR briefly but don't have good idea about the typical use cases yet. A few questions I am wondering so far:

- is this map necessary? can we limit the jvm debugger to be one at a time? or it is normal to have multiple debuggers in jvm world?
- could it be better to refactor all ports to be a port map instead of creating a map for debugging port and leave others as is?

jerrypeng: @nwangtw 
Your first question:

The reason why we pass multiple debugger ports is because the heron_executor.py can potentially starts up multiple heron instances based on the physical plan thus we need multiple debugger port assignments to be passed to it.

Second question:

That is essentially what #2516 does.  Puts all ports in a map

heron/executor/src/python/heron_executor.py
```diff
@@ -55,7 +57,7 @@ def print_usage():
       " <scheduler_classpath> <scheduler_port> <python_instance_binary>"
       " <metricscachemgr_classpath> <metricscachemgr_masterport> <metricscachemgr_statsport>"
       " <is_stateful> <ckptmgr_classpath> <ckptmgr_port> <stateful_config_file> "
-      " <healthmgr_mode> <healthmgr_classpath> <cpp_instance_binary>")
+      " <healthmgr_mode> <healthmgr_classpath> <cpp_instance_binary> <jvm_remote_debugger_ports>")
```
huijunw: do we need to add one more param in config/heron.aurora

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
```diff
@@ -257,6 +260,12 @@ public static void main(String[] args) throws IOException {
     String overrideConfigFile
         = commandLine.getOptionValue(CommandLineOptions.OVERRIDE_CONFIG_FILE);
 
+    Integer remoteDebuggerPort = null;
+    if (commandLine.hasOption(CommandLineOptions.REMOTE_DEBUGGER_PORT)) {
+      remoteDebuggerPort = Integer.parseInt(
```
huijunw: once heron-instance listens to this port, what requests will it accept?

jerrypeng: a user can attach a remote debugger to the port

heron/executor/src/python/heron_executor.py
jerrypeng: currently remote debugging is implemented for localscheduler and kubernetes scheduler.

srkukarni:
@nwangtw Now that the port mapping pr has been submitted, this pr is simplified and ready for another review. Thanks!

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
```diff
@@ -266,8 +275,13 @@ public static void main(String[] args) throws IOException {
...
 
     // Create the protobuf Instance
-    PhysicalPlans.InstanceInfo instanceInfo = PhysicalPlans.InstanceInfo.newBuilder().
-        setTaskId(taskId).setComponentIndex(componentIndex).setComponentName(componentName).build();
+    PhysicalPlans.InstanceInfo.Builder instanceInfoBuilder
+        = PhysicalPlans.InstanceInfo.newBuilder().setTaskId(taskId)
+        .setComponentIndex(componentIndex).setComponentName(componentName);
+    if (remoteDebuggerPort != null) {
+      instanceInfoBuilder.setRemoteDebuggerPort(remoteDebuggerPort);
+    }
```
nwangtw: Please add the port info in the logging message below if it is not there yet:

LOG.info("\nStarting instance " + instanceId + " for topology " + topologyName
         + " and topologyId " + topologyId + " for component " + componentName		         + " and topologyId " + topologyId + " for component " + componentName
         + " with taskId " + taskId + " and componentIndex " + componentIndex		         + " with taskId " + taskId + " and componentIndex " + componentIndex
         + " and stmgrId " + streamId + " and stmgrPort " + streamPort		         + " and stmgrId " + streamId + " and stmgrPort " + streamPort
         + " and metricsManagerPort " + metricsPort);		         + " and metricsManagerPort " + metricsPort);
 		 
     LOG.info("System Config: " + systemConfig);		     LOG.info("System Config: " + systemConfig);

heron/proto/physical_plan.proto
```diff
@@ -33,6 +33,7 @@ message InstanceInfo {
   required int32 component_index = 2; // specific to this component
   required string component_name = 3;
   repeated string params = 4;
+  optional int32 remote_debugger_port = 5;
```
srkukarni: Can you add a little more color to what this port is?

heron/instance/src/java/com/twitter/heron/instance/HeronInstance.java
jerrypeng: I have added this

jerrypeng:
@srkukarni @nwangtw thanks for the review.  I have addressed your comments

PR #2485, opened 2017-10-24
Acking topologies for Java documentation

lucperkins:


examples/src/java/com/twitter/heron/examples/api/AckingTopology.java
```diff
@@ -103,18 +108,20 @@ public void close() {
     }
 
     public void nextTuple() {
-      // We explicitly slow down the spout to avoid the stream mgr to be the bottleneck
+      // Explicitly slow down the spout to prevent the stream manager from being  the bottleneck
```
srkukarni: this is not necessary

examples/src/java/com/twitter/heron/examples/api/AckingTopology.java
```diff
@@ -103,18 +108,20 @@ public void close() {
...
       Utils.sleep(1);
 
       final String word = words[rand.nextInt(words.length)];
 
-      // To enable acking, we need to emit tuple with MessageId, which is an object
+      // To enable acking, we need to emit each tuple with a MessageId, which is an Object
```
srkukarni: We technically need to create a new message id for every new message. Maybe some explanation about that and why we are doing the way we are doing in this example

examples/src/java/com/twitter/heron/examples/api/AckingTopology.java
```diff
@@ -103,18 +108,17 @@ public void close() {
     }
 
     public void nextTuple() {
-      // We explicitly slow down the spout to avoid the stream mgr to be the bottleneck
-      Utils.sleep(1);
-
       final String word = words[rand.nextInt(words.length)];
 
-      // To enable acking, we need to emit tuple with MessageId, which is an object
+      // To enable acking, we need to emit each tuple with a MessageId, which is an Object
```
srkukarni: Can you add the following line as well
Technically each new message that we emit should be annottated with a unique message id. This will allow the spout to keep track of which messages should be acked back to the producer/retried when appropriate ack/fail happens. However for the sake of simplicitity we are tagging all tuples in this example to have the same message id.

PR #2486, opened 2017-10-24
Updated the workspace after we upgraded to newer requests

srkukarni:


kramasamy:
This will not work because the link is not correct.

PR #2487, opened 2017-10-24
Make Streamlets and KVStreamets independent of each other

srkukarni:
In the current Streamlet world, a KVStreamlet is also a Streamlet. This means that all of Streamlet's operations(like map/flatmap/filter, etc) are also available to it. However, the result of these operations are all Streamlets which means to access the KVStreamlet operations(like join), one needs to mapToKV again.
This pr changes that and makes Streamlet and KVStreamlet independent of each other. It also adds all the usual operators(map/flatmap/filter, etc) to the KVStreamlet


srkukarni:
@jerrypeng

lucperkins:
@srkukarni I'm experimenting with this locally and I've noted a regression in that the developer can no longer use `setName` and `setNumPartitions` on either `Streamlet`s or `KVStreamlet`s. It looks like those methods have been somehow depublicized in this implementation.

lucperkins:
@srkukarni Never mind. This was completely an error on my part. My experiments with this modified API have been 100% successful thus far.

jerrypeng:
We have the mapToKv operator for Streamlet -> KVStreamlet but what about vice versa?

jerrypeng:
@srkukarni does it make sense for Streamlet.reduceByWindow to return a KVStreamlet<Window, R> ?

Shouldn't it return another Streamlet? Something like Streamlet<Tuple2<Window, R>> ?

jerrypeng:
@srkukarni I am also noticing that the API might not be sufficient for processing windows in a more generic fashion.  What if I wanted to more generic processing of windows other than just a reduceByWindow?  For example the sliding average?  Perhaps we need the concept of a WindowedStreamlet?  So there is another set of operators that can operate on WindowedStreamlets?  Operators that consume windows of data instead of single tuples.

srkukarni:
wrt mapToKV equivalent, do we need that in KVStreamlet? All operations on Streamlet can be done on KVStreamlet. Can you give an example when this might be needed

srkukarni:
wrt Streamlet.reduceByWindow returning KVStreamlet, imo its cleaner to explicitly recognize that the return streamlet is a KVStreamlet rather than dance around that with a Tuple2. @lucperkins What do you think?

srkukarni:
WRT windowing operator, that might be true. Maybe you can explore that in a seperate pr?

lucperkins:
@srkukarni I'm 100% in favor of *not* using a tuple for that. To me, one of the core benefits of the Streamlet API is that you can ditch tuples entirely.

jerrypeng:
@srkukarni in regards an operator to transform KVStreamlet -> Streamlet.  A user might want to do a reduceByWindow instead of a reduceByWindowAndKey.  I think we should allow users to be able to convert if they wanted to.

jerrypeng:
@srkukarni everything else LGTM

lucperkins:
@jerrypeng +1 to allowing for windowed operations for non-KV streamlets

srkukarni:
@jerrypeng Should we instead add a reduceByWindow operator to KVStreamlet?

jerrypeng:
@srkukarni imo we should add a KVStreamlet -> Streamlet operator.  The reduceByWindow is just one example, I can also image a user would want convert KVStreamlet ->Streamlet and then print then log result or use other operators that operates on Streamlets and not KVStreamlets

srkukarni:
@lucperkins One of the intentions of this pr was to have all functionalities replicated in each other(KVStreamlet and Streamlet). Thus KVStreamlet has log, etc operation just like Streamlet. Thats why I mentioned adding reduceByWindow to it as well

PR #2495, opened 2017-10-27
Fix local scheduler without core uri paths.

cckellogg:
This patch will expand the HERON_HOME env variable if found to
a full path.


cckellogg:
Will fix in another pr.

PR #2496, opened 2017-10-27
include heron.aurora file in the distribution

kramasamy:


heron/config/src/yaml/BUILD
```diff
@@ -14,7 +14,7 @@ filegroup(
 
 filegroup(
     name = "conf-yaml",
-    srcs = glob(["conf/**/*.yaml"]),
+    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"]),
```
huijunw: there is `conf-aurora-yaml` why not use that

objmagic:
LGTM.

heron/config/src/yaml/BUILD
kramasamy: @huijunw - this target ```conf-yaml``` includes every yaml file and also every aurora file.

PR #2498, opened 2017-10-28
healthmgr: add processingdataskew detector for slow instance diagnosor

huijunw:
fix a bug found by @avflor : add the processing rate skew detector for slow instance diagnoser


avflor:
Looks good to me..

PR #2501, opened 2017-10-30
Java streamlet examples

lucperkins:
This is ready for review.


examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
```diff
@@ -0,0 +1,141 @@
...
+     * actually processed. In this case, each incoming element is converted
+     * to a byte array and written to the temporary file (successful writes
+     * are also logged). Any exceptions are converted to RuntimeExceptions,
+     * which will effectively kill the topology.
+     */
+    public void put(T element) {
+      byte[] bytes = String.format("%s\n", element.toString()).getBytes();
+
+      try {
+        Files.write(tempFilePath, bytes, StandardOpenOption.APPEND);
```
srkukarni: Does Files.write do an open/write/close? Isnt that inefficient to do it per element

examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
```diff
@@ -0,0 +1,141 @@
+//  Copyright 2017 Twitter. All rights reserved.
```
srkukarni: I believe a better example is a Copy example where in you read from one File Source and copy it to another File Sink

examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
lucperkins: It probably is less efficient than a lot of other alternatives, but it's only happening every half second and I like that it's just a one liner, which keeps the code clean

examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
lucperkins: We could do that, but there is a separate example topology that covers sources. Up to you.

srkukarni: It's ok for now. When we will have say a twitter connector handy, we could think of converting this example to reading twitter and writing to a file

lucperkins: Agreed. That would be very cool. We can iterate on these a lot in the coming weeks.

examples/src/java/com/twitter/heron/examples/streamlet/ImpressionsAndClicksTopology.java
```diff
@@ -0,0 +1,197 @@
...
+            impression -> impression.getUserId(),
+            // Key extractor for the clicks streamlet
+            click -> click.getUserId(),
+            // Window configuration for the join operation
+            WindowConfig.TumblingCountWindow(25),
+            // Join type (inner join means that all elements from both streams will be included)
+            JoinType.INNER,
+            // For each element resulting from the join operation, a value of 1 will be provided
+            // if the user IDs match between the elements (or a value of 0 if they don't).
+            (user1, user2) -> (user1.getAdId().equals(user2.getAdId())) ? 1 : 0
```
srkukarni: Do you want to instead join on the adid/userid combo?

examples/src/java/com/twitter/heron/examples/streamlet/ImpressionsAndClicksTopology.java
```diff
@@ -0,0 +1,197 @@
...
+        )
+        // The reduce function counts the number of ad clicks per user.
+        .reduceByKeyAndWindow(
```
srkukarni: this reduce will not needed once you change the key in the join

examples/src/java/com/twitter/heron/examples/streamlet/IntegerProcessingTopology.java
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * All Heron topologies require a main function that defines the topology's behavior
+   * at runtime
+   */
+  public static void main(String[] args) throws Exception {
+    Builder builder = Builder.createBuilder();
+
+    Streamlet<Integer> zeroes = builder.newSource(() -> 0);
+
+    builder.newSource(() -> ThreadLocalRandom.current().nextInt(1, 11))
```
srkukarni: We need a better example for repartition.

examples/src/java/com/twitter/heron/examples/streamlet/ImpressionsAndClicksTopology.java
lucperkins: Okay, yeah, I see the issue now. Will change up the processing logic accordingly.

examples/src/java/com/twitter/heron/examples/streamlet/SmartWatchTopology.java
```diff
@@ -0,0 +1,138 @@
...
+        .map(keyWindow -> {
+          // The per-key result of the previous reduce step
+          long totalFeetRun = keyWindow.getValue();
+
+          // The amount of time elapsed
+          long startTime = keyWindow.getKey().getWindow().getStartTime();
+          long endTime = keyWindow.getKey().getWindow().getEndTime();
+          long timeLengthMillis = endTime - startTime; // Cast to float to use as denominator
+
+          LOG.info(
```
srkukarni: might be better to pipe this to the log operator

examples/src/java/com/twitter/heron/examples/streamlet/IntegerProcessingTopology.java
lucperkins: It is a bit hackneyed. Feel free to suggest other ideas. I’m really not sure how people will use this in the wild.

examples/src/java/com/twitter/heron/examples/streamlet/SmartWatchTopology.java
lucperkins: It’s eventually logged in the consume operator. In this case I’m using it mostly for debugging so it can be removed.

PR #2502, opened 2017-10-30
Added a generalized reduceByKeyAndWindow operator inside the Streamlet

srkukarni:


srkukarni:
@jerrypeng

heron/api/src/java/com/twitter/heron/streamlet/KVStreamlet.java
```diff
@@ -151,4 +151,15 @@
...
+
+  /**
+   * Return a new Streamlet in which for each time_window, all elements are belonging to the
+   * same key are reduced using the BiFunction/identity and the result is emitted.
+   * @param windowCfg This is a specification of what kind of windowing strategy you like to have.
+   * Typical windowing strategies are sliding windows and tumbling windows
+   * @param reduceFn The reduce function that you want to apply to all the values of a key.
+   */
+  <VR> KVStreamlet<KeyedWindow<K>, VR> reduceByKeyAndWindow(WindowConfig windowCfg,
+                            VR identity,
```
lucperkins: I'd suggest that we brainstorm some alternative names for `identity`. Maybe `initialValue` or something like that?

heron/api/src/java/com/twitter/heron/streamlet/impl/KVStreamletImpl.java
```diff
@@ -290,4 +291,21 @@ public void toSink(Sink<? super KeyValue<? super K, ? super V>> sink) {
     addChild(retval);
     return retval;
   }
+
+  /**
+   * Return a new Streamlet in which for each time_window, all elements are belonging to the
+   * same key are reduced using the BiFunction/identity and the result is emitted.
+   * @param windowCfg This is a specification of what kind of windowing strategy you like to have.
+   * Typical windowing strategies are sliding windows and tumbling windows
+   * @param reduceFn The reduce function that you want to apply to all the values of a key.
```
jerrypeng: Can you update the java doc to reflect the change in method signature?

heron/api/src/java/com/twitter/heron/streamlet/KVStreamlet.java
```diff
@@ -151,4 +151,15 @@
    */
   KVStreamlet<KeyedWindow<K>, V> reduceByKeyAndWindow(WindowConfig windowCfg,
                                                       SerializableBinaryOperator<V> reduceFn);
+
+  /**
+   * Return a new Streamlet in which for each time_window, all elements are belonging to the
+   * same key are reduced using the BiFunction/identity and the result is emitted.
+   * @param windowCfg This is a specification of what kind of windowing strategy you like to have.
+   * Typical windowing strategies are sliding windows and tumbling windows
+   * @param reduceFn The reduce function that you want to apply to all the values of a key.
```
jerrypeng: Can you update the java doc to reflect the change in method signature?

lucperkins:
I've tried this out locally and it works as expected. Just made one comment re: variable naming. Another thing I'd recommend is enabling the same thing for `reduceByWindow`. There doesn't seem to be a compelling reason not to allow for that same logic there.

jerrypeng:
@srkukarni I think the existing:

@Override
   public KVStreamlet<KeyedWindow<K>, V>
       reduceByKeyAndWindow(WindowConfig windowCfg, SerializableBinaryOperator<V> reduceFn) 

should called the more generic version with the identity parameter.  The identity parameter can just be passed as null or something.  This way you don't have two implementations of the same thing essentially

srkukarni:
@jerrypeng That makes sense. I've re-implemented it as one version

srkukarni:
@lucperkins I'm modeling the naming on native java functions which have essentially the same concept. For example look at https://docs.oracle.com/javase/tutorial/collections/streams/reduction.html.

srkukarni:
@lucperkins I've added the reduceByWindow general function as you requested

heron/api/src/java/com/twitter/heron/streamlet/impl/operators/ReduceByWindowOperator.java
```diff
@@ -49,14 +52,10 @@ public void prepare(Map map, TopologyContext topologyContext, OutputCollector ou
   @SuppressWarnings("unchecked")
   @Override
   public void execute(TupleWindow inputWindow) {
-    I reducedValue = null;
+    T reducedValue = identity;
```
jerrypeng: shouldn't a null check be conducted here? if identity is null, it should be set to to the first tuple or something

heron/api/src/java/com/twitter/heron/streamlet/impl/operators/GeneralReduceByKeyAndWindowOperator.java
```diff
@@ -0,0 +1,87 @@
...
+    Window window = new Window(startWindow, endWindow, inputWindow.get().size());
+    for (K key : reduceMap.keySet()) {
+      KeyedWindow<K> keyedWindow = new KeyedWindow<>(key, window);
+      collector.emit(new Values(new KeyValue<>(keyedWindow, reduceMap.get(key))));
+    }
+  }
+
+  private void addMap(Map<K, VR> reduceMap, KeyValue<K, V> tup) {
+    if (!reduceMap.containsKey(tup.getKey())) {
+      reduceMap.put(tup.getKey(), identity);
```
jerrypeng: shouldn't a null check be conducted here? if identity is null, it should be set to to the first tuple or something

lucperkins:
@srkukarni Got it. If that's standard usage, I'd say go for it.

jerrypeng:
@srkukarni it would be nice if you can also write some unit tests to test out the logic of the ReduceByKeyAndWindowOperator and ReduceByWindowOperator

srkukarni:
@jerrypeng Added the unittests that you requested.

PR #2503, opened 2017-10-30
Issue 2500: Provide a yaml file to deploy bookkeeper in a StatefulSet for k8s installation

sijie:
- provide a yaml file for deploying bookkeeper in a StatefulSet (both general and gcp)
- update README with a note about installation on DC/OS K8S.


cckellogg:
Do we need to add pod affinity so bookies are not scheduled on the same node?

sijie:
@cckellogg ideally yes. I can add one here.

sijie:
@cckellogg added pod affinity

PR #2508, opened 2017-11-01
Use Stateful windows for streamlets

srkukarni:
@jerrypeng


srkukarni:
This is actually not needed.

PR #2511, opened 2017-11-01
fixing formating in examples

jerrypeng:


jerrypeng:
@maosongfu @srkukarni

PR #2512, opened 2017-11-01
increasing memory for examples

jerrypeng:
Saw OutOfMemoryErrors when running example topologies.

### DIFF
--- a/examples/src/java/com/twitter/heron/examples/api/ExampleResources.java
+++ b/examples/src/java/com/twitter/heron/examples/api/ExampleResources.java
@@ -17,7 +17,7 @@ import com.twitter.heron.common.basics.ByteAmount;
 
 public final class ExampleResources {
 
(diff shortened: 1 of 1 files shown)


cckellogg:
Should we reduce the container padding too? Config.setContainerRamPadding

ashvina:
Hi @jerrypeng 
At one point the resources for examples were reduced (see: 1f5b57b905e7a8ac56c14895c55b07667ceeb358). I think some users had requested that change to be able to run low footprint topologies. 

What do you think about using configurable resources instead of increasing for all the example topologies? Also did you see the OOME in existing examples?

jerrypeng:
@ashvina I saw the OOME immediately after starting the AckingTopology.  Perhaps, we can just increase the resource request for the AckingTopology and leave the rest the same as before.  Would you prefer that?

PR #2516, opened 2017-11-02
 improve passing of executor ports

jerrypeng:
Ports are passed around as a list in which each position represents a port for a specific purpose.  This is difficult to have optional ports.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
@@ -20,7 +20,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -41,7 +40,7 @@ import com.twitter.heron.spi.statemgr.SchedulerStateManagerAdaptor;
 import com.twitter.heron.spi.utils.ShellUtils;
 
 public final class SchedulerUtils {
-  public static final int PORTS_REQUIRED_FOR_EXECUTOR = 9;
+  public static final int PORTS_REQUIRED_FOR_EXECUTOR = ExecutorPorts.getRequiredNumOfPorts();
   public static final int PORTS_REQUIRED_FOR_SCHEDULER = 1;
   public static final String SCHEDULER_COMMAND_LINE_PROPERTIES_OVERRIDE_OPTION = "P";
 
@@ -50,18 +49,83 @@ public final class SchedulerUtils {
   /**
    * Enum that defines the type of ports that an heron executor needs
    */
-  public enum ExecutorPorts {
-    MASTER_PORT,
-    TMASTER_CONTROLLER_PORT,
-    TMASTER_STATS_PORT,
-    SHELL_PORT,
-    METRICS_MANAGER_PORT,
-    SCHEDULER_PORT,
-    METRICS_CACHE_MASTER_PORT,
-    METRICS_CACHE_STATS_PORT,
-    CHECKPOINT_MANAGER_PORT;
+  public static class ExecutorPorts extends HashMap<ExecutorPorts.ExecutorPortNames, Object> {
+    private static final long serialVersionUID = 3594142561871208879L;
 
-    public static Object getRequired(Map<ExecutorPorts, Object> ports, ExecutorPorts type) {
+    public enum ExecutorPortNames {
+      MASTER_PORT("master-port"),
+      TMASTER_CONTROLLER_PORT("tmaster-controller-port"),
+      TMASTER_STATS_PORT("tmaster-stats-port"),
+      SHELL_PORT("shell-port"),
+      METRICS_MANAGER_PORT("metrics-manager-port"),
+      SCHEDULER_PORT("scheduler-port"),
+      METRICS_CACHE_MASTER_PORT("metrics-cache-port"),
+      METRICS_CACHE_STATS_PORT("metrics-cache-stats-port"),
+      CHECKPOINT_MANAGER_PORT("checkpoint-manager-port");
+
+      private final String name;
+
+      ExecutorPortNames(String name) {
+        this.name = name;
+      }
+
+      public String toString() {
+        return this.name;
+      }
+
+      public static String[] getPortNames() {
+        List<String> names = new LinkedList<>();
+        for (ExecutorPortNames executorPortNames : ExecutorPortNames.values()) {
+          names.add(executorPortNames.toString());
+        }
+
+        return (String[]) names.toArray();
+      }
+    }
+
+    public static String[] getRequiredPorts() {
+      return new String[]{
+          ExecutorPortNames.MASTER_PORT.toString(),
+          ExecutorPortNames.TMASTER_CONTROLLER_PORT.toString(),
+          ExecutorPortNames.TMASTER_STATS_PORT.toString(),
+          ExecutorPortNames.SHELL_PORT.toString(),
+          ExecutorPortNames.METRICS_MANAGER_PORT.toString(),
+          ExecutorPortNames.SCHEDULER_PORT.toString(),
+          ExecutorPortNames.METRICS_CACHE_MASTER_PORT.toString(),
+          ExecutorPortNames.METRICS_CACHE_STATS_PORT.toString(),
+          ExecutorPortNames.CHECKPOINT_MANAGER_PORT.toString()
+      };
+    }
+
+    public static int getRequiredNumOfPorts() {
+      return getRequiredPorts().length;
+    }
+
+    public static ExecutorPorts withRequiredPorts(List<Integer> portsToUse) {
+      ExecutorPorts executorPorts = new ExecutorPorts();
+      executorPorts.setRequiredPorts(portsToUse);
+      return executorPorts;
+    }
+
+    public void setRequiredPorts(List<Integer> portsToUse) {
+      if (portsToUse.size() < getRequiredNumOfPorts()) {
+        throw new IllegalArgumentException("Did not provide enough Ports. "
+            + "Number of ports provided: " + portsToUse.size()
+            + ". Number of ports required: " + getRequiredNumOfPorts());
+      }
+
+      for (int i = 0; i < portsToUse.size(); i++) {
+        int port = portsToUse.get(i);
+        if (port == -1) {
+          throw new IllegalArgumentException("Invalid Port: "
+              + port + ". Failed to find enough ports for executor");
+        } else {
+          this.put(ExecutorPortNames.values()[i], port);
+        }
+      }
+    }
+
+    public static Object getRequired(ExecutorPorts ports, ExecutorPortNames type) {
       Object ret = getOptional(ports, type);
       if (ret == null) {
         throw new IllegalArgumentException("Required executor port "
@@ -70,7 +134,7 @@ public final class SchedulerUtils {
       return ports.get(type);
     }
 
-    public static Object getOptional(Map<ExecutorPorts, Object> ports, ExecutorPorts type) {
+    public static Object getOptional(ExecutorPorts ports, ExecutorPortNames type) {
       return ports.get(type);
     }
   }
@@ -164,24 +228,13 @@ public final class SchedulerUtils {
       Config config,
       Config runtime,
       int containerIndex,
-      List<Integer> freePorts) {
+      ExecutorPorts freePorts) {
     // First let us have some safe checks
     if (freePorts.size() < PORTS_REQUIRED_FOR_EXECUTOR) {
       throw new RuntimeException("Failed to find enough ports for executor");
     }
-    for (int port : freePorts) {
-      if (port == -1) {
-        throw new RuntimeException("Failed to find available ports for executor");
-      }
-    }
 
-    // Convert port to string
-    List<String> ports = new LinkedList<>();
-    for (int port : freePorts) {
-      ports.add(Integer.toString(port));
-    }
-
-    return getExecutorCommand(config, runtime, containerIndex, ports);
+    return getExecutorCommand(config, runtime, containerIndex, freePorts);
   }
 
   /**
@@ -197,7 +250,7 @@ public final class SchedulerUtils {
       Config config,
       Config runtime,
       int containerIndex,
-      List<String> ports) {
+      ExecutorPorts ports) {
     List<String> commands = new ArrayList<>();
     commands.add(Context.executorBinary(config));
     commands.add(createCommandArg(ExecutorFlag.Shard, Integer.toString(containerIndex)));
@@ -218,26 +271,27 @@ public final class SchedulerUtils {
    * @return String[] representing the arguments to start heron-executor
    */
   public static String[] executorCommandArgs(
-      Config config, Config runtime, Map<ExecutorPorts, Object> freePorts) {
+      Config config, Config runtime, ExecutorPorts freePorts) {
     TopologyAPI.Topology topology = Runtime.topology(runtime);
 
-    String masterPort = (String) ExecutorPorts.getRequired(freePorts, ExecutorPorts.MASTER_PORT);
-    String tmasterControllerPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.TMASTER_CONTROLLER_PORT);
-    String tmasterStatsPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.TMASTER_CONTROLLER_PORT);
-    String shellPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.SHELL_PORT);
-    String metricsmgrPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.METRICS_MANAGER_PORT);
-    String schedulerPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.SCHEDULER_PORT);
-    String metricsCacheMasterPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.METRICS_CACHE_MASTER_PORT);
-    String metricsCacheStatsPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.METRICS_CACHE_STATS_PORT);
-    String ckptmgrPort = (String) ExecutorPorts.getRequired(
-        freePorts, ExecutorPorts.CHECKPOINT_MANAGER_PORT);
+    String masterPort = ExecutorPorts.getRequired(freePorts,
+        ExecutorPorts.ExecutorPortNames.MASTER_PORT).toString();
+    String tmasterControllerPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.TMASTER_CONTROLLER_PORT).toString();
+    String tmasterStatsPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.TMASTER_CONTROLLER_PORT).toString();
+    String shellPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.SHELL_PORT).toString();
+    String metricsmgrPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.METRICS_MANAGER_PORT).toString();
+    String schedulerPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.SCHEDULER_PORT).toString();
+    String metricsCacheMasterPort =  ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.METRICS_CACHE_MASTER_PORT).toString();
+    String metricsCacheStatsPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.METRICS_CACHE_STATS_PORT).toString();
+    String ckptmgrPort = ExecutorPorts.getRequired(
+        freePorts, ExecutorPorts.ExecutorPortNames.CHECKPOINT_MANAGER_PORT).toString();
 
     List<String> commands = new ArrayList<>();
     commands.add(createCommandArg(ExecutorFlag.TopologyName, topology.getName()));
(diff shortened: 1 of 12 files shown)


jerrypeng:
@cckellogg @srkukarni Can you take a look

srkukarni:
@huijunw Can you check if this covers for aurora as well

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+        List<String> names = new LinkedList<>();
+        for (ExecutorPortNames executorPortNames : ExecutorPortNames.values()) {
+          names.add(executorPortNames.toString());
+        }
+
+        return (String[]) names.toArray();
+      }
+    }
+
+    public static String[] getRequiredPorts() {
```
ashvina: Is this method duplicating the responsibility of `getPortNames()`

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+      METRICS_CACHE_STATS_PORT("metrics-cache-s"),
+      CHECKPOINT_MANAGER_PORT("ckptmgr");
+
+      private final String name;
+
+      ExecutorPortNames(String name) {
+        this.name = name;
+      }
+
+      public String toString() {
```
ashvina: Instead of overloading `toString`, I think it would be better to add a `getter` method

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+  public static final int PORTS_REQUIRED_FOR_EXECUTOR = ExecutorPorts.getRequiredNumOfPorts();
   public static final int PORTS_REQUIRED_FOR_SCHEDULER = 1;
   public static final String SCHEDULER_COMMAND_LINE_PROPERTIES_OVERRIDE_OPTION = "P";
 
   private static final Logger LOG = Logger.getLogger(SchedulerUtils.class.getName());
 
+  /**
+   * Enum that defines the type of ports that an heron executor needs
+   */
+  public static class ExecutorPorts extends HashMap<ExecutorPorts.ExecutorPortNames, Object> {
```
ashvina: Why is `Object` used here instead of `Integer`?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+          ExecutorPortNames.METRICS_CACHE_STATS_PORT.toString(),
+          ExecutorPortNames.CHECKPOINT_MANAGER_PORT.toString()
+      };
+    }
+
+    public static int getRequiredNumOfPorts() {
+      return getRequiredPorts().length;
+    }
+
+    public static ExecutorPorts withRequiredPorts(List<Integer> portsToUse) {
```
ashvina: I am not sure if I understand the need of a static method and a non-static method for the same function?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+      ExecutorPorts executorPorts = new ExecutorPorts();
+      executorPorts.setRequiredPorts(portsToUse);
+      return executorPorts;
+    }
+
+    public void setRequiredPorts(List<Integer> portsToUse) {
+      if (portsToUse.size() < getRequiredNumOfPorts()) {
```
ashvina: Should we check for extra ports provided? i.e. `portsToUse == requiredPorts`

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+        throw new IllegalArgumentException("Did not provide enough Ports. "
+            + "Number of ports provided: " + portsToUse.size()
+            + ". Number of ports required: " + getRequiredNumOfPorts());
+      }
+
+      for (int i = 0; i < portsToUse.size(); i++) {
```
ashvina: iterator or lambda would be better here

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+        int port = portsToUse.get(i);
+        if (port == -1) {
```
ashvina: Any negative value is invalid

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: I am also unsure what is the significance of extending a `HashMap` here. Seems unnecessary here.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
+    public static Object getRequired(ExecutorPorts ports, ExecutorPortNames type) {
+      Object ret = getOptional(ports, type);
+      if (ret == null) {
+        throw new IllegalArgumentException("Required executor port "
+            + type.name() + " is not found");
+      }
+      return ports.get(type);
+    }
+
+    public static Object getOptional(ExecutorPorts ports, ExecutorPortNames type) {
```
ashvina: The method name and the return type do not match. Moreover could you please explain the use of this method.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: This is done to be generic so that some options in the future like ports for remote debugging of the jvm can be a list of integers as well:

https://github.com/twitter/heron/pull/2484

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: Currently it contains all the defined port types, but this is not going to be the case in the future.  For example, the ports for remote debugging as part of https://github.com/twitter/heron/pull/2484 is optional and with not be part of the required.  And currently we pass in all the ports that any possible executor will need which is also unnecessary.  For instance only container-0 need the ports related to the tmaster.  That cleanup can be done in a follow up PR.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: That is what the existing implementation implies.  SysUtils.getFreePort will return a -1 if it cannot find a free port.  We can do the check here or when we call SysUtils.getFreePort in the respective schedulers.  I am ok with either way

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: Sorry i don't quite follow how I can use an iterator or lambda here.  I am iterating portsToUse and ExecutorPortNames.values() at the same time.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: I think its more nature to override toString.  If I was to print ExecutorPortNames, I can just do 

System.out.println(ExecutorPortNames.MASTER_PORT) instead of something like System.out.println(ExecutorPortNames.MASTER_PORT.getName())

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: We can, but I am not sure it is the place to check for something like that since currently it will work fine even if there are additional ports but not if there are not enough ports.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: This method is just used for getting ports that are not required thus I don't throw an exception if I don't find the port in the map unlike getRequired

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: This method is just so that in the scheduler code I don't need to:

ExecutorPorts executorPorts = new ExecutorPorts();
executorPorts.setRequiredPorts(portsToUse);

instead I can just :
ExecutorPorts.withRequiredPorts(portsToUse)

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: its simpler because if I didn't I would have to have a class variable that is a map which will hold the contents of the portname to port mappings.  I would also need to write wrappers such as "put" to expose the those methods to users or make the variable public.  It the same reason why heron Config extends HashMap as well

jerrypeng:
@ashvina The reason for this PR is to move away of passing in ports for executors like positional arguments which make it very difficult to read and hard to pass in optional ports.  Also in a lot of places in the various schedulers, there exists a list of port names and then another list of ports.  Why not just have one object that is a map that has the info for both and gives you the flexibility of having optional arguments.  I don't think this solution is that complex.  Can you share your thoughts on why this solution is too complex?

maosongfu:
@huijunwu @objmagic Can we check will it break AuroraScheduler?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: I am looking at #2484. It seems remote debugging ports are required by the scripts and proto files. If that is correct, are you planning to change that? 

It also seems remote debugging ports will get promoted to `required` if remote debugging is enabled?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: I will not come to know that `toString` is returning the id of the port without looking at the code. One might also assume `System.out.println(ExecutorPortNames.MASTER_PORT)` prints port name and value too.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: Based on the current code `put` is the only method that would need to be exposed. As of now other methods from HashMap are not used anywhere. This class is not altering the behavior of the base class either. In which case I would prefer `composition` over `inheritance`

ashvina: I think this added flexibility makes use of this class complex, requiring casting to the correct type. The object could be anything, a `String`, a `Topology`. The user will need knowledge of all references before any modification can be made.

ashvina:
@jerrypeng Thanks for your comments. 
First, I agree that the way ports are currently managed is error prone. Thanks for taking this up.
However, this PR is addressing some future needs and I don't understand them well. Personally I find this solution overfitting. I think it would be better for one more reviewer to take a look.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: I thought `Optional` referred to `java.util.Optional`.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
ashvina: `static` methods are harder to test. On that note, there are no test cases in the PR.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: remote debugging ports will only be need if the user sets a flag in there topology.  This will be an optional port .  The scheduler will decide how to get these ports and pass them to SchedulerUtils.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
billonahill: I agree that ports should be Integers. Using `Object` for flexibility is almost always an anti-pattern. I would suggest using integer and if we find later that there's a requirement for multiple port bound to a single name (which seems like something to avoid IMO), then we would consider adding support for lists of integers.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +40,105 @@
...
 
   private static final Logger LOG = Logger.getLogger(SchedulerUtils.class.getName());
 
+  /**
+   * Enum that defines the type of ports that an heron executor needs
+   */
+  public static class ExecutorPorts extends HashMap<ExecutorPorts.ExecutorPortNames, Object> {
+    private static final long serialVersionUID = 3594142561871208879L;
+
+    public enum ExecutorPortNames {
```
billonahill: This enum of port names provides little more than a list of strings. I suggest instead having an enum of ports (not names) and include metadata of each, like whether it's required or not and it's default value. Then enum itself could then have a getter for fetching just the required ones for example.

One benefit of the proposed approach is that to add a new port you have to modify the enum and explicitly set whether it's required or not when doing it. As written someone needs to add a port name and know that there's also another place in code to modify to represent required or not.

Also enum names should be singular not plural. That way you work with a singular enum (e.g., `doSomething(ExecutorPortName name)`) or the with all of them (e.g. `ExecutorPortName[] all = ExecutorPortName.values()`).

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
billonahill: Please do not use the value of `toString()` for any business logic. The `toString` value should only ever be used to log messages **and nothing else**. Overriding `toString()` for logic is an anti-pattern. Doing so violates encapsulation and provides and unclear API to consumers. Also future developers could find that the logging of the object is unclear (e.g., logging just an int for a Port enum might not be desirable) and change it. This would break functionality. I have been bit by this and have spent hours debugging such design flaws. Please don't do this.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
billonahill: If we have an `ExecutorPort` enum which includes whether it's required or optional, then the caller doesn't need to know which of the two methods to call here. They can just call `getPort(ExecutorPort port)` and the method can return `Integer` or throw if a required port isn't found.

jerrypeng:
@billonahill @ashvina thanks for the code review and suggestions.  I have refactored the code.  Can you guys have another look.  Thanks!

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -127,29 +177,14 @@ private SchedulerUtils() {
    * @param config The static Config
    * @param runtime The runtime Config
    * @param containerIndex the executor/container index
-   * @param freePorts list of free ports
+   * @param ports list of free ports
```
ashvina: The java doc needs to be updated for `ports`

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -167,7 +202,7 @@ private SchedulerUtils() {
       Config config,
       Config runtime,
       int containerIndex,
-      List<String> ports) {
+      Map<ExecutorPort, String> ports) {
```
ashvina: Please update the javadoc for `ports`. 
Also one of these two methods can be deleted, `getExecutorCommand` and `executorCommand`. I am not sure if the intent is to consume `<String, Integer>` type map in one of the methods.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
```diff
@@ -417,11 +418,13 @@ protected ObjectNode getContainerSpec(ObjectMapper mapper,
   protected ArrayNode getPorts(ObjectMapper mapper) {
     ArrayNode ports = mapper.createArrayNode();
 
-    for (int i = 0; i < KubernetesConstants.PORT_NAMES.length; i++) {
+    for (Map.Entry<ExecutorPort, String> entry
```
ashvina: Converting `KubernetesConstants.EXECUTOR_PORTS` to a `<String, Integer>` map would reduce one unnecessary `Integer.parseInt` below.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -39,12 +39,62 @@
...
+    public static String getPort(ExecutorPort executorPort,
+                                  Map<ExecutorPort, String> portMap) {
+      if (!portMap.containsKey(executorPort) && executorPort.isRequired()) {
+        throw new RuntimeException("Required port " + executorPort.getName() + " not provided");
+      }
+
+      return portMap.get(executorPort);
+    }
+
+    public static int getRequiredNumOfPorts() {
```
ashvina: In the current approach each scheduler seeks for a specific port and adds its value. The scheduler will not fail If a new `required` port is added and the scheduler is not updated. I have a suggestion. Lets add a method to return all required ports, `Set<ExecutorPort> getRequiredPorts`. All schedulers would iterate on this set to provide the required values. If a new port is added, the scheduler code will fail with an error.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesConstants.java
```diff
@@ -74,10 +74,18 @@ private KubernetesConstants() {
   public static final String METRICS_CACHE_STATS_PORT = "6008";
   public static final String CHECKPOINT_MGR_PORT = "6009";
 
-  public static final String[] PORT_LIST = new String[]{
-      MASTER_PORT, TMASTER_CONTROLLER_PORT, TMASTER_STATS_PORT,
-      SHELL_PORT, METRICSMGR_PORT, SCHEDULER_PORT, METRICS_CACHE_MASTER_PORT,
-      METRICS_CACHE_STATS_PORT, CHECKPOINT_MGR_PORT };
+  public static final Map<ExecutorPort, String> EXECUTOR_PORTS = new HashMap<>();
+  static {
+    EXECUTOR_PORTS.put(ExecutorPort.MASTER_PORT, MASTER_PORT);
```
ashvina: I suggested a method `getRequiredPorts` in one of the comments. Please see if using that would be better. If a required port is not `put` into the `EXECUTOR_PORTS` throw an error.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesScheduler.java
jerrypeng: I am not sure this is a good idea.  For some of the schedulers, e.g. Marathon Scheduler, the ports are actually strings, e.g. $PORT1.  Thus, its probably simpler if we just keep everything as a String.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: In the current implementation, a exception will be thrown if the scheduler doesn't set all of the required ports.  The method "executorCommandArgs" calls the "getPort" method in "ExecutorPort" to get all the ports.  If the port is required and missing a runtime exception will be thrown:

https://github.com/twitter/heron/pull/2516/files#diff-2f8f616b79cedacafb019678d260770fR81

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: will update

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: I will delete the executorCommand method.  It was my intention to pass the port map where the value is a string since some schedulers specify their ports as a string such as the Marathon Scheduler.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesConstants.java
jerrypeng: Please see me previous comment concerning the current implementation will already throw an exception if a required port is missing.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: If a new port required port is added here and retrieved in the "executorCommandArgs" method then an exception will be thrown.

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/KubernetesConstants.java
jerrypeng: I will replace getNumOfRequiredPorts with getRequiredPorts as it is a more generic method

jerrypeng:
@ashvina thanks for your review.  I have addressed your comments

PR #2519, opened 2017-11-03
Simplified Streamlet interface

srkukarni:
Removed KVStreamlet and consolidated all functionality into one Streamlet interface. We get reduceByKey/join in Streamlet by letting users specify a keyExtractor function.


srkukarni:
@jerrypeng

PR #2522, opened 2017-11-05
Allow users to register timer tasks in topologies

jerrypeng:
This is done sone that:
1. Users don't have to use additional threads in their topology to use a timers.
2. Collectors are not thread safe so use other threads to emit tuples will cause race conditions.  Using timer events part of the slave event loop with prevent that since its the same thread.
3. Tick tuples are cumbersome to use to trigger events and you can only have tick tuples emitted at one frequency.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -666,14 +666,14 @@ public class Config extends HashMap<String, Object> {
    * Registers a timer event that executes periodically
    * @param conf the map with the existing topology configs
    * @param name the name of the timer
-   * @param timerDuration the frequency in which to run the task
+   * @param interval the frequency in which to run the task
    * @param task the task to run
    */
   @SuppressWarnings("unchecked")
   public static void registerTopologyTimerEvents(Map<String, Object> conf,
-                                                 String name, Duration timerDuration,
+                                                 String name, Duration interval,
                                                  Runnable task) {
-    if (timerDuration.isZero() || timerDuration.isNegative()) {
+    if (interval.isZero() || interval.isNegative()) {
       throw new IllegalArgumentException("Timer duration needs to be positive");
     }
     if (!conf.containsKey(Config.TOPOLOGY_TIMER_EVENTS)) {
@@ -686,6 +686,6 @@ public class Config extends HashMap<String, Object> {
     if (timers.containsKey(name)) {
       throw new IllegalArgumentException("Timer with name " + name + " already exists");
     }
-    timers.put(name, Pair.of(timerDuration, task));
+    timers.put(name, Pair.of(interval, task));
   }
 }
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -34,6 +34,7 @@
 package com.twitter.heron.api;
 
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -265,6 +266,8 @@ public class Config extends HashMap<String, Object> {
    * This variable contains Map<String, String>
    */
   public static final String TOPOLOGY_ENVIRONMENT = "topology.environment";
+
+  public static final String TOPOLOGY_TIMER_EVENTS = "topology.timer.events";
 
   private static final long serialVersionUID = 2550967708478837032L;
   // We maintain a list of all user exposed vars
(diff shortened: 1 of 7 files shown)


jerrypeng:
@nwangtw to address your questions

**What are the typical use cases of this timer event? Is it for users or for internal cases?**

Mentioned in the description of the PR, the use cases are:

1 . Users don't have to use additional threads in their topology to use a timers.
2. Collectors are not thread safe so use other threads to emit tuples will cause race conditions. Using timer events part of the slave event loop with prevent that since its the same thread.
3. Tick tuples are cumbersome to use to trigger events and you can only have tick tuples emitted at one frequency.

This could be used by the user writing the topology or by a bolt we implement for the user such as the windowed bolt.

**Timer event is for topology? Will it be potentially in Spout/Bolt level?**

This could be used in a spout or bolt

**Is it possible that we may need an per-timer on/off feature in future?**

I don't see this as a valid use case

**Is there a limit of how many timers can we have and how long/short can it be? Suggest to check the duration to be at least positive.**

How many timers or the duration of timers should be up to the user.  The user should already understand the obvious performance consequences if there are too many timers or the timer frequency is too often.  We can check that the duration is positive though.  I will add that.

**tickTupleFreqMs becomes hidden. It might be better to be more explicit (like sharing TimerEvent code but having a standalone PrepareTickTupleTimer function).**

We can, but to be honest, this feature more or less deprecates tick tuples. Tick tuples are not very easy to use for triggering timer events.  This feature makes it much easier and more natural

**I am not familiar with this looper object and . Is it possible that tmerevent can cause**

Not sure what your question is

**needs unit tests.**

We can add some unit tests

jerrypeng:
@nwangtw thanks for the review.  I have addressed your comments

heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -652,4 +661,31 @@ public void setTopologyStatefulCheckpointIntervalSecs(int secs) {
...
+  /**
+   * Registers a timer event that executes periodically
+   * @param conf the map with the existing topology configs
+   * @param name the name of the timer
+   * @param timerDuration the frequency in which to run the task
+   * @param task the task to run
+   */
+  @SuppressWarnings("unchecked")
+  public static void registerTopologyTimerEvents(Map<String, Object> conf,
+                                                 String name, Duration timerDuration,
```
nwangtw: I am wondering if "interval" is more clear than "timerDuration". Open to other thoughts.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```diff
@@ -298,20 +300,16 @@ public void activate() {
   public void deactivate() {
   }
 
-  private void PrepareTickTupleTimer() {
```
nwangtw: need to be public?

heron/api/tests/java/com/twitter/heron/api/ConfigTest.java
```diff
@@ -0,0 +1,48 @@
+//  Copyright 2017 Twitter. All rights reserved.
```
nwangtw: Formatting looks different.

heron/api/tests/java/BUILD
```diff
@@ -34,7 +34,8 @@ java_tests(
     "com.twitter.heron.streamlet.impl.operators.ReduceByKeyAndWindowOperatorTest",
     "com.twitter.heron.streamlet.impl.operators.GeneralReduceByWindowOperatorTest",
     "com.twitter.heron.streamlet.impl.operators.GeneralReduceByKeyAndWindowOperatorTest",
-    "com.twitter.heron.api.HeronSubmitterTest"
+    "com.twitter.heron.api.HeronSubmitterTest",
+    "com.twitter.heron.api.ConfigTest"
```
nwangtw: Sort

nwangtw:
Thanks for the responses. Never mind about the "tmerevent can cause" question. I was wondering about racing condition/deadlock at beginning but it didn't look risky later.

heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
jerrypeng: nope I will change it back

heron/api/src/java/com/twitter/heron/api/Config.java
jerrypeng: I will change the name

jerrypeng:
@nwangtw I have addressed your comments

Issue #2523, opened 2017-11-06
heronstreaming.io out of date wrt compilation instructions


srkukarni:
CPP bindings introduced cereal library which needs cmake-2.2 which is not the default version of cmake that comes with ubuntu. Thus to compile latest heron, we need to update the instructions to upgrade the cmake.

Code0x58:
keeping these up to date feels a bit redundant. `./bazel_configure.py` seems to be a good tools if you want to compile anywhere supported, otherwise there is the vagrant setup, or Dockerfiles - it may be good to refer to those to keep things DRY - as people will have a copy of the codebase anyway if compiling.

joshfischer1108:
Agreed. Thanks @Code0x58 .

PR #2524, opened 2017-11-07
New cmake installation instructions

lucperkins:
Addressing #2523

### DIFF
--- a/website/content/docs/developers/compiling/linux.md
+++ b/website/content/docs/developers/compiling/linux.md
@@ -21,8 +21,20 @@ $ sudo apt-get upgrade -y
 ### Step 2 --- Install required libraries
 
 ```bash
-$ sudo apt-get install git build-essential automake cmake libtool-bin zip \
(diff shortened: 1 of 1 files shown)


website/content/docs/developers/compiling/linux.md
```diff
@@ -21,10 +21,22 @@ $ sudo apt-get upgrade -y
 ### Step 2 --- Install required libraries
 
 ```bash
-$ sudo apt-get install git build-essential automake cmake libtool-bin zip \
+$ sudo apt-get install git build-essential automake libtool-bin zip \
   libunwind-setjmp0-dev zlib1g-dev unzip pkg-config python-setuptools -y
 ```
 
+In addition, you'll also need to install [cmake](https://cmake.org/) version
+2.2.3 (this is *not* the default version installed by `apt-get`):
```
srkukarni: actually the better way to do this would be to add edison backports and just do apt-get install.

kramasamy:
@lucperkins - no need this for anymore since cmake dependency has been removed.

PR #2527, opened 2017-11-08
Apply config translation for Bolt and Spout

nwangtw:
Bolt and Spout configs need to be translated first before assigning to Heron components.

### DIFF
--- a/storm-compatibility/src/java/backtype/storm/topology/BoltDeclarerImpl.java
+++ b/storm-compatibility/src/java/backtype/storm/topology/BoltDeclarerImpl.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import backtype.storm.grouping.CustomStreamGrouping;
 import backtype.storm.grouping.CustomStreamGroupingDelegate;
 import backtype.storm.tuple.Fields;
+import backtype.storm.utils.ConfigUtils;
 import backtype.storm.utils.Utils;
 
 public class BoltDeclarerImpl implements BoltDeclarer {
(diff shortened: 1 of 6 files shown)


storm-compatibility/src/java/backtype/storm/topology/BoltDeclarerImpl.java
```diff
@@ -35,13 +35,18 @@ public BoltDeclarerImpl(com.twitter.heron.api.topology.BoltDeclarer delegate) {
   @Override
   @SuppressWarnings({"rawtypes", "unchecked"})
   public BoltDeclarer addConfigurations(Map conf) {
-    delegate.addConfigurations((Map<String, Object>) conf);
+    // Translate config to heron config and then apply.
```
srkukarni: Currently the translation occurs at the end of the step, in StormSubmitter. Any reason  not to do it there? That way there will be just one centralized place of doing it

nwangtw: The translation in StormSubmitter processes topology level configs. Components have their own configs and component level configs are merged later when Spouts and Bolts are created.


Without this change, the raw Storm configs for Spouts and Bolts are passed into Heron layer. As the result, TOPOLOGY_TICK_TUPLE_FREQ_MS is missing in Bolt level because it is generated in the translation by Config.setTickTupleFrequency().

nwangtw: Doing all the translation in StormSubmitter.submitTopology() is one option, but it would require iterating all components and replacing their configs(so there will be heron configs in storm objects, which is not ideal). Overall, doing translation in *DeclarerImpl seems to be cleaner to us.

maosongfu:
👍  LGTM

PR #2531, opened 2017-11-09
Explicity setting heron apiserver to run as cluster-admin

jerrypeng:
Since RBAC is enabled by default in Kubernetes 1.8

### DIFF
--- a/deploy/kubernetes/general/apiserver.yaml
+++ b/deploy/kubernetes/general/apiserver.yaml
@@ -48,7 +48,7 @@ spec:
       serviceAccountName: heron-apiserver
       containers:
         - name: heron-apiserver
-          image: heron/heron:0.16.2
+          image: heron/heron:latest
           command: ["sh", "-c"]
           args:
             - >-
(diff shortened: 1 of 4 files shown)


jerrypeng:
@cckellogg can you test this out. Thanks

deploy/kubernetes/general/apiserver.yaml
```diff
@@ -1,17 +1,51 @@
...
   replicas: 1
   template:
     metadata:
       labels:
         app: heron-apiserver
     spec:
+      serviceAccountName: heron-apiserver
       containers:
         - name: heron-apiserver
           image: heron/heron:0.16.2
```
cckellogg: Can we update the image here too heron/heron:latest?

cckellogg:
this file needs to be updated too (https://github.com/twitter/heron/blob/master/deploy/kubernetes/gcp/bookkeeper-apiserver.yaml)

deploy/kubernetes/general/apiserver.yaml
jerrypeng: sure

jerrypeng:
@cckellogg thanks for the review.  I have addressed your comments

PR #2534, opened 2017-11-10
Change heron api dependency

aahmed-se:
This change separates the core low level ui(the ui that the instance targets) and the high level streamlet interface. The aim of this is to ensure that dependent librarires(like heron-storm) only deal with the core api and the higher level streamlet api can morph pretty rapidly without worrying compatibility issues.

### DIFF
--- a/examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
+++ b/examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
@@ -23,7 +23,6 @@ import java.nio.file.StandardOpenOption;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.logging.Logger;
 
-import com.twitter.heron.api.utils.Utils;
 import com.twitter.heron.examples.streamlet.utils.StreamletUtils;
 import com.twitter.heron.streamlet.Builder;
 import com.twitter.heron.streamlet.Config;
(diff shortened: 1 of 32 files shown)


heron/api/src/java/BUILD
```diff
@@ -20,13 +20,22 @@ api_deps_files =  \
         "//heron/common/src/java:basics-java",
     ]
 
+# Legacy Low Level Api
```
nwangtw: why "Legacy"? My understanding is that this API should be suggested in some cases. Same for the functional comment. It is not "External".

examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
```diff
@@ -115,7 +114,7 @@ public static void main(String[] args) throws Exception {
         .newSource(() -> {
           // This applies a "brake" that makes the processing graph write
           // to the temporary file at a reasonable, readable pace.
-          Utils.sleep(500);
```
nwangtw: Not sure if this is necessary.

1. StreamletUtils is a class in examples directory. I feel it is reasonable for heron to provide a general function like sleep.
2. Utils looks like a low level object, correct? Why high level classes should avoid using it?

srkukarni: I think one of attempts of these examples is to show that one can use just the high level api without relying on anything from the low level api.

heron/api/src/java/BUILD
aahmed-se: fixed

examples/src/java/com/twitter/heron/examples/streamlet/FilesystemSinkTopology.java
nwangtw: Make sense. Thanks.

PR #2535, opened 2017-11-10
Since Heron does not use kryo for anything, there is no need to shade it out

srkukarni:


kramasamy:
if it is not used - should we remove the library from the jar? If the application programs link against this with different version of kryo then you might have a clash.

srkukarni:
It is not included in the instance jar either. This snippet of code seems to be a leftover from prev where kryo was part of instance/api.

huijunw:
good to know that Heron does not use kryo

PR #2538, opened 2017-11-10
Use builder pattern in Streamlet API

lucperkins:
At the moment, there are two classes in the Streamlet API (`Config` and `Resources`) that could benefit from a more user-friendly builder pattern (I've also added constructors that explicitly use the defaults). I've updated these classes to use this pattern as well as existing Streamlet API examples. Here's an example:

```java
Resources resources = Resources.defaultResources();

Config config = new Config.Builder()
        .setNumContainers(3)
        .setContainerResources(resources)
        .build();
```


lucperkins:
@srkukarni Awesome! In the future, I think we should strive to offer this pattern for pretty much all user-facing constructors.

PR #2540, opened 2017-11-11
Introduce Kryo Serializer to Streamlet

srkukarni:
Currently Streamlet relies only on Java serialization which has very bad performance. This pr allows topologies to use Kryo serialization.


srkukarni:
@jerrypeng

heron/api/src/java/com/twitter/heron/streamlet/KeyValue.java
```diff
@@ -31,6 +31,10 @@
     return new KeyValue<R, T>(k, v);
   }
 
+  public KeyValue() {
+    // nothing really
```
nwangtw: what is this dummy ctor for? Is it better to have mis-usage protection if this is required by kryo?

heron/api/src/java/com/twitter/heron/streamlet/Config.java
```diff
@@ -44,7 +45,7 @@ public static Config defaultConfig() {
         .build();
   }
 
-  com.twitter.heron.api.Config getHeronConfig() {
+  public com.twitter.heron.api.Config getHeronConfig() {
```
nwangtw: Is this change necessary?

heron/api/src/java/com/twitter/heron/streamlet/Config.java
```diff
@@ -107,6 +108,19 @@ public Builder setUserConfig(String key, Object value) {
...
 
+    /**
+     * Sets the topology to use the Kryo serializer for serializing
+     * streamlet elements
+     */
+    public Builder useKryoSerializer() {
+      try {
+        config.setSerializationClassName(new KryoSerializer().getClass().getName());
+      } catch (NoClassDefFoundError e) {
+        throw new RuntimeException("Please link with kryo");
```
nwangtw: Suggest to be more explicit in error message, like "Linking with kryo is needed because useKryoSerializer is used".

heron/api/src/java/com/twitter/heron/streamlet/Config.java
srkukarni: Not needed. Reverted

heron/api/src/java/com/twitter/heron/streamlet/Config.java
srkukarni: Changed

heron/api/src/java/com/twitter/heron/streamlet/KeyValue.java
srkukarni: This is primarily for Kryo. Made it non public

PR #2546, opened 2017-11-13
remove yaml cpp binary and use bazel native rules

kramasamy:
Purpose of this pull request
Make the compilation faster for yaml cpp

How would this pull request resolve the issue
This pull request untars the binary tar ball into actual files and directly uses bazel to compile. Hence yaml cpp does not depend on make/automake/cmake/autoconfig etc. Since bazel uses parallelism underneath, it speeds up compilation.

What would be when the pull request is merged: (Issues resolved? What's the expected results?)
The expected result is in the increase faster compilation. This is a part of the broader effort to remove tar balls out of the code base. If Twitter were to mirror selective github repos, we can further simplify the PR using new_git_repository rules. It can be done in a later PR once repos are setup.


srkukarni:
I didn't understand the purpose of this change. Why is it better?

kramasamy:
It is incremental benefit as we move towards removing tar balls out of the source code base
yaml cpp does not make use make/cmake/autoconf/automake tools to setup and compile. Instead, it directly uses bazel - because of that bazel can parallelize the compilation of the yaml cpp files - thus reducing the compilation time. While it might be small in yaml-cpp, as we convert other projects it will get better and better.

maosongfu:
@kramasamy 
Can we have following parts in the description?
Current issue:
The purpose of this pull request:
How would this pull request resolve the issue:
What would be when the pull request is merged: (Issues resolved? What's the expected results?)

kramasamy:
Purpose of this pull request
- Make the compilation faster for yaml cpp

How would this pull request resolve the issue
- This pull request untars the binary tar ball into actual files and directly uses bazel to compile. Hence yaml cpp does not depend on make/automake/cmake/autoconfig etc. Since bazel uses parallelism underneath, it speeds up compilation

What would be when the pull request is merged: (Issues resolved? What's the expected results?)
- The expected result is in the increase faster compilation. This is a part of the broader effort to remove tar balls out of the code base. If Twitter were to mirror selective github repos, we can further simplify the PR using new_git_repository rules. It can be done in a later PR once repos are setup.

nicknezis:
Closing this pull request. If you believe the PR is still needed, please leave a comment.

PR #2547, opened 2017-11-13
Refactoring windowing

jerrypeng:
- change to using timers rather than tick tuples
- removing unneccessary locks


jerrypeng:
@srkukarni

Issue #2530, opened 2017-11-08
Current heron print_usage does not match the actual usages 


maosongfu:
In https://github.com/twitter/heron/pull/2471,the command-line arguments were changed to be passed in a key-value fashion.
However, the print_usage does not reflect this change:
https://github.com/twitter/heron/blob/master/heron/executor/src/python/heron_executor.py#L47

The `print_usage()` should be updated as well.

maosongfu:
@jerrypeng @srkukarni

nwangtw:
seems completed.

PR #2548, opened 2017-11-14
Correct the print_usage functionality in Heron Executor

srkukarni:
#2530


maosongfu:
The change looks good but why the travis-ci fails continuously?

srkukarni:
unrelated to this change and its been failing for all prs now.

PR #2549, opened 2017-11-14
Make Kryo the default serializer for the Streamlet API

lucperkins:
This PR makes Kryo the default serializer when using the Streamlet API. You can select the Java serializer like this:

```java
Config config = new Config.Builder()
    .setSerializer(Config.Serializer.JAVA)
    .build();
```


heron/api/tests/java/com/twitter/heron/streamlet/impl/StreamletImplTest.java
```diff
@@ -23,6 +23,7 @@
 import org.junit.Test;
 
 import com.twitter.heron.api.topology.TopologyBuilder;
+import com.twitter.heron.streamlet.Config;
```
nwangtw: seems not necessary?

heron/api/src/java/com/twitter/heron/streamlet/Config.java
```diff
@@ -65,8 +65,10 @@ public static Config defaultConfig() {
 
   public static class Builder {
     private com.twitter.heron.api.Config config;
+    private boolean useKryo;
```
nwangtw: Hmm. Do we really want to make Kryo as default?

Also I don't like this new boolean. If in future we need/have a new serializer, this code could become messy. In fact it already has issues If user calls: builder. useKryoSerializer(false). useKryoSerializer().

I am wondering if an enum could be cleaner here.

lucperkins: I conferred with @srkukarni on this and he agreed that the performance improvements associated with using Kryo are significant enough (with no apparent drawbacks) that Kryo should be the default. As for an enum, that could make things trickier in the future when enabling people to use any serializer they'd like.

lucperkins:
@nwangtw I took your advice and changed the serializer into an enum. Kryo is still set as the default. I'll continue to confer with project leads on that.

heron/api/src/java/com/twitter/heron/streamlet/Config.java
```diff
@@ -109,19 +148,24 @@ public Builder setUserConfig(String key, Object value) {
...
      * streamlet elements
+     * @param topologySerializer The serializer to be used in this topology
      */
-    public Builder useKryoSerializer() {
-      try {
-        config.setSerializationClassName(new KryoSerializer().getClass().getName());
-      } catch (NoClassDefFoundError e) {
-        throw new RuntimeException("Linking with kryo is needed because useKryoSerializer is used");
+    public Builder setSerializer(Serializer topologySerializer) {
+      serializer = topologySerializer;
```
nwangtw: Suggest to have two stages.

setSerializer() function set the variable. And have another applySerializer() to be called in build() function.

heron/api/src/java/com/twitter/heron/streamlet/Config.java
```diff
@@ -109,19 +148,24 @@ public Builder setUserConfig(String key, Object value) {
...
+      if (serializer == Serializer.KRYO) {
+        try {
+          config.setSerializationClassName(new KryoSerializer().getClass().getName());
+        } catch (NoClassDefFoundError e) {
+          throw new RuntimeException("Linking with kryo is needed because setSerializer is used. " +
+              "You may be attempting to set the serializer more than once.");
```
nwangtw: Not sure if this error message is accurate or not.

heron/api/src/java/com/twitter/heron/streamlet/Config.java
lucperkins: Good idea. Done.

heron/api/src/java/com/twitter/heron/streamlet/Config.java
lucperkins: You’re right. I was misidentifying the problem. Using KryoSerializer.class instead of instantiating a KryoSerializer seems to have fixed the issue. You can still set the serializer more than once but again, I believe this is an issue in all builder pattern classes.

PR #2551, opened 2017-11-14
Move topology config translations into a different function

nwangtw:
To avoid component configs overwriting topology configs.

The root cause is that currently the translation function generate configs when a key does NOT exist.


srkukarni:
Sorry, but I'm having a hard time understanding the rationale behind this. Could you please detail more?

nwangtw:
@srkukarni 

So we added config translation for spout and bolt because it was missing last week. When spouts and bolts are created they merge topology configs and component configs.

say if topology config has {
k1: 1
k2: 2
ke: 3
}
and bolt config has {
k1: 10
k4: 40
}
The final config for the bolt would be {
k1: 10
k2: 2
ke: 3
k4: 40
}

This looks right as long as all they key/value pairs are expected.

I missed on case in the doStormTranslation in my previous PR.

Most part in the function are like this and they are safe:
if (config.contains(storm_key)) {
  config.put(heron_key, value)
}

However in the function, one config is different from others: TOPOLOGY_ACKER_EXECUTORS

The code is like:
if (config.contains(storm.TOPOLOGY_ACKER_EXECUTORS)) {
  config.put(heron. TopologyReliabilityMode, value1)
} else {
  config.put(heron. TopologyReliabilityMode, value2)  // This is dangerous.
}

Two cases:
- when the bolt config does have "TOPOLOGY_ACKER_EXECUTORS" key, the output config would have a heron. TopologyReliabilityMode key with value1. This is ok.
- when the bolt config DOESN'T have "TOPOLOGY_ACKER_EXECUTORS" key, the output config would still have a heron.TOPOLOGY_ACKER_EXECUTORS key, with value2. And when the config is merged with topology config, the heron. TopologyReliabilityMode will overwrite the topology config and this is bad.

This PR moves the "if ... else ..." block out of the shared translation function to a topology only translation function, so that the config merging is safe.

nwangtw:
sure thing!

PR #2553, opened 2017-11-15
fix bugs in AuroraHeronShellController

huijunw:
Fixed 2 bugs in AuroraHeronShellController
1. stmgr index error
2. delegate to aurora client when heron-shell/kill fails

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
@@ -102,6 +102,10 @@ class AuroraHeronShellController implements AuroraController {
     try {
       NetworkUtils.sendHttpPostRequest(con, "X", payload.getBytes());
       return NetworkUtils.checkHttpResponseCode(con, 200);
+    } catch (ConnectException e) {
+      // if heron-shell command fails, delegate to aurora client
+      LOG.info("heron-shell killexecutor failed; try aurora client ..");
+      cliController.restart(containerId);
     } finally {
       con.disconnect();
     }
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
@@ -82,10 +82,18 @@ class AuroraHeronShellController implements AuroraController {
       return false;
     }
 
-    int index = containerId - 1; // stmgr container starts from 1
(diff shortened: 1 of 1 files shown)


maosongfu:
@nwangtw could you please review this?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
```diff
@@ -74,26 +74,38 @@ public boolean killJob() {
...
+    String host = "x";
+    int port = -1;
+    int offset = "stmgr-".length();
+    for (StMgr sm : stateMgrAdaptor.getPhysicalPlan(topologyName).getStmgrsList()) {
+      Integer id = Integer.valueOf(sm.getId().substring(offset));
+      LOG.info("comparing container " + offset + " " + sm.getId() + " with " + containerId);
+      if (containerId.equals(id)) {
+        host = sm.getHostName();
+        port = sm.getShellPort();
+        break;
```
nwangtw: What if the id is not found? It would be better to have protective code instead of creating a connection to "http://x:-1/killexecutor".

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
```diff
@@ -74,26 +74,38 @@ public boolean killJob() {
...
       return false;
     }
 
-    int index = containerId - 1; // stmgr container starts from 1
-    StMgr contaienrInfo = stateMgrAdaptor.getPhysicalPlan(topologyName).getStmgrs(index);
-    String host = contaienrInfo.getHostName();
-    int port = contaienrInfo.getShellPort();
+    String host = "x";
+    int port = -1;
+    int offset = "stmgr-".length();
```
nwangtw: Suggest to move this logic into a function and compare the names in the reversed way:

StMgr searchContainer(Integer id) {
  prefix = "stmgr-" + id;
  for (StMgr sm : stateMgrAdaptor.getPhysicalPlan(topologyName).getStmgrsList()) {
    if (sm.getId().startsWith(prefix)) return sm;
  }
  return null;
}

It is more readable and easier to test the function.

huijunw: done

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraHeronShellController.java
huijunw: done

PR #2554, opened 2017-11-15
Retry when failing to kill an Aurora topology

maosongfu:
A lot of Twitter internal topology owners complained heron failing to kill an aurora topology because
the Aurora service can be unavailable or unstable for a while, or it can sometimes provide a false negative response.

This pull request adds retry mechanism to kill an Aurora topology with more confidence. It can increase the robustness of AuroraScheduler regardless of the Aurora Service People can also specify the retry attempts and interval via config.

This pull request also re-format some coding-styles for `RuntimeManagerRunner.java`.

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
@@ -22,6 +22,9 @@ import com.twitter.heron.spi.common.Context;
 public final class AuroraContext extends Context {
   public static final String JOB_LINK_TEMPLATE = "heron.scheduler.job.link.template";
   public static final String JOB_TEMPLATE = "heron.scheduler.job.template";
+  public static final String JOB_MAX_KILL_ATTEMPTS = "heron.scheduler.job.max.kill.attempts";
+  public static final String JOB_KILL_RETRY_INTERVAL_MS =
+      "heron.scheduler.job.kill.retry.interval.ms";
 
   private AuroraContext() {
   }
@@ -33,5 +36,13 @@ public final class AuroraContext extends Context {
   public static String getHeronAuroraPath(Config config) {
     return config.getStringValue(JOB_TEMPLATE,
         new File(Context.heronConf(config), "heron.aurora").getPath());
+  }
+
+  public static int getJobMaxKillAttempts(Config config) {
+    return config.getIntegerValue(JOB_MAX_KILL_ATTEMPTS, 5);
+  }
+
+  public static long getJobKillRetryIntervalMs(Config config) {
+    return config.getLongValue(JOB_KILL_RETRY_INTERVAL_MS, 2000);
   }
 }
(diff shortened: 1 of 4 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
```diff
@@ -34,4 +37,12 @@ public static String getHeronAuroraPath(Config config) {
     return config.getStringValue(JOB_TEMPLATE,
         new File(Context.heronConf(config), "heron.aurora").getPath());
   }
+
+  public static int getJobMaxKillAttempts(Config config) {
+    return config.getIntegerValue(JOB_MAX_KILL_ATTEMPTS, 5);
```
huijunw: do we put them in yaml?

maosongfu: If people want to override the value, they would put it in the yaml file. 
Currently, we just put `5` as the default value and use it.

huijunw: i would suggest put an example config in yaml, otherwise how can user know there is a config like this in the code ?

nwangtw: true. better to have an example.

maosongfu: Added. Thanks.

maosongfu:
Tested with actual Twitter distributed env.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraContext.java
kramasamy: This value is hard coded, probably not a good idea. Since you already have the yaml config, you can use this value - whatever this value is. Since most of the time this value is not changed, it automatically becomes the default value. 

Also the logic seems to be incorrect. If you are allowing for overrides, the overridden value should be used - but is that there somewhere?

maosongfu: - This value is hard coded, probably not a good idea.
We need to handle backward compatibility; this is why we have this value in source code. Different places would use their own configs, not necessarily sync up with OSS Heron repo. This implementation can prevent bringing down running Heron service.

- Since you already have the yaml config, you can use this value - whatever this value is. Since most of the time this value is not changed, it automatically becomes the default value.
We need to handle backward compatibility. This implementation also provides the flexibility about what you mentioned.

- If you are allowing for overrides, the overridden value should be used - but is that there somewhere?
`config.getIntegerValue(JOB_MAX_KILL_ATTEMPTS, 5);` means we would try to loop up the value for the key `JOB_MAX_KILL_ATTEMPTS` in the config. If people supply value for this key heron will use it, otherwise, it will use the default value 5. This implementation allows overrides, and the sample aurora/scheduler.yaml does the overrides.

PR #2555, opened 2017-11-15
add SCHEDULER_IS_SERVICE config for zk tool

huijunw:
Fix null exception for zk tool

```
==> State Manager root path: /storm/heron/states
Exception in thread "main" java.lang.NullPointerException
at com.twitter.heron.statemgr.zookeeper.curator.CuratorStateManager.initialize(CuratorStateManager.java:73)
at com.twitter.heron.statemgr.FileSystemStateManager.doMain(FileSystemStateManager.java:288)
at com.twitter.heron.statemgr.zookeeper.curator.CuratorStateManager.main(CuratorStateManager.java:431)
```


maosongfu:
We can also add a default value here:
`https://github.com/twitter/heron/blob/8ce680d14a6fc6070f7e0d635cb933058d3fce5d/heron/spi/src/java/com/twitter/heron/spi/common/Context.java#L109`,
like: 
`cfg.getBooleanValue(Key.SCHEDULER_IS_SERVICE, false);`

huijunw:
@maosongfu i would prefer setting the config in tool main. it is hard to say the default is false rather than true. for auroa, it is false. for local, it is true. both aurora and local are often seen scenarios.

PR #2559, opened 2017-11-15
Standardize builder interfaces and remove Resources class

lucperkins:
We currently have two builder patterns active:

```java
// Pattern 1
T t = new T.Builder()
    .build();

// Pattern 2
T t = T.newBuilder()
    .build();
```

This PR standardizes on pattern 2 for the `Config` and `Resources` classes in the Streamlet API. If there are remaining usages of pattern 1, please let me know.

This PR also removes the `Resources` class from the Streamlet API. In the future it may prove to be worthwhile to have a separate config class for that (if resource configuration becomes far more complex) but for now I think that this greatly simplifies the API. Now you can create a topology config like this:

```java
Config topologyConfig = Config.newBuilder()
    .setContainerRamInGB(10)
    .setContainerCpu(10.0f)
    .setDeliverySemantics(DeliverySemantics.ATMOST_ONCE)
    .build();
```


examples/src/java/com/twitter/heron/examples/streamlet/IntegerProcessingTopology.java
```diff
@@ -16,10 +16,10 @@
 
 import java.util.concurrent.ThreadLocalRandom;
 
+import com.twitter.heron.common.basics.ByteAmount;
```
srkukarni: Can we pl not do this. I specifically want no dep with anything heron other than api.

lucperkins:
@srkukarni This is again ready for review. I've removed the `Resources` class but have left the `ByteAmount` class (this is an issue that I'll deal with soon, but it's beyond the scope of this PR).

PR #2560, opened 2017-11-16
Refactor kubernetes scheduler to use statefulsets.

cckellogg:


cckellogg:
@jerrypeng will you review this too

deploy/kubernetes/minikube/apiserver.yaml
```diff
@@ -48,7 +48,7 @@ spec:
       serviceAccountName: heron-apiserver
       containers:
         - name: heron-apiserver
-          image: heron/heron:latest
+          image: heron/heron:lastest
```
srkukarni: What is this?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -179,19 +179,37 @@ private SchedulerUtils() {
...
+   * @param shardId the executor/container index
    * @param ports a map of ports to use where the key indicate the port type and the
    * value is the port
    * @return String[] representing the command to start heron-executor
    */
   public static String[] getExecutorCommand(
       Config config,
       Config runtime,
-      int containerIndex,
+      int shardId,
```
srkukarni: Can this be done better?

cckellogg: The flag we set is called shard so i updated the variable name to match. We can probably do this better but that is probably a bigger change that we can address in another pr.

PR #2567, opened 2017-11-17
Streamlet API cleanup

lucperkins:
This PR removes some redundant code from the Streamlet API and streamlines things a little bit, for the sake of future maintainability.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
+++ b/heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
@@ -15,7 +15,6 @@
 package com.twitter.heron.streamlet.impl;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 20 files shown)


heron/api/src/java/com/twitter/heron/streamlet/impl/streamlets/ConsumerStreamlet.java
```diff
@@ -38,13 +39,7 @@ public ConsumerStreamlet(StreamletImpl<R> parent, SerializableConsumer<R> consum
 
   @Override
   public boolean doBuild(TopologyBuilder bldr, Set<String> stageNames) {
-    if (getName() == null) {
-      setName(defaultNameCalculator("consumer", stageNames));
-    }
-    if (stageNames.contains(getName())) {
-      throw new RuntimeException("Duplicate Names");
-    }
-    stageNames.add(getName());
```
erenavsarogullari: Hi Luc. I was also thinking about this changes. Thanks for it. Adding streamlet name to stagesNames logic looks missed for `ConsumerStreamlet`. Other Streamlets look ok. I think this logic can also be part of the abstracted `setDefaultNameIfNone` method after streamlet name is uniqueness verification.

heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
```diff
@@ -125,6 +125,16 @@ public String getName() {
     return name;
   }
 
+  protected void setDefaultNameIfNone(String prefix, Set<String> stageNames) {
```
erenavsarogullari: I think definition of Streamlet/StagePrefixNameType as enum can be useful by supporting type-safety and all types can sit in single place.

PR #2572, opened 2017-11-18
add exception handling for metricscache process

huijunw:
add exception handling for metricscache thread 
so that the process can quit when exception is thrown from thread.

### DIFF
--- a/heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/MetricsCacheManager.java
+++ b/heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/MetricsCacheManager.java
@@ -18,7 +18,6 @@ package com.twitter.heron.metricscachemgr;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.util.concurrent.TimeUnit;
-import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -76,8 +75,6 @@ public class MetricsCacheManager {
 
   private TopologyMaster.MetricsCacheLocation metricsCacheLocation;
 
-  private final long mainThreadId;
-
   /**
    * Constructor: MetricsCacheManager needs 4 type information:
    * 1. Servers: host and 2 ports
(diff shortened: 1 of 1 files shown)


heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/MetricsCacheManager.java
```diff
@@ -126,6 +129,61 @@ public MetricsCacheManager(String topologyName,
...
+  public class DefaultExceptionHandler implements Thread.UncaughtExceptionHandler {
+
+    /**
+     * Handler for uncaughtException
+     */
+    public void uncaughtException(Thread thread, Throwable exception) {
+      // Add try and catch block to prevent new exceptions stop the handling thread
+      try {
+        // Delegate to the actual one
+        handleException(thread, exception);
```
maosongfu: I don't think we need to make it so complicated. We will not do any recovering here,
so we can simply:
```
Log(...);
// The ReleaseResourcesExplicitlyIfNeeded() can be a bit tricky. 
// If you want to release resources explicitly, make sure set a timeout to void unexpected hanging
// Also, it is possible that it can have other unexpected behaviors
// If all resources can eventually be released by the death of a process, 
// we are ok not to clean it explicitly
ReleaseResourcesExplicitlyIfNeeded(); 
Runtime.getRuntime().halt(1);
```

heron/metricscachemgr/src/java/com/twitter/heron/metricscachemgr/MetricsCacheManager.java
```diff
@@ -126,6 +126,34 @@ public MetricsCacheManager(String topologyName,
...
+        LOG.log(Level.SEVERE,
+            "Exception caught in thread: " + thread.getName()
+            + " with thread id: " + thread.getId(),
+            exception);
+        Runtime.getRuntime().halt(1);
+
+        // SUPPRESS CHECKSTYLE IllegalCatch
+      } catch (Throwable t) {
+        LOG.log(Level.SEVERE, "Failed to handle exception. Process halting", t);
+        Runtime.getRuntime().halt(1);
```
maosongfu: You can put this in `finally`

huijunw: updated

PR #2580, opened 2017-11-22
fix a bug for `deactivated heron-instance` during `heron update`

huijunw:
We saw below events by time sequence during `heron update` scaling up:
1. the old bolt is deactivated, its in-stream-queue drains and is empty
2. the new added bolt is deactivated, its in-stream-queue is full due to instance did not start().
3. the new added bolt does not receive pplan from stmgr due to in-stream-queue full.
4. Then this bolt never receives the `activate` tuple

This PR fix the above bug by:
1. let instance start()

### DIFF
--- a/heron/instance/src/java/com/twitter/heron/instance/spout/SpoutInstance.java
+++ b/heron/instance/src/java/com/twitter/heron/instance/spout/SpoutInstance.java
@@ -187,7 +187,7 @@ public class SpoutInstance implements IInstance {
     // Add spout tasks for execution
     addSpoutsTasks();
 
-    topologyState = TopologyAPI.TopologyState.RUNNING;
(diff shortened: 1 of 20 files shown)


heron/instance/src/java/com/twitter/heron/instance/spout/SpoutInstance.java
```diff
@@ -186,8 +186,6 @@ public void init(State<Serializable, Serializable> state) {
   public void start() {
     // Add spout tasks for execution
     addSpoutsTasks();
-
-    topologyState = TopologyAPI.TopologyState.RUNNING;
```
nwangtw: this is expected?

huijunw: no need to use `topologyState ` any more

PR #2582, opened 2017-11-22
fix mismatch container id between packing alg and aurora

huijunw:
1, Currently, Heron has the issues ....
We found the mismatch container id between packing algorithm and aurora during `heron update`

2, This pull request does ....
The PR updates the packing plan container id with aurora container id after 'aurora allocates container' and before 'update packing plan in zk'.
This PR changes spi/IScalable interface addContainers().

3, With this fix, Heron can ....
IScalable. addContainers() returns the new container id. Heron can re-map the real container id from Aurora to packing plan.

4, Test has been done .....
test done in Aurora datacenter.
testcase: in-consecutive container id test case.
testcase: append container, append after in-consecutive id container
testcase: remove container at the end and in the middle of id range


heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -157,19 +158,39 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+          scalableScheduler.get().addContainers(containerDelta.getContainersToAdd());
+      // Update the PackingPlan with new mapped container-ids
+      if (remapping != null) {
+        LOG.info("Doing remapping. ");
+        for (PackingPlan.ContainerPlan cp : proposedPackingPlan.getContainers()) {
+          if (remapping.containsKey(cp.getId())) {
+            // Replace with the actual mapping one
+            updatedContainers.remove(cp);
+            updatedContainers.add(remapping.get(cp.getId()));
+          }
```
nwangtw: what if remapping doesn't have the key?

heron/spi/src/java/com/twitter/heron/spi/scheduler/IScalable.java
```diff
@@ -30,8 +31,10 @@
    * Requests new containers for scaling a topology
    *
    * @param containersToAdd Set of containers to be added by the scheduler
+   * @return Map from container-id assigned by packing algorithm, to the actual deployed container-id
```
nwangtw: - the actual deployed container-id is a ContainerPlan object? the comment is a little confusing.
- if the mapping is not needed in a specific scheduler, the function should return empty map?

suggest to have a comment about it.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +112,33 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+    return Arrays.asList(containerIdStr.split(", "))
+        .stream().map(x->Integer.valueOf(x)).collect(Collectors.toSet());
   }
 
-  // Utils method for unit tests
-  @VisibleForTesting
-  boolean runProcess(List<String> auroraCmd) {
-    StringBuilder stdout = new StringBuilder();
-    StringBuilder stderr = new StringBuilder();
+  boolean runProcess(List<String> auroraCmd, StringBuilder stdout, StringBuilder stderr) {
```
nwangtw: should this function be VisibleForTesting too? Seems more suitable for unit tests.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: then keep the original key

heron/spi/src/java/com/twitter/heron/spi/scheduler/IScalable.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: updated with test

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +112,33 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
 
     LOG.info(String.format("Requesting %s new aurora containers %s", count, auroraCmd));
-    if (!runProcess(auroraCmd)) {
+    StringBuilder stderr = new StringBuilder();
+    if (!runProcess(auroraCmd, null, stderr)) {
       throw new RuntimeException("Failed to create " + count + " new aurora instances");
     }
+
+    String stdoutStr = stderr.toString();
+    String pattern = "Querying instance statuses: [";
```
nwangtw: Not good to match against "Querying instance statuses: ". I think it is better to match against regex ".*\[(\S*)\]" and extract the string within [] and then parse it.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -157,19 +158,39 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+        }
+      }
     }
 
+    PackingPlan updatedPackingPlan =
+        new PackingPlan(proposedPackingPlan.getId(), updatedContainers);
+    PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
+    PackingPlans.PackingPlan updatedProtoPackingPlan = serializer.toProto(updatedPackingPlan);
+    LOG.info("The updated Packing Plan: " + updatedProtoPackingPlan);
```
maosongfu: This should be `fine` Level Logging

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -157,19 +158,39 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+        new HashSet<>(proposedPackingPlan.getContainers());
     // request new resources if necessary. Once containers are allocated we should make the changes
     // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
     if (newContainerCount > 0 && scalableScheduler.isPresent()) {
-      scalableScheduler.get().addContainers(containerDelta.getContainersToAdd());
+      Map<Integer, PackingPlan.ContainerPlan> remapping =
+          scalableScheduler.get().addContainers(containerDelta.getContainersToAdd());
+      // Update the PackingPlan with new mapped container-ids
+      if (remapping != null) {
+        LOG.info("Doing remapping. ");
```
maosongfu: This should be `fine` Level Logging

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,43 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
   }
 
   @Override
-  public void addContainers(Integer count) {
+  public Set<Integer> addContainers(Integer count) {
```
maosongfu: Should we return a list to preserve order or set?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,43 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+        jobSpec + "/0", count.toString(), "--verbose"));
 
     LOG.info(String.format("Requesting %s new aurora containers %s", count, auroraCmd));
-    if (!runProcess(auroraCmd)) {
+    StringBuilder stderr = new StringBuilder();
+    if (!runProcess(auroraCmd, null, stderr)) {
       throw new RuntimeException("Failed to create " + count + " new aurora instances");
     }
+
+    if (stderr.length() <= 0) { // no container was added
```
maosongfu: If we need to add containers, and `stderr.length() <= 0`, should we consider it as an error and throw an exception?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,43 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+      LOG.info("empty output by Aurora");
+      return new HashSet<Integer>();
+    }
+    String stdoutStr = stderr.toString();
+    String pattern = "Querying instance statuses: [";
+    int idx1 = stdoutStr.indexOf(pattern) + pattern.length();
```
maosongfu: Good to make this parsing logic as a method below

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -187,8 +188,22 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
...
   @Override
-  public void addContainers(Set<PackingPlan.ContainerPlan> containersToAdd) {
-    controller.addContainers(containersToAdd.size());
+  public Map<Integer, PackingPlan.ContainerPlan> addContainers(
+      Set<PackingPlan.ContainerPlan> containersToAdd) {
+    // Do the actual containers adding
+    LinkedList<Integer> diffs = new LinkedList<>(
+        controller.addContainers(containersToAdd.size()));
+    Map<Integer, PackingPlan.ContainerPlan> remapping = new HashMap<>();
+    // Do the remapping
```
maosongfu: Good to add more comments to illustrate the remapping logic.

heron/spi/src/java/com/twitter/heron/spi/scheduler/IScalable.java
```diff
@@ -30,8 +31,12 @@
    * Requests new containers for scaling a topology
    *
    * @param containersToAdd Set of containers to be added by the scheduler
+   * @return Map from container-id assigned by packing algorithm, 
+   *             to ContainerPlan with the actual deployed container-id.
```
maosongfu: It is good to comment clearly:
- Will the map-to-return include all container-ids to add, no matter whether requesting a remapping? 
Or
- Will the map-to-return include only the container-ids requesting a remapping?
Or
- Will the map-to-return include at least all container-ids requesting a remapping, but also potentially include container-ids not requesting a remapping?

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: i prefer set because aurora returns unique ids. aurora does not enforce id order

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: `[]` is not enough, the stdout includes another [] list which is not container id list.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +112,33 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+        jobSpec + "/0", count.toString(), "--verbose"));
 
     LOG.info(String.format("Requesting %s new aurora containers %s", count, auroraCmd));
-    if (!runProcess(auroraCmd)) {
+    StringBuilder stderr = new StringBuilder();
+    if (!runProcess(auroraCmd, null, stderr)) {
       throw new RuntimeException("Failed to create " + count + " new aurora instances");
     }
+
+    String stdoutStr = stderr.toString();
```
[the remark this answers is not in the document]
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: updated

heron/spi/src/java/com/twitter/heron/spi/scheduler/IScalable.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
maosongfu: But don't you want a list of ids with increasing order? 

It could potentially help the remapping with less items.

huijunw: e. how can it help remapping with less items

huijunw:
@ashvina This PR is trying to map from `logical-container-id` to `aurora-container-id`. However Aurora does not accept metadata. 
Aurora gives container id on its own implementation. We do not know what container id Aurora will give before Aurora actually does container allocation. 
Thus this PR does a container-id remapping to align `logical-container-id` to the real `aurora-container-id` after Aurora does container allocation.

ashvina:
Currently, packing algorithm is executed before container allocation and a logical-container id is pre-allocated by the packing algorithm. The user needs to know this logical-container-id in cases like deadlocks where container restarts are required. The packing algorithm also makes resource (cpu, ram) suggestions. 
 
The actual resource allocated by schedulers can be different for from the required resources. For e.g. all Aurora containers are of same size and YARN containers cannot be smaller than the limit configured by the YARN admin. Currently, the packing plan created by packing algorithm is updated by the scheduler based on actual allocation. I think we could employ the same reasoning and pattern for container id too. The container id suggested by packing algorithm is updated by the scheduler based on actual allocation by the scheduler. I believe the logical-container-id is not used anywhere else. 

In long term, it would be useful to find a way to assign logical container id to Aurora containers. Each container should fetch details of the instances it needs to execute. I think @srkukarni proposed something similar earlier.

huijunw:
talked with @ashvina , and @nwangtw @maosongfu 
will update according to @ashvina suggestion:

1. Change SPI to :
```
// @return Set of added containers with real ids and resources from scheduler
Set<PackingPlan.ContainerPlan> addContainers(Set<PackingPlan.ContainerPlan> containersToAdd);
```
2. Also, add the returned containers to current packing plan and persist that in ZK.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,45 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+
+    if (stderr.length() <= 0) { // no container was added
+      throw new RuntimeException("empty output by Aurora");
+    }
+    return extractContainerIds(stderr.toString());
+  }
+
+  private Set<Integer> extractContainerIds(String auroraStdoutStr) {
+    String pattern = "Querying instance statuses: [";
+    int idx1 = auroraStdoutStr.indexOf(pattern) + pattern.length();
```
nwangtw: I think this line might have a bug.

idx1 is not -1 when the string is not found. Instead, it is -1 + pattern.length(). So you cannot compare it with 0.

Suggest to add a unit test for the function to make sure it works correctly in both successful and failure cases.

huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,46 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
       throw new RuntimeException("Failed to create " + count + " new aurora instances");
     }
+
+    if (stderr.length() <= 0) { // no container was added
+      throw new RuntimeException("empty output by Aurora");
+    }
+    return extractContainerIds(stderr.toString());
+  }
+
+  private Set<Integer> extractContainerIds(String auroraStdoutStr) {
```
nwangtw: Suggest to have a unit test for this function, not a blocker though.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -111,29 +113,46 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
-    if (!runProcess(auroraCmd)) {
+    StringBuilder stderr = new StringBuilder();
+    if (!runProcess(auroraCmd, null, stderr)) {
       throw new RuntimeException("Failed to create " + count + " new aurora instances");
     }
+
+    if (stderr.length() <= 0) { // no container was added
+      throw new RuntimeException("empty output by Aurora");
+    }
+    return extractContainerIds(stderr.toString());
```
nwangtw: the function signature is "String auroraStdoutStr" and here the parameter is stderr. Maybe the parameter should be renamed.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -183,8 +185,23 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
...
-    controller.addContainers(containersToAdd.size());
+  public Set<PackingPlan.ContainerPlan> addContainers(
+      Set<PackingPlan.ContainerPlan> containersToAdd) {
+    // Do the actual containers adding
+    LinkedList<Integer> newAddedContainerIds = new LinkedList<>(
+        controller.addContainers(containersToAdd.size()));
+    Set<PackingPlan.ContainerPlan> remapping = new HashSet<>();
+    // Do the remapping:
+    // use the `newAddedContainerIds` to replace the container id in the `containersToAdd`
+    for (PackingPlan.ContainerPlan cp : containersToAdd) {
```
nwangtw: I think it should be better to test containersToAdd.size() == newAddedContainerId.size().

The controller.addContainers() function could return empty set or smaller set, and we need to handle the two cases here.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: update to `auroraOutputStr`

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: there was a unit test already.
added one more unit test

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: done

PR #2588, opened 2017-11-26
Add the ability to mount volumes for topologies in k8s.

cckellogg:
This patch add the ability to mount volumes for topologies in kubernetes. It also fixes a couple of submit bugs with the apiserver and cli client. The cli was not sending the override params and the apiserver was not applying the overrides correctly.

### DIFF
--- a/heron/tools/common/src/python/utils/config.py
+++ b/heron/tools/common/src/python/utils/config.py
@@ -382,18 +382,33 @@ def defaults_cluster_role_env(cluster_role_env):
 ################################################################################
 # Parse the command line for overriding the defaults
 ################################################################################
-def parse_override_config(namespace):
-  """Parse the command line for overriding the defaults"""
+def parse_override_config_and_write_file(namespace):
+  """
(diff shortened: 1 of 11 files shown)


heron/tools/common/src/python/utils/config.py
```diff
@@ -382,20 +382,35 @@ def defaults_cluster_role_env(cluster_role_env):
...
+  """
+  overrides = parse_override_config(namespace)
   try:
     tmp_dir = tempfile.mkdtemp()
     override_config_file = os.path.join(tmp_dir, OVERRIDE_YAML)
     with open(override_config_file, 'w') as f:
-      for config in namespace:
-        f.write("%s\n" % config.replace('=', ': '))
+      for k, v in overrides.items():
+        f.write("%s:%s\n" % (k, v))
```
huijunw: there is a space after `:`, which breaks our jeckins ci for two days. @maosongfu  @nwangtw

PR #2593, opened 2017-11-28
Add ability for users to allow tuple drops upon backpressure detection

srkukarni:


heron/common/src/cpp/network/client.cpp
```diff
@@ -218,3 +218,8 @@ void Client::StartBackPressureConnectionCb(Connection* conn) {
 void Client::StopBackPressureConnectionCb(Connection* conn) {
   // Nothing to be done here. Should be handled by inheritors if they care about backpressure
 }
+
+bool Client::HasCausedBackPressure() const {
+  Connection* conn = static_cast<Connection*>(conn_);
```
nwangtw: Do we need to check state&pointer here? Can conn_ be NULL in some cases/states?

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -356,6 +358,10 @@ void InstanceServer::DrainTupleStream(proto::stmgr::TupleStreamMessage* _message
   TaskIdInstanceDataMap::iterator iter = instance_info_.find(task_id);
   if (iter == instance_info_.end() || iter->second->conn_ == NULL) {
     LOG(ERROR) << "task_id " << task_id << " has not yet connected to us. Dropping...";
+  } else if (droptuples_upon_backpressure_ && iter->second->conn_->hasCausedBackPressure()) {
+    LOG(ERROR) << "task_id " << task_id << " is causing backpressure, so Dropping tuples to him";
```
nwangtw: suggest to add "and droptuples_upon_backpressure is set to true" in log.

Also, how many logs are we talking about here?

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -356,6 +358,10 @@ void InstanceServer::DrainTupleStream(proto::stmgr::TupleStreamMessage* _message
...
+    // Ideally we would have counted the numbers lost, but we cannot since the tuples are in
```
nwangtw: Do we have any information for the package we can log? Like time window, etc? Sorry if this is a silly question.

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -371,16 +377,24 @@ void InstanceServer::SendToInstance2(sp_int32 _task_id,
 void InstanceServer::DrainTupleSet(sp_int32 _task_id,
                                 proto::system::HeronTupleSet2* _message) {
   TaskIdInstanceDataMap::iterator iter = instance_info_.find(_task_id);
-  if (iter == instance_info_.end() || iter->second->conn_ == NULL) {
-    LOG(ERROR) << "task_id " << _task_id << " has not yet connected to us. Dropping...";
-    if (_message->has_control()) {
+  if (iter == instance_info_.end() || iter->second->conn_ == NULL
+      || (droptuples_upon_backpressure_ && iter->second->conn_->hasCausedBackPressure())) {
+    LOG(ERROR) << "task_id " << _task_id
+               << " has not yet connected to us or is not keeping up, Dropping...";
```
nwangtw: same, suggest to add "and droptuples_upon_backpressure is set to true" in log.

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -463,6 +483,13 @@ void InstanceServer::StopBackPressureConnectionCb(Connection* _connection) {
   sp_string instance_name = GetInstanceName(_connection);
   CHECK_NE(instance_name, "");
 
+  LOG(INFO) << "We don't observe back pressure now on sending data to instance " << instance_name;
```
nwangtw: I am just wondering if this should be a WARNING log. Not a blocker.

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -463,6 +483,13 @@ void InstanceServer::StopBackPressureConnectionCb(Connection* _connection) {
...
+
+  if (droptuples_upon_backpressure_) {
```
nwangtw: Suggest to add a log here like "droptuples_upon_backpressure is set to true, hence no backpressure mechanism is initialized"

heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -192,6 +194,12 @@ bool StMgrClient::SendTupleStreamMessage(proto::stmgr::TupleStreamMessage& _msg)
   if (IsConnected()) {
     SendMessage(_msg);
     return true;
+  } else if (droptuples_upon_backpressure_ && HasCausedBackPressure()) {
+    if (++ndropped_messages_ % 100 == 0) {
+      LOG(INFO) << "Dropping " << ndropped_messages_ << "th tuple message to stmgr "
```
nwangtw: The drop count is for both not connected and backpressure, so the log could be misleading. Just a concern to me, not a real issue. Your call.

heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -192,6 +194,12 @@ bool StMgrClient::SendTupleStreamMessage(proto::stmgr::TupleStreamMessage& _msg)
   if (IsConnected()) {
     SendMessage(_msg);
```
nwangtw: Just to confirm. The drop upon backpressure logic is handled in the "(!IsConnected())" branch. Is it expected? Do we need it in the "(IsConnected)" branch?

heron/stmgr/src/cpp/manager/stmgr-client.cpp
```diff
@@ -211,13 +219,17 @@ void StMgrClient::StartBackPressureConnectionCb(Connection* _connection) {
   // Ask the StMgrServer to stop consuming. The client does
   // not consume anything
 
-  client_manager_->StartBackPressureOnServer(other_stmgr_id_);
+  if (!droptuples_upon_backpressure_) {
+    client_manager_->StartBackPressureOnServer(other_stmgr_id_);
```
nwangtw: I don't know this part of code yet. Just to confirm, the StartBackPressureOnServer() call doesn't affect the HasCausedBackpressure() result, right?

srkukarni:
@nwangtw This flag makes sense even in ATLEAST ONCE because the tuples will time out and will retry later

nwangtw:
Got it. Thanks!

heron/common/src/cpp/network/client.cpp
srkukarni: Not really, but I've added some checks just for safety

heron/stmgr/src/cpp/manager/instance-server.cpp
srkukarni: Expanded comment and also log every so often

heron/stmgr/src/cpp/manager/instance-server.cpp
srkukarni: Addressed by adding bytes dropped

heron/stmgr/src/cpp/manager/instance-server.cpp
srkukarni: Added

heron/stmgr/src/cpp/manager/instance-server.cpp
srkukarni: Sounds good

heron/stmgr/src/cpp/manager/instance-server.cpp
srkukarni: Added

heron/stmgr/src/cpp/manager/stmgr-client.cpp
srkukarni: Redid the logic

heron/stmgr/src/cpp/manager/stmgr-client.cpp
srkukarni: I think as long as we are counting drops, we should be ok

heron/stmgr/src/cpp/manager/stmgr-client.cpp
srkukarni: Nopes

nwangtw:
Cool. Overall LGTM.

One last comment is to add the config in storm compatibility as well to keep them in sync.

srkukarni:
@nwangtw 1. Does Storm have an equivalent call?

nwangtw:
Good question. 

I thought that the config as well and the translation should be added to the storm layer. Or maybe it shouldn't since the storm API shouldn't have the new features? I think I am ok either way.

srkukarni:
Unless there is an equivalent storm api, there is no point adding this flag in the storm layer. As you mentioned storm layer is strictly for emulating the storm api, not adding func to it.

nwangtw:
Make sense. Thanks.

Issue #2576, opened 2017-11-20
`heron update` re-packing exception


huijunw:
how to reproduce:
1. on github master
`~/bin/heron submit --deploy-deactivated --verbose local ~/.heron/examples/heron-api-examples.jar com.twitter.heron.examples.api.WordCountTopology WordCountTopology 3 `
2. update it
`~/bin/heron update --verbose --dry-run local WordCountTopology --component-parallelism consumer:4`

IMO, the `heron update` should be able to update the above example topology successfully by simply adding one more container.

The `heron update` throws exception and fails to update. See below.
```
[2017-11-20 14:08:21 -0800] [FINE] com.twitter.heron.scheduler.RuntimeManagerMain: Exception when submitting topology 
com.twitter.heron.spi.packing.PackingException: Could not initialize containers using existing packing plan
	at com.twitter.heron.packing.builder.PackingPlanBuilder.initContainers(PackingPlanBuilder.java:259)
	at com.twitter.heron.packing.builder.PackingPlanBuilder.addInstance(PackingPlanBuilder.java:153)
	at com.twitter.heron.packing.builder.PackingPlanBuilder.addInstance(PackingPlanBuilder.java:141)
	at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.placeFFDInstance(FirstFitDecreasingPacking.java:312)
	at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.assignInstancesToContainers(FirstFitDecreasingPacking.java:265)
	at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.getFFDAllocation(FirstFitDecreasingPacking.java:246)
	at com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking.repack(FirstFitDecreasingPacking.java:180)
	at com.twitter.heron.scheduler.RuntimeManagerRunner.buildNewPackingPlan(RuntimeManagerRunner.java:304)
	at com.twitter.heron.scheduler.RuntimeManagerRunner.updateTopologyHandler(RuntimeManagerRunner.java:183)
	at com.twitter.heron.scheduler.RuntimeManagerRunner.call(RuntimeManagerRunner.java:81)
	at com.twitter.heron.scheduler.RuntimeManagerMain.callRuntimeManagerRunner(RuntimeManagerMain.java:448)
	at com.twitter.heron.scheduler.RuntimeManagerMain.manageTopology(RuntimeManagerMain.java:396)
[...]

PR #2600, opened 2017-12-02
RoundRobinPacking implements IRepacking

huijunw:
Walk around https://github.com/twitter/heron/issues/2576
Some legacy topologies use RoundRobinPacking lacking configurations for neither ResourceCompliantRRPacking nor FirstFitDecreasingPacking, which results in exceptions.
This PR enables IRepacking implementation for RoundRobinPacking.

Tested with local scheduler
`~/bin/heron update --verbose --dry-run --config-property heron.class.repacking.algorithm=com.twitter.heron.packing.roundrobin.RoundRobinPacking local WordCountTopology --component-parallelism consumer:4`

### DIFF
--- a/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
@@ -373,7 +373,7 @@ public class RoundRobinPacking implements IPacking, IRepacking {
       Integer count = e.getValue();
 
       if (componentChanges.containsKey(componentName)) {
-        count += componentChanges.get(componentName);
+        count = componentChanges.get(componentName);
       }
 
       newComponentParallelism.put(componentName, count);
--- a/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//    http://www.apache.org/licenses/LICENSE-2.0
+// http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
@@ -30,6 +30,7 @@ import com.twitter.heron.common.basics.ByteAmount;
 import com.twitter.heron.spi.common.Config;
 import com.twitter.heron.spi.common.Context;
 import com.twitter.heron.spi.packing.IPacking;
+import com.twitter.heron.spi.packing.IRepacking;
 import com.twitter.heron.spi.packing.InstanceId;
 import com.twitter.heron.spi.packing.PackingException;
 import com.twitter.heron.spi.packing.PackingPlan;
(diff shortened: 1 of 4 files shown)


heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -343,4 +353,32 @@ private void validatePackingPlan(PackingPlan plan) throws PackingException {
       }
     }
   }
+
+  @Override
+  public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> componentChanges)
+      throws PackingException {
+    int initialNumContainer = TopologyUtils.getNumContainers(topology);
+    int initialNumInstance = TopologyUtils.getTotalInstance(topology);
+    double initialNumInstancePerContainer = (double) initialNumInstance / initialNumContainer;
```
maosongfu: This implementation tries to keep the ratio of `#(instances)/#(containers)` always the same during re-scaling.

The issue here is:
During each re-scale, we would update only # of instances in topology def, and # of containers will not be updated. So the # of containers fetched by `TopologyUtils.getNumContainers(topology)` is not accurate. So if we continues scaling-up a topology, we would end up:
**increasing the # of instances in one single container boundlessly**.

One thought is always getting # of instances and containers based on `currentPackingPlan`. This will reflect the correct values.
But it also has some issues: we will try to keep the ratio of `#(instances)/#(containers)` the same with last `currentPackingPlan`, rather than the original one. And every time we would ceil the # of containers, since the # of containers has to be an integer. And after several times of scale-up, it can lead to:
**The # of instances in per container decreases and eventually there will be only one instance per container.**

The ideal case is to always keep the ratio of `#(instances)/#(containers)` the same as the **original** ratio. But it seems our current state updating is kind of messy, and I am not sure whether there is a simple way to get it. Or one thought is to insert this original ratio as a config during submission.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -232,10 +242,11 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
    *
    * @return containerId -&gt; list of InstanceId belonging to this container
    */
-  private Map<Integer, List<InstanceId>> getRoundRobinAllocation() {
+  private Map<Integer, List<InstanceId>> getRoundRobinAllocation(
+      int numContainer, Map<String, Integer> parallelismMap) {
     Map<Integer, List<InstanceId>> allocation = new HashMap<>();
-    int numContainer = TopologyUtils.getNumContainers(topology);
-    int totalInstance = TopologyUtils.getTotalInstance(topology);
+    int totalInstance =
```
nwangtw: TopologyUtils.getTotalInstance() doesn't work? Maybe worth adding this code to TopologyUtils as a util function?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -343,4 +353,32 @@ private void validatePackingPlan(PackingPlan plan) throws PackingException {
       }
     }
   }
+
+  @Override
+  public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> componentChanges)
```
nwangtw: Please add comment about what the logic below is doing and why. This function is not very straightforward.

Also are there unit tests for this class?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
nwangtw: I am curious about how the original number of container is evaluated? Is it configured or calculated? If it is configured, maybe better to pass a parameter? If it is calculated based on topology, can we run the same calculation with the new topology config to have a good number of container? I agree with maosong that the ratio may or may not be the best option here.

maosongfu: - I am curious about how the original number of container is evaluated? 
It is set by customers by is put into the topology-level config: https://github.com/twitter/heron/blob/master/heron/api/src/java/com/twitter/heron/api/Config.java#L356


One issue is there is a bug when updating the topology: https://github.com/twitter/heron/blob/master/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java#L317. It will update only the new # of instances, but not the new # of containers.
Also, I don't think it is a good idea to update the topology def here; I think it should be unchanged once a topology is submitted. All runtime changes should apply to the PhycicalPlan.


- I agree with maosong that the ratio may or may not be the best option here.
Ratio is ok if we can keep the ratio same as the original one. 
Another thought is we could also accept `updated count of containers` when updating the topology, rather than accept `updated count of instances for component`. To make it, we can reuse existing routine: the component name is `container` or `stmgr`, like:
`--component-parallelism container:100`

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: +1 for unit tests. Particularly, the unit tests should guarantee situations below not to happen:
- count of instances in a container grows boundlessly
- count of instances in a container reduces to 1 eventually

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
maosongfu: The implementation of `TopologyUtils.getTotalInstance()`, https://github.com/twitter/heron/blob/57310fe58a808ee9cf88b91121d9d8822698bee1/heron/api/src/java/com/twitter/heron/api/utils/TopologyUtils.java#L130, seems same?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: how about count the max#(instance) and min#(instance) per container from the current packing plan, and keep both counts during scaling ?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: getTotalInstance() needs a TopologyAPI.Topology which is different from Map<String, Integer>
or we may refactor getTotalInstance()? what do you think

nwangtw: I think refactoring might be better. Easier to read and easier to test.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: make the ratio remains same during topology lifecycle #2602

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: refactored

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: added comment

huijunw: added unit test

billonahill:
When `IRepacking` was developed we made the decision not to implement it in `RoundRobinPacking` for a couple of reasons.

The first is that RRPacking chooses to take the requested resources as hints and obey the number of containers requested. This has led to unexpected behavior for users who increase the parallelism for a given component to (via changing settings and redeploying) and end up getting fewer resources per component instance. This resulted in hard-to-diagnose issues. The `ResourceCompliantRRPacking` classes were written to always give the resources the user expects/requests and to adjust the number of containers as needed. This approach seemed more ideal so the plan was to deprecate `RRPacking`.

The second reason is because conceptually there is a disconnect when implementing scaling of an algorithm that ignores requested resources in favor of a number of containers setting. When scaling, the algorithm has to dynamically figure our how many containers should be added/removed to a topology that previously had a fixed number of them. This implies a shift from container-based to resource-based sizing which is not easy to rationalize and is hence prone to misunderstanding or troubleshooting challenges. Or it means we just add/remove containers and the numContainers setting that was the source of truth during deploy becomes not so.

I would recommend continuing with the plan to deprecate this class in favor of the resource-base ones.

huijunw:
1. The present resource based packing algorithms requires changes in user topology code. Meanwhile, it is not prod verified yet so a lot of testing and bug-fixes will be involved. The migration will not be trivial and it can take 2+ quarters.
2.The users would like to use `heron update` now but RRpacking does not support it yet and it is a blocker since people can not repack with FirstFitPackingAlgorithm.  The implementation in this PR and #2602 handle the mentioned `end up getting fewer resources per component instance` issue.
3. Make RRrepacking extends IRepack can add an option for user; it will not hurt the migration to the new packing algorithm in future.

billonahill:
Thanks for the clarification. Fair enough.

huijunw:
Thanks @billonahill , could you also review #2602, which is related to this PR.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -343,4 +351,37 @@ private void validatePackingPlan(PackingPlan plan) throws PackingException {
       }
     }
   }
+
+  /*
+   * read the current packing plan with update parallelism to calculate a new packing plan
```
nwangtw: is "current packing plan" accurate?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -343,4 +351,37 @@ private void validatePackingPlan(PackingPlan plan) throws PackingException {
...
+   */
+  @Override
+  public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> componentChanges)
+      throws PackingException {
+    int initialNumContainer = TopologyUtils.getNumContainers(topology);
+    int initialNumInstance = TopologyUtils.getTotalInstance(topology);
+    double initialNumInstancePerContainer = (double) initialNumInstance / initialNumContainer;
+
+    Map<String, Integer> currentComponentParallelism = currentPackingPlan.getComponentCounts();
+    Map<String, Integer> newComponentParallelism = new HashMap<>();
```
nwangtw: I am wondering if it is easier to understand this way:

newComponentParallelism = currentComponentParallelism;
for (e: componentChanges) {
  newComponentParallelism.put(e.getKey, e.getValue);
}

int newNumInstance = TopologyUtils.getTotalInstance(newComponentParallelism);

nwangtw: also, I am wondering if we should verify that the value needs to be positive?

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: this packing plan is the latest from zk.

heron/packing/src/java/com/twitter/heron/packing/roundrobin/RoundRobinPacking.java
huijunw: newComponentParallelism is supposed to include the old and new components, while the suggested code includes new components only.

huijunw: added positive parallelism check in RuntimeManagerRunner.java

huijunw: adopt @nwangtw suggestion code

srkukarni:
@billonahill Does the pr look good to you? If so can you pl approve? Thanks!

billonahill:
Looks like travis is failing for this PR.

The approach seems to be to regenerate the packing plan from scratch when doing the repacking. This means that if the right component is scaled, the entire topology set of topology instances will be shuffled. This can be a very disruptive approach. Have you considered other algorithms for deciding where to place new instances, or how to remove them that are less disruptive?

If you look at the other IRepacking algos there is a scoring framework utilized which allows the ability to decide if you want to scale for least disruption vs trying to remove containers vs homogeneity, etc. That could be leveraged here for consistency with the other IRepackings.

huijunw:
@billonahill  the unit test has been fixed.
I would suggest have this PR merged to make `update` work at present. Then, we may evolve the algorithm to be less disruptive.

billonahill:
👍

PR #2601, opened 2017-12-04
 adding Nomad scheduler support for Heron

jerrypeng:
This PR contains the core code for running Heron on Nomad.

In a subsequent PR, I will add additional scripts to allow users to easily deploy a standalone cluster running Heron.

Something like this:

heron standalone set masters

heron standalone set slaves

heron standalone start cluster

### DIFF
--- a/heron/config/src/yaml/BUILD
+++ b/heron/config/src/yaml/BUILD
@@ -14,7 +14,7 @@ filegroup(
 
 filegroup(
     name = "conf-yaml",
-    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"]),
+    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"] + ["conf/**/*.sh"]),
 )
 
 filegroup(
--- a/heron/config/src/yaml/conf/nomad/statemgr.yaml
+++ b/heron/config/src/yaml/conf/nomad/statemgr.yaml
@@ -1,0 +1,23 @@
+# local state manager class for managing state in a persistent fashion
+heron.class.state.manager: com.twitter.heron.statemgr.zookeeper.curator.CuratorStateManager
+
+# local state manager connection string
+heron.statemgr.connection.string:  127.0.0.1:2181
+
+# path of the root address to store the state in a local file system
+heron.statemgr.root.path: "/heron"
+
+# create the zookeeper nodes, if they do not exist
+heron.statemgr.zookeeper.is.initialize.tree: True
+
+# timeout in ms to wait before considering zookeeper session is dead
+heron.statemgr.zookeeper.session.timeout.ms: 30000
+
+# timeout in ms to wait before considering zookeeper connection is dead
+heron.statemgr.zookeeper.connection.timeout.ms: 30000
+
+# timeout in ms to wait before considering zookeeper connection is dead
+heron.statemgr.zookeeper.retry.count: 10
+
+# duration of time to wait until the next retry
+heron.statemgr.zookeeper.retry.interval.ms: 10000
--- a/heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
+++ b/heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
@@ -34,7 +34,6 @@ public final class DownloadRunner {
     final Path topologyDestination = Paths.get(destination);
 
     final File file = topologyDestination.toFile();
-
     if (!file.exists()) {
       file.mkdirs();
     }
--- a/heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
+++ b/heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
@@ -34,6 +34,7 @@ public final class DownloadRunner {
     final Path topologyDestination = Paths.get(destination);
 
     final File file = topologyDestination.toFile();
+
     if (!file.exists()) {
       file.mkdirs();
     }
--- a/heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
+++ b/heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
@@ -1,0 +1,26 @@
+//  Copyright 2017 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.downloader;
+
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Path;
+
+public class FileDownloader implements Downloader {
+  @Override
+  public void download(URI uri, Path destination) throws Exception {
+    final URL url = uri.toURL();
+    Extractor.extract(url.openStream(), destination);
+  }
+}
--- a/heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
+++ b/heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
@@ -17,6 +17,10 @@ import java.net.URI;
 import java.net.URL;
 import java.nio.file.Path;
 
+/**
+ * Used to download files via heron_downloader that has a URI prefix of "file://"
+ * E.g. ./heron_dowloader file:///foo/bar /path/location
+ */
 public class FileDownloader implements Downloader {
   @Override
   public void download(URI uri, Path destination) throws Exception {
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -50,7 +50,9 @@ def print_usage():
       "Usage: ./heron-executor --shard=<shardid> --topology-name=<topname>"
       " --topology-id=<topid> --topology-defn-file=<topdefnfile>"
       " --state-manager-connection=<state_manager_connection>"
-      " --state-manager-root=<state_manager_root> --tmaster-binary=<tmaster_binary>"
+      " --state-manager-root=<state_manager_root>"
+      " --state-manager-config-file=<state_manager_config_file>"
+      " --tmaster-binary=<tmaster_binary>"
       " --stmgr-binary=<stmgr_binary> --metrics-manager-classpath=<metricsmgr_classpath>"
       " --instance-jvm-opts=<instance_jvm_opts_in_base64> --classpath=<classpath>"
       " --master-port=<master_port> --tmaster-controller-port=<tmaster_controller_port>"
@@ -173,6 +175,7 @@ class HeronExecutor(object):
     self.topology_defn_file = parsed_args.topology_defn_file
     self.state_manager_connection = parsed_args.state_manager_connection
     self.state_manager_root = parsed_args.state_manager_root
+    self.state_manager_config_file = parsed_args.state_manager_config_file
     self.tmaster_binary = parsed_args.tmaster_binary
     self.stmgr_binary = parsed_args.stmgr_binary
     self.metrics_manager_classpath = parsed_args.metrics_manager_classpath
@@ -277,6 +280,7 @@ class HeronExecutor(object):
     parser.add_argument("--topology-defn-file", required=True)
     parser.add_argument("--state-manager-connection", required=True)
     parser.add_argument("--state-manager-root", required=True)
+    parser.add_argument("--state-manager-config-file", required=True)
     parser.add_argument("--tmaster-binary", required=True)
     parser.add_argument("--stmgr-binary", required=True)
     parser.add_argument("--metrics-manager-classpath", required=True)
@@ -978,7 +982,9 @@ class HeronExecutor(object):
     Receive updates to the packing plan from the statemgrs and update processes as needed.
     """
     statemgr_config = StateMgrConfig()
-    statemgr_config.set_state_locations(configloader.load_state_manager_locations(self.cluster))
+    statemgr_config.set_state_locations(configloader.load_state_manager_locations(
+        self.cluster, state_manager_config_file=self.state_manager_config_file,
+        overrides={"heron.statemgr.connection.string": self.state_manager_connection}))
     try:
       self.state_managers = statemanagerfactory.get_all_state_managers(statemgr_config)
       for state_manager in self.state_managers:
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
@@ -189,6 +189,24 @@ public final class SchedulerUtils {
       Config runtime,
       int shardId,
       Map<ExecutorPort, String> ports) {
+    return getExecutorCommand(config, runtime, Integer.toString(shardId), ports);
+  }
+
+  /**
+   * Utils method to construct the command to start heron-executor
+   *
+   * @param config The static config
+   * @param runtime The runtime config
+   * @param shardId the executor/container index
+   * @param ports a map of ports to use where the key indicate the port type and the
+   * value is the port
+   * @return String[] representing the command to start heron-executor
+   */
+  public static String[] getExecutorCommand(
+      Config config,
+      Config runtime,
+      String shardId,
+      Map<ExecutorPort, String> ports) {
     List<String> commands = new ArrayList<>();
     commands.add(Context.executorBinary(config));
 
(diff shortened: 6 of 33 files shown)


jerrypeng:
@srkukarni @maosongfu @nwangtw

heron/executor/src/python/heron_executor.py
```diff
@@ -978,7 +982,9 @@ def start_state_manager_watches(self):
     Receive updates to the packing plan from the statemgrs and update processes as needed.
     """
     statemgr_config = StateMgrConfig()
-    statemgr_config.set_state_locations(configloader.load_state_manager_locations(self.cluster))
+    statemgr_config.set_state_locations(configloader.load_state_manager_locations(
```
srkukarni: can you comment on what's going on here?

jerrypeng: I add this because specifying an override via commandline for the config "heron.statemgr.connection.string" does not get propagated into the heron_executor.  Thus, if I want to specify the location of zookeeper via command line, I cannot.

heron/config/src/yaml/BUILD
```diff
@@ -14,7 +14,7 @@ filegroup(
 
 filegroup(
     name = "conf-yaml",
-    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"]),
+    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"] + ["conf/**/*.sh"]),
```
nwangtw: Hmm. Is it possible to avoid the .sh files here?

heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
```diff
@@ -34,6 +34,7 @@ public static void main(String[] args) throws Exception {
     final Path topologyDestination = Paths.get(destination);
 
     final File file = topologyDestination.toFile();
+
```
nwangtw: revert this change

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
```diff
@@ -0,0 +1,26 @@
...
+package com.twitter.heron.downloader;
+
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Path;
+
+public class FileDownloader implements Downloader {
+  @Override
+  public void download(URI uri, Path destination) throws Exception {
+    final URL url = uri.toURL();
```
nwangtw: do we need try/cache here?

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
```diff
@@ -0,0 +1,26 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.downloader;
+
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Path;
+
+public class FileDownloader implements Downloader {
```
nwangtw: Why a new downloader is added? HttpDownloader is not enough in this case? The implementation seems to be the same.

If there is a new case, please add comment to explain. My concern is that in future people dont know which one to use or could mis-use them.

heron/schedulers/src/java/BUILD
```diff
@@ -53,6 +53,25 @@ kubernetes_deps_files = \
...
+          "@commons_codec_commons_codec//jar",
+          "@commons_logging_commons_logging//jar",
+          "@org_apache_httpcomponents_http_client//jar",
+          "@org_apache_httpcomponents_http_core//jar",
+          "@org_bouncycastle_bcpkix_jdk15on//jar",
+          "@org_bouncycastle_bcprov_jdk15on//jar",
+      ]
+    
+nomad_deps_files = \
+    scheduler_deps_files + nomad_sdk_deps +[
```
nwangtw: space between "+" and "["

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/AppsV1beta1Controller.java
```diff
@@ -228,7 +228,7 @@ boolean isStatefulSet() {
     final Config runtimeConfiguration = getRuntimeConfiguration();
     final String[] executorCommand =
         SchedulerUtils.getExecutorCommand(configuration, runtimeConfiguration,
-            containerId, ports);
+            Integer.valueOf(containerId), ports);
```
cckellogg: This needs to be a string because we use an environment variable for the container id

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -189,29 +189,10 @@ private SchedulerUtils() {
...
-   * Utils method to construct the command to start heron-executor
-   *
-   * @param config The static config
-   * @param runtime The runtime config
-   * @param shardId the executor/container index
-   * @param ports a map of ports to use where the key indicate the port type and the
-   * value is the port
-   * @return String[] representing the command to start heron-executor
-   */
-  public static String[] getExecutorCommand(
```
cckellogg: Kubernetes needs to be able to send a string as the shard id

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
jerrypeng: So I added this because the heron_downloader couldn't handle URIs like: "file:///foo/bar".  This is the same implementation as the HttpDownloader, but I am not sure if we should point URIs with a prefix of "file" to the the http downloader or just have a new file downloader.  The reason to why we need to support copy of files by the heron_downloader is because there could be instances where the user is distributing JARs via some sort of network storage and we need to copy and extract the jars from the network storage location into the local heron directory structure

heron/schedulers/src/java/BUILD
jerrypeng: will add

heron/downloaders/src/java/com/twitter/heron/downloader/DownloadRunner.java
jerrypeng: will revert

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
jerrypeng: though this is really done once on the start of the topology

heron/config/src/yaml/BUILD
jerrypeng: We need the heron_nomad.sh script to start heron_executors in nomad.  Not sure where else would be more appropriate to put this since the script needs to be available on where you install the heron cli.  The heron.aurora script is placed in the aurora config folder.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
jerrypeng: will change back

heron/schedulers/src/java/com/twitter/heron/scheduler/kubernetes/AppsV1beta1Controller.java
jerrypeng: will change back

jerrypeng:
@srkukarni @nwangtw @cckellogg thanks for your reviews!  I have addressed your comments.

heron/config/src/yaml/conf/nomad/statemgr.yaml
```diff
@@ -0,0 +1,23 @@
+# local state manager class for managing state in a persistent fashion
+heron.class.state.manager: com.twitter.heron.statemgr.zookeeper.curator.CuratorStateManager
+
+# local state manager connection string
+heron.statemgr.connection.string:  127.0.0.1:2181
```
nwangtw: why 2181? I think it is the default port for zk so maybe better to avoid if possible.

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
nwangtw: kk.

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
nwangtw: Got it. thx. Please add a comment about the use case. An example could be good enough.

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  @After
+  public void after() throws Exception {
+    scheduler.close();
+  }
+
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+
```
nwangtw: Remove the empty line.

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+
```
nwangtw: Remove the empty line.

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
```diff
@@ -0,0 +1,366 @@
...
+    List<Job> jobList = Arrays.asList(jobs);
+
+    // Fail to restart
+    PowerMockito.when(NomadScheduler.getTopologyJobs(Mockito.any(NomadApiClient.class),
+        Mockito.anyString())).thenReturn(jobList);
+    PowerMockito.doThrow(new RuntimeException()).when(NomadScheduler.class,
+        "restartJobs", Mockito.any(), anyVararg());
+    Assert.assertFalse(scheduler.onRestart(restartTopologyRequest));
+
+    // succeed to restart
```
nwangtw: nit: s/succeed/Succeed

There are a few of them in this file

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
```diff
@@ -0,0 +1,366 @@
...
+        = Scheduler.KillTopologyRequest.newBuilder()
+        .setTopologyName(TOPOLOGY_NAME)
+        .build();
+
+    scheduler.initialize(this.mockConfig, this.mockRuntime);
+
+    Job[] jobs = {new Job(), new Job(), new Job()};
+    List<Job> jobList = Arrays.asList(jobs);
+
+    // fail to kill
```
nwangtw: nit: s/fail/Fail

heron/schedulers/tests/java/BUILD
```diff
@@ -68,6 +68,26 @@ kubernetes_deps_files = [
     "//third_party/java:kubernetes-java-client",
 ]
 
+nomad_sdk_deps = [
+          "@com_hashicorp_nomad//jar",
```
nwangtw: nit: bad indent.

heron/config/src/yaml/BUILD
nwangtw: heron.aurora might be ok (maybe not ideal though) because aurora file is a config and used by aurora cli.

.sh file is worse than aurora file. :(  So I am wondering if it is possible to avoid this .sh file and have a config to be used by nomad CLI?

I don't know nomad enough and I am open to other thoughts/suggestions. Also this is not a blocker if this is not feasible.

jerrypeng: The .sh file is actually read by the scheduler and submitted to nomad as part of a job description in json format.  I can inline the contents of the .sh script in the scheduler code, but that might not be as clear to the user as just having a file with comments in it

heron/downloaders/src/java/com/twitter/heron/downloader/FileDownloader.java
jerrypeng: sure thing

heron/config/src/yaml/conf/nomad/statemgr.yaml
jerrypeng: 2181 is the default port number for zookeeper.  If I was to run whole setup (scheduler, topology, and zookeeper) on a single node, this is the default I would use.  The cli and all executors will be able to contact the local zookeeper on 127.0.0.1:2181

heron/schedulers/tests/java/BUILD
jerrypeng: will fix

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
jerrypeng: will fix

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
jerrypeng: will fix

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
jerrypeng: will fix

heron/schedulers/tests/java/com/twitter/heron/scheduler/nomad/NomadSchedulerTest.java
jerrypeng: will fix

jerrypeng:
@nwangtw thanks for the additional review!  I have addressed your comments

heron/config/src/yaml/conf/nomad/statemgr.yaml
nwangtw: got it. thx.

heron/config/src/yaml/BUILD
nwangtw: kk.

heron/executor/src/python/heron_executor.py
```diff
@@ -277,6 +280,7 @@ def parse_args(args):
     parser.add_argument("--topology-defn-file", required=True)
     parser.add_argument("--state-manager-connection", required=True)
     parser.add_argument("--state-manager-root", required=True)
+    parser.add_argument("--state-manager-config-file", required=True)
```
huijunw: config/yaml/heron.aurora needs update to add this param.

nwangtw: :( required. Is it possible to be optional?

nwangtw: I don't see any scheduler passes this flag. Why it is not causing errors?

jerrypeng: it is added in SchedulerUtils that every scheduler calls except aurora I believe

nwangtw: I see. That could be the case.

Another note is that it seems all the other schedulers have statemanager configs explicitly in the following file. It may make sense to have this new config as well, especially since it is a required field.

heron/config/src/yaml/conf/kubernetes/statemgr.yaml
heron/config/src/yaml/conf/marathon/statemgr.yaml
heron/config/src/yaml/conf/yarn/statemgr.yaml
and so on.

nwangtw: Never mind. I see what you mean. You want to override the config file itself.

nwangtw: @jerrypeng  Cant add you as reviewer. :(  Here is the PR for aurora when you get a minute. I am testing the change internally.

dadgar:
@jerrypeng I am curious why not have heron submit a job to a running Nomad cluster?

jerrypeng:
@dadgar heron can submit to an existing nomad scheduler, which is this PR focuses.  This PR allows users to submit heron jobs to existing nomad clusters.  In another PR, after this one was merged, we introduced a standalone cluster/distributed mode for heron using Nomad as the scheduler.  Before this mode was introduced, users would have to had setup Aurora, YARN, Kubernetes, etc. to run a Heron job using many nodes if they didn't already have a cluster of such available to them.  This was a high barrier for entry to users since many of these schedulers are not easy to setup, if not impossible, if you do not have sudo access.  Since Nomad is so simple to install and start up, we decided to use Nomad as the de facto scheduler/cluster manager for Heron standalone cluster mode.  The code to for setting up a standalone cluster is mainly scripts to start a nomad cluster and copying and templating files.  Like I said before, if a user already has Nomad cluster, they can absolutely use their existing deployment to run Heron jobs.

jerrypeng:
@dadgar here is some instructions on how to start a standalone cluster via nomad:

https://twitter.github.io/heron/docs/operators/deployment/schedulers/standalone/

We still have to write some docs on how to deploy to existing nomad cluster

dadgar:
@jerrypeng Ah thanks for the great answer! I saw the PR and then looked at the docs so I thought this was implementing the standalone mode! So there are just missing docs for running against an existing Nomad cluster.

That is awesome that you all chose to run the standalone cluster mode on Nomad!

jerrypeng:
@dadgar yup!

@lucperkins if you have some time let's write some docs on how to deploy to an existing nomad cluster and also a blog to announce nomad support and standalone cluster mode.

PR #2602, opened 2017-12-05
keep topology defn during update

huijunw:
We found that the topology defn is changed during update process, which causes the messy state. The original topology definition should remain same during the whole topology lifecycle, while the packing plan shows the latest topology information during update/repack.

This PR keeps topology defn during update.

Tested in local scheduler with #2600.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -147,37 +147,26 @@ public class UpdateTopologyManager implements Closeable {
     TopologyAPI.Topology topology = getTopology(stateManager, topologyName);
     boolean initiallyRunning = topology.getState() == TopologyAPI.TopologyState.RUNNING;
 
-    System.out.println("debug - getUpdatedTopology");
     // fetch the topology, which will need to be updated
     TopologyAPI.Topology updatedTopology =
         getUpdatedTopology(topologyName, proposedPackingPlan, stateManager);
 
-    System.out.println("debug - deactivateTopology");
     // deactivate and sleep
     if (initiallyRunning) {
       // Update the topology since the state should have changed from RUNNING to PAUSED
       updatedTopology = deactivateTopology(stateManager, updatedTopology, proposedPackingPlan);
     }
 
-    System.out.println("debug - addContainers");
     // request new resources if necessary. Once containers are allocated we should make the changes
     // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
     if (newContainerCount > 0 && scalableScheduler.isPresent()) {
       scalableScheduler.get().addContainers(containerDelta.getContainersToAdd());
     }
 
-    // update parallelism in updatedTopology since TMaster checks that
-    // Sum(parallelism) == Sum(instances)
-    // debug: check if topo defn update needed
-// logInfo("Update new Topology: %s",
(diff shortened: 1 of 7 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -163,10 +163,6 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
       scalableScheduler.get().addContainers(containerDelta.getContainersToAdd());
     }
 
-    // update parallelism in updatedTopology since TMaster checks that
-    // Sum(parallelism) == Sum(instances)
-    logInfo("Update new Topology: %s", stateManager.updateTopology(updatedTopology, topologyName));
```
maosongfu: If we remove the topology-update on stateManager, it seems we could remove or clean a lot of codes, like:
```
private TopologyAPI.Topology getUpdatedTopology(...)
static TopologyAPI.Topology mergeTopology(...)
private static boolean updateComponent(...)
```
And a lot of unit tests.
Can we also do that?

huijunw: removed

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -147,14 +145,10 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
 
-    // fetch the topology, which will need to be updated
-    TopologyAPI.Topology updatedTopology =
-        getUpdatedTopology(topologyName, proposedPackingPlan, stateManager);
-
     // deactivate and sleep
     if (initiallyRunning) {
       // Update the topology since the state should have changed from RUNNING to PAUSED
-      updatedTopology = deactivateTopology(stateManager, updatedTopology, proposedPackingPlan);
+      deactivateTopology(stateManager, topology, proposedPackingPlan);
```
maosongfu: Good to have a comment here saying:
`Will throw exceptions internally if failed to xxxx`

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -177,7 +167,7 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
       // delete the packing plan. Instead we could message tmaster to invalidate the physical plan
       // and/or possibly even update the packing plan directly
       SysUtils.sleep(Duration.ofSeconds(10));
-      reactivateTopology(stateManager, updatedTopology, removableContainerCount);
+      reactivateTopology(stateManager, topology, removableContainerCount);
```
maosongfu: Good to have a comment here saying:
`Will throw exceptions internally if failed to xxxx`

maosongfu:
Also, plz make sure no unit tests will break

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: done

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: done

PR #2608, opened 2017-12-08
make a single heron package for each platform, removed api, tools, cl…

kramasamy:
…ient and instead a single package called heron


scripts/packages/BUILD
```diff
@@ -50,19 +48,15 @@ filegroup(
 filegroup(
     name = "tarpkgs",
     srcs = [
-        ":heron-api",
         ":heron-core",
-        ":heron-tools",
-        ":heron-client",
+        ":heron",
```
nwangtw: Hmm. API is removed? This will affect Twitter internal use case. Is it possible for you to add it back?

PR #2610, opened 2017-12-08
Make single linux binary

aahmed-se:


Code0x58:
It looks like the changes were merged elsewhere, or the is practically a null change with a PR that doesn't match the change, so I am guessing this is a candidate for closing.

aahmed-se:
This can be closed I didn't get a chance to a proper PR , we are still making different binaries for centos and ubuntu we will keep that for now.

PR #2611, opened 2017-12-09
Multi node standalone cluster for heron

jerrypeng:
I have added scripts and features to allow users to easily stand up a mutli node Heron cluster.

Users can start a cluster using the heron CLI as follows

```
# Set Roles
$ heron standalone set masters
$ heron standalone set slaves
$ heron standalone set zookeepers # users will need to start their own zookeeper cluster
# Start cluster
$ heron standalone cluster start
# Submit topology
$ heron submit standalone  ~/workspace/heron/bazel-bin/examples/src/java/api-examples-unshaded_deploy.jar com.twitter.heron.examples.api.AckingTopology test1
# Kill topology
$ heron kill standalone test1
# Stop Cluster
$ heron standalone cluster stop
```

Add an Http uploader for uploading topology packages.
Add endpoints to API Server so that topology packages can be uploaded to it and dowloaded by executors 

@kramasamy contributed to this PR by implementing the packaging of the Nomad in the heron install script


heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
```diff
@@ -310,8 +310,10 @@ Task getTask(String taskName, int containerIndex,
         throw new RuntimeException(msg, e);
       }
       Map<String, String> metaData = jobActual.getMeta();
-      if (metaData.get(NomadConstants.NOMAD_TOPOLOGY_NAME).equals(topologyName)) {
-        ret.add(jobActual);
+      if (metaData != null) {
+        if (metaData.get(NomadConstants.NOMAD_TOPOLOGY_NAME).equals(topologyName)) {
```
cckellogg: could metaData.get here ever return null?

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -0,0 +1,153 @@
...
+    return Config.toLocalMode(builder.build());
+  }
+
+  // save uploaded file to new location
+  private void writeToFile(InputStream uploadedInputStream,
+                           String uploadedFileLocation) {
+    File file = new File(uploadedFileLocation);
+    file.getParentFile().mkdirs();
+
+    try {
```
cckellogg: Maybe use a util here? FileHelper.copy(in, path)?

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -0,0 +1,153 @@
...
+  @Path("/upload")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  public Response uploadFile(
+      @FormDataParam("file") InputStream uploadedInputStream,
+      @FormDataParam("file") FormDataContentDisposition fileDetail) {
+
+    Config config = createConfig();
+
+    if (uploadedInputStream == null) {
+      return Response.status(Response.Status.BAD_REQUEST).build();
```
cckellogg: Can we add some useful error messages here? There are some private helper methods (that could be refactored out) in the TopologyResources to help with this.

heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
jerrypeng: Yes if there is some other nomad job running that does not have this user defined meta data set e.g. API server

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
jerrypeng: I can move the function to FileHelper but the existing copy only copies files already on disk

jerrypeng:
@cckellogg thanks for the review! I have addressed your comments

aahmed-se:
@kramasamy @srkukarni 
We tested this is multi node virtual machine env , there is problem in detection of localhost execution whereas if we want to run the installer on a node and still make it part of the cluster, things don't work as expected it needs to be resolved before we merge the pr.

heron/config/src/yaml/BUILD
```diff
@@ -14,7 +14,11 @@ filegroup(
 
 filegroup(
     name = "conf-yaml",
-    srcs = glob(["conf/**/*.yaml"] + ["conf/**/*.aurora"] + ["conf/**/*.sh"]),
+    srcs = glob(["conf/**/*.yaml"]
```
nwangtw: nit: suggest moving "[conf/**.*.yaml"] to next line, and suggest sorting them.

heron/config/src/yaml/conf/standalone/roles/master_servers.txt
```diff
@@ -0,0 +1 @@
+# Please provide a list of Master servers. Please provide at least one server.
```
nwangtw: suggest adding an example here.

heron/config/src/yaml/conf/standalone/roles/slave_servers.txt
```diff
@@ -0,0 +1 @@
+# Please provide a list of slave servers. Please provide at least one server.
```
nwangtw: suggest adding an example here.

heron/config/src/yaml/conf/standalone/roles/zookeeper_servers.txt
```diff
@@ -0,0 +1 @@
+# Please provide a list of zookeeper servers (optionally with the port to contact ZK on. Default if 2181).
```
nwangtw: suggest adding an example here.

heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
```diff
@@ -310,8 +310,10 @@ Task getTask(String taskName, int containerIndex,
         throw new RuntimeException(msg, e);
       }
       Map<String, String> metaData = jobActual.getMeta();
-      if (metaData.get(NomadConstants.NOMAD_TOPOLOGY_NAME).equals(topologyName)) {
-        ret.add(jobActual);
+      if (metaData != null && metaData.containsKey(NomadConstants.NOMAD_TOPOLOGY_NAME)) {
```
nwangtw: Maybe a stupid question, can we use if (metaData != null && metaData.get(NomadConstants.NOMAD_TOPOLOGY_NAME) == topologyName) here?

Also if we need to break it into two if statements, I think if (metaData != null ) { if (...) } might be better.

heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
```diff
@@ -356,12 +358,12 @@ static void startJobs(NomadApiClient apiClient, Job... jobs) {
...
-        LOG.info("response: " + response);
+        LOG.fine("response: " + response);
       } catch (IOException | NomadException e) {
         String msg = "Failed to submit job "
-            + job.getId() + " with error: " + e.getMessage()
-            + " with job spec: " + job.toString();
-        LOG.log(Level.SEVERE, msg, e);
+            + job.getId() + " with error: " + e.getMessage();
+
+        LOG.log(Level.FINE, msg + " with job spec: " + job.toString());
```
nwangtw: 1. Why change to FINE? This error looks serious.
2. why append strings in two places? One in msg =, another in LOG.log()?

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/utils/FileHelper.java
```diff
@@ -110,6 +111,23 @@ private static void addFileToArchive(TarArchiveOutputStream archiveOutputStream,
     }
   }
 
+  // save uploaded file to new location
+  public static void writeToFile(InputStream uploadedInputStream,
+                           String uploadedFileLocation) throws IOException {
```
nwangtw: Just to confirm that this indent is expected or not.

heron/tools/cli/src/python/main.py
```diff
@@ -82,7 +83,8 @@ def get_command_handlers():
       'restart': restart,
       'submit': submit,
       'update': update,
-      'version': version
+      'version': version,
+      'standalone': standalone
```
nwangtw: sort

heron/tools/cli/src/python/main.py
```diff
@@ -330,11 +331,15 @@ def execute(handlers, local_commands):
...
+  # set log level
+  log.set_logging_level(command_line_args)
+  Log.debug("Input Command Line Args: %s", command_line_args)
+
+# command to be execute
   command = command_line_args['subcommand']
   is_local_command = command in local_commands
 
-  if command == 'version':
+  if command == 'version' or command == 'standalone':
```
nwangtw: suggest to put the standalone command to the local command list, instead of here as an or operation. It is  more complicated than the version command.

See below in main() function,
local_commands = ('help', 'version', 'config')

heron/config/src/yaml/conf/standalone/roles/master_servers.txt
jerrypeng: will add

heron/config/src/yaml/conf/standalone/roles/slave_servers.txt
jerrypeng: will add

heron/config/src/yaml/conf/standalone/roles/zookeeper_servers.txt
jerrypeng: will add

heron/config/src/yaml/BUILD
jerrypeng: will do

heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
jerrypeng: We cannot do "metaData.get(NomadConstants.NOMAD_TOPOLOGY_NAME) == topologyName)" because first in java to compare strings you have to your .equals instead of "==".  "==" will check if its the same instance rather if it has to same value.  If you use .equals we need to make sure that its not null or NPE will be thrown.

heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
jerrypeng: To answer you first question:

The job spec can be quite long that is why I would rather print it explicitly if the user specified --verbose via the command line.

To answer your second question:

The reason why I append if so that the RuntimeException thrown only contains the error message and not the whole job spec but I would log the error message and the job spec if verbose is set

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/utils/FileHelper.java
jerrypeng: I will fix the incorrect indent

heron/tools/cli/src/python/main.py
jerrypeng: will do

jerrypeng:
@nwangtw thanks for the review!  I have addressed your comments

PR #2612, opened 2017-12-09
Add back api packaging

nwangtw:
It could be easier to have an api package for releasing API as a library.

### DIFF
--- a/scripts/packages/BUILD
+++ b/scripts/packages/BUILD
@@ -2,6 +2,8 @@ package(default_visibility = ["//visibility:public"])
 
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")
 
+load("/tools/rules/heron_api", "heron_api_files")
+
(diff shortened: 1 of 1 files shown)


kramasamy:
@nwangtw - do you need this target inside Twitter?

nwangtw:
Yeah. It could be helpful.

In fact I have a walkaround without this target in oss heron, but I feel it could be useful for other users as well.

kramasamy:
@nwangtw - currently we publish the jars to maven automatically. Furthermore, python api's are being published to pypi. Hence heron api with just tar balls of jars might not make sense.

nwangtw:
So all other modules are tar balls but APIs are using different language specific packages? Not really a problem for me, but I still feel adding the python api files into the api tar ball makes more sense then keep all api files as is. I think your concern would be that users would have to build all of languages by default.

Anyway, I am ok to leave the targets as is and drop this change. We can revisit the question if other users have more questions about it.

Thanks again for taking a look!

PR #2617, opened 2017-12-12
Fix api server create config cluster mode.

cckellogg:
- Use load config with cluster mode if we are not running locally.


cckellogg:
@jerrypeng will this effect nomad?

jerrypeng:
@cckellogg ya it will since the FileResource should resolve ${HOME}/.herondata/repository/topologies/ to be ~/.herondata/repository/topologies/ which I believe is what is resolved for local cluster.  How was the all working before?  What changed to cause the break?

jerrypeng:
nvm I think this is ok +1

PR #2627, opened 2017-12-16
[Streamlet] Reinforce usage of StreamletNamePrefixes for more type-safety

erenavsarogullari:

- Reinforce usage of `StreamletNamePrefixes` for more type-safety
- Small fix for the documentation.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
+++ b/heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
@@ -157,7 +157,7 @@ public abstract class StreamletImpl<R> implements Streamlet<R> {
    * @param prefix The name prefix of this streamlet
    * @param stageNames The collections of created streamlet/stage names
    */
-  protected void setDefaultNameIfNone(String prefix, Set<String> stageNames) {
+  protected void setDefaultNameIfNone(StreamletNamePrefixes prefix, Set<String> stageNames) {
     if (getName() == null) {
       setName(defaultNameCalculator(prefix, stageNames));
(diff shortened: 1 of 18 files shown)


erenavsarogullari:
cc @lucperkins @srkukarni

heron/api/src/java/com/twitter/heron/streamlet/impl/StreamletImpl.java
```diff
@@ -157,7 +157,7 @@ public String getName() {
    * @param prefix The name prefix of this streamlet
    * @param stageNames The collections of created streamlet/stage names
    */
-  protected void setDefaultNameIfNone(String prefix, Set<String> stageNames) {
+  protected void setDefaultNameIfNone(StreamletNamePrefixes prefix, Set<String> stageNames) {
```
lucperkins: Could you rename this enum to `StreamletNamePrefix`? I think it makes more sense that way.

erenavsarogullari: Firstly, thanks for review.
Sure, what about `StreamletNamePrefixType` by emphasizing type? In general, most of the enums have type suffix (e.g: `JoinType`)

lucperkins: In this case I would say no, because you’re not specifying a type per se but rather just providing a string

erenavsarogullari: Addressed and ready for re-review ;)

PR #2630, opened 2017-12-19
Add new state-manager-config-file argument for aurora

nwangtw:


jerrypeng:
@nwangtw do you need to add STATEMGR_YAML to:

https://github.com/twitter/heron/blob/master/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java#L211

as well?

nwangtw:
Yeah. You are right. Need to update a few other files as well.

jerrypeng:
+1

PR #2631, opened 2017-12-20
Add python encoding

nwangtw:
Most python files are not specifying python version and encoding.

Added these lines at beginning of all python files I can find.

#!/usr/bin/env python2.7
# -*- encoding: utf-8 -*-


srkukarni:
One effect of this change is that it forces python2.7. One of the significant efforts that we have made is for us to be compatible with python3. This pr is at odds with that effort

nwangtw:
I see. The target of the PR is to add the encoding so I am open to suggestions on the python version.

- it might be better to be consistent across the code base I think. But it may not be a big deal?
- With explicitly version number we may be able to avoid some unexpected behavior and it could be very hard to debug (I don't know python that well and don't know if it is a good enough reason though).

Anyway, if you think "#!/usr/bin/env python" is better, I can make the change.

srkukarni:
IMO there are two issues here that possibly should be separated
1. Python version
2. encoding.
Can we fix just the encoding here and put the other one for discussion?

nwangtw:
@srkukarni sure, let me do it.

nwangtw:
@srkukarni updated.

added "#!/usr/bin/env python" for all files that are missing the line, kept the 2.7 version number for all that had specified it and moved the env line to the beginning of every file. Basically version specification should be the same as before.

PR #2632, opened 2017-12-21
Refactor aurora scheduler to use SchedulerUtils to construct arguments

nwangtw:
Refactor aurora scheduler code to be more consistent with other schedulers.

Topology arguments are hardcoded in aurora config file currently instead of using the shared functions in SchedulerUtils. It is hard to maintain and it is causing compatibility issue (since config file is in client and the setup code is in core).

fyi @jerrypeng


nwangtw:
Christmas cleaning!

PR #2633, opened 2017-12-21
 Refactoring and improving standalone mode

jerrypeng:
- Put the standalone mode cli into a separate heron-admin cli
- allow users to submit topologies to standalone cluster via apiserver
- Users can submit a topology via a service-url or without.  If you do not provide a service url the configs in ~/.heron/conf/standalone needs to correspond to the cluster.  These configs will be set for you if you start a standalone cluster.
- Being able to user a service-url allows users to easy share access with other users since all you need to submit a topology to the cluster is the service-url

To set cluster roles:

`heron-admin standalone set`

To start cluster:

`heron-admin standalone cluster start`

To get service-url:

`heron-admin standalone get service-url`

To submit topology with service-url:

`heron submit --service-url=http://10.128.0.5:9000 standalone  ~/heron/bazel-bin/examples/src/java/api-examples-unshaded_deploy.jar com.twitter.heron.examples.api.AckingTopology --verbose test3`

To submit topology without service-url:

`heron submit standalone  ~/heron/bazel-bin/examples/src/java/api-examples-unshaded_deploy.jar com.twitter.heron.examples.api.AckingTopology --verbose test5`

To kill:

`heron kill --service-url=http://10.128.0.5:9000 standalone test2`

or 

`heron kill standalone test2`

Stop Cluster:

`heron-admin standalone cluster stop`

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
@@ -47,6 +47,7 @@ import com.twitter.heron.scheduler.UpdateTopologyManager;
 import com.twitter.heron.scheduler.utils.Runtime;
 import com.twitter.heron.scheduler.utils.SchedulerUtils;
 import com.twitter.heron.spi.common.Config;
+import com.twitter.heron.spi.common.ConfigLoader;
 import com.twitter.heron.spi.common.Context;
 import com.twitter.heron.spi.packing.PackingPlan;
 import com.twitter.heron.spi.scheduler.IScheduler;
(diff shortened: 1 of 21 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/nomad/NomadScheduler.java
```diff
@@ -64,7 +65,12 @@
   @Override
   public void initialize(Config config, Config runtime) {
     this.localConfig = config;
-    this.clusterConfig = Config.toClusterMode(config);
+    this.clusterConfig = Config.toClusterMode(Config.newBuilder()
```
nwangtw: It seems cluster config is used only for downloading topology after this change? If this is true, maybe add comment or rename variable. The variable itself gives me an impression that there are a local mode and a cluster mode.

heron/tools/admin/src/python/main.py
```diff
@@ -0,0 +1,178 @@
+# Copyright 2016 Twitter. All rights reserved.
```
nwangtw: s/2016/2017.

This is a new tool? Is there a doc about it?

jerrypeng: There isn't currently.  Will right one in the future, but the you can always type "heron-admin help" and it will give you a detailed explanation of all commands and subcommads

PR #2643, opened 2017-12-26
Add a --heron-home option into heron-cli to specify lib/config directory

nwangtw:
Inside Twitter we need to support multi heron version and we need to be able to override paths to heron libs/configs in the internal tools.

This PR just refactor the two get_*_heron_dir() functions slightly so that the paths are available to be overridden beforehand if needed.


heron/tools/common/src/python/utils/config.py
```diff
@@ -58,6 +58,12 @@
 ROLE_KEY = "role.required"
 ENVIRON_KEY = "env.required"
 
+# global variabls that can be overriden when necessary.
+# Note that overriden has to happen before the first call
```
huijunw: who overrides them

nwangtw: Users who want to customize the tools.

nwangtw: I can also add an option to override heron client data files in cli.

nwangtw:
I am planning to add an option in heron client to specify this path. It could be useful for all users and it doesn't have side effects.

nwangtw:
Probably the last PR for this year~

nwangtw:
Dropping this PR in favor of easier installation of older build. Heron tools package is self contained basically so installation is easy.

PR #2649, opened 2018-01-02
bookkeeper using emptyDir

kramasamy:


mjschmidt:
What are these changes for? What does using emptyDir mean?

kramasamy:
@mjschmidt - emptyDir is a stateful set concept that k8s uses. When the pod migrates to another machine it gets a fresh empty directory.

mjschmidt:
@kramasamy @sijie That is awesome! I love that! Great work guys!

PR #2651, opened 2018-01-02
Add wait for exit function in WakeableLooper and fix exiting code in SocketChannelHelper

nwangtw:
Add waitForExit() function in WakeableLooper and use it to avoid racing condition in SocketChannelHelper's forceFlushWithBestEffort() function.

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
+++ b/heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
@@ -21,6 +21,7 @@ import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 import com.twitter.heron.common.basics.ByteAmount;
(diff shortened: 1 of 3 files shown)


heron/common/src/java/com/twitter/heron/common/network/SocketChannelHelper.java
```diff
@@ -213,6 +214,10 @@ public void write() {
   // Force to flush all data in underneath buffer queue to socket with best effort
   // It is most likely happen when we are handling some unexpected cases, such as exiting
   public void forceFlushWithBestEffort() {
+    looper.exitLoop();
+    // Wait for NIO loop to confirm stopping process.
+    looper.waitForExit(10, TimeUnit.SECONDS);
```
huijunw: what happens if looper does not stop after 10 seconds

nwangtw: The function stops waiting and returns false.

In this case, the code works pretty much the same way as before after the waiting. Except exitLoop() set a flag to true so the looper won't do any real work afterwards.

PR #2652, opened 2018-01-02
added a quick fix

kramasamy:


mjschmidt:
@kramasamy what does the empty dir bookkeeper do?

Issue #2647, opened 2017-12-31
Particular Windowing use case- Are EvictionPolicies and TriggerPolicies Pluggable?


dancollins34:
Hello,

So, I'm trying to figure out a bit of a different windowing methodology than by count or timestamp.  While looking through the codebase, I found the com.twitter.heron.api.windowing.{triggers, evictors} package, and was wondering whether there is support for custom EvictionPolicies and TriggerPolicies?

First, some background on the use case in case I am going about this in an overly complex way.  I am looking to build a topology which will accurately process events with in-event timestamps coming roughly in order, but with a possible overall event lag and some slightly out of order timestamps.  In addition, the use case would require reprocessing of the same events at a later date, hopefully with a faster throughput than they originally arrived.

Looking through the current windowing policies, it doesn't look like any of the 4 would allow for both of these use cases to be processed with the same topology in a similar way.  Instead, what I am attempting to do is write a trigger/eviction policy with three parameters: A window periodicity (i.e. 1 second, 2 seconds, 1 minute, etc.), a number of windows to wait before evicting a given window(i.e. wait for events with timestamps in the next two windows of timestamps in the 12:30:20-12:30:29.99 and 12:30:30-12:30:39.99 to arrive before processing the 12:30:10-12:30:19.99 window), and a default real-time timeout in case no future events arrive.

I've attempted to implement these policies, and have attached draft code below.  However, I'm unable to find a way to actually use these policies with either the topology or (preferrably) streamlet apis without changes to the source code.  Which brings me to my actual question- is there a way to use a custom EvictionPolicy and TriggerPolicy for a bolt window?

Thank you,

dancollins34

TriggerPolicy
```java
package com.twitter.heron.api.windowing.triggers;

import java.io.Serializable;
import java.util.Iterator;
import java.util.Map;
import java.util.NoSuchElementException;
import java.util.TreeMap;

import com.twitter.heron.api.windowing.DefaultEvictionContext;
import com.twitter.heron.api.windowing.Event;
import com.twitter.heron.api.windowing.TriggerHandler;
import com.twitter.heron.api.windowing.TriggerPolicy;
import com.twitter.heron.api.windowing.evictors.StagesTimeoutEvictionPolicy;


public class StagesTimeoutTriggerPolicy<T extends Serializable> implements TriggerPolicy<T, StagesTimeoutTriggerPolicy.StagesTimeoutTracker>{
    class StagesTimeoutTracker{
[...]

PR #2653, opened 2018-01-03
Add support for custom evictor and trigger policies

dancollins34:
Hello,

Pursuant to my question in #2647 , I've added support for custom eviction and trigger policies.  The modifications that I have made is to:

-Add the ability to add a custom EvictionPolicy and TriggerPolicy to a WindowingConfigs
-Add an if statement to check for the existence of custom policies in the WindowingConfigs passed to a WindowedBoltExecutor
-Add the ability to add a custom trigger or evictor policy to a BaseWindowedBolt
-Add support for a custom trigger and evictor policy WindowConfig and WindowConfigImpl in the streamlet api

Thoughts? This is my first time interacting with the underlying api, but I tried to keep the methods/implementation as similar as possible to what already existed.

Daniel

-------------

Add support in both the streamlet and topology apis for the usage of custom eviction policies extending com.twitter.heron.api.windowing.EvictionPolicy and trigger policies extending com.twitter.heron.api.windowing.TriggerPolicy to enable user-defined windowing schemes.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
+++ b/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
@@ -238,15 +238,17 @@ public class WindowedBoltExecutor implements IRichBolt,
     validate(topoConf, windowLengthCount, windowLengthDurationMs, slidingIntervalCount,
         slidingIntervalDurationMs);
 
-    if(topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR)){
-      evictionPolicy = (EvictionPolicy<Tuple, ?>) topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR);
-    }else{
+    if (topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR)) {
+      evictionPolicy = (EvictionPolicy<Tuple, ?>)
+              topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR);
+    } else {
       evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
     }
 
-    if(topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER)){
-      triggerPolicy = (TriggerPolicy<Tuple, ?>) topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER);
-    }else{
+    if (topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER)) {
+      triggerPolicy = (TriggerPolicy<Tuple, ?>)
+              topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER);
+    } else {
       triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
               evictionPolicy, topoConf);
     }
--- a/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
+++ b/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
@@ -250,9 +250,9 @@ public class WindowedBoltExecutor implements IRichBolt,
       validate(topoConf, windowLengthCount, windowLengthDurationMs, slidingIntervalCount,
               slidingIntervalDurationMs);
 
+      evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
       triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
               evictionPolicy, topoConf);
-      evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
     } else {
       throw new IllegalArgumentException(
               "If either a custom TriggerPolicy or EvictionPolicy is defined, both must be."
--- a/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
+++ b/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
@@ -237,9 +237,21 @@ public class WindowedBoltExecutor implements IRichBolt,
     // validate
     validate(topoConf, windowLengthCount, windowLengthDurationMs, slidingIntervalCount,
         slidingIntervalDurationMs);
-    evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
-    triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
-        evictionPolicy, topoConf);
+
+    if(topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR)){
+      evictionPolicy = (EvictionPolicy<Tuple, ?>) topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_EVICTOR);
+    }else{
+      evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
+    }
+
+    if(topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER)){
+      triggerPolicy = (TriggerPolicy<Tuple, ?>) topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER);
+    }else{
+      triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
+              evictionPolicy, topoConf);
+    }
+
+
     manager.setEvictionPolicy(evictionPolicy);
     manager.setTriggerPolicy(triggerPolicy);
     // restore state if there is existing state
--- a/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
+++ b/heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
@@ -251,13 +251,17 @@ public class WindowedBoltExecutor implements IRichBolt,
               slidingIntervalDurationMs);
 
       evictionPolicy = getEvictionPolicy(windowLengthCount, windowLengthDurationMs);
-      triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
-              evictionPolicy, topoConf);
+      triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs);
     } else {
       throw new IllegalArgumentException(
               "If either a custom TriggerPolicy or EvictionPolicy is defined, both must be."
       );
     }
+
+    triggerPolicy.setEvictionPolicy(evictionPolicy);
+    triggerPolicy.setTopologyConfig(topoConf);
+    triggerPolicy.setTriggerHandler(manager);
+    triggerPolicy.setWindowManager(manager);
 
     manager.setEvictionPolicy(evictionPolicy);
     manager.setTriggerPolicy(triggerPolicy);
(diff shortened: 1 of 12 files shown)


srkukarni:
@jerrypeng

dancollins34:
As an aside, the travis build appears to be failing due to style issues.  I can make those changes, but I'm unsure if I'm allowed to sign the CLA, and therefore contribute, given my current terms of employment.  I'll find that out tomorrow.

srkukarni:
@dancollins34 I believe the reason why travis is failing is checkstyle issues.
If you are building on mac, please use 
bazel build --config=darwin heron/...
this command should also throw the above check style issues

dancollins34:
Thank you for the advice. I've removed the style issues.

heron/api/src/java/com/twitter/heron/api/windowing/WindowingConfigs.java
```diff
@@ -35,6 +35,9 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import com.twitter.heron.api.tuple.Tuple;
+
+
```
jerrypeng: nit.  Please remove extra space

heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
```diff
@@ -237,9 +237,23 @@ protected void validate(Map<String, Object> topoConf, Count windowLengthCount, L
...
+
+    if (topoConf.containsKey(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER)) {
+      triggerPolicy = (TriggerPolicy<Tuple, ?>)
+              topoConf.get(WindowingConfigs.TOPOLOGY_BOLTS_WINDOW_CUSTOM_TRIGGER);
+    } else {
+      triggerPolicy = getTriggerPolicy(slidingIntervalCount, slidingIntervalDurationMs, manager,
+              evictionPolicy, topoConf);
+    }
+
+
```
jerrypeng: Place remove extra line

kramasamy:
It will be nice to provide some examples of how to use this feature.

heron/api/src/java/com/twitter/heron/api/bolt/WindowedBoltExecutor.java
```diff
@@ -237,9 +237,23 @@ protected void validate(Map<String, Object> topoConf, Count windowLengthCount, L
     // validate
     validate(topoConf, windowLengthCount, windowLengthDurationMs, slidingIntervalCount,
```
jerrypeng: If a custom trigger or evictor, validate should not be called.  Arguments would be directly passed into the custom evictor and trigger when they are intialized

jerrypeng:
Generally looks good, but I would like to bring a of point for discussion:

Even when using custom triggers and evictors, should users still be be required to use the existing API (e.g.  setTopologyBoltsWindowLengthDurationMs(long value) in WindowingConfigs) to set window lengths and sliding intervals?  or should such parameters be passed directly into the custom trigger and evictors objects e.g. passing these parameter via their respective constructors?  I can see pros and cons in each approach

dancollins34:
I don’t think there’s any way you could prescribe the universe of all possible windowing parameters in that way. I think it makes a lot more sense to make the extent of the requirements to be “subclass triggerPolicy/evictionPolicy” instead of trying to restrict the subset of what each of those could be.

To your comment about validate, I think it might make sense to make this a 3 condition if statement. I.E. if eviction and trigger policy are defined in the config, use them. If neither eviction nor trigger policy are defined, run validate and run the code currently in the master branch, and if only one of them is defined, raise an exception.

I’ll fix the blank lines, make this above change and try to write some example code tonight.

kramasamy:
@jerrypeng - can you please re-review again?

dancollins34:
So, I've noticed something else about TriggerPolicies that I had missed on my first go round.  There is no way for the API user to actually instantiate an instance of a custom TriggerPolicy to put it into the WindowConfig object because they would first need to have a reference to the WindowManager to act as a TriggerPolicy.  I'm writing a commit attempting to address this now.

kramasamy:
@dancollins34 - some errors have been occurring in the unit tests. Can you please fix them?

dancollins34:
Built on my machine that’s weird. I’ll fix it tonight.

dancollins34:
Looks like it just timed out. Re running build

dancollins34:
Yes, there are failures. Attempting to address them now.

dancollins34:
It looks like this one failed on a c++ test? I’m not sure how that’s possible I haven’t modified the c++ code in any way.

dancollins34:
Example usage

```java
public class RandomTriggerPolicy<T extends Serializable> extends AbstractBaseTriggerPolicy<T, Boolean> {
    Float triggerChance;
    Random sampler;

    public RandomTriggerPolicy(Float triggerChance){
        super(false, false, false);

        if(triggerChance > 1.0 || triggerChance < 0.0){
            throw new IllegalArgumentException("Cannot have a likelihood outside of 0 to 1 range");
        }

        this.triggerChance = triggerChance;
        this.sampler = new Random();
    }

    @Override
    public void track(Event<T> event) {
        if (started && !event.isWatermark()) {
            if (sampler.nextDouble() < triggerChance){
                this.handler.onTrigger();
            }
        }
    }

    @Override
    public void reset() {
        // NOOP
    }

    @Override
    public void shutdown() {
        // NOOP
    }

    @Override
    public Boolean getState() {
        return false;
    }

    @Override
    public void restoreState(Boolean state) {
        // NOOP
    }
}

public class AllEvictionPolicy<T extends Serializable> implements EvictionPolicy<T, Boolean> {
    private EvictionContext context;

    @Override
    public Action evict(Event<T> event) {
        return Action.PROCESS;
    }

    @Override
    public void track(Event<T> event) {
        // NOOP
    }

    @Override
    public void setContext(EvictionContext context) {
        this.context = context;
    }

    @Override
    public EvictionContext getContext() {
        return context;
    }

    @Override
    public void reset() {
        // NOOP
    }

    @Override
    public Boolean getState() {
        return true;
    }

    @Override
    public void restoreState(Boolean state) {
        // NOOP
    }
}

public class Main{
    public static void main(String[] args){
        Builder builder = Builder.newBuilder();

        builder.newSource(() -> 1).reduceByKeyAndWindow(
            value -> value,
            value -> value,
            WindowConfig.CustomWindow(new RandomTriggerPolicy<>(.5f), new AllEvictionPolicy<>()),
            (x, y) -> x + y
        ).log();
    }
}
```

srkukarni:
Not sure what the build failure was. Have kicked the build again.

kramasamy:
@dancollins34 - can you add examples of how to use these under examples/ and storm-compatibility-examples? This will help developers give same code that they can understand and also quickly copy and modify if needed.

heron/api/src/java/com/twitter/heron/api/windowing/TriggerPolicy.java
```diff
@@ -80,4 +80,32 @@
...
+   * @param triggerHandler the trigger handler
+   */
+  void setTriggerHandler(TriggerHandler triggerHandler);
+
+  /**
+   * Sets the window manager that uses this trigger policy
+   *
+   * @param windowManager the window manager
+   */
+  void setWindowManager(WindowManager<T> windowManager);
```
jerrypeng: I think we should remove this from the interface.  Having this interface is confusing to users since WindowManager implements the TriggerHandler interface.  WindowManager is also not a interface that users can implement thus it shouldn't be part of an interface that users will have to implement. I think having "setTriggerHandler" should suffice.  For build in triggers (e.g. WatermarkCountTriggerPolicy, WatermarkTimeTriggerPolicy), we can just pass in the WindowManager via their constructor.

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.Serializable;
+import java.util.Map;
+
+import com.twitter.heron.api.windowing.EvictionPolicy;
+import com.twitter.heron.api.windowing.TriggerHandler;
+import com.twitter.heron.api.windowing.TriggerPolicy;
+import com.twitter.heron.api.windowing.WindowManager;
+
+
+public abstract class AbstractBaseTriggerPolicy<T extends Serializable, S>
```
jerrypeng: Not sure if this interface is necessary, if we remove WindowManager from the TriggerPolicy interface

jerrypeng:
@dancollins34 thanks for your work!  Generally looks good.  I just have few comments for you

heron/api/src/java/com/twitter/heron/api/windowing/TriggerPolicy.java
dancollins34: The reasoning behind this was that the WindowManager, TriggerHandler (which is the WindowManager) and TopologyConfig are three things the users may not have access to directly.  I was aiming to provide access to the same structures used in the internal TriggerPolicies to custom TriggerPolicies, should users wish to use them.  The idea was that, if someone wanted to, it should be possible to use the existing trigger policies in this manner.

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
dancollins34: Regardless, you would still need a way to get the TriggerHandler (WindowPolicy) attached to the custom TriggerPolicy that was passed to the config.  AbstractBaseTriggerPolicy was just made as a way to standardize how these attachments were handled for the internal classes, and provide a way to use those same attachment variables in custom classes.

dancollins34:
Will address @kramasamy comments and add to /examples following resolution of concerns raised by @jerrypeng

kramasamy:
@dancollins34 - it will be nice if you could add some documentation on how to use this in a separate PR.

jerrypeng:
@dancollins34 thanks for replying to my comments.  I had a question regarding the example you provided above for a custom trigger.  The TriggerHandler is never set but you call the handler in the method track: "this.handler.onTrigger();" How is that going to work? or am I missing something?

heron/api/src/java/com/twitter/heron/api/windowing/TriggerPolicy.java
jerrypeng: Looking more closely at the code I am confused how the a custom trigger policy will work.  All the existing trigger policies require the WindowManager to actually trigger windows. How will a user implementing a custom TriggerPolicy do this.  They have no reference to the WindowManager.  Shouldn't the windowManager be passed into the custom trigger implicitly?  As of right now, trigger polices can't trigger windows without the WindowManager.  That is why I suggested removing the setWindowManager interface, since trigger policies need the window manager regardless of what kind of trigger policy it is.

dancollins34: Take a look at the changes i’ve made to windowedBoltExecutor.  Essentially, in order to allow these to attach to arbitrary triggerPolicies, Ive replaced their inclusion in the constructor with the setters you see in the interface for all trigger policies.

dancollins34:
Yes it is- in the windowedBoltExecutor. Take a look at those changes.

dancollins34:
@jerrypeng

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
```diff
@@ -0,0 +1,128 @@
...
+  protected Map<String, Object> topoConf;
+
+  private boolean requiresEvictionPolicy = false;
+  private boolean requiresWindowManager = false;
+  private boolean requiresTopologyConfig = false;
+
+  /**
+  * Set the requirements in the constructor
+  */
+  public AbstractBaseTriggerPolicy(boolean requiresEvictionPolicy,
```
jerrypeng: Not sure if we need these checks.  I think it should be up to the user to make sure to pass in the correct objects he or she needs for a custom trigger.  Having to pass these booleans into the constructor also makes the interface confusing to some reading to code.  If you really want these checks in place, I would suggest using setters to set these.  Then it is clear to the reader of the code what is being set.

heron/api/src/java/com/twitter/heron/api/windowing/TriggerPolicy.java
jerrypeng: I see

jerrypeng:
@dancollins34 thanks for the clarification.  I am +1 on this PR

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
dancollins34: @jerrypeng These checks were in place to keep the existing trigger policies as restrictive as they were before wrt which of the objects were attached.  I could provide an alternate no-arg constructor that sets them all to false, then use setters instead?  However, since this is the abstract class which already partially implements the TriggerPolicy interface, I would think it would be okay to perform some checks, but I have no issue with removing them either.

kramasamy:
@jerrypeng - Can I merge this?

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
jerrypeng: lets remove them to keep the interface as simple and clean as possible

jerrypeng:
@kramasamy we can once the last comment is addressed

heron/api/src/java/com/twitter/heron/api/windowing/triggers/AbstractBaseTriggerPolicy.java
dancollins34: @jerrypeng will do

jerrypeng:
+1

@kramasamy we can merge

PR #2657, opened 2018-01-06
Fixing  the slurm scheduler

supunkamburugamuve:
Fixing the slurm scheduler, the shard id is generated by the slurm.sh script, so we are not including that in the command generated from the java code.


supunkamburugamuve:
Working on the test error

kramasamy:
@supunkamburugamuve - is this ready to go?

supunkamburugamuve:
Yes, it is ready.

PR #2659, opened 2018-01-12
Use the minimal ram size for components instead of the last one

nwangtw:
Currently the component ram size calculation assumes all instances for a component use the same amount of memory, when generating component ram map. However this is not true for some packingplan algorithms, like RoundRobin. As the result, the bigger ram size can be used and the total memory used by instances in more crowded container can be more than the container ram size.

Example:
each container has 2G for instances.

container 1 has 1 "spout" and 1 "bolt", the ram size is 1G per instance
container 2 has 1 "bolt", the ram size is 2G per instance
The component ram map might be "spout:1g,bolt:1g" or "spout:1g,bolt:2g". In the second case, the instances in container 1 would use 3G memory, because the instances in the container would have the following jvm arguments:
spout 1: "-xmx 1g"
bolt 1: "-xmx 2g"

### DIFF
--- a/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
+++ b/heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
@@ -123,11 +123,15 @@ public class PackingPlan {
    * @return String describing component ram distribution
    */
   public String getComponentRamDistribution() {
+    // Generate a map with the minimal ram size for each component
     Map<String, ByteAmount> ramMap = new HashMap<>();
-    // The implementation assumes instances for the same component require same ram
     for (ContainerPlan containerPlan : this.getContainers()) {
       for (InstancePlan instancePlan : containerPlan.getInstances()) {
-        ramMap.put(instancePlan.getComponentName(), instancePlan.getResource().getRam());
+        ByteAmount newRam = instancePlan.getResource().getRam();
+        ByteAmount currentRam = ramMap.get(instancePlan.getComponentName());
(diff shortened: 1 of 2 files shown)


heron/spi/src/java/com/twitter/heron/spi/packing/PackingPlan.java
```diff
@@ -123,11 +123,15 @@ public Integer getInstanceCount() {
...
   public String getComponentRamDistribution() {
+    // Generate a map with the minimal ram size for each component
     Map<String, ByteAmount> ramMap = new HashMap<>();
-    // The implementation assumes instances for the same component require same ram
     for (ContainerPlan containerPlan : this.getContainers()) {
       for (InstancePlan instancePlan : containerPlan.getInstances()) {
-        ramMap.put(instancePlan.getComponentName(), instancePlan.getResource().getRam());
+        ByteAmount newRam = instancePlan.getResource().getRam();
+        ByteAmount currentRam = ramMap.get(instancePlan.getComponentName());
+        if (currentRam == null || currentRam.asBytes() > newRam.asBytes()) {
```
huijunw: why newRam != currentRam

nwangtw: It happens to all algorithms in fact, but it may not be a problem because some algorithms requires per component resources specified.

And it is impossible to be handled in RoundRobin. Different container has different set of components. RoundRobin calculates the components with specified Ram sizes first and allocate the memory left to the rest components. As the result, instances in different containers quite often have different memory size. The example in the description is the simplest case.

PR #2669, opened 2018-01-18
moved eco to the top level 

kramasamy:
and packaged to be shipped with storm compatibility and also moved eco examples to the top level directory.


joshfischer1108:
@kramasamy the yaml files to assemble the topologies are in`heron/examples/src/resources/*yaml`.    

If you want to move them, now would be the time, but then you will also have to modify 
`tools/rules/heron_examples.bzl`

kramasamy:
Will move it

joshfischer1108:
@kramasamy Looks good to me.

PR #2673, opened 2018-01-20
Scalaapi

skanjila:
Added the first set of scala files and bazel build file that links in java implementations.


heron/api/src/scala/BUILD
```diff
@@ -0,0 +1,52 @@
+licenses(["notice"])
```
erenavsarogullari: **General BUILD Comment**:
`*.scala` files are not referenced here. Bazel supports Scala via 
https://github.com/bazelbuild/rules_scala 
so i think we need to use scala_library to compile scala files by referencing
`com/twitter/heron/streamlet/scala/**/*.scala` as `src`.

heron/api/src/scala/Builder.scala
```diff
@@ -0,0 +1,46 @@
+//
```
erenavsarogullari: **General Comment:**
Currently, All files look living under `heron/api/src/scala`. 
- Related `*.scala` files should be under package: `heron/api/src/scala/com/twitter/heron/streamlet/scala`  (except `BUILD` file)
e.g: Currently, _Java Streamlet API_ lives under `heron/api/src/java/com/twitter/heron/streamlet`

heron/api/src/scala/Builder.scala
```diff
@@ -0,0 +1,46 @@
...
+  * of all the starting points of the computation dag and uses this
+  * information to build the topology
+  */
+trait Builder {
+
+  /**
+    * All sources of the computation should register using addSource.
+    * @param supplier The supplier function that is used to create the streamlet
+    */
+  def newSource[R](supplier: SerializableSupplier[R]): Streamlet[R]
```
erenavsarogullari: In the light of Scala User perspective and in my opinion, All exposed Scala `Traits` should be **pure scala**. This means Scala traits should accept `Scala Functions` instead of `FunctionalInterface` here:

`def newSource[R](supplier: () => R): Streamlet[R]`

instead of 

`def newSource[R](supplier: SerializableSupplier[R]): Streamlet[R]`

heron/api/src/scala/SerializableBiFunction.scala
```diff
@@ -0,0 +1,28 @@
+//
```
erenavsarogullari: **General Serializable-FunctionalInterfaces Comment**
I think all Serializable Functional Interfaces(SerializableFunction, SerializableConsumer etc...) should be hidden from Scala Users. This means Scala API should just accept pure Scala functions instead of them. In the light of this, i think we do not need to have these Serializable interfaces' s Scala versions. These can be wrapped at Scala StreamletImpl level by transforming incoming `Scala Function` to related `SerializableFunctionalInterface`

heron/api/src/scala/Streamlet.scala
```diff
@@ -0,0 +1,231 @@
...
+    * Gets the number of partitions of this Streamlet.
+    * @return the number of partitions of this Streamlet
+    */
+  def getNumPartitions(): Int
+
+  /**
+    * Return a new Streamlet by applying mapFn to each element of this Streamlet
+    * @param mapFn The Map Function that should be applied to each element
+    */
+  def map[T](mapFn: SerializableFunction[R, _ <: T]): Streamlet[T]
```
erenavsarogullari: In the light of my previous comment, as a sample for `map` function, i think this should be as follows.
`def map[T](mapFn: R => _ <: T): Streamlet[T]`
instead of 
`def map[T](mapFn: SerializableFunction[R, _ <: T]): Streamlet[T]`

Same comment for the following functions.

heron/api/src/scala/com/twitter/heron/streamlet/Config.scala
```diff
@@ -0,0 +1,260 @@
...
+
+}
+
+/**
+  * Config is the way users configure the execution of the topology.
+  * Things like streamlet delivery semantics, resources used, as well as
+  * user-defined key/value pairs are passed on to the topology runner via
+  * this class.
+  */
+@SerialVersionUID(6204498077403076352L)
```
nwangtw: UIDs seem to be the same as Java version. Just to confirm if this what we want or not.

heron/api/src/scala/com/twitter/heron/streamlet/Window.scala
```diff
@@ -0,0 +1,53 @@
...
+import scala.beans.{BeanProperty, BooleanBeanProperty}
+
+
+/**
+  * Window is a container containing information about a particular window.
+  * Transformations that depend on Windowing, pass the window information
+  * inside their streamlets using this container.
+  */
+@SerialVersionUID(5103471810104775854L)
+class Window() extends Serializable {
```
nwangtw: another candidate to be case class.

heron/api/src/scala/com/twitter/heron/streamlet/KeyValue.scala
```diff
@@ -0,0 +1,53 @@
...
+  var value: V = _
+
+  def this(k: K, v: V) = {
+    this()
+    this.key = k
+    this.value = v
+  }
+
+  override def toString(): String =
+    "{ " + String.valueOf(key) + " : " + String.valueOf(value) +
```
sijie: can't we just use `"{ " + key + " : " + value + " }"`?

heron/api/src/scala/com/twitter/heron/streamlet/KeyedWindow.scala
```diff
@@ -0,0 +1,47 @@
...
+  var window: Window = _
+
+  def this(key: T, window: Window) = {
+    this()
+    this.key = key
+    this.window = window
+  }
+
+  override def toString(): String =
+    "{ Key: " + String.valueOf(key) + " Window: " + window.toString +
```
sijie: do we really need `String.valueOf` or `toString`? if you just `"{ key: " + key + " window: " + window + " }"`, it is better than explicitly call `toString`, because it handles null field, while `window.toString` would throw NPE if `window` is null.

heron/api/src/scala/com/twitter/heron/streamlet/Streamlet.scala
```diff
@@ -0,0 +1,231 @@
...
+    * Gets the number of partitions of this Streamlet.
+    * @return the number of partitions of this Streamlet
+    */
+  def getNumPartitions(): Int
+
+  /**
+    * Return a new Streamlet by applying mapFn to each element of this Streamlet
+    * @param mapFn The Map Function that should be applied to each element
+    */
+  def map[T](mapFn: SerializableFunction[R, _ <: T]): Streamlet[T]
```
sijie: `SerializableFunction` is a java function, not a scala function. which I think you can't really use all those scala syntactic sugar. if we are generating a scala oriented api, it would be better to think about taking scala function, rather than using java functions.

PR #2675, opened 2018-01-22
JSON output for heron-admin standalone cluster info command

lucperkins:
At the moment, running `heron-admin standalone info` command outputs a formatted string. This PR changes the output to pretty-printed JSON so that the output can be easily stored in JSON files and used with tools like `jq`.


jerrypeng:
LGTM +1

@lucperkins Thanks for fixing these things!

PR #2680, opened 2018-01-23
Add Download hostname override config

aahmed-se:


heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -89,10 +96,15 @@ public Response uploadFile(
     }
 
     String uploadDir = config.getStringValue(FILE_SYSTEM_DIRECTORY);
+    String downloadHostNameOverride = config.getStringValue(DOWNLOAD_HOSTNAME_OVERRIDE);
 
-    String fileName = UUID.randomUUID() + "-" + fileDetail.getFileName();
+    if (StringUtil.isNotBlank(downloadHostNameOverride)) {
```
cckellogg: let's add a method that we can use below when forming the uri. I don't like setting a static variable here. 
private String getHostnameOrIp(Config config);
Also you can call config.getString(key, defaultValue) so we don't need the string util.

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -90,9 +97,13 @@ public Response uploadFile(
 
     String uploadDir = config.getStringValue(FILE_SYSTEM_DIRECTORY);
 
-    String fileName = UUID.randomUUID() + "-" + fileDetail.getFileName();
+    if (StringUtil.isNotBlank(getDownloadHostName())) {
```
cckellogg: can we do a little refactoring and have method call getHostnameOrIp() so we are not setting a static variable like this. Then lines 119/120 becomes:
String uri = String.format("http://%s:%s/api/v1/file/download/%s", getHostnameOrIp(), getPort(), fileName);

aahmed-se:
@cckellogg refactored to method

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -143,4 +153,12 @@ private Config createConfig() {
...
 
+  private String getHostNameOrIP() {
+    // Override hostname if provided in flags
+    if (StringUtil.isNotBlank(getDownloadHostName())) {
+      hostname = getDownloadHostName();
+    }
+    return (hostname != null) ? hostname : ip.toString();
+  }
+
 }
```
cckellogg: I think we should not be setting static state here.
final String downloadHostName = getDownloadHostName();
if (StringUtil.isNotBlank(downloadHostName) {
  return downloadHostName;
}

return hostname != null ? hostname : ip;

aahmed-se: Changed it and added a check for ip being null

PR #2681, opened 2018-01-23
enable write_env_exec_file method

wxl24life:
As is noted in https://twitter.github.io/heron/docs/contributors/community/, in order to bootstrap an intellij idea project, `$ ./scripts/setup-intellij.sh` should run first. It uses `./scripts/compile_env/env_exec.sh` file which should be created by `./bazel_configure.py`.

### DIFF
--- a/bazel_configure.py
+++ b/bazel_configure.py
@@ -398,7 +398,7 @@ def main():
   env_map['STRIP'] = discover_tool_default('strip', "strip", 'STRIP', '/usr/bin/strip')
 
   # write the environment executable file
-  # write_env_exec_file(platform, env_map)
+  write_env_exec_file(platform, env_map)
 
 if __name__ == '__main__':
   main()


kramasamy:
@wxl24life - env_exec.sh file is not needed anymore. If you could remove the if block in 

https://github.com/streamlio/streamlio/blob/master/deploy/kubernetes/helm/templates/broker.yaml#L84

and send a PR that will be awesome!

wxl24life:
@kramasamy sorry, 404 error found when opening your link?

kramasamy:
Sorry wrong link - check if this works
https://github.com/twitter/heron/blob/master/scripts/setup-intellij.sh#L4

wxl24life:
@kramasamy Thanks. I have tested it. After removing the heading`if` block, `./scripts/setup-intellij.sh` can run smoothly. I think for history reason, as was noted in PR #1429 , that heading`if` block was added to check the existence of `env_exec.sh`. But for now, the `env_exec.sh` file is not needed anymore. I will close this thread and send a new PR;-)

PR #2684, opened 2018-01-25
make heron core distribution - symbolic linkable

kramasamy:
for running local topologies rather than curling. This has the advantage of 

- Removing the curl dependency
- Reduces the disk space
- Capable for a single docker

### DIFF
--- a/scripts/packages/template_bin.sh
+++ b/scripts/packages/template_bin.sh
@@ -97,7 +97,7 @@ if [ -d "${base}" -a -x "${base}/bin/heron" ]; then
 fi
 
 mkdir -p ${bin} ${base} ${base}/etc
-echo -n .
+echo "..done"
 
 unzip -q -o "${BASH_SOURCE[0]}" -d "${base}"
 untar ${base}/heron.tar.gz ${base}
@@ -120,9 +120,13 @@ ln -s "${base}/bin/heron-explorer" "${bin}/heron-explorer"
 ln -s "${base}/bin/heron-tracker" "${bin}/heron-tracker"
 ln -s "${base}/bin/heron-ui" "${bin}/heron-ui"
 ln -s "${base}/bin/heron-apiserver" "${bin}/heron-apiserver"
-echo -n .
+
+echo "Uncompressing."
+untar ${base}/dist/heron-core.tar.gz ${base}/dist
 
 rm "${base}/heron.tar.gz"
+rm "${base}/dist/heron-core.tar.gz"
+rm "${base}/dist/release.yaml"
 
 cat <<EOF
 
--- a/heron/config/src/yaml/conf/local/client.yaml
+++ b/heron/config/src/yaml/conf/local/client.yaml
@@ -1,2 +1,5 @@
-# location of the core package
-heron.package.core.uri:                      file://${HERON_DIST}/heron-core.tar.gz
+# do not use core package URI
+heron.package.use_core_uri:                  false
+
+# core package directory to use
+heron.package.core.directory:                ${HERON_DIST}/heron-core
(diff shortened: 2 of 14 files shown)


maosongfu:
LGTM.

scripts/packages/template_bin.sh
```diff
@@ -120,9 +120,13 @@ ln -s "${base}/bin/heron-explorer" "${bin}/heron-explorer"
 ln -s "${base}/bin/heron-tracker" "${bin}/heron-tracker"
 ln -s "${base}/bin/heron-ui" "${bin}/heron-ui"
 ln -s "${base}/bin/heron-apiserver" "${bin}/heron-apiserver"
-echo -n .
+
+echo "Uncompressing."
+untar ${base}/dist/heron-core.tar.gz ${base}/dist
 
 rm "${base}/heron.tar.gz"
+rm "${base}/dist/heron-core.tar.gz"
```
jerrypeng: This changes the expected behavior of the install and will break a lot of things.  For example, the nomad scheduler expected that heron core to be at:

file://${HERON_DIST}/heron-core.tar.gz

you unTARing it here and removing the archive will break that

jerrypeng: this will Aurora too

kramasamy: @jerrypeng - are you using the heron-install.sh script or just untaring the heron.tar.gz?

jerrypeng: For nomad/standalone, the user will use heron-install.sh to install on a bootstrap machine.  Then the .heron directory will be copied to all the machines in the standalone cluster as is

jerrypeng: It will break Aurora as well:

heron_core_release_uri = '{{CORE_PACKAGE_URI}}'
heron_topology_jar_uri = '{{TOPOLOGY_PACKAGE_URI}}'
core_release_file = "heron-core.tar.gz"
topology_package_file = "topology.tar.gz"

# --- processes ---
fetch_heron_system = Process(
  name = 'fetch_heron_system',
  cmdline = 'curl %s -o %s && tar zxf %s' % (heron_core_release_uri, core_release_file, core_release_file)
)

kramasamy: @jerrypeng - for nomad case, we can add a config that in nomad/scheduler.yaml - similar to local and it will work. For aurora nothing will break since the core_package_uri will be in a URL that is stored somewhere else.

jerrypeng: I don't understand how adding just a config will work?  Both in Nomad and Aurora start scripts it expects a archive and will execute "tar zxf" on it to untar it. We will have to change that as welll

kramasamy: the core logic - can either curl or use symbolic - which is controlled by a config. It is a question of telling nomad which one to use. For Aurora, the deployment is that - heron-core.tar.gz is served from a file server (like webserver or packer)

jerrypeng: ok

jerrypeng: gotcha, we can probably use the same method of creating a symbolic link in heron-nomad.sh

Yitian-Zhang: I was confused about this change. When I used Aurora after this change and using HDFS as the uploader as the same time, how should I configure the path to the heron-core file in Client.yaml? Should I copy the ./heron/dist/heron-core directory to the HDFS as when I using 0.17.1 version? (When I used the 0.17.1 version, I copied the heron-core.tar.gz file to HDFS and configured the HDFS path in Client.yaml, then I modified the curl command in heron.aurora file). And I don't know what should I do now.

PR #2685, opened 2018-01-25
Add skip runtime validation option

nwangtw:
Internal users reported an issue that a kill command is not fully successful. As the result there are some left over data in zk. It seems clearState() function deletes 8 nodes one by one and the delete calls don't guarantee to be successful.

Because of these left over data, users can't deploy the topology.
And because some key data has been deleted, "heron kill" command throws a "Topology 'test' does not exist" error when users tried to run it again.

This new "skip-runtime-validation" option can be useful when a topolgoy is not "fully killed",
for example, topology is killed but data clean up is not successful,
to skip the runtime validation in order to run the kill command again

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -153,13 +153,13 @@ public class RuntimeManagerMain {
         .build();
 
     Option dryRun = Option.builder("u")
-        .desc("run in dry-run mode")
+        .desc("Run in dry-run mode")
         .longOpt("dry_run")
         .required(false)
         .build();
 
     Option dryRunFormat = Option.builder("t")
-        .desc("dry-run format")
+        .desc("Dry-run format")
         .longOpt("dry_run_format")
         .hasArg()
         .required(false)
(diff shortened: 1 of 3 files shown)


kramasamy:
@nwangtw - this is probably not the right approach to implement this. Rather than adding more options - we could do the following -

- when heron kill issued for a topology, write a znode under that topology that topology is going to be killed. This is an atomic operation

- Once the znode is written, subsequent submit can clear the state before launching the topology.

nwangtw:
@kramasamy I considered a few options and a new option seems to be the most reliable and straightforward one.

You are suggesting leaving all the data in zookeeper for a topology after it is killed or still call clearState() after the flag node is created (I prefer the second one)? this could be an option if it is ok with us. I was assuming that we want a topology to start from a clean state hence trying to avoid calling the function again in submit command. You are right that one major issue is that we need an atomic operation instead of 8 unreliable steps.

In fact, if it is ok for us to call clearState() again during submit, we may use topology name node as the flag so we don't need a new one.

nwangtw:
Create another PR https://github.com/twitter/heron/pull/2689 based on @kramasamy's suggestion.

nwangtw:
Discard PR. We are going for a different solution: https://github.com/twitter/heron/pull/2730

PR #2688, opened 2018-01-25
[Streamlet Scala API] Add initial Build Files

erenavsarogullari:

- Base PR for Scala `Streamlet` API
- Scala API Bazel Build Support
- Source Trait as pure Scala version of Java interface
- UT Coverage by approving initial Scala Tests passed

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -881,11 +881,11 @@ new_http_archive(
 rules_scala_version="5cdae2f034581a05e23c3473613b409de5978833" # update this as needed
 
 http_archive(
-             name = "io_bazel_rules_scala",
-             url = "https://github.com/bazelbuild/rules_scala/archive/%s.zip"%rules_scala_version,
-             type = "zip",
-             strip_prefix= "rules_scala-%s" % rules_scala_version
-             )
+    name = "io_bazel_rules_scala",
+    url = "https://github.com/bazelbuild/rules_scala/archive/%s.zip"%rules_scala_version,
+    type = "zip",
+    strip_prefix= "rules_scala-%s" % rules_scala_version
+)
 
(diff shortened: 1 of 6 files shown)


WORKSPACE
```diff
@@ -876,3 +876,18 @@ new_http_archive(
     urls = ["https://releases.hashicorp.com/nomad/0.7.0/nomad_0.7.0_linux_amd64.zip"],
     build_file = "third_party/nomad/nomad.BUILD",
 )
+
+# scala integration
+rules_scala_version="5cdae2f034581a05e23c3473613b409de5978833" # update this as needed
+
+http_archive(
+             name = "io_bazel_rules_scala",
```
kramasamy: We use 4 spaces indentation.

erenavsarogullari: Addressed.

kramasamy:
Ship it!

heron/api/src/scala/com/twitter/heron/streamlet/scala/Source.scala
```diff
@@ -0,0 +1,34 @@
...
+import java.io.Serializable
+import com.twitter.heron.streamlet.Context
+
+/**
+  * Source is how Streamlet's originate. The get method
+  * invocation returns new element that form the tuples of the streamlet.
+  * setup/cleanup is where the generator can do any one time setup work, like
+  * establishing/closing connection to sources, etc.
+  */
+trait Source[T] extends Serializable {
```
skanjila: can we add a method that allows one to collect state associated with a streamlet's state, also I would like to return a meaningful data structure from these methods, maybe a map[String,Int] that corresponds with a status and a meaningful error message

PR #2689, opened 2018-01-25
Call cleanState() before submitting topology in case previous kill co…

nwangtw:
…mmand wasn't fully successful and there is left-over data

Internal users reported an issue that a kill command is not fully successful. As the result there are some left over data in zk and users won't be able to submit the topology again. And because the key data has been deleted, "heron kill" command throws a "Topology 'test' does not exist" error when users tried to run it again.

It seems clearState() function deletes 8 nodes one by one and the delete calls don't guarantee to be successful. Now we call clearState() again in case there is any left over (unexpected) data. The validation code in submitMain is also updated to be consistent to the validation in RuntimeManager.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
@@ -35,6 +35,7 @@ import com.twitter.heron.api.utils.TopologyUtils;
 import com.twitter.heron.common.basics.DryRunFormatType;
 import com.twitter.heron.common.basics.SysUtils;
 import com.twitter.heron.common.utils.logging.LoggingHelper;
+import com.twitter.heron.proto.system.ExecutionEnvironment;
 import com.twitter.heron.scheduler.dryrun.SubmitDryRunResponse;
 import com.twitter.heron.scheduler.utils.DryRunRenders;
 import com.twitter.heron.scheduler.utils.LauncherUtils;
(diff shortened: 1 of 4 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/SubmitterMain.java
```diff
@@ -425,6 +428,10 @@ public void submitTopology() throws TopologySubmissionException {
       // Check if topology is already running
       validateSubmit(adaptor, topology.getName());
 
+      // Try to clear left over data in case the previous kill operation
```
kramasamy: How do you distinguish the case between 
- Normal running topology
- Topology in a partial state

nwangtw: if a topology has defn and executionState data, it is considered as "running", otherwise it is considered as "not running". This is the validation logic in RuntimeManager, and i updated the validateSubmit() called right above to follow the same condition in this PR.

I was thinking of using defn data as the source of truth, but executionState is very important in RuntimeManager validation. So here I am keep both. Another option is to use executionState only as source of truth, if you think it is better that way.

When the topology is "not running", it should be safe to call cleanState().

huijunw:
At present, if cleaning zk fails, we suggest user :
```
do 
   `heron kill xxx`
until success
```
which is similar as this pr. This pr suggests:
```
do
  `heron submit xxx`
until success
```
This pr does not solve the root of cause 'cleaning zk fails'.

I prefer `heron kill xx` until success because it does what it is named.

nwangtw:
Discard because we are going to a different solution in heron kill command: https://github.com/twitter/heron/pull/2730

PR #2691, opened 2018-01-26
Fix for localhost metrics manager

jrcrawfo:
What?
- Fixes for getting local hostname from within a Docker container

Why?
- Metrics management doesn't work correctly if running Heron using Docker containers deployed in a cluster environment

### DIFF
--- a/heron/common/src/java/com/twitter/heron/common/utils/misc/PhysicalPlanHelper.java
+++ b/heron/common/src/java/com/twitter/heron/common/utils/misc/PhysicalPlanHelper.java
@@ -111,7 +111,11 @@ public class PhysicalPlanHelper {
     }
 
     try {
-      this.hostname = InetAddress.getLocalHost().getHostName();
+      if (System.getenv("HOST") != null) {
+        this.hostname = System.getenv("HOST");
+      } else {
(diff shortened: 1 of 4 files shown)


maosongfu:
Two comments:
1. Do you think we could make:
```
 if (System.getenv("HOST") != null) {
        this.hostname = System.getenv("HOST");
     } else {
        this.hostname = InetAddress.getLocalHost().getHostName();
     }
``` 
as an utlity method rather than duplicate the code multiple times?

2. Use a more specific system env name like `HERON_SERVER_HOST` rather than the generic `HOST` to avoid any unexpected system injection?

jrcrawfo:
#1. Yes, we should do that. I’ll make that change. 

#2. The reason for using HOST is because it is commonly passed in by the scheduler to be an actual resolvable hostname in a Docker environment (where hostname otherwise is the container id). So, it’s best to leave this as is, as there is also a helper function in the python executor code to do the same logic.

maosongfu:
@jrcrawfo 
Fair. Will give a ship it once:
1. The utility method is added.
2. The Travis Ci build can pass.

kramasamy:
@jrcrawfo - any update on this PR?

nlu90:
@jrcrawfo Ping for this PR, any updates?

PR #2692, opened 2018-01-26
Making emit, ack, and fail thread safe

jerrypeng:
We have some use cases in which users will want to emit and ack or fail in different threads.  I have done some performance comparisons and there seems to not be a performance difference but perhaps others can chime in.

### DIFF
--- a/heron/instance/src/java/org/apache/heron/instance/AbstractOutputCollector.java
+++ b/heron/instance/src/java/org/apache/heron/instance/AbstractOutputCollector.java
@@ -17,6 +17,7 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.ReentrantLock;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Message;
@@ -40,6 +41,7 @@ public class AbstractOutputCollector {
   private long totalTuplesEmitted;
   private long totalBytesEmitted;
   private PhysicalPlanHelper helper;
+  public final ReentrantLock lock = new ReentrantLock();
 
   /**
    * The SuppressWarnings is only until TOPOLOGY_ENABLE_ACKING exists.
@@ -73,7 +75,7 @@ public class AbstractOutputCollector {
       }
     }
 
-    this.outputter = new OutgoingTupleCollection(helper, streamOutQueue);
+    this.outputter = new OutgoingTupleCollection(helper, streamOutQueue, lock);
   }
 
   public void updatePhysicalPlanHelper(PhysicalPlanHelper physicalPlanHelper) {
--- a/heron/instance/src/java/org/apache/heron/instance/OutgoingTupleCollection.java
+++ b/heron/instance/src/java/org/apache/heron/instance/OutgoingTupleCollection.java
@@ -15,6 +15,7 @@
 package org.apache.heron.instance;
 
 import java.io.Serializable;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Message;
@@ -55,7 +56,7 @@ public class OutgoingTupleCollection {
   private HeronTuples.HeronControlTupleSet.Builder currentControlTuple;
 
   // Total data emitted in bytes for the entire life
-  private long totalDataEmittedInBytes;
+  private AtomicLong totalDataEmittedInBytes = new AtomicLong();
 
   // Current size in bytes for data types to pack into the HeronTupleSet
   private long currentDataTupleSizeInBytes;
@@ -72,7 +73,7 @@ public class OutgoingTupleCollection {
         SerializeDeSerializeHelper.getSerializer(helper.getTopologyContext().getTopologyConfig());
 
     // Initialize the values in constructor
-    this.totalDataEmittedInBytes = 0;
+    this.totalDataEmittedInBytes.set(0);
     this.currentDataTupleSizeInBytes = 0;
 
     // Read the config values
@@ -81,7 +82,7 @@ public class OutgoingTupleCollection {
     this.controlTupleSetCapacity = systemConfig.getInstanceSetControlTupleCapacity();
   }
 
-  public void sendOutTuples() {
+  public synchronized void sendOutTuples() {
     flushRemaining();
   }
 
@@ -90,7 +91,7 @@ public class OutgoingTupleCollection {
    * @param state instance's state
    * @param checkpointId the checkpointId
    */
-  public void sendOutState(State<Serializable, Serializable> state,
+  public synchronized void sendOutState(State<Serializable, Serializable> state,
                            String checkpointId) {
     // flush all the current data before sending the state
     flushRemaining();
@@ -114,7 +115,7 @@ public class OutgoingTupleCollection {
     outQueue.offer(storeRequest);
   }
 
-  public void addDataTuple(
+  public synchronized void addDataTuple(
       String streamId,
       HeronTuples.HeronDataTuple.Builder newTuple,
       long tupleSizeInBytes) {
@@ -132,10 +133,10 @@ public class OutgoingTupleCollection {
     currentDataTuple.addTuples(newTuple);
 
     currentDataTupleSizeInBytes += tupleSizeInBytes;
-    totalDataEmittedInBytes += tupleSizeInBytes;
+    totalDataEmittedInBytes.getAndAdd(tupleSizeInBytes);
   }
 
-  public void addAckTuple(HeronTuples.AckTuple.Builder newTuple, long tupleSizeInBytes) {
+  public synchronized void addAckTuple(HeronTuples.AckTuple.Builder newTuple, long tupleSizeInBytes) {
     if (currentControlTuple == null
         || currentControlTuple.getFailsCount() > 0
         || currentControlTuple.getAcksCount() >= controlTupleSetCapacity) {
@@ -144,10 +145,10 @@ public class OutgoingTupleCollection {
     currentControlTuple.addAcks(newTuple);
 
     // Add the size of data in bytes ready to send out
-    totalDataEmittedInBytes += tupleSizeInBytes;
+    totalDataEmittedInBytes.getAndAdd(tupleSizeInBytes);
   }
 
-  public void addFailTuple(HeronTuples.AckTuple.Builder newTuple, long tupleSizeInBytes) {
+  public synchronized void addFailTuple(HeronTuples.AckTuple.Builder newTuple, long tupleSizeInBytes) {
     if (currentControlTuple == null
         || currentControlTuple.getAcksCount() > 0
         || currentControlTuple.getFailsCount() >= controlTupleSetCapacity) {
@@ -156,7 +157,7 @@ public class OutgoingTupleCollection {
     currentControlTuple.addFails(newTuple);
 
     // Add the size of data in bytes ready to send out
-    totalDataEmittedInBytes += tupleSizeInBytes;
+    totalDataEmittedInBytes.getAndAdd(tupleSizeInBytes);
   }
 
   private void initNewDataTuple(String streamId) {
@@ -210,18 +211,18 @@ public class OutgoingTupleCollection {
   }
 
   public long getTotalDataEmittedInBytes() {
-    return totalDataEmittedInBytes;
+    return totalDataEmittedInBytes.get();
   }
 
   // Clean the internal state of OutgoingTupleCollection
-  public void clear() {
+  public synchronized void clear() {
     currentControlTuple = null;
     currentDataTuple = null;
 
     outQueue.clear();
   }
 
-  public void updatePhysicalPlanHelper(PhysicalPlanHelper physicalPlanHelper) {
+  public synchronized void updatePhysicalPlanHelper(PhysicalPlanHelper physicalPlanHelper) {
     this.helper = physicalPlanHelper;
   }
 }
--- a/heron/instance/src/java/org/apache/heron/instance/OutgoingTupleCollection.java
+++ b/heron/instance/src/java/org/apache/heron/instance/OutgoingTupleCollection.java
@@ -16,6 +16,7 @@ package org.apache.heron.instance;
 
 import java.io.Serializable;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Message;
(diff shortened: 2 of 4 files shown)


srkukarni:
Is there a perf implication?

jerrypeng:
@srkukarni I don't think so

nwangtw:
I am a bit concerned about it and we may need to talk first before merging

kramasamy:
@nwangtw - Let us chat on tuesday so that we can discuss its implications. It is very common for people to ack/fail in another thread. If we don't support it - it is a limitation.

nwangtw:
@kramasamy " It is very common for people to ack/fail in another thread."

Yeah. I believe it makes sense in these cases. I think we need to treat user functions like addActTuple, addDataTuple and sinking functions like sendOutTuple separately. For high throughput topologies, sinking functions could take a long time and it is not ideal to block user functions/threads.

maosongfu:
@jerrypeng 
1. Can we have some performance tests and show the number? For instance, against the exclamation topology for the `throughput/per cpu`

2. I am thinking whether this implementation is doing in the right way: it changes the interfaces' expected behaviors. In other words, it would break the interfaces, in some non-functional requirements, like performance, or design philosophy. 
I have some other thoughts to achieve the thread-safe support, for instance: adding new interfaces providing methods `safeAck(...), safeFail(...), safeEmit(...), etc.` and people can call those methods if they need thread-safety.

@srkukarni @kramasamy @nwangtw Thoughts?

srkukarni:
IMO adding new apis is not the right way to do this. Storm's acking mechanism is thread safe and so should we. This will not break any existing clients since it actually relaxes constraints. The only issue is performance. And even there I'm not sure if this is a huge hit esp since the contention between the two threads is very little.

nwangtw:
Personally I am not against adding the protection in the needed functions. If storm assumes that, then keeping the same API sounds more reasonable.

My concerns are:

1. even it is a relax, this is an API change, so it is better to be careful. What exactly is the problem? Users need to call "OutputCollector.sendTuple" in multiple threads? Any other functions are needed?
2. OutgoingTupleCollection is not an interface layer class and it is shared by both bolt and spout. I can understand the bolt case, but do we need the same protection for spout? Is this the best place to synchronize compare to AbsOutputCollector or BoltOutputCollectorImpl?
3. it seems the current bolt process is: 1. check there is room in outQueue, 2. execute bolt logic, 3. sendout*. We should be clear if add*tuple functions are allowed to be called out of the time window of step 2. If no, then the sendOut* functions are not needed to be synchronized. If yes, I think we have a lot  more to worry about than making these functions to be synchronized.

jerrypeng:
@nwangtw To address your concerns:

1. Yes there are use cases when users will emit and ack tuples from different threads.  For example, there are use cases when a user will put tuples on a queue and ack them later.

2. We do need this synchronization for spouts as well.  A user can very well want to call emit from different threads.

3. I am not sure I understand you question.  Are you saying that because users can emit from a different thread, that it might overflow the outQueue?  The current check doesn't prevent a user from overflowing the queue to begin with right?  Even if I am not emitting from a different thread, I can still emit as many times as I want which can potentially overflow the outqueue.

jerrypeng:
some numbers:


  | Emit Count (10min) | Latency (ms)
-- | -- | --
Vanilla | 112342934 | 7001.68
With Change | 104313910 | 6328.94

  | Emit Count (10min) | Latency (ms)
-- | -- | --
Vanilla | 112342934 | 7001.68
With Change | 104313910 | 6328.94

jerrypeng:
For ExclamationTopology:

  | Emit Count (10min) | 10min avg tuples/sec
-- | -- | --
Vanilla | 587,239,062 | 978,732
With Change | 642,861,133 | 1,071,435

nwangtw:
Thanks for the responses.

Regarding the 3rd concern, I talked with Maosong later and confirmed the queue doesn't have a hard cap. So it should be fine.

I guess the only concern we have is @maosongfu's concern about stateful calculation, or the execution order?

kramasamy:
@srkukarni - can you address @maosongfu concerns?

srkukarni:
This change shouldn't affect stateful semantics.

maosongfu:
@skanjila 
Suppose an instance is expected to receive data in such sequences: tuple_A, tuple_B, tuple_C, stateful-checkpoint-request, tuple_D, ...
Can we guarantee the instance will do a stateful-checkpoint exactly after the completion of processing tuple_A, tuple_B, tuple_C and before the start of processing tuple_D, if tuple_A, tuple_B, tuple_C, tuple_D are processed in different threads?

srkukarni:
Heron effectively once assumes that tuples are processed in the order they are received. Thus (leaving aside this pr), in today's code, if tuples are processed out of bounds, the contract of effectively once does not apply. This pr does not change that invariant.

maosongfu:
@srkukarni 
Previously heron-instance processed data in a single thread so it will process data in the order it receives, unless users create multi-threads by themselves. Then it is fair to ask them to take care of the order.
But after this PR heron-instance allows process data in multi-threads by itself but asking people to guarantee the output order. Then the current effectively-once implementation does not sound like in a good shape or a complete solution. 

It sounds like:
Previously we provide a `queue` implementation without guaranteeing any thread-safety so if people want concurrency they have to take care both `add()` and `remove`. But now we offer a concurrent queue solution/implementation providing thread-safe `add()` without thread-safe `remove()`. Definitely, we could ask customers to guarantee the `remove()` thread-safe on their own but, but the implementation does not sound like an actual complete `concurrent queue.`

srkukarni:
@maosongfu I think the concern is misplaced. Lets take two cases
1. Users use only a single thread to do process/ack. In this case, both the current version and the new pr work the same.
2. Users use multiple threads. Then in the current version, even if they acked in the right order, the system does not ensure effectively once, since the system could do checkpointing before tuple_A is acked for example. This doesn't change with the new pr either.
Thus in both cases, the current semantics still hold good. 
Heron's chandy lamport algorithm implementation requires that users take up next tuple only after processing the old tuple. If users deviate from that, effectively once cannot be ensured. This is true both in the current version and the new pr.

kramasamy:
@maosongfu - can you approve the PR after @srkukarni has addressed the concerns?

maosongfu:
@srkukarni @jerrypeng 
We need to stop all data processing in a heron-instance and then do a checkpoint, and then recover the data processing.

More specifically:
In: https://github.com/twitter/heron/blob/c65e312c078e52222b6ea797662715dc03674b36/heron/instance/src/java/com/twitter/heron/instance/bolt/BoltInstance.java
```
    if (bolt instanceof IStatefulComponent) {
      ((IStatefulComponent) bolt).preSave(checkpointId);
    }

    collector.sendOutState(instanceState, checkpointId);
```
And:
In https://github.com/twitter/heron/blob/28478557d50718725491c2b6a631dea1dc56f485/heron/instance/src/java/com/twitter/heron/instance/spout/SpoutInstance.java
```
    if (spout instanceof IStatefulComponent) {
      ((IStatefulComponent) spout).preSave(checkpointId);
    }

    collector.sendOutState(instanceState, checkpointId);
```
We need to make sure these two blocks will not be executed concurrently with data-processing. Otherwise, we will get a dirty state.

srkukarni:
Two points
1. This is not related to this pr. This concurrency can happen today as well.
2. More importantly, if there are user threads doing execution, its not possible to even know when its good to do the checkpoint. Do we have to wait for all pending tuples to be acked(this may not happen based on user logic).

nwangtw:
I think Maosong's concern is that this concurrent threads support is not complete.

Here is my understanding:

For stateful topologies, user can't use concurrent threads. It is the same as today and existing topologies should work fine. However today users can't have concurrent threads at all, hence it is less an issue; if we add the concurrent threads support, we have to either handle the synchronization or  block concurrent threads in stateful topologies.

srkukarni:
Users can(and do) have concurrent threads today. SummingBird for instance has internal threads that do all kind of processing. The only thing that they have to do is to do their acks/fails in the same thread as process thread. Thus in the present scenario imagine this situation
1. User Bolt has two 'worker' threads that actually do some work and use some kind of queue to put their acks/fails in the main process thread.
2. tupleA, tupleB, ckptTuple, tupleC is the order as seen by the system
3. say tupleA is handled by threadA and tupleB is handled by threadB.
4. process thread receives ckptTuple. It calls preSave, followed by state collection. If the user's dont carefully program their preSave, we wont have guarantee that users will record tupleA and tupleB's execution state. 
5. When things go bad and state is rewinded, ckpt is restored at the bolt level  and system continues from tupleC.
Note that in this case, because users are doing out of band/out of order processing, tupleA and tupleB are not acked/failed before ckpt is handled. Thus it is left to the preSave functionality for them to save their state properly.
Now whether the acks/fails are done using some kind of sync between user thread and worker threads or in the worker threads themselves has no bearing upon this state of affairs. Thus this pr does not in any way alter the way stateful topologies are handled.
It would be great if you can provide an example of why a stateful topology would break with this pr.
Thanks!

maosongfu:
One the one hand:
Currently, we claim all interfaces are not thread-safe so users will take full responsibility to do the multi-threads, for instance, getting wrong results. Admittedly, merging this pull request will not break the single-thread case, but this new pull request claims some interfaces thread-safe. So it sounds like a queue implementation has only: 1. thread-safe `push()` without thread-safe `pop()` 2. will still work in the single-thread case, and it claims itself a thread-safe concurrent queue.

On the other hand:
one thing I mentioned above is:
```
if (spout instanceof IStatefulComponent) {
      ((IStatefulComponent) spout).preSave(checkpointId);
    }

    collector.sendOutState(instanceState, checkpointId);
```

Above block should be executed atomically. So we would need to make sure the state is flushed out before any furthermore processing. And currently based on this implementation, it seems **no way** to guarantee that inside the topology if users do emit in multi-threads. So users will **no way** to implement a correct effective-once topology with the support of thread-safety in this pull request.

An example is: an instance completes a checkpoint, but some later tuples are processed and emitted in another thread - users have **no way** to guarantee the checkpoint is flushed before the flush of later processed data tuples' emits, since this execution is out of the scope of the topology. And then leading a wrong state of downstream instances.

Meanwhile, it is achievable (achieved in some other stream processing engines) to prevent the concurrent execution of `data processing` and `checkpoint` in a heron-instance level. As I mentioned above, 
> This can be achieved by using `lock/mutex/semaphore/etc`.

I like the idea supporting thread-safe interfaces but we should carefully design & implement it and should not break any existing features (like effective-once).

kramasamy:
@maosongfu @srkukarni - is this PR has come to a conclusion?

nwangtw:
Seems not. Maosong's concern hasn't been replied yet.  @jerrypeng @skanjila

jerrypeng:
@nwangtw @maosongfu I have refactored the code to include synchronization with saving the state.  Can you please take a look. Thanks!

srkukarni:
@nlu90 can you please take a look as well

bornej:
Hi, 
while investigating the possibility of using multi-threaded bolt/spouts, I stumbled upon this PR.
I don't know if this is the appropriate place to ask but I'm stuck with some questions...

From what I understand, in each Heron instance there is only one slave thread doing bolt/spout computation (and a gateway thread used for communications with the stream manager). But after reading this discussion I'm a bit confused

@jerrypeng 
> We have some use cases in which users will want to emit and ack or fail in different threads.

@srkukarni 
> Users can(and do) have concurrent threads today. 
> ... imagine this situation
> 1. User Bolt has two 'worker' threads that actually do some work and use some kind of queue to put 
> their acks/fails in the main process thread.
 
Is it currently possible for a user to create a multi-threaded Bolt/spout? (Without altering HeronInstance implementation-> FixedThreadPool size) 
If so, could someone help me figure out how to achieve that?

I'm very new to Heron so I hope you will forgive me if my question seems stupid.

I understand that Heron philosophy is to use "single"-threaded Instances (bolt/spouts) and why this design choice was made but I'm wondering if some computation tasks could benefit from multi-threading...

Thanks for your help and your work.
Best Regards.
Jonathan.

nlu90:
@bornej It is possible to utilizing multithreads in Heron even without this patch. The only requirement is that you need to ack/fail the processed tuple in the original main thread for correctness. If your topology is AT-MOST-ONCE semantics, then you don't even need to worry about this problem.

@srkukarni After reviewing this long converstaion and the changes made, it looks good to me. But I think we need to mention the thread-safe acking to users very carefully, ensures them that EFFECTLY-ONCE might be broken if they don't implement `preSave` carefully.

bornej:
Thanks you very much,
Is there any example of multi-threaded heron bolt/spout/topology out there?

jerrypeng:
@nlu90 just to be clear, with out this patch its not even possible to guarantee at-most once.  An NPE will very likely to be thrown if you emit, ack, or fail from different threads

jerrypeng:
@bornej here is an example of a bolt that acks and emits from a different thread

```
public static class ExclamationBolt extends BaseRichBolt {
    private static final long serialVersionUID = -2267338658317778214L;
    private OutputCollector collector;
    private long nItems;
    private long startTime;
    private LinkedBlockingQueue<Tuple> queue = new LinkedBlockingQueue<>();

    @Override
    @SuppressWarnings("rawtypes")
    public void prepare(Map conf, TopologyContext context, OutputCollector acollector) {
      collector = acollector;
      nItems = 0;
      startTime = System.currentTimeMillis();

      Thread thread1 = new Thread(new Runnable() {
        private long pItems;

        @Override
        public void run() {
          System.out.println("start thread 1 ....");
          while (true) {
            Tuple tuple = null;
            try {
              tuple = queue.take();
            } catch (InterruptedException e) {
              e.printStackTrace();
            }
            collector.ack(tuple);

            ++pItems;

            if (pItems % 10000 == 0) {
              long latency = System.currentTimeMillis() - startTime;
              System.out.println("Bolt - 1 processed " + pItems + " tuples in " + latency + " ms");
            }
          }
        }
      });

      thread1.setName("thread-1");
      thread1.start();
    }

    @Override
    public void execute(Tuple tuple) {
      try {
        queue.put(tuple);
      } catch (InterruptedException e) {
        e.printStackTrace();
      }

      ++nItems;
      if (nItems % 10000 == 0) {
        long latency = System.currentTimeMillis() - startTime;
        System.out.println("Bolt - 0 processed " + nItems + " tuples in " + latency + " ms");
      }
    }
  }
```

bornej:
@jerrypeng Thank you so much!

I will do experiments with heron multithreading in the next weeks. I hope to have some performance comparisons to provide soon.

jerrypeng:
@maosongfu can you approve? or do you have anymore concerns?

nwangtw:
I think maosong's concern is still valid. For stateful process, we have to make sure the checkpoint are made after all user threads have done their acks.  Which is why Neng said above "But I think we need to mention the thread-safe acking to users very carefully, ensures them that EFFECTLY-ONCE might be broken if they don't implement preSave carefully."

jerrypeng:
@nwangtw I agree that effectively-once semantics might not be honored if users decide process tuples asynchronously, but it is possible if the user uses the presave function as a trigger to create synchrony among user threads.  I think it is matter of education and documentation to educate users on this, but I think this PR is good go.

nwangtw:
@jerrypeng agreed. The concern is that users CAN use the function in the wrong way. Avoiding it in code would be nice; protecting the case might be helpful and acceptable too; otherwise very clear documentation and education is needed (we should have it anyway though). I think maosong prefers to have one of the first 2 options available.

Maybe we should have an example code for presave() function if we want to cover stateful process?

PR #2693, opened 2018-01-26
Nwang/count stmgr connection failures in all clients

nwangtw:
We found out that when a bad stmgr is experiencing connection issue to
all other stmgrs, it takes the stmgr a lot longer to reach the per stmgr
client max attempt and quite becaus bad connections cause extra delays.
In our case, instead of quitting after 5 minutes (1s reconnect timeout,
300 max attempts), the bad stmgr quitted after more than 2 hours. At the
same time, all other good stmgrs were quitting every 5 minutes.

A total reconnection attempt checking is added in order to make the bad
stmgr quit sooner.

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
@@ -35,10 +35,10 @@ namespace stmgr {
 // New connections made with other stream managers.
 const sp_string METRIC_STMGR_NEW_CONNECTIONS = "__stmgr_new_connections";
 
-StMgrClientMgr::StMgrClientMgr(EventLoop *eventLoop, const sp_string &_topology_name,
-                               const sp_string &_topology_id, const sp_string &_stmgr_id,
-                               StMgr *_stream_manager,
-                               heron::common::MetricsMgrSt *_metrics_manager_client,
+StMgrClientMgr::StMgrClientMgr(EventLoop* eventLoop, const sp_string& _topology_name,
+                               const sp_string& _topology_id, const sp_string& _stmgr_id,
+                               StMgr* _stream_manager,
+                               heron::common::MetricsMgrSt* _metrics_manager_client,
                                sp_int64 _high_watermark, sp_int64 _low_watermark,
                                bool _droptuples_upon_backpressure)
     : topology_name_(_topology_name),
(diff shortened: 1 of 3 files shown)


nwangtw:
This is just an initial PR to demo an issue and a quick solution. I have a few other thoughts but would like to get some more ideas from you guys first. Thanks in advance.

heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
```diff
@@ -186,6 +190,27 @@ void StMgrClientMgr::HandleStMgrClientRegistered() {
...
+  // The default client reconnect time is 1s and the default max attempt is 300.
+  // Therefore if a connection is broken, stmgr would be restartd in 5 minutes by stmgr client.
+  // Although we want to restart stmgr sooner if there are more broken connections,
+  // it could be risky if the threshold is too agressive and causing stmgr to restart too soon.
+  // Here the max attempt across all clients is set to: per client max attempt * client count / 8.
+  // Hence if all connections are broken, the lower bound of restart time is about 40s.
+  // However in reality, broken connections cause extra delays hence less reconnects are made.
+  // In one case, it takes more than 120 minutes for a stmgr client to reach the max attempt and
+  // restart. So the stmgr is expected to be restarted in 10 to 20 minutes with the / 8 factor.
+  // The mininum max attempt is per client max attempt * 2 in case number of clients is too low.
```
huijunw: how is your proposal different from 'change 300 to 300/8'

nwangtw: The issue here is for a bad stmgr, all connections are failing and it is more likely to be a local problem. with a global max attempt, stmgr doesn't need to wait per client failure count to reach 300. Because network failure can be very slow, bad stmgr tends to restart a lot slower than good stmgr.

when there is only one bad connection, it is more likely to be a remote problem, so we don't  want the stmgr to kill itself too soon. changing 300 to 300/8 makes good stmgrs to quit a lot sooner and this is not what we want.

huijunw: cool, got it

nwangtw: The PR is a simple demo and I don't really like this solution myself. I just think we need a check based on global view. Any ideas are welcome.

nwangtw:
Dropping this PR in favor of the new solution: https://github.com/twitter/heron/pull/2711

PR #2697, opened 2018-01-30
Make scala dependencies from rules_scala explicit

nwangtw:
Currently some internal dependencies are used directly and can not be preprocessed.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -875,10 +875,12 @@ new_http_archive(
 
 # scala integration
 rules_scala_version="5cdae2f034581a05e23c3473613b409de5978833" # update this as needed
+scala_version = "2.11"
+scala_point_version = scala_version +".11"
 
 http_archive(
     name = "io_bazel_rules_scala",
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -883,7 +883,74 @@ http_archive(
...
-load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repositories")
-scala_repositories()
-load("@io_bazel_rules_scala//specs2:specs2_junit.bzl","specs2_junit_repositories")
-specs2_junit_repositories()
\ No newline at end of file
+SCALA_BUILD_FILE = """
+# scala.BUILD
+java_import(
+    name = "scala-xml",
+    jars = ["lib/scala-xml_2.11-1.0.5.jar"],
```
erenavsarogullari: Hi @nwangtw,

Thanks for working on this.

For long-term maintenance perspective, i think the following constants can help us when Scala version is upgraded with just small change (if makes sense)
```
scala_version = "2.11"
scala_point_version = scala_version +".11"
```

nwangtw:
dropping this PR for now. Looking for different ways.

PR #2701, opened 2018-01-31
Necessary software for single docker image

kramasamy:


tools/rules/heron_tools.bzl
```diff
@@ -4,6 +4,15 @@
...
+def heron_tools_bin_all_files():
+    return [
+        "//heron/tools/tracker/src/python:heron-tracker",
+        "//heron/tools/ui/src/python:heron-ui",
+        "//heron/tools/apiserver/src/shell:heron-apiserver",
+        "//heron/tools/cli/src/python:heron",
+        "//heron/tools/explorer/src/python:heron-explorer",
+    ]
+
 def heron_tools_bin_files():
```
cckellogg: is this needed anymore?

kramasamy: @cckellogg - you mean heron_tools_bin_files - probably leave it for now until we don't need it.

cckellogg: Yes the heron_tools_bin_files it looks like the new one replaces it.

kramasamy: yes for now - Let us leave the old one and remove it later.

kramasamy: @cckellogg - we need both since one is used for heron binary packaging while the other is used for docker container.

PR #2706, opened 2018-02-01
[metricscache] reduce exception stack trace to first 2 lines

huijunw:
Our customers requires to reduce the exception stack trace size.
I found the spi comment (https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/metricsmgr/metrics/ExceptionInfo.java#L37) showing 'first 2 lines', but the present tmaster-metricscollector and metricscache, they include all lines. This PR reduces the metricscachesink to first 2 lines while keeping tmaster-metricscolletor untouched.


nwangtw:
2 might be a bit aggressive. How many lines are generated currently?

Also just to confirm: first 2 lines are the most recent ones?

huijunw:
The line count depends. As far as I saw, they vary from 40 to 90.
The first 2 lines are shows what the exception is (the first line) and where the exception is thrown (the second line)

PR #2707, opened 2018-02-01
[metricscache] align metricscache toggle switch

huijunw:
Since the present metricscache works only for healthmgr/dhalion, our customers would like to align them.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -509,9 +509,9 @@ class HeronExecutor(object):
         '--ckptmgr_port=%s' % str(self.checkpoint_manager_port)]
     retval["heron-tmaster"] = tmaster_cmd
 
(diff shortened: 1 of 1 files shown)


heron/executor/src/python/heron_executor.py
```diff
@@ -509,9 +509,9 @@ def _get_tmaster_processes(self):
         '--ckptmgr_port=%s' % str(self.checkpoint_manager_port)]
     retval["heron-tmaster"] = tmaster_cmd
 
-    retval["heron-metricscache"] = self._get_metrics_cache_cmd()
 
     if self.health_manager_mode.lower() != "disabled":
+      retval["heron-metricscache"] = self._get_metrics_cache_cmd()
```
nwangtw: Suggest to add a comment here.

huijunw: added

PR #2711, opened 2018-02-03
Move stmgr quitting logic from stmgr client to tmaster client

nwangtw:
We found out that when a bad stmgr is experiencing connection issue to
all other stmgrs, it takes the stmgr a lot longer to reach the per stmgr
client max attempt and quite becaus bad connections cause extra delays.
In our case, instead of quitting after 5 minutes (1s reconnect timeout,
300 max attempts), the bad stmgr quitted after more than 2 hours. At the
same time, all other good stmgrs were quitting every 5 minutes.

Here is a more detailed document about the issue:
https://docs.google.com/document/d/1WHNc2NEp2gVL9ge2QVKp9t4Hpd4U9sAbzBqCu4-iDUM/edit#

### DIFF
--- a/heron/common/src/cpp/config/heron-internals-config-reader.cpp
+++ b/heron/common/src/cpp/config/heron-internals-config-reader.cpp
@@ -234,8 +234,8 @@ sp_int32 HeronInternalsConfigReader::GetHeronStreammgrXormgrRotatingmapNbuckets(
   return config_[HeronInternalsConfigVars::HERON_STREAMMGR_XORMGR_ROTATINGMAP_NBUCKETS].as<int>();
 }
 
-sp_int32 HeronInternalsConfigReader::GetHeronStreammgrClientReconnectMaxAttempts() {
-  return config_[HeronInternalsConfigVars::HERON_STREAMMGR_CLIENT_RECONNECT_MAX_ATTEMPTS].as<int>();
+sp_int32 HeronInternalsConfigReader::GetHeronStreammgrClientReconnectTmasterMaxAttempts() {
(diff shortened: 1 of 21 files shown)


nwangtw:
From @skanjila 

I could't add comments to the document, thus am posting my comments to the
mailing list

One more approach could be to do the current measurement as it is, but
instead of leaving the quitting decision to the stmgtclient, have
stmgrclientmgr do the decision. Thus everytime a stmgr client detects
connection issues, inform that to stmgrclientmgr which keeps a map of
peerstmgrid to error count. Thus it is able to decide things like am i
seeing connection errors from all stmgrs or if only a few of them are
having issues. Then it can take the decisions better.

Reply:

Yeah. That is an option too. In fact it was my first try:
https://github.com/twitter/heron/pull/2693 (just an initiative, not
completed, a count map should be used instead of a single total count)

In most cases, I think both solutions should have the same result. A few
reasons I changed to a tmaster check:
- with tmaster, there is only one source of truth and tmaster is more
critical anyway. If the tmaster link is not healthy, stmgrs won't work
correctly: topology may have created replacement nodes but the disconnected
nodes could keep going by themselves.
- it is more straightforward. The logic is the same as the current one. One
the other side, if we use an array for all remote stmgrs, we could have a
smarter logic (which is good) but it could make stmgrs more complicated and
less straightforward (bad). I left the stmgr counters there so if in future
we decide to add this feature, it should be easy to add. There is a gap
between "errors from all" and "errors from a few" and this is not a
simple/quick question.

nwangtw:
thanks!

PR #2717, opened 2018-02-06
deprecate RRPacking in default config

huijunw:
use ResourceCompliantRRPacking as default in config to make re-packing compatible with packing.

ps: safest is to use the same alg for both packing and repacking

### DIFF
--- a/heron/config/src/yaml/conf/aurora/packing.yaml
+++ b/heron/config/src/yaml/conf/aurora/packing.yaml
@@ -1,2 +1,3 @@
 # packing algorithm for packing instances into containers
-heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.RoundRobinPacking
+heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.ResourceCompliantRRPacking
+heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking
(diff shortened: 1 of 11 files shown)


heron/config/src/yaml/conf/aurora/packing.yaml
```diff
@@ -1,2 +1,3 @@
 # packing algorithm for packing instances into containers
-heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.RoundRobinPacking
+heron.class.packing.algorithm:    com.twitter.heron.packing.roundrobin.ResourceCompliantRRPacking
+heron.class.repacking.algorithm:  com.twitter.heron.packing.binpacking.FirstFitDecreasingPacking
```
kramasamy: @huijunw - thanks for the PR - should we use the same algorithm for both?

huijunw: done

nwangtw:
LGTM.

Related question: if RoundRobin is not good, should we deprecate it? Like removing it or adding a warning when user uses it? I am wondering if there are some cases that RoundRobin is better than the others.

huijunw:
I would prefer keeping the RR alg since it is tested in production and simple enough as a reference for new developers.

nwangtw:
SGTM. Thanks

PR #2720, opened 2018-02-06
fix a RRrepacking bug

huijunw:
The argument to SPI repacking is delta value rather than target value.


maosongfu:
Do we have unit tests for this?

PR #2724, opened 2018-02-07
Allowing API Server to distribute heron core package

jerrypeng:


heron/config/src/yaml/conf/nomad/client.yaml
```diff
@@ -2,3 +2,7 @@
 heron.package.core.uri:                      file://${HERON_DIST}/heron-core.tar.gz
```
kramasamy: Do we need this - if we are changing the config later in the same file?

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -124,27 +122,38 @@ public Response uploadFile(
...
+    rb.header("content-disposition", "attachment; filename = "
+        + file);
+    return rb.build();
+  }
+
+  /**
+   * Endpoint for downloading Heron Core
+   */
+  @GET
+  @Path("/downloadcore")
```
kramasamy: should we have something like /download/core rather /downloadcore

kramasamy:
Now the question is what happens if the heron api server moves to different node because nomad choose to?

jerrypeng:
@kramasamy you can run the api server by itself and not necessarily in nomad.  If a user already has a nomad cluster, we can assume they will use some sort of service discovery mechanism (Consul) to  find api server anyways

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
jerrypeng: I rather it not be under the same endpoint as the topology endpoint in case we have classes

heron/config/src/yaml/conf/nomad/client.yaml
jerrypeng: we are not changing it in the file.  The later config is commented an serves just as a reminder that user can do that as well

kramasamy:
@jerrypeng - if is not about the service discovery mechanism - it is about the core file. Should core file be available in the same node as api server - if that it the case when the api server moves - how are we going to move the core file or the assumption is that the core file needs to be available at neutral location - accessible from all nodes.

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
jerrypeng: I can make that change

jerrypeng:
@kramasamy I see that can be a problem. You can also pin the api server to run only on a set of nodes or not run in Nomad as well.  Not sure what else we can do about that.  The user also don't have to use the API server for distributing core.  They can also just upload to S3 and expose a HTTP link

jerrypeng:
@kramasamy thanks for the review.  Do you have any more concerns?

PR #2728, opened 2018-02-09
[integration test] add flush to AggregatorBolt

huijunw:
We found the AggregatorBolt may fail to emit tuples to Http server in some scenarios.
This PR flushes the tuples before its death.


huijunw:
Close this PR. Let's find a better way to do this

PR #2730, opened 2018-02-13
Allow user to run kill command when some topology data is missing

nwangtw:
When kill or submit command is not fully successful, it is possible that there are some incomplete data for the topology. We need to allow users to run kill command in this case instead of rejecting it because of missing data.

Previous tries:
https://github.com/twitter/heron/pull/2685
https://github.com/twitter/heron/pull/2689

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -377,7 +377,7 @@ public class RuntimeManagerMain {
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      boolean foundData = validateRuntimeManage(adaptor, topologyName);
+      boolean hasRunningData = validateRuntimeManage(adaptor, topologyName);
 
       // 2. Try to manage topology if valid
       // invoke the appropriate command to manage the topology
@@ -392,7 +392,7 @@ public class RuntimeManagerMain {
       // Create a ISchedulerClient basing on the config
       ISchedulerClient schedulerClient = getSchedulerClient(runtime);
 
-      callRuntimeManagerRunner(runtime, schedulerClient, foundData);
+      callRuntimeManagerRunner(runtime, schedulerClient, hasRunningData);
     } finally {
       // 3. Do post work basing on the result
       // Currently nothing to do here
@@ -417,8 +417,8 @@ public class RuntimeManagerMain {
       String topologyName) throws TopologyRuntimeManagementException {
     // Check whether the topology has already been running
     Boolean isTopologyRunning = adaptor.isTopologyRunning(topologyName);
-    boolean found = isTopologyRunning != null && isTopologyRunning.equals(Boolean.TRUE);
-    if (!found) {
+    boolean hasRunningData = isTopologyRunning != null && isTopologyRunning.equals(Boolean.TRUE);
+    if (!hasRunningData) {
       if (command == Command.KILL) {
         LOG.warning(String.format("Topology '%s' is not found or not running", topologyName));
       } else {
@@ -454,17 +454,17 @@ public class RuntimeManagerMain {
             topologyName, currentState, configState));
       }
     }
-    return found;
+    return hasRunningData;
   }
 
   protected void callRuntimeManagerRunner(
       Config runtime,
       ISchedulerClient schedulerClient,
-      boolean foundData)
+      boolean hasRunningData)
     throws TopologyRuntimeManagementException, TMasterException, PackingException {
     // create an instance of the runner class
     RuntimeManagerRunner runtimeManagerRunner =
-        new RuntimeManagerRunner(config, runtime, command, schedulerClient, foundData);
+        new RuntimeManagerRunner(config, runtime, command, schedulerClient, hasRunningData);
 
     // invoke the appropriate handlers based on command
     runtimeManagerRunner.call();
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
@@ -377,7 +377,7 @@ public class RuntimeManagerMain {
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      boolean hasRunningData = validateRuntimeManage(adaptor, topologyName);
+      boolean hasExecutionData = validateRuntimeManage(adaptor, topologyName);
 
       // 2. Try to manage topology if valid
       // invoke the appropriate command to manage the topology
@@ -392,7 +392,7 @@ public class RuntimeManagerMain {
       // Create a ISchedulerClient basing on the config
       ISchedulerClient schedulerClient = getSchedulerClient(runtime);
 
-      callRuntimeManagerRunner(runtime, schedulerClient, hasRunningData);
+      callRuntimeManagerRunner(runtime, schedulerClient, !hasExecutionData);
     } finally {
       // 3. Do post work basing on the result
       // Currently nothing to do here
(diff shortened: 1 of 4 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -48,13 +48,18 @@
...
   private final ISchedulerClient schedulerClient;
+  private final boolean running;
 
-  public RuntimeManagerRunner(Config config, Config runtime,
-                              Command command, ISchedulerClient schedulerClient) {
+  public RuntimeManagerRunner(Config config,
+                              Config runtime,
+                              Command command,
+                              ISchedulerClient schedulerClient,
+                              boolean running) {
```
maosongfu: 2 minor comments:
1. Why pass the flag `running` in the constructor, rather than in the `Config runtime`?
2. Name the boolean as `isRunning` will be more expressive and more java style.

nwangtw: Thanks!

For 1, "config" is for the topology configuration. running flag is not a config.
For 2, will do.

maosongfu: OK Fair enough.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -90,6 +95,7 @@ public void call()
    * Handler to activate a topology
    */
   private void activateTopologyHandler(String topologyName) throws TMasterException {
+    assert isRunning;
```
kramasamy: Here the topology might not be in the running state?

nwangtw: "Running" means there are both topology defn and execution state data.

If any of them is not available, it means the topology is not "running". This could be caused by a failed submit command or a failed kill command.

Like we discussed before, in this state, kill command won't work. User has to learn how heron stores data in different places and checks/clears them manually, which is not good. In this PR, kill command is allowed with warning messages.

nwangtw: And for all other commands, validation code shouldn't allow them to be executed. So in this line, isRunning flag should always be true.

maosongfu: Karthik mentioned a good point. `Running` can be confusing since we have used it already for topology state `Running`, `Paused`, etc.
Maybe we could rename it to `isExisting`?

nwangtw: That is true. Let me think about it. Thanks.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -157,8 +165,12 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     // clean up the state of the topology in state manager
     cleanState(topologyName, Runtime.schedulerStateManagerAdaptor(runtime));
 
-    // Clean the connection when we are done.
-    LOG.fine("Scheduler killed topology successfully.");
+    if (foundData) {
+      LOG.fine(String.format("Scheduler killed topology %s successfully.", topologyName));
+    } else {
+      LOG.warning(String.format("Topology %s data is not found, but scheduler has tried to"
```
kramasamy: What is this message for?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -378,7 +377,7 @@ public void manageTopology()
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      validateRuntimeManage(adaptor, topologyName);
+      boolean foundData = validateRuntimeManage(adaptor, topologyName);
```
kramasamy: foundData - does not sound like a good name. Any other choices?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
nwangtw: I think if topology data is not available, it is possible that the topology was not running.  The message "Scheduler killed topology %s successfully."  can be misleading. So I think in this case, "scheduler has tried to clean up" is more accurate.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
nwangtw: sigh.

Maosong suggested "exist" but I feel it is not good either. :(

This is a trick case and naming is tricky too.

kramasamy: this data is all about executionState - is it correct? if so we can name it as
```
executionStateExists
```
will do?

nwangtw: - dataFound
- topologyFound
- topologyExist
or reverse them:
- topologyNotFound
- topologyNotExist

Which one sounds better?

nwangtw: it is not only about executionState.

We got a report later that executionStateExists is available but definition is not. Basically none of them is safe.

This flag is from this logic:
Boolean isTopologyRunning = adaptor.isTopologyRunning(topologyName);
boolean found = isTopologyRunning != null && isTopologyRunning.equals(Boolean.TRUE);

That is why I used isRunning at beginning, which is not ideal for sure.

Another option is to skip the validation altogether for kill command and no need to worry about which data is available. The checks and flags here are used to show different message, except envronment check if execution state is available (which isn't critical for this command to me).

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -378,7 +377,7 @@ public void manageTopology()
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      validateRuntimeManage(adaptor, topologyName);
+      boolean hasRunningData = validateRuntimeManage(adaptor, topologyName);
```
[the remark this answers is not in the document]
kramasamy: alternatively a better one is 

```
staleExecutionData
```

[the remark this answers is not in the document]
nwangtw: kk. sgtm. let me make the change.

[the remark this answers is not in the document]
nwangtw: It seems "ExecutionState" is a bit unclear here. One zk node is for "ExecutionState", the code is:

ExecutionEnvironment.ExecutionState executionState = adaptor.getExecutionState(topologyName);

I am going to use "hasExecutionData" in this function and hasStaleExecutionData in Runner class. Let me know if I missed something.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -403,46 +402,69 @@ public void manageTopology()
...
-          "cluster/role/environ does not match. Topology '%s' is running at %s, not %s",
-          topologyName, currentState, configState));
+      if (command == Command.KILL) {
+        LOG.warning(String.format("Topology execution state for '%s' is not found", topologyName));
+      } else {
+        throw new TopologyRuntimeManagementException(
+            String.format("Failed to get execution state for topology %s", topologyName));
+      }
+    } else {
+      // Execution state is available, validate configurations.
```
kramasamy: @nwangtw - can you make this as a function?

nwangtw: sounds good. let me do it.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -157,8 +165,12 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     // clean up the state of the topology in state manager
     cleanState(topologyName, Runtime.schedulerStateManagerAdaptor(runtime));
 
-    // Clean the connection when we are done.
-    LOG.fine("Scheduler killed topology successfully.");
+    if (hasStaleExecutionData) {
+      LOG.warning(String.format("Topology %s data is not found, but scheduler has tried to"
+          + " clean up data for it anyway.", topologyName));
+    } else {
+      LOG.fine(String.format("Scheduler killed topology %s successfully.", topologyName));
```
kramasamy: this message might not be misleading - since we are not killing the topology I think.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -378,7 +377,7 @@ public void manageTopology()
       // TODO(mfu): timeout should read from config
       SchedulerStateManagerAdaptor adaptor = new SchedulerStateManagerAdaptor(statemgr, 5000);
 
-      validateRuntimeManage(adaptor, topologyName);
+      boolean hasExecutionData = validateRuntimeManage(adaptor, topologyName);
```
kramasamy: is this variable 

```
hasStaleExecutionData
```

Issue #1177, opened 2016-07-28
Heron on Mesos


wking1986:
Hi maosongfu:
     heron on mesos  appears a problem :

objmagic:
Which version of Heron are you using, or what git commit id of the Heron you built?

wking1986:
![image](https://cloud.githubusercontent.com/assets/4937857/17201033/0a6d226c-54bd-11e6-8b6d-dfa2c30d3943.png)

heron version is ‘master’
git clone date is : 2016.7.27

kramasamy:
@wking1986 - it looks like a PEX issue. When pex is exploding the zip file, it seems to be writing into a directory where permission is denied. Are you running Mesos locally? Can you do ```ls -al``` of the directory where you have heron-executor.stderr logs. You should see a directory called .pex there. If that is not there, there is some issue.

objmagic:
@kramasamy actually it is because that the OS call that sets process group lead does not have permission. See [heron-executor.py#L501](https://github.com/twitter/heron/blob/816813f1838f538c041f690eef1380b2e99facd6/heron/executor/src/python/heron-executor.py#L501) and [setpgrp manual](https://www.cl.cam.ac.uk/cgi-bin/manpage?2+setpgrp)

kramasamy:
thanks @objmagic - @wking1986 - what user do you run your topologies as?

akozich:
I faced the same issue. I'm running:
- heron 0.17.4
- mesos 1.4.1
- CentOS 7.4

This exception "Operation not permitted" is mapped from underlying EPERM system error code. This code means that current process is session master and by POSIX standard is not permitter to change process group.

I added some debug logging just before executing 'setpgrp' to print parent processes with corresponding process id, process group id, session id, parent process id and here's the result:

PID		PGID	SESS	PPID	ARGS
31845	31845	31845	31835	python2.7 ./heron-core/bin/heron-executor --topology-name=WordCountTopology ...
31835	31800	31800	31800	mesos-executor --launcher_dir=/usr/libexec/mesos
31800	31800	31800	6110	/usr/libexec/mesos/mesos-containerizer launch
6110	6110	6110	1		/usr/sbin/mesos-slave --master=zk://zookeeper:2181/mesos --log_dir=/var/log/mesos --containerizers=mesos --hostname_lookup=false --work_dir=/var/lib/mesos

As you can see heron-executor is started by mesos-executor in a new session and is a master of that session.

PR #2731, opened 2018-02-13
Check that current process is not session master

akozich:
According to posix specification it is not permitted to change process group of session master. In this case setpgrp will fail with EPERM. Before calling setpgrp we should first check that current process is not a session master.

https://github.com/twitter/heron/issues/1177


kramasamy:
@akozich - thanks for the PR. waiting for the ci to finish.

kramasamy:
@jrcrawfo - wondering if you have some feedback.

PR #2733, opened 2018-02-13
[hold]restore the initial state when `aurora job add` command fails during `heron update`

huijunw:
When the `aurora job add xxx` fails due to some reason, the activated topology stays `deactivated`. The PR fixes the issue by putting it in a try/finally.

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -153,39 +153,41 @@ public class UpdateTopologyManager implements Closeable {
       deactivateTopology(stateManager, topology, proposedPackingPlan);
     }
 
-    Set<PackingPlan.ContainerPlan> updatedContainers =
-        new HashSet<>(proposedPackingPlan.getContainers());
-    // request new resources if necessary. Once containers are allocated we should make the changes
-    // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
-    if (newContainerCount > 0 && scalableScheduler.isPresent()) {
-      Set<PackingPlan.ContainerPlan> containersToAdd = containerDelta.getContainersToAdd();
-      Set<PackingPlan.ContainerPlan> containersAdded =
-          scalableScheduler.get().addContainers(containersToAdd);
-      // Update the PackingPlan with new container-ids
-      if (containersAdded != null) {
-        updatedContainers.removeAll(containersToAdd);
-        updatedContainers.addAll(containersAdded);
+    try {
+      Set<PackingPlan.ContainerPlan> updatedContainers =
+          new HashSet<>(proposedPackingPlan.getContainers());
+      // request new resources if necessary. Once containers are allocated we should make the changes
+      // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
+      if (newContainerCount > 0 && scalableScheduler.isPresent()) {
+        Set<PackingPlan.ContainerPlan> containersToAdd = containerDelta.getContainersToAdd();
+        Set<PackingPlan.ContainerPlan> containersAdded =
+            scalableScheduler.get().addContainers(containersToAdd);
+        // Update the PackingPlan with new container-ids
+        if (containersAdded != null) {
+          updatedContainers.removeAll(containersToAdd);
+          updatedContainers.addAll(containersAdded);
+        }
       }
-    }
-
-    PackingPlan updatedPackingPlan =
-        new PackingPlan(proposedPackingPlan.getId(), updatedContainers);
-    PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
-    PackingPlans.PackingPlan updatedProtoPackingPlan = serializer.toProto(updatedPackingPlan);
-    LOG.fine("The updated Packing Plan: " + updatedProtoPackingPlan);
-
-    // update packing plan to trigger the scaling event
-    logInfo("Update new PackingPlan: %s",
-        stateManager.updatePackingPlan(updatedProtoPackingPlan, topologyName));
-
-    // reactivate topology
-    if (initiallyRunning) {
-      // wait before reactivating to give the tmaster a chance to receive the packing update and
-      // delete the packing plan. Instead we could message tmaster to invalidate the physical plan
-      // and/or possibly even update the packing plan directly
-      SysUtils.sleep(Duration.ofSeconds(10));
-      // Will throw exceptions internally if tmaster fails to deactivate
-      reactivateTopology(stateManager, topology, removableContainerCount);
+  
+      PackingPlan updatedPackingPlan =
+          new PackingPlan(proposedPackingPlan.getId(), updatedContainers);
+      PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
+      PackingPlans.PackingPlan updatedProtoPackingPlan = serializer.toProto(updatedPackingPlan);
+      LOG.fine("The updated Packing Plan: " + updatedProtoPackingPlan);
+  
+      // update packing plan to trigger the scaling event
+      logInfo("Update new PackingPlan: %s",
+          stateManager.updatePackingPlan(updatedProtoPackingPlan, topologyName));
+    } finally {
+      // reactivate topology
+      if (initiallyRunning) {
+        // wait before reactivating to give the tmaster a chance to receive the packing update and
+        // delete the packing plan. Instead we could message tmaster to invalidate the physical plan
+        // and/or possibly even update the packing plan directly
+        SysUtils.sleep(Duration.ofSeconds(10));
+        // Will throw exceptions internally if tmaster fails to deactivate
+        reactivateTopology(stateManager, topology, removableContainerCount);
+      }
     }
 
     if (removableContainerCount > 0 && scalableScheduler.isPresent()) {
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
@@ -156,8 +156,9 @@ public class UpdateTopologyManager implements Closeable {
     try {
       Set<PackingPlan.ContainerPlan> updatedContainers =
           new HashSet<>(proposedPackingPlan.getContainers());
-      // request new resources if necessary. Once containers are allocated we should make the changes
-      // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
+      // request new resources if necessary. Once containers are allocated we should make the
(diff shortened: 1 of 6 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -153,39 +153,41 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+      }
+  
+      PackingPlan updatedPackingPlan =
+          new PackingPlan(proposedPackingPlan.getId(), updatedContainers);
+      PackingPlanProtoSerializer serializer = new PackingPlanProtoSerializer();
+      PackingPlans.PackingPlan updatedProtoPackingPlan = serializer.toProto(updatedPackingPlan);
+      LOG.fine("The updated Packing Plan: " + updatedProtoPackingPlan);
+  
+      // update packing plan to trigger the scaling event
+      logInfo("Update new PackingPlan: %s",
```
maosongfu: Need to think about whether to rollback the packing_plan on zk if failure happens.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -153,39 +153,41 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+          stateManager.updatePackingPlan(updatedProtoPackingPlan, topologyName));
+    } finally {
+      // reactivate topology
+      if (initiallyRunning) {
```
maosongfu: Would be much better if we can print meaningful log to notify users when unrecoverable bad things happen, like:
```
Failing to update topology. The topology can be in a strange stage. Please check carefully or redeploy the topology.
```

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -153,39 +153,41 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
-        updatedContainers.addAll(containersAdded);
+    try {
+      Set<PackingPlan.ContainerPlan> updatedContainers =
+          new HashSet<>(proposedPackingPlan.getContainers());
+      // request new resources if necessary. Once containers are allocated we should make the changes
+      // to state manager quickly, otherwise the scheduler might penalize for thrashing on start-up
+      if (newContainerCount > 0 && scalableScheduler.isPresent()) {
+        Set<PackingPlan.ContainerPlan> containersToAdd = containerDelta.getContainersToAdd();
+        Set<PackingPlan.ContainerPlan> containersAdded =
+            scalableScheduler.get().addContainers(containersToAdd);
```
maosongfu: Need to check whether we successfully add containers.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: There are two scenarios: recoverable and unrecoverable. Aurora may be in unknown state when it fails, resulting unrecoverable situation.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: The update procedure may throw several exceptions which are hard to be differentiated in the upper function. I added the suggested print msg in the present exception msg.

maosongfu:
1. Can we have pre-check whether the role has enough resources before doing the actual scale-up/down (Careful, by-pass this step if the environment is not prod)?

2. If the role does not have enough resources, we would reject the request directly, rather than potentially to put the topology in a wired state.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -112,10 +123,122 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+      return true;
+    }
+    return false;
+  }
+
   @Override
   public Set<Integer> addContainers(Integer count) {
-    //aurora job add <cluster>/<role>/<env>/<name>/<instance_id> <count>
-    //clone instance 0
+    try {
```
maosongfu: I think this `check quota` logic should be at the upper level rather than at this level. i.e. in `AuroraScheduler.addContainers(...)`

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -112,10 +123,122 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+      JsonNode jsonStatus = askAuroraStatus();
```
maosongfu: We can simply get the env from `Context` (https://github.com/twitter/heron/blob/master/heron/spi/src/java/com/twitter/heron/spi/common/Context.java#L36). No need to ask Aurora Status.

huijunw: i do not think so. the aurora tier is not same with env in job spec key

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: do you mean adding `check quota` to spi/scheduler interface

maosongfu: I mean, we can do this 
```
try {
      JsonNode jsonStatus = askAuroraStatus();
     ...
}
```
in `AuroraScheduler.addContainers(...)`

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
maosongfu: I mean, only when users try to update a topology in `prod` env, which can be parsed from the `Context`, we will try to `askAuroraQuota()`; we don't need the method `askAuroraStatus()`

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -112,10 +123,122 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
...
+      String msg = String.format(
+          "not enough disk quota: %d disk per container x %d container > %d available",
+          containerDiskMb, count, availableDiskMb);
+      throw new TopologyUpdateRecoverableException(msg);
+    }
+    return true;
+  }
+
+  private boolean isProd(JsonNode jsonStatus) {
+    String tier = jsonStatus.get(0).get("active").get(0).get("assignedTask").get("task").get("tier")
```
nwangtw: This is not safe.

Although "tier = preferred" is the current config for production jobs,  I would expect that "production: true" is still being used in many other companies.

Also tier is customizable, so I think there is no safe way to check isProd().

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
nwangtw: It is not a safe assumption.

maosongfu: @nwangtw it should be safe; this should be verified ahead to come to the actual update stage.
Furthermore, topology-name is unique for a cluster, so there will be not topologies with the same name in one cluster.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: if we cannot know if topo is prod anyway, shall we use cluster/role/`env` to judge isProd() as maosong suggested previously. so that the code will be simple and stupid .. what do you think @maosongfu  @nwangtw

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -112,10 +123,122 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
     }
   }
 
+  private JsonNode askAuroraQuota() throws JsonProcessingException, IOException {
```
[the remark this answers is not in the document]
maosongfu: I agree it is a bit overkill, if we are to use this implementation.

But the point is: it is pretty common that users forget requesting more resources and then fail with insufficient resources. It is a super bad experience.

Another thought is: before the actual update, we could remind customers of providing enough resources. For instance, to print: `Please make sure there are sufficient resources to update this job? [Y/N]` and proceed when users input `Y`

@nwangtw @huijunw Thoughts?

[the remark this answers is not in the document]
huijunw: If there is prompt, do we need this pr still? I assume user types Y and takes the risk

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
nwangtw: the env value is less safe. :(

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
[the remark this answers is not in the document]
nwangtw: I agree with maosong that this is a quite common issue and we should see what we can do. I think a warning could be helpful for users to be aware of this issue. They are free to take the risk but it is our problem that they don't know the risk.

Ideally we should be able to tell user how many new containers (no need to be core/ram) will be added so user can verify either there is enough quota or there is no quota cap before proceeding.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
nwangtw: I mean from aurora point of view.

From aurora doc it seems to me that environments are configurable. For example, inside Twitter we call it "prod" and in apache doc it is called "production".

Another one is that users might config their prod jobs to be preemptible if the SLA is low, which doesn't need quota although it is a "prod" job. This requires user to customize the aurora configs, but it is quite easy from heron design.

huijunw:
replaced by #2735

PR #2735, opened 2018-02-16
add warning when `aurora job add` command fails during `heron update`

huijunw:
replace #2733

After discussion with maosong and ning, we agree:
1 Warn users clearly before the actual update process. So they can know the risks. (done by prompt)
2 Warn users clearly after failure. And let users know the topology can enter a wired state, and a re-deployment is recommended. 

Since the warning tells the users to assure the topology's running state, heron does not try to activate it.

Example:
```
$ ~/bin/heron update --config-property heron.command.update.prompt=prompt \
--verbose local ExclamationTopology --component-parallelism word:4
```
Default config value `disabled`

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -14,7 +14,6 @@
 
 package com.twitter.heron.scheduler;
 
-import java.io.Console;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -14,6 +14,7 @@
 
 package com.twitter.heron.scheduler;
 
+import java.io.Console;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
@@ -161,6 +162,25 @@ public class RuntimeManagerRunner {
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+    String fmt =
+        "The present aurora job has %d containers. After update there will be %d containers. ";
+    Console c = System.console();
+    if (c == null) {
+      LOG.warning("No console to prompt user");
+      System.err.println(String.format(fmt, oldContainerCount, newContainerCount));
+      return true;
+    }
+
+    String userInput = c.readLine(
+        String.format(fmt + " Please make sure there are sufficient resources to update this job."
+            + " Continue update? [Y/n]: ", oldContainerCount, newContainerCount));
+    if (userInput.equalsIgnoreCase("n")) {
+      return false;
+    }
+    return true;
+  }
+
   /**
    * Handler to update a topology
    */
@@ -188,6 +208,15 @@ public class RuntimeManagerRunner {
       PackingPlan oldPlan = deserializer.fromProto(currentPlan);
       PackingPlan newPlan = deserializer.fromProto(proposedPlan);
       throw new UpdateDryRunResponse(topology, config, newPlan, oldPlan, changeRequests);
+    }
+    
+    int newContainerCount = proposedPlan.getContainerPlansCount();
+    int oldContainerCount = currentPlan.getContainerPlansCount();
+    if (newContainerCount > oldContainerCount) {
+      if (!consolePrompt(oldContainerCount, newContainerCount)) {
+        LOG.fine("Scheduler updated topology canceled.");
+        return;
+      }
     }
 
     Scheduler.UpdateTopologyRequest updateTopologyRequest =
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -163,7 +163,7 @@ public class RuntimeManagerRunner {
   }
 
   boolean confirmWithUser(int oldContainerCount, int newContainerCount) {
-    System.out.print(String.format("The present aurora job has %d containers. "
+    LOG.info(String.format("The present aurora job has %d containers. "
         + "After update there will be %d containers. "
         + "Please make sure there are sufficient resources to update this job. "
         + "Continue update? [y/N]: ", oldContainerCount, newContainerCount));
@@ -206,7 +206,7 @@ public class RuntimeManagerRunner {
 
     int newContainerCount = proposedPlan.getContainerPlansCount();
     int oldContainerCount = currentPlan.getContainerPlansCount();
-    if (newContainerCount > oldContainerCount && Context.updatePrompt(config)) {
+    if (newContainerCount > oldContainerCount && "prompt".equals(Context.updatePrompt(config))) {
       if (!confirmWithUser(oldContainerCount, newContainerCount)) {
         LOG.warning("Scheduler updated topology canceled.");
         return;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -162,23 +162,23 @@ public class RuntimeManagerRunner {
     LOG.fine("Scheduler killed topology successfully.");
   }
 
-  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+  boolean confirmWithUser(int oldContainerCount, int newContainerCount) {
     String fmt =
-        "The present aurora job has %d containers. After update there will be %d containers. ";
+        "The present aurora job has %d containers. After update there will be %d containers.";
     Console c = System.console();
     if (c == null) {
-      LOG.warning("No console to prompt user");
+      LOG.warning("No console to prompt user. Proceeding.");
       System.err.println(String.format(fmt, oldContainerCount, newContainerCount));
       return true;
     }
 
     String userInput = c.readLine(
         String.format(fmt + " Please make sure there are sufficient resources to update this job."
-            + " Continue update? [Y/n]: ", oldContainerCount, newContainerCount));
-    if ("n".equalsIgnoreCase(userInput)) {
-      return false;
+            + " Continue update? [y/N]: ", oldContainerCount, newContainerCount));
+    if ("y".equalsIgnoreCase(userInput)) {
+      return true;
     }
-    return true;
+    return false;
   }
 
   /**
@@ -213,8 +213,8 @@ public class RuntimeManagerRunner {
     int newContainerCount = proposedPlan.getContainerPlansCount();
     int oldContainerCount = currentPlan.getContainerPlansCount();
     if (newContainerCount > oldContainerCount) {
-      if (!consolePrompt(oldContainerCount, newContainerCount)) {
-        LOG.fine("Scheduler updated topology canceled.");
+      if (!confirmWithUser(oldContainerCount, newContainerCount)) {
+        LOG.warning("Scheduler updated topology canceled.");
         return;
       }
     }
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -175,7 +175,7 @@ public class RuntimeManagerRunner {
     String userInput = c.readLine(
         String.format(fmt + " Please make sure there are sufficient resources to update this job."
             + " Continue update? [Y/n]: ", oldContainerCount, newContainerCount));
-    if (userInput.equalsIgnoreCase("n")) {
+    if ("n".equalsIgnoreCase(userInput)) {
       return false;
     }
     return true;
@@ -209,7 +209,7 @@ public class RuntimeManagerRunner {
       PackingPlan newPlan = deserializer.fromProto(proposedPlan);
       throw new UpdateDryRunResponse(topology, config, newPlan, oldPlan, changeRequests);
     }
-    
+
     int newContainerCount = proposedPlan.getContainerPlansCount();
     int oldContainerCount = currentPlan.getContainerPlansCount();
     if (newContainerCount > oldContainerCount) {
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -206,7 +206,8 @@ public class RuntimeManagerRunner {
 
     int newContainerCount = proposedPlan.getContainerPlansCount();
     int oldContainerCount = currentPlan.getContainerPlansCount();
-    if (newContainerCount > oldContainerCount && "prompt".equals(Context.updatePrompt(config))) {
+    if (newContainerCount > oldContainerCount
+        && "prompt".equalsIgnoreCase(Context.updatePrompt(config))) {
       if (!confirmWithUser(oldContainerCount, newContainerCount)) {
         LOG.warning("Scheduler updated topology canceled.");
         return;
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
@@ -227,9 +227,10 @@ public class RuntimeManagerRunner {
 
     LOG.fine("Sending Updating topology request: " + updateTopologyRequest);
     if (!schedulerClient.updateTopology(updateTopologyRequest)) {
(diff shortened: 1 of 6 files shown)


heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
...
+    if (c == null) {
+      LOG.warning("No console to prompt user");
+      System.err.println(String.format(fmt, oldContainerCount, newContainerCount));
+      return true;
+    }
+
+    String userInput = c.readLine(
+        String.format(fmt + " Please make sure there are sufficient resources to update this job."
+            + " Continue update? [Y/n]: ", oldContainerCount, newContainerCount));
+    if ("n".equalsIgnoreCase(userInput)) {
```
nwangtw: I think it might be better to check "y" instead of "n". don't forget to update "[Y/n]" to "[y/N]"

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+    String fmt =
+        "The present aurora job has %d containers. After update there will be %d containers. ";
+    Console c = System.console();
+    if (c == null) {
+      LOG.warning("No console to prompt user");
```
nwangtw: suggest: "No console to prompt user. Proceeding."

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+    String fmt =
+        "The present aurora job has %d containers. After update there will be %d containers. ";
```
nwangtw: looks like you dont need the space at the end of fmt.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
```
nwangtw: maybe rename to something like confirmWithUser(). How do you think?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -190,6 +210,15 @@ void updateTopologyHandler(String topologyName, String newParallelism)
       throw new UpdateDryRunResponse(topology, config, newPlan, oldPlan, changeRequests);
     }
 
+    int newContainerCount = proposedPlan.getContainerPlansCount();
+    int oldContainerCount = currentPlan.getContainerPlansCount();
+    if (newContainerCount > oldContainerCount) {
+      if (!consolePrompt(oldContainerCount, newContainerCount)) {
+        LOG.fine("Scheduler updated topology canceled.");
```
nwangtw: maybe a warning to be more explicit

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -161,6 +161,13 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
       Set<PackingPlan.ContainerPlan> containersToAdd = containerDelta.getContainersToAdd();
       Set<PackingPlan.ContainerPlan> containersAdded =
           scalableScheduler.get().addContainers(containersToAdd);
+      if (containersAdded.size() != containersToAdd.size()) {
```
nwangtw: I am wondering if it is possible to move this check to RuntimeManagerRunner.java.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
```diff
@@ -112,6 +112,9 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
     }
   }
 
+  private static final String ERR_PROMPT =
```
nwangtw: This error message is declared and used in a few different places. Is it possible to clean it up?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
```diff
@@ -161,6 +161,13 @@ private void updateTopology(final PackingPlans.PackingPlan existingProtoPackingP
...
+        updatedContainers.addAll(containersAdded);
+        throw new RuntimeException("Scheduler failed to add expected countainer count: added "
```
nwangtw: "Scheduler failed to add requested containers. Requested X, added Y. ..."

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+    String fmt =
+        "The present aurora job has %d containers. After update there will be %d containers. ";
+    Console c = System.console();
```
maosongfu: In general, it is a bad idea to require a console to be assigned as this very much impacts the flexibility of your application. Many ways of invoking Java will not assign a console (i.e. `System.console()` returns null), and your application is unusable in those instances.
(https://stackoverflow.com/questions/20982664/why-does-system-console-return-null-for-a-command-line-app)

You can always use `System.in` and `System.out` instead, as follows:

```
String qq;
Scanner scanner = new Scanner(System.in);
qq = scanner.nextLine();
System.out.println(qq);
```

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: removed

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: how to?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraCLIController.java
huijunw: this file in schedulers directory, the others are in scheduler-core. what do you suggest?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -190,6 +210,15 @@ void updateTopologyHandler(String topologyName, String newParallelism)
       throw new UpdateDryRunResponse(topology, config, newPlan, oldPlan, changeRequests);
     }
 
+    int newContainerCount = proposedPlan.getContainerPlansCount();
+    int oldContainerCount = currentPlan.getContainerPlansCount();
+    if (newContainerCount > oldContainerCount) {
+      if (!consolePrompt(oldContainerCount, newContainerCount)) {
```
maosongfu: I would suggest having an option (can be overridden by `config-override`) to bypass the prompt so it is easier to invoke the `update` command in other scripts or process, like ci, Jenkins.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,25 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean consolePrompt(int oldContainerCount, int newContainerCount) {
+    String fmt =
+        "The present aurora job has %d containers. After update there will be %d containers. ";
+    Console c = System.console();
+    if (c == null) {
+      LOG.warning("No console to prompt user");
+      System.err.println(String.format(fmt, oldContainerCount, newContainerCount));
```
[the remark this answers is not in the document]
huijunw: gone with Console

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: will do in another pr

huijunw: added the config to pass ci

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
nwangtw: The function addContainers() above the if statement is in AuroraCLIController, the code looks like:

public Set<Integer> addContainers(Integer count) {
  ....
  return extractContainerIds(stderr.toString());
}

You should be able to check container added vs count in there before returning I think. Then no changes will be needed in this file.

heron/spi/src/java/com/twitter/heron/spi/common/Key.java
```diff
@@ -170,7 +170,10 @@
   SHELL_BINARY          ("heron.binaries.shell",           "${HERON_BIN}/heron-shell"),
   PYTHON_INSTANCE_BINARY("heron.binaries.python.instance", "${HERON_BIN}/heron-python-instance"),
   CPP_INSTANCE_BINARY   ("heron.binaries.cpp.instance",    "${HERON_BIN}/heron-cpp-instance"),
-  DOWNLOADER_BINARY     ("heron.binaries.downloader",      "${HERON_BIN}/heron-downloader");
+  DOWNLOADER_BINARY     ("heron.binaries.downloader",      "${HERON_BIN}/heron-downloader"),
+
+  // keys for `heron` command line
+  UPDATE_PROMPT         ("heron.command.update.prompt", "disabled");
```
nwangtw: The config name is not very straghtforward. Suggest to add a comment for it. Also include the values:
disabled => no prompt
prompt => enable prompt.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/UpdateTopologyManager.java
huijunw: i think this check is for all schedulers, not only for aurora , as suggested by maosong

nwangtw: kk. that could be useful.

heron/spi/src/java/com/twitter/heron/spi/common/Key.java
nwangtw: still not very clear. How about let's do something like this:

// Prompt user when more containers are required so that user has another chance to double check quota is available. To enable it, change the config from "disabled" to "prompt".

huijunw: updated

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,19 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
     LOG.fine("Scheduler killed topology successfully.");
   }
 
+  boolean confirmWithUser(int oldContainerCount, int newContainerCount) {
+    LOG.info(String.format("The present aurora job has %d containers. "
```
kramasamy: This code is in scheduler core - you cannot assume that it is aurora specific. It is used by other schedulers. Please change this

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -161,6 +162,19 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
...
+        + "After update there will be %d containers. "
+        + "Please make sure there are sufficient resources to update this job. "
+        + "Continue update? [y/N]: ", oldContainerCount, newContainerCount));
+    Scanner scanner = new Scanner(System.in);
```
kramasamy: Confirmation with the user should not be for all cases. If the user knows the behavior - it is annoying everytime to say yes/no - furthermore think about the fact that heron update is running in CI/CD - yes/no will not work very well.

kramasamy: This verification can be done at the cli level itself - rather than doing it in the core.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -198,9 +222,10 @@ void updateTopologyHandler(String topologyName, String newParallelism)
 
     LOG.fine("Sending Updating topology request: " + updateTopologyRequest);
     if (!schedulerClient.updateTopology(updateTopologyRequest)) {
-      throw new TopologyRuntimeManagementException(String.format(
+      throw new TopologyRuntimeManagementException(
           "Failed to update topology with Scheduler, updateTopologyRequest="
-              + updateTopologyRequest));
+              + updateTopologyRequest + "The topology can be in a strange stage. "
+                  + "Please check carefully or redeploy the topology !!");
```
kramasamy: Do you need this another warning?

kramasamy:
@huijunw @nwangtw - not sure what is the intention of this PR. Ideally instead of doing all the warning etc - heron update itself should take care of itself. If heron update has an issue in terms of getting resources etc, it should revert itself - making the operation as though it did not change anything.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
kramasamy: furthermore, there is no way for the user to verify whether the resources are available. Resource checking should be automatic.

huijunw: There is a toggle switch for this prompt. The default value is off, no change to the existing behavior.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: will remove `aurora`

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
huijunw: The user usually assumes the topology keeps same as before the command is issued, however, the topology may be in unknown state. This warning is explicitly to tell user the check the topology state.

kramasamy:
@huijunw - talked with @nwangtw about a cleaner approach. we can discuss it offline.

PR #2736, opened 2018-02-16
include eco into storm compatibility jar

kramasamy:


kramasamy:
@joshfischer1108 - here is the fix.

joshfischer1108:
Looks fine to me.

PR #2740, opened 2018-02-20
[Streamlet Scala API] Add Scala StreamletImpl Support

erenavsarogullari:

Adding `Scala StreamletImpl` by providing abstraction for Scala API specific logic
Adding Scala Streamlet `map` function support
Adding `ScalaToJavaConverter` in order to transform following logics:
 - Scala `Supplier` to `SerializableSupplier`
 - Scala `Function` to `SerializableFunction`
 - Scala `Sink` to Java `Sink`
Adding UT Coverages for both `Scala StreamletImpl` and `ScalaToJavaConverter`

### DIFF
--- a/heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
+++ b/heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
@@ -1,0 +1,45 @@
+//  Copyright 2018 Twitter. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.streamlet.scala.converter
+
+import com.twitter.heron.streamlet.Context
+import com.twitter.heron.streamlet.scala.Sink
+
+/**
+  * This class transforms passed User defined Scala Functions, Sources, Sinks
+  * to related Java versions
+  */
+object ScalaToJavaConverter {
+
+  def toSerializableSupplier[T](f: () => T) =
+    new com.twitter.heron.streamlet.SerializableSupplier[T] {
+      override def get(): T = f()
+    }
+
+  def toSerializableFunction[R, T](f: R => _ <: T) =
+    new com.twitter.heron.streamlet.SerializableFunction[R, T] {
+      override def apply(r: R): T = f(r)
+    }
+
+  def toJavaSink[T](sink: Sink[T]): com.twitter.heron.streamlet.Sink[T] = {
+    new com.twitter.heron.streamlet.Sink[T] {
+      override def setup(context: Context): Unit = sink.setup(context)
+
+      override def put(tuple: T): Unit = sink.put(tuple)
+
+      override def cleanup(): Unit = sink.cleanup()
+    }
+  }
+
+}
--- a/heron/api/src/scala/com/twitter/heron/streamlet/scala/impl/StreamletImpl.scala
+++ b/heron/api/src/scala/com/twitter/heron/streamlet/scala/impl/StreamletImpl.scala
@@ -11,33 +11,52 @@
 //  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
-package com.twitter.heron.streamlet.scala
+package com.twitter.heron.streamlet.scala.impl
 
 import com.twitter.heron.streamlet.{KeyValue, KeyedWindow}
+import com.twitter.heron.streamlet.{
+  JoinType,
+  SerializableTransformer,
+  WindowConfig
+}
+import com.twitter.heron.streamlet.scala.{Sink, Streamlet}
+import com.twitter.heron.streamlet.scala.converter.ScalaToJavaConverter._
 
-// TODO: This Java Streamlet API references will be changed with Scala versions when they are ready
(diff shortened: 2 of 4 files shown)


erenavsarogullari:
cc @kramasamy @skanjila @sijie @nwangtw

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
```diff
@@ -0,0 +1,45 @@
...
+package com.twitter.heron.streamlet.scala.converter
+
+import com.twitter.heron.streamlet.Context
+import com.twitter.heron.streamlet.scala.Sink
+
+/**
+  * This class transforms passed User defined Scala Functions, Sources, Sinks
+  * to related Java versions
+  */
+object ScalaToJavaConverter {
```
skanjila: I think we need to make this object be general purpose for all of the scala traits to use, for the Builder stuff lets get this code review moving before I work more on that because I will add more to this function for the Builder stuff

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
```diff
@@ -0,0 +1,45 @@
...
+    new com.twitter.heron.streamlet.SerializableSupplier[T] {
+      override def get(): T = f()
+    }
+
+  def toSerializableFunction[R, T](f: R => _ <: T) =
+    new com.twitter.heron.streamlet.SerializableFunction[R, T] {
+      override def apply(r: R): T = f(r)
+    }
+
+  def toJavaSink[T](sink: Sink[T]): com.twitter.heron.streamlet.Sink[T] = {
```
skanjila: It seems we are overriding the java Sink and plugging in our sink, however what happens if the sink is sufficiently complicated inside those methods, do we need to account for that

heron/api/src/scala/com/twitter/heron/streamlet/scala/impl/StreamletImpl.scala
```diff
@@ -0,0 +1,291 @@
...
+  WindowConfig
+}
+import com.twitter.heron.streamlet.scala.{Sink, Streamlet}
+import com.twitter.heron.streamlet.scala.converter.ScalaToJavaConverter._
+
+object StreamletImpl {
+
+  def toScalaStreamlet[R](
+      javaStreamlet: com.twitter.heron.streamlet.Streamlet[R]): Streamlet[R] =
+    new StreamletImpl[R](javaStreamlet)
```
skanjila: what does this imply when we pass the javaStreamlet as an arg and have it listed with a colon, the notation is a bit confusing here, can you add some comments

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
```diff
@@ -0,0 +1,45 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+package com.twitter.heron.streamlet.scala.converter
+
+import com.twitter.heron.streamlet.Context
+import com.twitter.heron.streamlet.scala.Sink
+
+/**
+  * This class transforms passed User defined Scala Functions, Sources, Sinks
```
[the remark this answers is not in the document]
erenavsarogullari: `Builder, Streamlet, Source, Sink` Traits are the main interfaces for Scala API and `ScalaToJavaConverter` is common container for these traits.

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
erenavsarogullari: `Builder` functions accept both `SerializableSupplier` and `Source`. You can use existing `toSerializableSupplier` for `SerializableSupplier`. For `Source`, `toJavaSource` function needs to be created as same as with existing `toJavaSink` function.

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
erenavsarogullari: `Sink` exposes the following functions(`setup, put and cleanup`) as _public_ and other logics which are covered internally are managed and exposed to low level Topology API by these functions. (e.g: `com.twitter.heron.streamlet.impl.sinks.ComplexSink` just calls its `setup` and `put` functions. This just provides wrapper.

heron/api/src/scala/com/twitter/heron/streamlet/scala/impl/StreamletImpl.scala
```diff
@@ -0,0 +1,291 @@
...
+package com.twitter.heron.streamlet.scala.impl
+
+import com.twitter.heron.streamlet.{KeyValue, KeyedWindow}
+import com.twitter.heron.streamlet.{
+  JoinType,
+  SerializableTransformer,
+  WindowConfig
+}
+import com.twitter.heron.streamlet.scala.{Sink, Streamlet}
+import com.twitter.heron.streamlet.scala.converter.ScalaToJavaConverter._
```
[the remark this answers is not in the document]
erenavsarogullari: Yes, this is `Scala Companion object` and object parts covers `static functions`. For example, `createSupplierStreamlet` function is static in `Java StreamletImpl` and used by its UTs.

heron/api/src/scala/com/twitter/heron/streamlet/scala/impl/StreamletImpl.scala
erenavsarogullari: After related `user defined function` is applied(e.g: `map`), it will return `Java Streamlet` object type and this needs to be transformed to `Scala Streamlet `object type again before exposing the user so in order to achieve this, `toScalaStreamlet` function accept `Java Streamlet` Interface and transforms it to `Scala Streamlet` Trait type.

heron/api/src/scala/com/twitter/heron/streamlet/scala/converter/ScalaToJavaConverter.scala
skanjila: Yes I was saying I would like this checked in so that I can add those functions that you are describing

PR #2741, opened 2018-02-21
[#2735 follow up] move out of core into aurora scheduler

huijunw:
This PR follows #2735, moving the code out of core into aurora scheduler

To rollback aurora will be in another PR

### DIFF
--- a/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
+++ b/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
@@ -185,27 +185,24 @@ public class AuroraScheduler implements IScheduler, IScalable {
     return true;
   }
 
-
-  boolean confirmWithUser(int newContainerCount) {
+  private static final String CONFIRMED_YES = "y";
+  boolean hasConfirmedWithUser(int newContainerCount) {
     LOG.info(String.format("After update there will be %d more containers. "
         + "Please make sure there are sufficient resources to update this job. "
         + "Continue update? [y/N]: ", newContainerCount));
     Scanner scanner = new Scanner(System.in);
     String userInput = scanner.nextLine();
-    if ("y".equalsIgnoreCase(userInput)) {
-      return true;
-    }
-    return false;
+    return CONFIRMED_YES.equalsIgnoreCase(userInput);
   }
 
   @Override
   public Set<PackingPlan.ContainerPlan> addContainers(
       Set<PackingPlan.ContainerPlan> containersToAdd) {
-    if ("prompt".equalsIgnoreCase(Context.updatePrompt(config))) {
-      if (!confirmWithUser(containersToAdd.size())) {
-        LOG.warning("Scheduler updated topology canceled.");
-        return null;
-      }
+    Set<PackingPlan.ContainerPlan> remapping = new HashSet<>();
+    if ("prompt".equalsIgnoreCase(Context.updatePrompt(config))
+        && !hasConfirmedWithUser(containersToAdd.size())) {
+      LOG.warning("Scheduler updated topology canceled.");
+      return remapping;
     }
 
     // Do the actual containers adding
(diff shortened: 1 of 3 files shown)


heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -184,9 +185,29 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
...
+  boolean confirmWithUser(int newContainerCount) {
+    LOG.info(String.format("After update there will be %d more containers. "
+        + "Please make sure there are sufficient resources to update this job. "
+        + "Continue update? [y/N]: ", newContainerCount));
+    Scanner scanner = new Scanner(System.in);
+    String userInput = scanner.nextLine();
+    if ("y".equalsIgnoreCase(userInput)) {
+      return true;
+    }
+    return false;
```
fred521: private static final String CONFIRMED_YES = "y";
,
return CONFIRMED_YES.equalsIgnoreCase(userInput);

fred521:
```
private static final String CONFIRMED_YES = "y";
,
return CONFIRMED_YES.equalsIgnoreCase(userInput);
```

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -184,9 +185,29 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
     return true;
   }
 
+
+  boolean confirmWithUser(int newContainerCount) {
```
fred521: I'd prefer to use the name pattern like "hasConfirmedWithUser(int newContainerCount )", but it is up to you

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -184,9 +185,29 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
...
+      return true;
+    }
+    return false;
+  }
+
   @Override
   public Set<PackingPlan.ContainerPlan> addContainers(
       Set<PackingPlan.ContainerPlan> containersToAdd) {
+    if ("prompt".equalsIgnoreCase(Context.updatePrompt(config))) {
+      if (!confirmWithUser(containersToAdd.size())) {
```
fred521: try to avoid if else or if if, maybe single statement is a little bit clear. But I am not sure the entire logic, I could be wrong.

```
Boolean hasNotConfirmedWithUser = "prompt".equalsIgnoreCase(Context.updatePrompt(config)) && !hasConfirmedWithUser(containersToAdd.size());
  if(hasNotConfirmedWithUser) LOG.warning("Scheduler updated topology canceled.");
```

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -184,9 +185,29 @@ public boolean onUpdate(Scheduler.UpdateTopologyRequest request) {
...
+        LOG.warning("Scheduler updated topology canceled.");
+        return null;
```
fred521: return null sometimes will cause an additional null check logic when using this method.
In the effective java design pattern, it should always avoid returning null.
Or you can return an empty object or initial the object at the very beginning.
ag, I don't know the whole logic, just some notes.

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: updated

heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraScheduler.java
huijunw: updated

Issue #2128, opened 2017-07-30
Provide a brew install script


aahmed-se:

kramasamy:
+1. We need this for every release.

aahmed-se:
Current formulae published , will figure out an automation strategy later,
https://github.com/streamlio/homebrew-formulae

PR #2742, opened 2018-02-22
Fix installation docs

lucperkins:
Heron installation has changed quite a bit recently. There is now just one installation script and one Homebrew recipe for macOS. This PR brings the docs in line with the current state of affairs.


spmason:
I just tried following these docs.  Heron doesn't seem to be available on homebrew by default - do you have to tap https://github.com/streamlio/homebrew-formulae first, as hinted to in https://github.com/twitter/heron/issues/2128#issuecomment-322059440? Or should it be available in the main homebrew repository at some point?

PR #2744, opened 2018-02-24
add a switch for metricscache

huijunw:
Added a switch to turn on/off metricscache. Config options:
```
heron.topology.metricscachemgr.mode: cluster
heron.topology.metricscachemgr.mode: disabled (default)
```

In some scenarios, the customers do not turn on healthmgr/dhalion, so the metricscache does not have consumer. This PR add a switch so that user can turn it off.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -239,7 +239,7 @@ class HeronExecutor(object):
     self.checkpoint_manager_port = parsed_args.checkpoint_manager_port
     self.stateful_config_file = parsed_args.stateful_config_file
     self.metricscache_manager_mode = parsed_args.metricscache_manager_mode \
-        if parsed_args.jvm_remote_debugger_ports else "disabled"
+        if parsed_args.metricscache_manager_mode else "disabled"
     self.health_manager_mode = parsed_args.health_manager_mode
     self.health_manager_classpath = '%s:%s'\
         % (self.scheduler_classpath, parsed_args.health_manager_classpath)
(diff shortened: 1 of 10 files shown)


Yaliang:
Just a general question: is there a test to test configuration settings?

heron/executor/src/python/heron_executor.py
```diff
@@ -238,6 +238,8 @@ def init_parsed_args(self, args):
     self.checkpoint_manager_classpath = parsed_args.checkpoint_manager_classpath
     self.checkpoint_manager_port = parsed_args.checkpoint_manager_port
     self.stateful_config_file = parsed_args.stateful_config_file
+    self.metricscache_manager_mode = parsed_args.metricscache_manager_mode \
+        if parsed_args.jvm_remote_debugger_ports else "disabled"
```
nwangtw: why "if parsed_args.jvm_remote_debugger_ports"? typo?

heron/spi/src/java/com/twitter/heron/spi/common/Context.java
```diff
@@ -339,4 +339,9 @@ public static final String statefulStorageCustomClassPath(Config cfg) {
     Object o = statefulStorageConfig.get(Key.STATEFUL_STORAGE_CUSTOM_CLASSPATH.value());
     return o == null ? "" : (String) o;
   }
+
+  public static String metricscacheMgrMode(Config cfg) {
+    return cfg.getStringValue(Key.METRICSCACHEMGR_MODE);
```
nwangtw: do we need a default value?

huijunw: there is a default value in Key.java

heron/executor/src/python/heron_executor.py
huijunw: typo, updated

PR #2747, opened 2018-02-25
Add a new tmaster endpoint for runtime config

nwangtw:
Add a new tmaster endpoint for topology/component runtime config:

/runtime_config

This is part of runtime config support. The design doc can be found here: https://docs.google.com/document/d/1VWVmnGouZMyQriKWwXn25fDeiHKd7nYCIcbAyb2jcak/edit#heading=h.nnj2pmfg1qm7


maosongfu:
It is a `POST` request changing topology state. Should we use a more expressive endpoint name, like `update_runtime_config`?

nwangtw:
I thought about both and chose this one. I feel it depends on the meaning of "config". If it is a verb here, then runtime_config is good enough. No objection to the "update_runtime_config" endpoint though. Let's get more idea from other people. I am happy to change it after we make a decision.

kramasamy:
probably /runtime_config/update - will be more apt. It gives more room to expand as well.

maosongfu:
`/runtime_config/update` sgtm.

nwangtw:
sounds good. let me do it.

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -54,25 +70,7 @@ sp_int32 TController::Start() { return http_server_->Start(); }
...
-    delete request;
-    return;
-  }
-  if (id != tmaster_->GetTopologyId()) {
-    LOG(ERROR) << "Topology id does not match";
-    http_server_->SendErrorReply(request, 400);
-    delete request;
-    return;
-  }
+  // Validation
```
kramasamy: @nwangtw - why is the validation code removed?

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -211,5 +185,125 @@ void TController::HandleCleanStatefulCheckpointRequestDone(IncomingHTTPRequest*
...
+ * Validate topology.
+ * - topology id matches
+ * - topology is initialized
+ * return true if topology is validated, false otherwise with error details stored in result object
+ */
+bool TController::ValidateTopology(const IncomingHTTPRequest* request, ValidationResult& result) {
+  const sp_string& id = request->GetValue("topologyid");
+  if (id == "") {
+    LOG(ERROR) << "Topologyid not specified in the request";
+    result.SetResult(400, "Missing topologyid argument in the request");
```
kramasamy: Spelling mistake - it should be "topology id"

heron/common/src/cpp/config/topology-config-helper.cpp
```diff
@@ -399,17 +399,105 @@ bool TopologyConfigHelper::DropTuplesUponBackpressure(const proto::api::Topology
...
+
+// Update component level config
+void TopologyConfigHelper::SetComponentConfig(proto::api::Topology* _topology,
+                                              const sp_string& _component_name,
+                                              const std::map<sp_string, sp_string>& _update) {
+  // We are assuming component names are unique and updating config for all instances
+  // with the specific component name.
+  for (sp_int32 i = 0; i < _topology->spouts_size(); ++i) {
+    proto::api::Component* comp = _topology->mutable_spouts(i)->mutable_comp();
+    if (comp->name() == _component_name && comp->has_config()) {
```
srkukarni: why the check comp->has_config?

heron/common/src/cpp/config/topology-config-helper.cpp
```diff
@@ -399,17 +399,105 @@ bool TopologyConfigHelper::DropTuplesUponBackpressure(const proto::api::Topology
...
+      proto::api::Config* config = comp->mutable_config();
+      UpdateConfigFromKVMap(config, _update);
+    }
+  }
+
+  for (sp_int32 i = 0; i < _topology->bolts_size(); ++i) {
+    proto::api::Component* comp = _topology->mutable_bolts(i)->mutable_comp();
+    if (comp->name() == _component_name && comp->has_config()) {
```
srkukarni: same as above

heron/common/src/cpp/config/topology-config-helper.cpp
```diff
@@ -399,17 +399,105 @@ bool TopologyConfigHelper::DropTuplesUponBackpressure(const proto::api::Topology
...
+    std::map<sp_string, sp_string>::const_iterator iter = _update.find(key);
+    if (iter != _update.end()) {
+      _config->mutable_kvs(i)->set_value(iter->second);
+    }
+  }
+}
+
 bool TopologyConfigHelper::GetBooleanConfigValue(const proto::api::Topology& _topology,
-                                                 const std::string& _config_name,
+                                                 const sp_string& _config_name,
```
srkukarni: all new code should use std::string. sp_string is legacy

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -211,5 +192,125 @@ void TController::HandleCleanStatefulCheckpointRequestDone(IncomingHTTPRequest*
   }
   delete request;
 }
+
+void TController::HandleRuntimeConfigRequest(IncomingHTTPRequest* request) {
```
srkukarni: HandleUpdateRuntimeConfigRequest

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -211,5 +192,125 @@ void TController::HandleCleanStatefulCheckpointRequestDone(IncomingHTTPRequest*
...
+    http_server_->SendErrorReply(request, 400, "Failed to validate runtime configs");
+    delete request;
+    return;
+  }
+
+  auto cb = [request, this](proto::system::StatusCode status) {
+    this->HandleRuntimeConfigRequestDone(request, status);
+  };
+
+  if (!tmaster_->RuntimeConfigTopology(config, std::move(cb))) {
```
srkukarni: tmaster_->UpdateRuntimeConfigTopology

heron/common/src/cpp/config/topology-config-helper.cpp
nwangtw: so that in case user has a typo, topology can reject the request.

heron/common/src/cpp/config/topology-config-helper.cpp
nwangtw: got it. will do.

heron/common/src/cpp/config/topology-config-helper.cpp
srkukarni: If I read the code correctly, if a particular component doest not have a comp specific configuration initially, I wont be able to add something specific there. Why this limitation?

nwangtw: Yeah. That is correct. This check is just for protection. I am open for other thoughts.

Here are a few cases,
- the topology has a config topology.config.1, and user can use command line tool to update the value. in case there is a typo like topolog.config.1, we can reject the config change.
- This could also be useful when two configs work together and need to be updated at the same time. If one of them has a typo, the update command should be cancelled.
- Or if the component name is wrong, the issue can be detected instead of waiting for the topology to react.

Like you said, the limit is that the config has to exist. I feel it is safer this way because if you want to keep the runtime config in future, you have to update the config manually in code. Requiring the config to exist could avoid some mistakes.

The check doesn't cover 100% cases and the limit is kinda artificial. If you guys prefer not having the check or any other idea, I am totally happy to consider.

heron/tmaster/src/cpp/manager/tcontroller.cpp
nwangtw: I am going to use UpdateRuntimeConfig for this one.

heron/common/src/cpp/config/topology-config-helper.cpp
nwangtw: On the other hand, we can skip the check and leave all runtime configs pass through. In addition to user mistakes, there might be some unused configs and it could be harder to track them.

nwangtw:
Just realize I forgot to append the ":runtime" postfix (described in doc: https://docs.google.com/document/d/1VWVmnGouZMyQriKWwXn25fDeiHKd7nYCIcbAyb2jcak/edit#). Adding it now.

nwangtw:
Changed the runtime config validation to verify component name only, not config name.

PR #2749, opened 2018-02-26
fixing bugs in the API server when downloading files

jerrypeng:


jerrypeng:
@aahmed-se

heron/tools/apiserver/src/java/com/twitter/heron/apiserver/resources/FileResource.java
```diff
@@ -164,11 +167,11 @@ private Config createConfig() {
 
   private String getHostNameOrIP() {
     // Override hostname if provided in flags
-    if (StringUtil.isNotBlank(getDownloadHostName())) {
```
cckellogg: Should we add a static util method isNotEmpty()?

jerrypeng: sure

jerrypeng: done

fred521: Duplicate code found here, better to have a method to reduce the code

getResponseByFile(String filePath)
```
  public Response downloadFile(final @PathParam("file") String file) {
    Config config = createConfig();
    String uploadDir = config.getStringValue(FILE_SYSTEM_DIRECTORY);
    String filePath = uploadDir + "/" + file;
    return getResponseByFile(filePath);
  }


  public Response downloadHeronCore() {
    String corePath = getHeronCorePackagePath();
    return getResponseByFile(corePath);
  }

  private Response getResponseByFile(String filePath){

    File file = new File(filePath);
    if (!file.exists()) {
      //Debug log if it is necessary
      return Response.status(Response.Status.NOT_FOUND).build();
    }
    String mimeType = new MimetypesFileTypeMap().getContentType(file);
    Response.ResponseBuilder rb = Response.ok(file, mimeType);
    rb.header("content-disposition", "attachment; filename = "
            + file.getName());
    return rb.build();

  }
```

jerrypeng: @fred521 thanks for point that out I have refactored the code

jerrypeng:
@fred521 Yup I experimented with using StringUtils.isNotBlank() but it will return true if the input is null for some reason even though the javadoc of it states it should return false.  Didn't want to spend too much time investigating why for such a trivial thing

PR #2752, opened 2018-02-27
Pass role and environment to java topoology components.

cckellogg:


kramasamy:
@cckellogg - wondering if the values are being read in the instance.

PR #2753, opened 2018-02-28
fixing bug in wakeable loop causing IndexOutOfBoundException

jerrypeng:
When there is a task in taskOnWakeup that clears the taskOnWakeup list i can cause the following exception:
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
       at java.util.ArrayList.rangeCheck(ArrayList.java:653)
       at java.util.ArrayList.get(ArrayList.java:429)
       at com.twitter.heron.common.basics.WakeableLooper.executeTasksOnWakeup(WakeableLooper.java:160)
       at com.twitter.heron.common.basics.WakeableLooper.runOnce(WakeableLooper.java:89)
       at com.twitter.heron.common.basics.WakeableLooper.loop(WakeableLooper.java:79)
       at com.twitter.heron.instance.Slave.run(Slave.java:180)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
       at java.lang.Thread.run(Thread.java:748)

This happens in stateful process.  When there is an error and the instance gets a message to restore to  state it will call the following:

https://github.com/twitter/heron/blob/master/heron/instance/src/java/com/twitter/heron/instance/Slave.java#L113

which initiates the restore and is also a task in taskOnWakeup.  The restore process also involves clearing the taskOnWakeup list.


heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -92,8 +92,16 @@ private void runOnce() {
...
   private void onExit() {
-    for (Runnable r : exitTasks) {
-      r.run();
+    // Using CopyOnWriteArrayList to prevent from tasks in exitTasks from clearing
+    // the exitTasks list and potentially causing IndexOutOfBoundsException.
+    // The method iterator() in CopyOnWriteArrayList return an iterator object that holds
+    // different copy of the elements, hence the term snapshot iterator. The snapshot iterator
+    // doesn’t allow modifying the list while traversing, and it will not throw
+    // ConcurrentModificationException if the list is being modified by other thread during
+    // the traversal, and the read and write operations work on different copies of elements.
```
nwangtw: two questions:
1. is modifying the list at the time expected? If not, maybe a more proper fix is to solve issue in the appending code?
2. will some exit tasks and tasks on wake up being missed with copyOnWrite list?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -40,11 +40,11 @@
  */
 public abstract class WakeableLooper {
   // The tasks could only be added but not removed
-  private final List<Runnable> tasksOnWakeup;
+  private final CopyOnWriteArrayList<Runnable> tasksOnWakeup;
```
maosongfu: WakeableLooper is designed not to be thread-safe; concurrent scenarios should be handled in outer logic.

If the purpose of this pull request is to support thread-safety, using this concurrent structure is not safe enough, if thinking this from an application point of view, since the actual `Runnable` a removed reference pointing to might not run normally, for instance, running with a dirty state.

If the purpose is to avoid task-removal during execution, I would suggest not to use a `Concurrent` data structure to avoid confusion.

BTW, IIRC, previously we could only add a task to `tasksOnWakeup` when executing tasks from it, but we could not remove a task.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
maosongfu: IIRC, previously we could only add a task to such kind of `tasks-list` when executing tasks from it, but we could not remove a task.

On the other hand, WakeableLooper is designed not to be thread-safe; so it should be executed with only one thread.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
maosongfu: Also, if the purpose is to avoid task-removal during execution, why not find out such code snippets and correct them?

jerrypeng:
@maosongfu @nwangtw This exception doesn't arise from any concurrency situations.  Its just one task in list actually clears the list.  So in the previous for loop when we try to get the next task to run e.g. task i + 1 there will be an index out of bounds exception thrown since their is nothing in the list anymore.  When iterating a CopyOnWriteArrayList, a separate copy is made for the read and thus the clearing of the list will not effect us

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
jerrypeng: @maosongfu The code snippet of the task that clears the tasksOnWakeup list:

https://github.com/twitter/heron/blob/master/heron/instance/src/java/com/twitter/heron/instance/Slave.java#L113


Has to do with restoring state.

nwangtw:
Thanks for reply! It makes more sense now.

However I am afraid this fix is still not safe. If my understanding is correct, the clear call should clear all tuples immediately and cancel the pending actions. With CopyOnWrite list, the current content in the lists are still being executed afterwards.

maosongfu:
Thanks for the explanation. At the same time, I am also looking at the code and thinking about whether we would have better options for a fix.

jerrypeng:
@maosongfu @nwangtw I have made the fix simpler and with fewer changes.  Can you please take another look?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -92,8 +92,13 @@ private void runOnce() {
   }
 
   private void onExit() {
-    for (Runnable r : exitTasks) {
-      r.run();
+    int s = exitTasks.size();
```
nwangtw: I would suggest checking == 0, like

for (int i = 0; i < s; i ++) {
      exitTasks.get(i).run();
      // in case previous task cleared the list
      if (exitTasks.size() == 0) {
        break;
      }
}		    

Another qq, no tasks would be added when this function is called? I think onExit() might be safe but executeTasksOnWakeup() is a concern.

jerrypeng:
@nwangtw I have pushed another commit that adds a locking mechanism for adding new tasks and timers to the wakeable loop. During the state restore, we can lock the looper so that no new tasks can be added and after the state restore is completed we can unlock.  What do you think?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
jerrypeng: Though is checking tasksOnWakeup.size() <= i more defensive programming?  With this check an index out of bounds exception cannot happen while for some reason or another not all the tasks are cleared you can get into a situation where i > list size causing index out of bounds exception

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -157,6 +182,10 @@ private void executeTasksOnWakeup() {
     // We pre-get the size to avoid execute the tasks added during execution
     int s = tasksOnWakeup.size();
     for (int i = 0; i < s; i++) {
+      // in case previous task cleared the list
```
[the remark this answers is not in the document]
jerrypeng: This won't happen during the state restore since I just add the locks which will prevent new tasks from being add during the state restore.  And its only during the state restore that we clear the tasks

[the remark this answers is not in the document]
nwangtw: I think we need to answer two questions beforehand:
- where the tasks are added?
- when can tasks be added?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
nwangtw: it is more defensive, but confusing mathematically.

In fact I dont like either of them because it is not ideal to check the size an array when it can change.

Another option is to change the interface:
task.run() should return a boolean value indicate the loop should be continued or not. When it returns false, the loop should clear the list and return. Something like:

for (int i = 0; i < s; i ++) {
  if (!exitTasks.get(i).run()) {
    exitTasks.clear();
    break;
  }
}

This way we can avoid the clear() call in tasks.
How do you think?

nwangtw: It seems that it is possible and allowed to add new tasks when the array is being executed. :(

but i think the new tasks added by the tasks before restore should be cleared too. So the clear() call is still correct.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -44,22 +44,34 @@
...
-  private final ArrayList<Runnable> exitTasks;
+  private final List<Runnable> exitTasks;
 
   // For selector since there is bug in selector.select(timeout): we could not
   // use a timeout > 10 * Integer.MAX_VALUE
   // So here we set Integer.MAX_VALUE as the infinite future
   // We will also multiple 1000*1000 to convert mill-seconds to nano-seconds
   private static final Duration INFINITE_FUTURE = Duration.ofMillis(Integer.MAX_VALUE);
   private volatile boolean exitLoop;
+  private boolean lockAddNewTasks = false;
```
nwangtw: If we want to use this lock, it needs to be an atomicBoolean, or needs to be volatile.

my concern is that simple boolean might have code cache and can be inconsistent between threads.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -102,31 +119,39 @@ private void onExit() {
   public abstract void wakeUp();
 
   public void addTasksOnWakeup(Runnable task) {
-    tasksOnWakeup.add(task);
-    // We need to wake up the looper itself when we add a new task, otherwise, it is possible
-    // this task will never be executed due to the looper will never be wake up.
-    wakeUp();
+    if (!lockAddNewTasks) {
```
nwangtw: QQ: Can tasks be added by other thread?

If they can, then there might be a racing condition.
the existing tasks are being cleared, and a new package is received and some tasks might be added. A lock flag is not enough and we have to sync up the two thread.

jerrypeng:
@nwangtw @maosongfu I looked through code more thoroughly to find who is add adding or removing tasks from wakeable loop for the slave thread. Only methods that handle state operation i.e. com.twitter.heron.instance.Slave.handleControlMessage will potentially add tasks as a task exisiting in the queue. There is no concurrency problems since only tasks are add upon initializing the slave thread or by tasks run by the slave looper so everything is in sync.  Thus, we don't really need locks per se but just make sure we handle the situation where tasks add other tasks and tasks clearing the list correctly

jerrypeng:
@nwangtw I am not sure about changing the Runnable to a custom interface to support the return of a boolean to determine whether we should clear to list.  Seems like a heavy change since a lot things use the wakeable not just the slave thread

jerrypeng:
@nwangtw I add a boolean that gets set when tasks are cleared to break out of the loop.  What do think about that mechanism?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -92,9 +100,14 @@ private void runOnce() {
   }
 
   private void onExit() {
-    for (Runnable r : exitTasks) {
-      r.run();
+    int s = exitTasks.size();
+    for (int i = 0; i < s; i++) {
+      if (clearExitTasks) {
```
nwangtw: is it guaranteed that the clear() function is called in the same thread? the concern is that if the clear() function is called in a different thread, it might be called after the if() check and you can still get the wrong index error. Since this object might be used by new code in future so there may not be a guarantee even if there s no problem for now. I think it might be better to have a comment to be clear that this flag is not thread safe and the clear functions are required to be called ONLY in the same thread.

Or I am wondering would it be safer to make onExit, executeTasksOnWakeup, clearTasksOnWakeup and  clearExitTasks synchronized?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -44,36 +44,44 @@
...
-  private final ArrayList<Runnable> exitTasks;
+  private final List<Runnable> exitTasks;
 
   // For selector since there is bug in selector.select(timeout): we could not
   // use a timeout > 10 * Integer.MAX_VALUE
   // So here we set Integer.MAX_VALUE as the infinite future
   // We will also multiple 1000*1000 to convert mill-seconds to nano-seconds
   private static final Duration INFINITE_FUTURE = Duration.ofMillis(Integer.MAX_VALUE);
   private volatile boolean exitLoop;
+  private boolean clearTasksOnWakeup = false;
```
nwangtw: maybe better for the two flags to be initialized in constructor?

Another suggest is to rename to clearTasksOnWakeupFlag to avoid variable and function sharing the same name. Maybe not a big deal. Your call.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
jerrypeng: I have checked all code paths.  Ya I can confirm clear() is always called from the same thread.  I don't know how we can make nExit, executeTasksOnWakeup, clearTasksOnWakeup and clearExitTasks synchronized.  This would give rise to deadlock situations since tasks run in executeTasksOnWakeup() can in turn call clear().

jerrypeng:
@nwangtw thanks for reviewing the PR again.  I have addressed your comments.  Do you have any further concerns?

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
nwangtw: I was thinking of simply adding synchronized keyword to these functions. I believe Java synchronized blocks in Java are reentrant so it is not a deadlock.

jerrypeng: I see what you are saying, but should we add them? It gives a developer the sense that these operations are thread safe now.  Should we even allow that in the architecture.  I think that is a bigger discussion. I can't think of a use case for that now.  It also might not be wise to allow developers to call these from potentially other threads since heron's instance architecture is clearly defined to be 2 thread based, gateway and slave.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
jerrypeng: I have renamed

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -44,36 +44,62 @@
...
-  private final ArrayList<Runnable> exitTasks;
+  private final List<Runnable> exitTasks;
 
   // For selector since there is bug in selector.select(timeout): we could not
   // use a timeout > 10 * Integer.MAX_VALUE
   // So here we set Integer.MAX_VALUE as the infinite future
   // We will also multiple 1000*1000 to convert mill-seconds to nano-seconds
   private static final Duration INFINITE_FUTURE = Duration.ofMillis(Integer.MAX_VALUE);
   private volatile boolean exitLoop;
+  private boolean terminateAllTasksOnWakeup;
```
maosongfu: Can we add comments for `terminateAllTasksOnWakeup ` and `terminateAllExitTasks `? Though I can understand them after I go through the code, it is still better to have more comments.

heron/common/src/java/com/twitter/heron/common/basics/WakeableLooper.java
```diff
@@ -44,36 +44,62 @@
...
+    tasksOnWakeup = new ArrayList<>();
     timers = new PriorityQueue<TimerTask>();
     exitTasks = new ArrayList<>();
+    terminateAllTasksOnWakeup = false;
+    terminateAllExitTasks = false;
   }
 
+  /**
+   * Clears all tasks.
+   * This is expected to be call synchronously.  This call is not thread safe!
```
maosongfu: Can you remove this comment (and similar comments below), and add comments at the head to declare:
1. Most methods except the `onExit()` are not thread-safe.
2. People should handle the concurrent scenarios in their business logic rather than in this class.

nwangtw:
Good comments are definitely helpful for my concerns. PR is acceptable to me after addressing maosong's comments.

jerrypeng:
@maosongfu @nwangtw I have added the comments requested.  Thanks for the review!

PR #2760, opened 2018-03-02
Add runtime config in heron update command

nwangtw:
- New parameter --user-config in CLI
- User config handling in scheduler


heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -177,9 +185,29 @@ void killTopologyHandler(String topologyName) throws TopologyRuntimeManagementEx
...
    */
   @VisibleForTesting
-  void updateTopologyHandler(String topologyName, String newParallelism)
+  void updateTopologyHandler(String topologyName, Config updateConfig)
       throws TopologyRuntimeManagementException, PackingException, UpdateDryRunResponse {
     assert !potentialStaleExecutionData;
+    String newParallelism = updateConfig.getStringValue(RUNTIME_MANAGER_COMPONENT_PARALLELISM_KEY);
+    String userRuntimeConfig = updateConfig.getStringValue(RUNTIME_MANAGER_RUNTIME_CONFIG_KEY);
+
+    if (newParallelism != null && !newParallelism.isEmpty()) {
```
kramasamy: @nwangtw - what happens if both the update component parallelism and also user run time config is provided simultaneously.

heron/spi/src/java/com/twitter/heron/spi/utils/TMasterUtils.java
```diff
@@ -147,8 +165,29 @@ public static void transitionTopologyState(String topologyName,
...
+  public static void sendRuntimeConfig(String topologyName,
+                                       TMasterCommand topologyStateControlCommand,
+                                       SchedulerStateManagerAdaptor statemgr,
+                                       String[] configs,
+                                       NetworkUtils.TunnelConfig tunnelConfig)
+      throws TMasterException {
+    final String runtimeConfigKey = "runtime-config";
+    final String runtimeConfigUpdateEndpoint = "runtime_config/update";
+
+    String[] arguments = new String[configs.length];
```
kramasamy: use vector<string> and no need to allocate memory.

heron/tools/cli/src/python/update.py
```diff
@@ -30,58 +32,118 @@ def create_parser(subparsers):
...
+      help='Runtime configurations for topology and/or components'
+      + 'colon-delimited: [component:]<name>:<value>')
 
   parser.set_defaults(subcommand='update')
   return parser
 
+def build_extra_args_dict(cl_args):
+  """ Build extra args map """
+  # Check parameters
+  component_parallelisms = cl_args['component_parallelism']
```
kramasamy: just use component_parallelism instead of component_parallelisms

heron/tools/cli/src/python/update.py
```diff
@@ -30,58 +32,118 @@ def create_parser(subparsers):
...
+  args.add_service_url(parser)
+  args.add_verbose(parser)
+
+  # Special parameters for update command
   def parallelism_type(value):
     pattern = re.compile(r"^[\w\.-]+:[\d]+$")
     if not pattern.match(value):
       raise argparse.ArgumentTypeError(
-          'Invalid syntax for component parallelism (<component_name>:<value>): %s' % value)
+          "Invalid syntax for component parallelism (<name:value>): %s" % value)
```
kramasamy: leave it as component_name - so that it is informative.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerRunner.java
kramasamy: This seems to be the case across the board.

PR #2776, opened 2018-03-13
Add rate limiter in output collector

nwangtw:
The rate limiter is only available in Java. Python/C++ support will be done later.

Guava RateLimiter is used, which provides a continuous rate limiting.

For performance concerns, I used unit test to run a few tests on my laptop. When the test spout  (super simple generating logic) generates 1k tuples, rate limiter slows down process slightly, 140+ ms to 160+ ms. But when the spout generates 100k tuples, there is no difference.

In this PR, I am adding rate limiter in Slave and passing it to SpoutInstance/BoltInstance to be applied on OutputCollectors. The considerations are:
- It needs to be applied to OutputCollectors because tuples are emitted to it one by one. In other objects, tuples are batched.
- Ratelimiter is created in Slave because it can be shared by both Spout and Bolt.
I am open to other thoughts about the two considerations above.

### DIFF
--- a/heron/api/src/java/com/twitter/heron/api/Config.java
+++ b/heron/api/src/java/com/twitter/heron/api/Config.java
@@ -96,7 +96,7 @@ public class Config extends HashMap<String, Object> {
    */
   public static final String TOPOLOGY_MESSAGE_TIMEOUT_SECS = "topology.message.timeout.secs";
   /**
-   * The per componentparallelism for a component in this topology.
+   * The per component parallelism for a component in this topology.
    * Note:- If you are changing this, please change the utils.h as well
    */
   public static final String TOPOLOGY_COMPONENT_PARALLELISM = "topology.component.parallelism";
@@ -290,6 +290,11 @@ public class Config extends HashMap<String, Object> {
   public static final String TOPOLOGY_DROPTUPLES_UPON_BACKPRESSURE =
       "topology.droptuples.upon.backpressure";
 
+  /**
+   * The per component output tuple per second in this topology.
+   */
+  public static final String TOPOLOGY_COMPONENT_OUTPUT_TPS = "topology.component.output.tps";
+
   private static final long serialVersionUID = 2550967708478837032L;
   // We maintain a list of all user exposed vars
   private static Set<String> apiVars = new HashSet<>();
@@ -327,6 +332,7 @@ public class Config extends HashMap<String, Object> {
     apiVars.add(TOPOLOGY_UPDATE_REACTIVATE_WAIT_SECS);
     apiVars.add(TOPOLOGY_REMOTE_DEBUGGING_ENABLE);
     apiVars.add(TOPOLOGY_DROPTUPLES_UPON_BACKPRESSURE);
+    apiVars.add(TOPOLOGY_COMPONENT_OUTPUT_TPS);
   }
 
   public Config() {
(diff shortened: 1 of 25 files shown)


srkukarni:
One meta comment. Why is rate limiting being done at the spout level and not stmgr level
1. libevent already has a leaky bucket abstraction layer that will allow you to do this pretty easily
2. this will then automatically apply to other languages.

nwangtw:
Good question.

Stream manager is more convenient implementation-wise than instance. We thought about the option, but it has its own issues. Instances push tuples to stream manager, so we cant control the input stream. And if we rate limit the output stream, the buffer size would increase and cause back pressure. Therefore, we need a way for stream manager to control the processing speed. (Or is there a way to avoid the issue?) On the other side, the rate limiter in instance slows down processing directly and there is no side effect.

ashvina:
Hi @nwangtw. If I understand correctly, rate limit is configured at the topology level. Which means TPS of individual components cannot be controlled, and the max rate to be configured will depend on the component producing maximum tuples. In one of the use cases we were working with, we needed to tune output of each components differently. 

What do you think about extending this configuration to components? 

I also agree with @srkukarni. If the output is throttled, input queues of the component will fill up and cause back-pressure. Back-pressure will be seen as soon as rate limit is applied and I don't think it matters where the rate limit is applied. Supporting all the languages would be better than creating a language base feature compatibility matrix.

kramasamy:
+1 @ashvina and @srkukarni - comments

srkukarni:
I want to move a little bit higher level and understand what goals we are trying to achieve here. I was under the impression that we want to rate limit the sources(aka the spouts). Thus each spout(or combined together) will have a certain rate that will be allowed into the system and the rest will be backpressure. Is my understanding correct? If this is correct, then applying this at stmgr is not only convenient(because stmgr knows which are spout connections and can enforce the rate limits on only those connections) but also language agnostic

nwangtw:
@ashvina rate limiting is at component level. Backpressure on the instance itself is ok, but backpressure has sideeffects. Topology will stop all spouts in that case if my understanding is correct. So I am hoping to avoid backpressure triggering/handling. Or we can also handling spout backpressure differently from bolt backpressure?

@srkukarni Yeah. The scope is correct. We are limiting the same data. The pipeline is like:

spout -> spout_instance -> gateway -> stmgr -> EXTERNAL.

currently the rate limiter is applied on the connection between spout -> spout_instance which is done via the emit() function and the tuples are pushed one by one. The limiter basically slows down the spout.

My understanding is that you are suggesting rate limiting the connection between stmgr and EXTERNAL for the specific spouts. It pauses spouts by putting them into backpressure state. I am trying to avoid the backpressure to avoid confusion and extra handling logic, but maybe this should be fine? I can take another look at this idea.

nwangtw:
I am looking into stmgr again. Thanks for the suggestion.

srkukarni:
@nwangtw My suggestion is to rate limit between gateway and stmgr. And this is different from backpressure. Backpressure is a state that the stmgr gets into and at that time it stops reading from spouts. On the other hand, rate limiting is simply reading at a certain rate from the connection of spout instance inside the stmgr. This should be something that you do during connection setup. This doesnt require any kind of handshaking, etc between stmgr/spout instance. And because stmgr uses libevent as the read interface, its literally just a matter of calling one function on the connection. Checkout http://www.wangafu.net/~nickm/libevent-book/Ref6a_advanced_bufferevents.html and in particular the  rate limiting section.

nwangtw:
@srkukarni got it! Thanks. I think my impression about data pushing between gateway and stmgr might be wrong. Let me check the code again.

maosongfu:
+1 to do the rate limit on `stmgr` side. Or even do the rate limit on `gateway` side is better than on `output_collector`(`slave`) side.

nwangtw:
stmgr based solution is implemented in: https://github.com/twitter/heron/pull/2783

nwangtw:
Dropped in favor of #2783

PR #2783, opened 2018-03-15
Add rate limiter in stmgr

nwangtw:
Add rate limit support in Stmgr (InstanceServer/BaseConnection) and heron config.

This is an alternative solution of https://github.com/twitter/heron/pull/2776


nwangtw:
Two more things to add:
- convert rate limit to per instance limit
- add a real unit test (the one added in the previous change was a setup only)

PR #2785, opened 2018-03-16
Add a new escaping '(61)' for '=' because the original one '&equals;'…

nwangtw:
… is not friendly to bash

### DIFF
--- a/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
+++ b/heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
@@ -442,21 +442,24 @@ public final class SchedulerUtils {
    * Encode the JVM options
    * <br> 1. Convert it into Base64 format
    * <br> 2. Add \" at the start and at the end
-   * <br> 3. replace "=" with "&amp;equals;"
+   * <br> 3. replace "=" with "(61)" and "&amp;equals;"
+   * Note that '=' is escaped in two different ways. '(61)' is the new escaping.
+   * '&amp;equals;' was the original way but it is not friendly to bash and is
+   * causing issues in some schedulers. It is still left there for backward compatibility
+   * reason
    *
    * @return encoded string
    */
   public static String encodeJavaOpts(String javaOpts) {
     String javaOptsBase64 = DatatypeConverter.printBase64Binary(
         javaOpts.getBytes(StandardCharsets.UTF_8));
-
-    return String.format("\"%s\"", javaOptsBase64.replace("=", "&equals;"));
+    return String.format("\"%s\"", javaOptsBase64.replace("=", "(61)").replace("=", "&equals;"));
   }
 
   /**
    * Decode the JVM options
    * <br> 1. strip \" at the start and at the end
-   * <br> 2. replace "&amp;equals;" with "="
+   * <br> 2. replace "(61)" and "&amp;equals;" with "="
    * <br> 3. Revert from Base64 format
    *
    * @return decoded string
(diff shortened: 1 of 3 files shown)


jerrypeng:
@nwangtw thanks for looking into this for me!  My only question is why are we replacing "=" with something else to begin with.  Its already base64 encoded and in quotes.

nwangtw:
My guess is that "=" is a valid character in base64, and it might mess up configs potentially because configurations are in config=value format.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -442,21 +442,24 @@ public static boolean setSchedulerLocation(
...
+   * reason
    *
    * @return encoded string
    */
   public static String encodeJavaOpts(String javaOpts) {
     String javaOptsBase64 = DatatypeConverter.printBase64Binary(
         javaOpts.getBytes(StandardCharsets.UTF_8));
-
-    return String.format("\"%s\"", javaOptsBase64.replace("=", "&equals;"));
+    return String.format("\"%s\"", javaOptsBase64.replace("=", "(61)").replace("=", "&equals;"));
```
jerrypeng: do we need .replace("=", "&equals;")) here?  All the "=" would already be replaced by the previous replace statement right?

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
```diff
@@ -442,21 +442,24 @@ public static boolean setSchedulerLocation(
    * Encode the JVM options
    * <br> 1. Convert it into Base64 format
    * <br> 2. Add \" at the start and at the end
-   * <br> 3. replace "=" with "&amp;equals;"
+   * <br> 3. replace "=" with "(61)" and "&amp;equals;"
+   * Note that '=' is escaped in two different ways. '(61)' is the new escaping.
+   * '&amp;equals;' was the original way but it is not friendly to bash and is
+   * causing issues in some schedulers. It is still left there for backward compatibility
```
jerrypeng: can we add a comment on why we need to replace "=" for future reference

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
nwangtw: right. Technically it is not really necessary.

I left it here to be consistent with the decoding path and easier (just a little) for people to understand. In future we can remove both of them at the same time.

jerrypeng: gotcha can we just add some comments at the places this is done for future reference.  Thanks!

jerrypeng: or maybe a TODO for the future

jerrypeng:
LGTM just some minor comments

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
nwangtw: Sure. added.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/utils/SchedulerUtils.java
nwangtw: Thinking about it again. Now I don't think leaving the encoding part here is very helpful. Plus it looks wrong and could cause more confusion. I am removing this replace() call in the new update.

It is also easier to tell that the decode() part is for backward compatibility purpose only without the encode() part.

PR #2787, opened 2018-03-16
Reduce cpu packing count for local instance

aahmed-se:
@srkukarni this fixes the local instance issue


kramasamy:
@srkukarni @aahmed-se - you need to reduce the memory as well (for instances) so that we can run in a docker container.

PR #2793, opened 2018-03-20
Check for negative return from % operator

srkukarni:


heron/api/src/java/com/twitter/heron/streamlet/impl/groupings/JoinCustomGrouping.java
```diff
@@ -47,6 +47,9 @@ public void prepare(TopologyContext context, String component,
     List<Integer> ret = new ArrayList<>();
     V obj = (V) values.get(0);
     int index = keyExtractor.apply(obj).hashCode() % taskIds.size();
+    if (index < 0) {
+      index = (index + taskIds.size()) % taskIds.size();
```
jerrypeng: why are we doing this?

jerrypeng: how can index but less than 0?

srkukarni: hashCode can return negative numbers. % according to spec can be negative if dividend is negative(it really is implementation specific, but looks like atleast some versions of jvm return negative).

nwangtw: First of all, what is the impact? This could cause crashing? or bad results?

This line is a bit redundant. Just notes, not a blocker.

(index + taskIds.size()) % taskIds.size()
or 
if (index) {
  index = index + taskIds.size();
}

should be enough.

Another option might be 
Math.abs(keyExtractor.apply(obj).hashCode()) % taskIds.size();

srkukarni: The issue is that if index is negative, list.get(index) fails.

jerrypeng: Math.abs(hash). This can also give you negative numbers when hash is bigger than 2^31, and the best way would be to use a shift mask (key.hashCode() & 0x7fffffff) % M, where M is the table size.

nwangtw: Math.abs(hash) can give negative result? Sounds strange to me.

bitmask is dangerous because it assumes the var is 32 bits. In case someone changed the data type it can cause other problem.

nwangtw:
I added another PR with a similar change:

https://github.com/apache/incubator-heron/pull/2953

nicknezis:
Fixed by @2953. Closing the pull request.

PR #2797, opened 2018-03-21
Align MetricsCacheMgr location discovery in Stmgr

huijunw:
MetricsCacheMgr can be turned on or off by a switch config. When it is off, its location does not exist in StateMgr. The present Stmgr always tries to discover the MetricsCacheMgr location.

This PR disables the above discovery in Stmgr when MetricsCacheMgr is not running.


nwangtw:
lgtm. is this change backward compatible?

PR #2798, opened 2018-03-21
While emitting in spout, adhere to the batch size limit

srkukarni:


heron/instance/src/java/com/twitter/heron/instance/spout/SpoutInstance.java
```diff
@@ -337,6 +341,10 @@ protected void produceTuple() {
       if (currentTime - startOfCycle - instanceEmitBatchTime.toNanos() > 0) {
         break;
       }
+      if (!ByteAmount.fromBytes(newTotalBytesEmitted - totalBytesEmitted)
```
nwangtw: do we need to add the same logic in c++ and python?

srkukarni: @nwangtw Good catch!
It looks like python already does this but c++ doesn't. I've added that logic as well.

srkukarni:
retest this please

jerrypeng:
@nwangtw @maosongfu Is this PR good to go?

PR #2799, opened 2018-03-21
Add default component resource config

nwangtw:


heron/api/src/java/com/twitter/heron/api/Config.java
```diff
@@ -295,6 +295,22 @@
...
 
+  /**
+   * Default number of cpu cores per component to be reserved for this topology
+   */
+  public static final String TOPOLOGY_COMPONENT_DEFAULT_CPU = "topology.component.default.cpu";
+  /**
+   * Default amount of ram per component to be reserved for this topology.
+   * In bytes.
+   */
+  public static final String TOPOLOGY_COMPONENT_DEFAULT_RAM = "topology.component.default.ram";
```
ashvina: Is this statement accurate: all instances of a component together will get this amount of RAM, be default? For e.g. if a comp C has instance i1 and i2 and default RAM is 2 GB, i1 will be assigned 1GB of ram.

On the other hand, if this is instance default, how is different from current instance ram config?

jerrypeng:
@nwangtw in the future are we planning on having the non-default version of all the variables?  If that is the case I am not sure that is a good idea.  We will have something like TOPOLOGY_COMPONENT_DEFAULT_CPU as well as TOPOLOGY_COMPONENT_CPU.  Might be an overkill.  Can we just use one variable e.g. TOPOLOGY_COMPONENT_CPU and just set a default for it and it will be the default value for all components in a topology for that resource.  If the use want to overwrite they can do it on a per component basis.  This is whole it was done in Storm. 

Also in Storm we used file i.e. default.yaml to hold all the default values for variables.  On packaging storm, the defaults.yaml will be placed inside storm-core.jar. The defaults.yaml file will be read on startup for default values.  This make it easier to find and change defaults compared to hardcoding default values in code.

jerrypeng:
@nwangtw yes components can have different resource requirements, I am just suggesting we use one variable for each resource and just pre set a default value for that variable and allow the user to overwrite it in their topology main if they want.

for example 

TOPOLOGY_COMPONENT_CPU is set to default of 10 the "word" component

if the user doesn't overwrite it like:

SpoutDeclarer spout = builder.setSpout("word", new TestWordSpout(), 10).setCPULoad(20);

than the "word" component will has a CPU req of 10

jerrypeng:
On a side note, I think the API for resource specification will be more user friendly if users can specify resource requirements directly on the Spout/bolt object like:

SpoutDeclarer spout = builder.setSpout("word", new TestWordSpout(), 10).setCPULoad(20);

instead of separately declaring it:

conf.setComponentRam("word", 20);

Perhaps we can have both options

heron/api/src/java/com/twitter/heron/api/Config.java
ashvina: @kramasamy @nwangtw 
I am a bit confused. 
1. The config name is `COMPONENT_DEFAULT`, but it is use to configure `INSTANCE_DEFAULT`. Should this be renamed? 
1. Why are we replacing existing configs for cpu and disk, `heron.resources.instance.cpu` and `heron.resources.instance.disk`

nwangtw: 1. It is a good question. I think both makes sense, in a different way. If you consider the next step, per component resource config, component_default makes more sense. If you consider that all instances are using the config as default value, instance_default makes more sense. I am thinking instance_default might feel better now. Open to other thoughts.

2. replacing? existing configs are used as fallback values, unless there is a bug.

nwangtw:
@jerrypeng Yeah. that is stage two. https://github.com/twitter/heron/pull/2800

nwangtw:
dropping to avoid new configs.

https://github.com/twitter/heron/pull/2800 should allow users to do it. We will see if we need another helper function or not in future

PR #2800, opened 2018-03-22
Add per component resource config for cpu and disk

nwangtw:


kramasamy:
@nwangtw - can you do double instead of float consistently across C++ and Java.

nwangtw:
Sure. It seems the existing cpu configs are floats in java. let me change them all.

PR #2801, opened 2018-03-22
Do not include scribe sink when compiling on non x86 platforms.

kramasamy:
This is due to scribe dependency on thrift which needs to generate java files - the binaries used are only x86 platforms.

### DIFF
--- a/tools/platform/BUILD
+++ b/tools/platform/BUILD
@@ -7,15 +7,7 @@ config_setting(
 )
 
 config_setting(
-    name = "ubuntu",
-    values = {
-        "cpu": "k8",
-    },
-    visibility = ["//visibility:public"],
-)
-
-config_setting(
-    name = "centos",
+    name = "k8",
     values = {
         "cpu": "k8",
     },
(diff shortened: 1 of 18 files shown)


tools/platform/BUILD
```diff
@@ -7,15 +7,7 @@ config_setting(
...
-    name = "ubuntu",
-    values = {
-        "cpu": "k8",
-    },
-    visibility = ["//visibility:public"],
-)
-
-config_setting(
-    name = "centos",
+    name = "k8",
```
nwangtw: I am wondering if the change in this file is expected/needed in this PR. it seems unrelated and looks a little strange.

kramasamy: good catch it is not needed.

kramasamy: @nwangtw - i need to take my comment back. this is needed to ensure all linux variants are grouped under one target which is k8 - otherwise, it is ubuntu, centos and more variations.

nwangtw: I see. So now k8 is considered as a platform? A little unclear but not a blocker to me.

kramasamy:
@nwangtw - k8 represents all variants of linux in bazel.

PR #2807, opened 2018-03-23
remove three other roundings in cpu calculation

nwangtw:


nwangtw:
unit tests to update

nlu90:
What's the background of removing this `Math.rounding()`?

nwangtw:
It is more about "why there was a Math.rounding()".

The rounding logic doesn't seem necessary if we assume cpu # can be less than fractional. Also the logic have bugs that in some cases it can be floored to 0.

However when I was working on this PR, I found out that some parking algorithms rely on the rounding so this is not a quick fix. This PR needs to be revisited when I get more time.

PR #2811, opened 2018-03-26
Fix comment in heron_internals.yaml

ajorgensen:
The comment does not match directly with the configurations below it.


ajorgensen:
It looks like the same test is failing on master, given that this is a comment change it shouldn't have any impact on existing tests.

PR #2812, opened 2018-03-26
[Uploaders] Improve HttpUploader Part II

erenavsarogullari:
- Remove redundant `HttpClient`
- Adding checks for both `Uploader_Http_URI` and `Topology_Package_File` at initialization level as fail-fast.
- Using `try-with-resources` block to close `CloseableHttpClient` instance so `finally` block is being removed.
- Extending UT coverages by adding `6 negative cases`.

### DIFF
--- a/heron/uploaders/src/java/com/twitter/heron/uploader/http/HttpUploader.java
+++ b/heron/uploaders/src/java/com/twitter/heron/uploader/http/HttpUploader.java
@@ -21,6 +21,9 @@ import java.nio.charset.StandardCharsets;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import com.google.common.base.Preconditions;
+
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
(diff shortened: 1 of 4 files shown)


heron/uploaders/src/java/com/twitter/heron/uploader/http/HttpUploader.java
```diff
@@ -44,55 +48,57 @@
...
+  private static final String FILE = "file";
   private Config config;
   private String topologyPackageLocation;
   private HttpPost post;
 
   @Override
   @SuppressWarnings("HiddenField")
   public void initialize(Config config) {
+    String uploaderUri = HttpUploaderContext.getHeronUploaderHttpUri(config);
+    Preconditions.checkArgument(StringUtils.isNotBlank(uploaderUri),
```
jerrypeng: Thanks for adding these checks.  We need to add these checks all around in Heron.  We currently lack any sort of organized config validation

erenavsarogullari: @jerrypeng Yes, exactly. Thanks for review ;)

jerrypeng: I created a config validation system using annotations in storm back in the day:

https://github.com/apache/storm/blob/master/storm-client/src/jvm/org/apache/storm/Config.java

Perhaps worthwhile to think if its worth while to have in Heron.  Though in Storm all configs are centralized which is not the case for Heron

PR #2815, opened 2018-03-27
Add optional topology tag.

cckellogg:


cckellogg:
The use case for the tag is to provide more information and context to a topology's function. For example, a topology could be an analytics topology, data-ingestion topology, machine-learning topology, example topology etc... This tag can be surfaced in the ui and explorer to provide more context about the topology's purpose. Also, this could be used for metric aggregation across tags. For example, how much resources are my analytics topologies using. Thoughts?

ashvina:
Thanks  @cckellogg 

This is very interesting. However, in the use case you mentioned, the tag/category seems like external to heron and a topology. It is also possibly that a user would like to update topology's tag without restarting the topology. I can also imagine that the tag will be assigned by an operator who may not have the ability to update topology code. Hence, I think making the tag a `Config` may not address the requirement. 

@kramasamy ?

cckellogg:
@ashvina any suggestions how to implement this feature?

ashvina:
Since this is a UI/explorer feature and no other heron component will ever use this data, how about adding an UI action/command to "decorate" topology with labels? Tracker could persist the data in a new topology sub-node in state manager.

cckellogg:
This would solve some of the use cases but not the metrics one. For example, in prometheus you can attach labels to metrics. Exposing the tag in the config would allow an operator or developer to group metrics by tag/category.

ashvina:
Hi @cckellogg 
Even in the use case you just mentioned, a tag seems external to topology logic. In my opinion, using topology config for tag is odd. I believe a topology developer will not know how its metrics will be grouped. Topologies can be shared and deployed in different environments. A developer cannot be expected to think of all those use cases in advance. 

It is possible my understanding is flawed. I am hoping others will present their opinion and we can move forward.

@maosongfu @nwangtw , what do you think about this?

kramasamy:
@ashvina - the reason we proposed this approach is because we already have some config such as topology.project, topology.team, etc. Instead of changing the design approach we were just augmenting what was already there.

nwangtw:
config is convenient because the functions/tools/ui are all available. But @ashvina's concern makes sense too.

I don't know if config is the best option or not so far, but I think we may step back a bit and think about it from tracking point of view. To me it seems to be a feature in heron-tracker/heron-ui. It could b nice to treat the information as a "property". And @kramasamy is right that topology.project and topology.team are similar.

Just my 2 cents.

PR #2818, opened 2018-03-28
Add missing command line options for runtime config

nwangtw:


nwangtw:
'f' is definitely not ideal. but 'r' and 'c' are both used. I just feel 'f' is relative better than the other characters in "runtime-config". Any suggestion is welcome.

Also it seems there are quite a few options now and these characters are not very useful any more. :( I am wondering it might be better to keep abbreviation to only key options.

heron/scheduler-core/src/java/com/twitter/heron/scheduler/RuntimeManagerMain.java
```diff
@@ -115,6 +115,13 @@ private static Options constructOptions() {
         .argName("component parallelism")
         .build();
 
+    Option runtimeConfig = Option.builder("f")
```
cckellogg: what about no short name or "rc"? 

Also instead of doing a list you can do something like this
final Option property = Option.builder("rc")
        .argName("comp:property=value")
        .numberOfArgs(2)
        .valueSeparator()
        .desc("use value for given property")
        .build();
then you can pass values like this:
-rc word:prop1=value1 -rc word:prop2=value2

nwangtw: sounds good! Didn't even thinking about two characters. :D

nwangtw: We do want to keep the current format though to be consistent to component_parallelism.

[comp:]<name>:<value>,[comp:]<name>:<value>,.....

The number of component can be infinity, not really 2.

This is the argument for scheduler, not the CLI argument, which is in python. The CLI argument is in the format in your example.

PR #2831, opened 2018-03-31
Fix bug in dry-run argument for update command

ajorgensen:
Since `list_extra_args` is an array, the `+=` operator deconstructs a
string and add each character as an individual array element. This made
it so when the update command it would not detect if it was in dry-run
mode and always run the update command. By encapsulating the `--dry-run`
string in an list it is properly added as an argument and runs with the
expected behavior.


kramasamy:
+1

PR #2832, opened 2018-03-31
allow storm topologies to use the heron state interface

jerrypeng:
To use the heron state interface uses will need to add these configs:

conf.put(com.twitter.heron.api.Config.TOPOLOGY_RELIABILITY_MODE, com.twitter.heron.api.Config.TopologyReliabilityMode.EFFECTIVELY_ONCE);
    conf.registerSerialization(com.twitter.heron.api.state.HashMapState.class);

in addition to implementing the stateful component interface:

public static class AckingTestWordSpout extends BaseRichSpout implements IStatefulComponent<String, String>{
.
.
.
}

### DIFF
--- a/storm-compatibility/src/java/org/apache/storm/topology/TopologyBuilder.java
+++ b/storm-compatibility/src/java/org/apache/storm/topology/TopologyBuilder.java
@@ -25,6 +25,7 @@ import org.apache.storm.generated.StormTopology;
 
 import com.twitter.heron.api.HeronTopology;
 import com.twitter.heron.api.bolt.WindowedBoltExecutor;
+import com.twitter.heron.api.topology.IStatefulComponent;
 
 public class TopologyBuilder {
   private com.twitter.heron.api.topology.TopologyBuilder delegate =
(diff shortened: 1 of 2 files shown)


jerrypeng:
@kramasamy

kramasamy:
Ok. Can we have the two configs at heron also available in storm namespace - users in storm namespace should not use heron namespace.

Implementing a stateful interface is fine.

jerrypeng:
The configs don’t make sense for storm though

kramasamy:
It feels weird to use Heron namespace config - which is always needed for stateful topologies. 

The second call is needed for serializing the state? If the state is always a hash map - should we register it rather than letting the programmer to do it. Again we don't want to expose anything in Heron API at the storm level.

jerrypeng:
@kramasamy this is kinda of a hack to begin with.  Storm doesn't have exactly once state.  If users want exactly once state and eloquence of the API they should use the Heron API to begin with.  This just unblocks users that want to use the heron state api with the storm api

kramasamy:
@jerrypeng - my key argument the user needs to be aware that a configuration from Heron API required for using with Storm API - which is not a good developer experience. API namespace should be clean.

srkukarni:
@jerrypeng I'm not sure if this is the right approach to take. You are breaking storm compatibility as well as breaking abstractions

PR #2845, opened 2018-04-03
add config for apiserver overrides

jerrypeng:
currently a scheduler invoked by the apiserver has no way of getting the set of config overrides set when launching the apiserver.  Adding a config so that a scheduler can read in the override yaml file created by the apiserver for the config overrides specified in the apiserver launch command.


jerrypeng:
@cckellogg @kramasamy can you guys take a look? Thanks!

PR #2850, opened 2018-04-05
Add SSL support for stream manager connections

ajorgensen:
This commit adds ssl support for connection between stream managers. The
main purpose of this change is to protect data that is in-transit
through a heron topology from packet sniffing as such it is only
implemented between each stream manager. Communication to the tmaster,
metrics manager, or instance will not be encrypted.

This basic implementation adds the ability to optionally pass an ssl certificate
and private key to the libevent connection option. This change does not
introduce any keystores and relies on proper permissions and keys to
exist on the system or in the sandbox at the time of the deploy.

The introduction of openssl is specifically left as a dynamic link
instead of a static link. The reason for this is if there is a
vulnerability discovered in an openssl library it will but much easier
for the topology owner to simply upgrade the system version of openssl
to pick up the fix. If openssl was statically linked into heron it would
require re-releasing older heron versions statically linked to the
latest openssl version and all topology owners to redownload the release
which would make rolling out the fix much harder.

Closes https://github.com/apache/incubator-heron/issues/2773

# Open Question
Since bazel/cpp are not my specialty. I do have a few open questions that will likely need to be addressed before this is merged.

1. What happens in environments where openssl is not installed? Can we have a flag that optionally removes the ssl feature if necessary?
2. I need to write up documentation for this but I think would like to have it bake in a production environment for a bit to make sure there are no issues. I can write up a document but mark it as beta if we want.

TODO
=====
- [ ] Server and client trust stores
- [ ] Only link against ssl or include ssl if its available


ajorgensen:
@kramasamy yes the intention is for everything to function as is and *only* in the case that a certificate and private key path are provided will an ssl connection be established.

Anyone running a production environment should always have their openssl version up to date to make sure it has the latest patches, bug fixes, etc. As long as the version running on the production machine is ABI compatible with 1.1.0 then this should work as intended. I believe openssl tries to be as backwards compatible as possible (https://abi-laboratory.pro/tracker/timeline/openssl/). 1.1.0 was released in 2016 which was the largest breaking change they've made and until then its been close to 100%. I think anyone running a version older than 1.1.0 will have bigger problems then not being able to run heron with ssl but assuming the methods that we're using are still in that version it *should* just work.

ajorgensen:
@kramasamy yes, we just got a high volume topology running with the change yesterday. There are one or two tweaks to this I need to make and I will provide more detailed instructions. I am currently on vacation so I will not be able to provide that until mid next week but will get it to you asap.

nlu90:
@ajorgensen seems several unit tests failed. Could you update it and for a further review?

kramasamy:
@ajorgensen - any reason for closing this?

ajorgensen:
It's not complete and it has diverged enough from our implementation that I need to cherry-pick the commits again. I figured it'd be best to open a new PR with the full set of changes but I can push them here if you'd rather.

PR #2852, opened 2018-04-06
first of the connectors - twitter spout

kramasamy:


third_party/cereal/BUILD
```diff
@@ -0,0 +1,126 @@
+licenses(["notice"])
```
nlu90: We already have a cereal lib in the third_party dir [here](https://github.com/apache/incubator-heron/tree/master/third_party/cereal).

So please either remove the files here or send out another separate PR for changing this lib

connectors/heron-twitter/src/java/com/twitter/heron/twitter/spout/Twitter.java
```diff
@@ -0,0 +1,122 @@
+package com.streamlio.connectors.twitter;
```
nlu90: the package should be `org.apache.heron.connectors.twitter` now

nlu90:
@kramasamy  Ping

nlu90:
@kramasamy Let's try to merge this first connector and then keep adding more in the future. Two more suggestions for this PR:

1. remove the `cereal` related files in third_party

2. rename the directory `connectors` to `heron-connectors`

PR #2857, opened 2018-04-06
Add two backup urls for zookeeper

ajorgensen:
I've experiences some flakiness with a few of the zookeeper archive
urls. This adds two backup urls in the event that one of them is down or
a connection cannot be established.

Zookeeper was the only obvious candidate that had mirrors so we can start here and add other mirrors if people experience flakiness for other archives.

Closes https://github.com/apache/incubator-heron/issues/2855


nlu90:
👍

PR #2884, opened 2018-05-02
move downloader registry map to yaml

huijunwu:
I am revisiting https://github.com/apache/incubator-heron/pull/2274. The static config in the downloader registry looks hard to extend in our private repo. I propose to refactor the `static` configuration code to a yaml file.

in container (cluster):
`$ heron-downloader.sh -u file:///Users/xxx/.heron/dist/heron-core.tar.gz -f ./tmp/`
local:
`$ ~/.heron/bin/heron-downloader.sh -u file:///Users/xxx/.heron/dist/heron-core.tar.gz -f ~/tmp/ -d ~/.heron -p ~/.heron/conf/local -m local`

### DIFF
--- a/heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
+++ b/heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
@@ -76,7 +76,6 @@ public final class DownloadRunner {
         .longOpt(CliArgs.TOPOLOGY_PACKAGE_URI.text)
         .hasArgs()
         .argName(CliArgs.TOPOLOGY_PACKAGE_URI.text)
-        .required()
         .build();
 
     Option destination = Option.builder("f")
@@ -84,7 +83,6 @@ public final class DownloadRunner {
         .longOpt(CliArgs.EXTRACT_DESTINATION.text)
         .hasArgs()
         .argName(CliArgs.EXTRACT_DESTINATION.text)
-        .required()
         .build();
 
     Option heronHome = Option.builder("d")
@@ -172,6 +170,7 @@ public final class DownloadRunner {
 
     String uri = cmd.getOptionValue(CliArgs.TOPOLOGY_PACKAGE_URI.text, null);
     String destination = cmd.getOptionValue(CliArgs.EXTRACT_DESTINATION.text, null);
+    // make it compatible with old param format
     if (uri == null && destination == null) {
       String[] leftOverArgs = cmd.getArgs();
       if (leftOverArgs.length != 2) {
--- a/heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
+++ b/heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
@@ -17,8 +17,6 @@ import java.io.File;
 import java.net.URI;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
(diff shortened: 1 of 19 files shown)


heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
```diff
@@ -38,7 +184,14 @@ public static void main(String[] args) throws Exception {
...
 
-    final Downloader downloader = Registry.get().getDownloader(topologyLocation);
+    Map<String, Class<? extends Downloader>> downloaders = new HashMap<>();
+    for (Map.Entry<String, Object> e
+        : ((Map<String, Object>) config.get(Key.DOWNLOADER_PROTOCOLS)).entrySet()) {
+      Class clazz = Class.forName((String) e.getValue());
+      downloaders.put(e.getKey(), clazz);
+    }
+
+    final Downloader downloader = Registry.getDownloader(downloaders, topologyLocation);
```
nlu90: how about only creating the downloader that will be used in the topology? Others seem not necessary to be created.

huijunwu: updated

heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
```diff
@@ -38,7 +184,14 @@ public static void main(String[] args) throws Exception {
       file.mkdirs();
     }
 
-    final Downloader downloader = Registry.get().getDownloader(topologyLocation);
+    Map<String, Class<? extends Downloader>> downloaders = new HashMap<>();
+    for (Map.Entry<String, Object> e
```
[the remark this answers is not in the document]
huijunwu: updated

nlu90:
LGTM

nlu90:
@kramasamy  This change may break your downloader usage, please update your code accordingly using the new `downloader.yaml` file.

jerrypeng:
So the arguments are changed for the downloader right?  Won't this break existing usages of the downloader.  Heron standalone/nomad calls the downloader to download topology packages.  Can we also support calling the downloader with the existing parameters as well?

jerrypeng:
I feel the changes have made the downloader a lot more complicated.  Now we have to pass in a cluster mode or config path because we have to read the downloader.yaml.  However, I think the most common use case for the downloader is 

"downloader <topology-package-uri> <extract-destination>"

huijunwu:
@jerrypeng 
1. The current Downloader hard-code the registry map, which is not extendable.
2. The Downloader should work in a similar way as other code in the heron, such as Uploader which adopts the yaml config for extension.

jerrypeng:
@huijunwu my concern is that it will break existing code that assume the format for the downloader is 

```
downloader <topology-package-uri> <extract-destination>
```

heron/downloaders/src/java/org/apache/heron/downloader/DownloadRunner.java
```diff
@@ -18,17 +18,160 @@
...
+            .putAll(ConfigLoader.loadConfig(heronHome, configPath, null, null))
+            .build());
+        break;
+
+      default:
+        throw new IllegalArgumentException(
+            "Invalid mode: " + cmd.getOptionValue(CliArgs.MODE.text));
+    }
+
+    final String uri = cmd.getOptionValue(CliArgs.TOPOLOGY_PACKAGE_URI.text, null);
```
nwangtw: what happens if uri and destination are null? the old logic has protection and it seems new code doesn't. I think it might be better to have a check here.

huijunwu:
@jerrypeng updated to be compatible with `downloader <topology-package-uri> <extract-destination>`. could you review it again

PR #2886, opened 2018-05-03
submit topology file by uri

huijunwu:
replace #2274
after #2884

`$ heron submit [options] cluster/[role]/[env] topology-file-uri topology-class-name [topology-args]`

for example:
```
$ tar -czvf heron-api-examples.jar.tar.gz heron-api-examples.jar
$ heron submit [options] cluster/[role]/[env] file:///Users/xxx/tmp/heron-api-examples.jar.tar.gz topology-class-name [topology-args]
$ heron submit [options] cluster/[role]/[env] http://www.examples.com/topo.jar.tar.gz topology-class-name [topology-args]
```


heron/tools/cli/src/python/submit.py
```diff
@@ -366,6 +368,16 @@ def submit_cpp(cl_args, unknown_args, tmp_dir):
...
   return launch_topologies(cl_args, topology_file, tmp_dir)
 
+def download(uri):
+  current_path = os.path.realpath(__file__)
+  print current_path
+  tmp_dir = tempfile.mkdtemp()
+  subprocess.call([current_path + "downloader.sh", uri, tmp_dir])
+  suffix = (".jar", ".tar", ".tar.gz", ".pex", ".dylib", ".so")
+  for f in os.listdir(tmp_dir):
+    if f.endswith(suffix):
```
nlu90: If there are multiple files in this tmp directory, then this `download` method will return the first file it encountered ending with a valid suffix. But is this returned file truly a valid topology file?

If there will be only one file downloaded to this directory, I suggest we remove this `for` loop.

huijunw: we need to make an agreement on this with users. the users may assume there is only one file, or assume there are multiple files

PR #2891, opened 2018-05-04
[WIP] Refactor StatefulStorage

nwangtw:
- Refactor Stateful Storage to be more independent of topology
- Refactor data structures.

### DIFF
--- a/heron/proto/ckptmgr.proto
+++ b/heron/proto/ckptmgr.proto
@@ -82,7 +82,9 @@ message StatefulConsistentCheckpoints {
   repeated StatefulConsistentCheckpoint consistent_checkpoints = 1;
 }
 
-// tmaster -> stmgr messages
+/*
+ * tmaster <-> stmgr messages
+ */
 
 // Message sent to stmgrs by the tmaster to initiate checkpointing
 message StartStatefulCheckpoint {
@@ -140,7 +142,9 @@ message StartStmgrStatefulProcessing {
   required string checkpoint_id = 1;
 }
 
-// tmaster -> ckptmgr messages
+/*
+ * tmaster -> ckptmgr messages
+ */
 
 // This is the message that a tmaster sends
 // when it wants to register itself
@@ -171,7 +175,25 @@ message CleanStatefulCheckpointResponse {
   repeated string cleaned_checkpoint_ids = 2;
 }
 
-// stmgr -> ckptmgr messages
+/*
+ * stmgr -> ckptmgr messages
+ */
+
+// This message encapsulates the info associated with
+// state of an instance/partition
+message InstanceStateCheckpoint {
+  required string checkpoint_id = 1;
+  required bytes state = 2;
+  // A version string that can be specified by user. It can be used to translate
+  // checkpoint data from older versions to the latest version when necessary.
+  optional string data_version = 3;
+}
+
+// This message encapsulates the info associated with
+// checkpoint metadata of a component
+message CheckpointComponentMetadata {
+  required int32 parallelism = 1;
+}
 
 // This is the message that a stmgr sends
 // when it wants to register itself with
@@ -195,6 +217,9 @@ message SaveInstanceStateRequest {
   // Information about the instance whose state this is
   required heron.proto.system.Instance instance = 1;
   required InstanceStateCheckpoint checkpoint = 2;
+  // Stmgrs need to send meta data as well to the first instance of
+  // each component as well
+  optional CheckpointComponentMetadata metadata = 3;
 }
 
 // After successfully saving state, this is what
@@ -221,19 +246,14 @@ message GetInstanceStateResponse {
   optional InstanceStateCheckpoint checkpoint = 4;
 }
 
-// stmgr -> Instance messages
+/*
+ * stmgr -> Instance messages
+ */
 
 // This is the message that the stmgr sends to its
 // local tasks to begin initiating checkpoint
 message InitiateStatefulCheckpoint {
   required string checkpoint_id = 1;
-}
-
-// This proto encapsulates the info associated with
-// state of an instance
-message InstanceStateCheckpoint {
-  required string checkpoint_id = 1;
-  required bytes state = 2;
 }
 
 // This is the message that the instance sends
@@ -262,7 +282,9 @@ message StartInstanceStatefulProcessing {
   required string checkpoint_id = 1;
 }
 
-// stmgr -> stmgr messages
+/*
+ * stmgr -> stmgr messages
+ */
 
 // Once stmgr receives StoreInstanceStateCheckpoint from an instance,
 // it sends this message to all of that instance's 
--- a/heron/spi/src/java/org/apache/heron/spi/statefulstorage/IStatefulStorage.java
+++ b/heron/spi/src/java/org/apache/heron/spi/statefulstorage/IStatefulStorage.java
@@ -16,37 +16,65 @@ package org.apache.heron.spi.statefulstorage;
 
 import java.util.Map;
 
-import org.apache.heron.proto.system.PhysicalPlans;
-
+/**
+ * The interface of all storage classes for checkpoints.
+ * For each checkpoint, two types of data are stored:
+ * - Component Meta Data (one per component).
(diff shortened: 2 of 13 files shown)


heron/proto/ckptmgr.proto
```diff
@@ -171,7 +175,25 @@ message CleanStatefulCheckpointResponse {
...
+  required bytes state = 2;
+  // A version string that can be specified by user. It can be used to translate
+  // checkpoint data from older versions to the latest version when necessary.
+  optional string data_version = 3;
+}
+
+// This message encapsulates the info associated with
+// checkpoint metadata of a component
+message CheckpointComponentMetadata {
+  required int32 parallelism = 1;
```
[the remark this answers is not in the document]
nwangtw: kk. component name could be useful for viewing data. Will add.

Firstly, stateful data is organized around components (the schema should be the same for all instances/partitions of the component and could be different between components). It is overall a 3-level tree structure: topology is the root, component is the mid level and instance/partition is the leaf level. We need to store component level data in storage.

Secondly, the parallelism value is important to detect the plan changes and handle the change correctly. In our first state, we can drop the stateful data when parallelism doesn't match. In future, we should handle repartition based on the old and new parallelisms of the component.

heron/spi/src/java/org/apache/heron/spi/statefulstorage/IStatefulStorage.java
```diff
@@ -16,37 +16,65 @@
...
+   */
+  Checkpoint restoreCheckpoint(final CheckpointPartitionInfo info) throws StatefulStorageException;
+
+  /**
+   * Store medata data for component. Ideally this function should only be called once
+   * for each component.
+   * @param info The information (reference key) for the checkpoint partition.
+   * @param metadata The checkpoint metadata from a component.
+   */
+  void storeComponentMetaData(final CheckpointPartitionInfo info, final CheckpointMetadata metadata)
```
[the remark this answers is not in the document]
nwangtw: They will be used in CheckpointManager. I will add the calls and TODOs.

When a checkpoint is created, instance 0 of each component should be responsible to call this function to save component information to storage.

During a state restore, all instances should load component data first and see if it is possible to recover from the checkpoint (and in future how to recover), then load instance/partition data.

heron/proto/ckptmgr.proto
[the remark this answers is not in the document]
nlu90: kk

heron/spi/src/java/org/apache/heron/spi/statefulstorage/IStatefulStorage.java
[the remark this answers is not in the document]
nwangtw: Good point!

Then storage implementation need to make this decision.

Instance 0 should be responsible for the meta data in distributed storages; and for local storages, all instances need to export the data (and the data needs to be stored in the instance/partition directory). StatefulStorage implementation should know the information and handle the function correctly.

heron/proto/ckptmgr.proto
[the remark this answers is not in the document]
nwangtw: That is one option for distributed storage. Another option is to store the data along with instance 0. This is a question we need to answer when updating stateful storage implementations and adding this feature.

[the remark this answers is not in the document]
srkukarni: is there any documentation that describes whats this pr aims to do? Without it its very hard to understand whats going on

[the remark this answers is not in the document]
nwangtw: We have a brief doc internally, but it is not as detailed as this PR. I will clean up the doc and share it here.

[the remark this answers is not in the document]
nwangtw: @skanjila here is the doc: https://docs.google.com/document/d/1p443mcdD607Vf9aCnMq5ESVMFrEuneGlceXmtUSvVMQ/edit#heading=h.myl6mycadhfh

heron/proto/ckptmgr.proto
```diff
@@ -171,7 +175,26 @@ message CleanStatefulCheckpointResponse {
...
+/*
+ * stmgr -> ckptmgr messages
+ */
+
+// This message encapsulates the info associated with
+// state of an instance/partition
+message InstanceStateCheckpoint {
+  required string checkpoint_id = 1;
+  required bytes state = 2;
+  // A version string that can be specified by user. It can be used to translate
```
srkukarni: exactly what does this data version do?

heron/proto/ckptmgr.proto
```diff
@@ -195,6 +218,9 @@ message SaveInstanceStateRequest {
   // Information about the instance whose state this is
   required heron.proto.system.Instance instance = 1;
   required InstanceStateCheckpoint checkpoint = 2;
+  // Stmgrs need to send component metadata so that Checkpoint
+  // Mgrs know how to repartition if needed.
+  optional CheckpointComponentMetadata metadata = 3;
```
srkukarni: does this need to be passed to cjptmgr? Doesn't ckptmgr know this info

nwangtw: yeah. this info will be passed to ckptmgr, so that it knows if repartition is needed and how to repartition.

heron/proto/ckptmgr.proto
nwangtw: We are thinking about calculation change. Not sure how useful it is though. Please feel free to let us know if you feel this is overkill.

Normally the calculation in bolts won't change much and the version number doesn't need to change. However in case the equation is changed, for example,  the old value is a duration in seconds and it is changed to duration in millisecond, or it is possible the value is a data structure and a new field is added. User can increase the version number and provide a conversion function between the two versions. Then the topology would be able to load the stateful data stored before the update, or skip/reject the data if topology doesn't know how to convert between the two versions.

If topology cant tell the difference and just load the data, the state could be polluted without user knowing it. The current design seems to rely on users manually removing the data? it could be risky and hard since the topology needs to be running in order to do it I think?

srkukarni: IMO this really pollutes the abstraction. You have now introduced some kind of versioning in the system at such a low level to begin with without any kind of higher level api/rationale.

heron/proto/ckptmgr.proto
srkukarni: my comment was more that when you change the parallelism, wouldnt the ckptmgr be restarted also and thus it too knows the topology structure?

heron/proto/ckptmgr.proto
nwangtw: kk. Let me remove it for now until we have a better idea/support.

heron/proto/ckptmgr.proto
nwangtw: chptmgr is stateless and it doesn't have the physical plan, so it doesn't know the parallelism by itself.

nwangtw: Checked the communication between stmgr and ckptmgr, it is a valid option to pass the topology structure information during the registration to simplify the requests.

We don't need to make the decision for now about where the structure information should be passed. Therefore I am going to remove this field for now to keep it simple.

Thanks again for the comments!

heron/proto/ckptmgr.proto
```diff
@@ -171,7 +175,23 @@ message CleanStatefulCheckpointResponse {
...
 }
 
-// stmgr -> ckptmgr messages
+/*
+ * stmgr -> ckptmgr messages
+ */
+
+// This message encapsulates the info associated with
+// state of an instance/partition
+message InstanceStateCheckpointPartition {
```
srkukarni: Is this just instance or partition or instance and partition? Its unclear, especially since it does not have anything to do with partition right now

srkukarni: My suggesstion would be to leave the name of InstanceStateCheckpoint for now and rename later?

kramasamy:
@nwangtw - can you resolve conflicts if this is good to go.

heron/proto/ckptmgr.proto
nwangtw: kk. let me do it.

PR #2900, opened 2018-05-16
Fix Potential Double Delete Bug

ajorgensen:
This fixes a double free bug that periodically causes the stream manager to crash if other stream managers or are killed at the right time. 

In this case opkt has the potential to be freed twice in the event of a failed packet sent in both `InternalSendResponse` and `conn->sendPacket(opkt)`

### DIFF
--- a/heron/common/src/cpp/network/server.cpp
+++ b/heron/common/src/cpp/network/server.cpp
@@ -175,15 +175,11 @@ void Server::OnNewPacket(Connection* _connection, IncomingPacket* _packet) {
 void Server::InternalSendResponse(Connection* _connection, OutgoingPacket* _packet) {
   if (active_connections_.find(_connection) == active_connections_.end()) {
     LOG(ERROR) << "Trying to send on unknown connection! Dropping.. " << std::endl;
-    delete _packet;
     return;
   }
-  if (_connection->sendPacket(_packet) != 0) {
+  if (_connection->sendPacket(*_packet) != 0) {
     LOG(ERROR) << "Error sending packet to! Dropping... " << std::endl;
-    delete _packet;
-    return;
   }
-  return;
 }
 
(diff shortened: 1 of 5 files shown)


kramasamy:
We can merge it

ajorgensen:
@nlu90 fixed. Is there a bazel command just run checkstyle?

kramasamy:
@ajorgensen - here it is

bazel build --config=darwin <targets>

ajorgensen:
@kramasamy  Oh interesting... the checkstyles never run for me during a normal build. For example if I introduce the same checkstyle issue and run

` bazel build //heron/common/src/cpp/..` 

It builds successfully. Its not until I run the `scripts/packages` directive that the checkstyle actually fires.

heron/common/src/cpp/network/server.cpp
```diff
@@ -211,17 +207,16 @@ void Server::InternalSendRequest(Connection* _conn, google::protobuf::Message* _
   sp_int32 byte_size = _request->ByteSize();
   sp_uint32 sop = OutgoingPacket::SizeRequiredToPackString(_request->GetTypeName()) + REQID_size +
                   OutgoingPacket::SizeRequiredToPackProtocolBuffer(byte_size);
-  auto opkt = new OutgoingPacket(sop);
-  CHECK_EQ(opkt->PackString(_request->GetTypeName()), 0);
-  CHECK_EQ(opkt->PackREQID(rid), 0);
-  CHECK_EQ(opkt->PackProtocolBuffer(*_request, byte_size), 0);
+  OutgoingPacket opkt(sop);
```
nwangtw: Suggest to make the same change in SendResponse() and send SendMessage() in this file.

nicknezis:
@nwangtw It's been a while, but is this still something that should be merged in?

PR #2905, opened 2018-05-21
log the object when it cannot be serialized when emitting tuples

huijunwu:
We saw some topology may encounter the NotSerializableException and the heron-instance crashes. But the user may not be able to analyze why his tuple has NotSerializableException. This PR logs the tuple content to make the diagnosis easy.

### DIFF
--- a/heron/api/src/java/org/apache/heron/api/serializer/JavaSerializer.java
+++ b/heron/api/src/java/org/apache/heron/api/serializer/JavaSerializer.java
@@ -40,7 +40,7 @@ public class JavaSerializer implements IPluggableSerializer {
       oos.writeObject(object);
       oos.flush();
     } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


heron/api/src/java/org/apache/heron/api/serializer/JavaSerializer.java
```diff
@@ -40,7 +40,7 @@ public void initialize(Map<String, Object> config) {
       oos.writeObject(object);
       oos.flush();
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      throw new RuntimeException("NotSerializable object is: " + object.toString(), e);
```
nwangtw: Suggest: "Failed to serialize object: " + object.toString()

huijunw: updated

PR #2908, opened 2018-05-23
reorder operations in zk_client destructor

huijunwu:
fix https://github.com/apache/incubator-heron/issues/2907


maosongfu:
Had a discussion with Huijun, and met some new concerns. Hold it for now until we better understand it.

nwangtw:
Discussed offline. Update:

The fix is valid and helpful in the current implementation. The questions are about the implementation of queue and if we can handle zk events in a simpler way. So we agreed to submit the fix which could improve heron reliability, and then discuss about the zk client implementation in future.

In conclusion, we have a green light to merge~

PR #2910, opened 2018-05-30
make LocalFSStorage clean checkpoints before store

nlu90:
The current implementation of LocalFSStorage would cause unlimited checkpoints to be saved on local disk and thus out of disk space problem.

This PR fixes the problem by adding a limit of the number of checkpoints saved on local disk and clean unnecessary checkpoints every time before trying to store a new checkpoint. The most recent N checkpoints will be saved and any older checkpoints will be deleted.

Tested on local mac and worked as expected.

### DIFF
--- a/heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
+++ b/heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
@@ -20,6 +20,8 @@
 package org.apache.heron.statefulstorage.localfs;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Map;
 import java.util.logging.Logger;
 
@@ -32,20 +34,25 @@ import org.apache.heron.spi.statefulstorage.Checkpoint;
 import org.apache.heron.spi.statefulstorage.IStatefulStorage;
 import org.apache.heron.spi.statefulstorage.StatefulStorageException;
 
-public class LocalFileSystemStorage implements IStatefulStorage {
+public class  LocalFileSystemStorage implements IStatefulStorage {
   private static final Logger LOG = Logger.getLogger(LocalFileSystemStorage.class.getName());
 
   private static final String ROOT_PATH_KEY = "heron.statefulstorage.localfs.root.path";
+  private static final String MAX_CHECKPOINTS_KEY = "heron.statefulstorage.localfs.max.checkpoints";
 
   private String checkpointRootPath;
+  private int maxCheckpoints;
 
   @Override
   public void init(Map<String, Object> conf) throws StatefulStorageException {
     checkpointRootPath = (String) conf.get(ROOT_PATH_KEY);
+    maxCheckpoints = (int) conf.get(MAX_CHECKPOINTS_KEY);
     if (checkpointRootPath != null) {
-      checkpointRootPath = checkpointRootPath.replaceFirst("^~", System.getProperty("user.home"));
+      checkpointRootPath =
+          checkpointRootPath.replaceFirst("^~", System.getProperty("user.home"));
     }
-    LOG.info("Initialing... Checkpoint root path: " + checkpointRootPath);
+    LOG.info("Initialing LocalFileSystemStorage with Checkpoint root path: "
+        + checkpointRootPath + " Max checkpoints: " + maxCheckpoints);
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
```diff
@@ -32,20 +34,25 @@
...
   private static final String ROOT_PATH_KEY = "heron.statefulstorage.localfs.root.path";
+  private static final String MAX_CHECKPOINTS_KEY = "heron.statefulstorage.localfs.max.checkpoints";
 
   private String checkpointRootPath;
+  private int maxCheckpoints;
 
   @Override
   public void init(Map<String, Object> conf) throws StatefulStorageException {
     checkpointRootPath = (String) conf.get(ROOT_PATH_KEY);
+    maxCheckpoints = (int) conf.get(MAX_CHECKPOINTS_KEY);
```
nwangtw: could be helpful to have a default value?

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
```diff
@@ -134,6 +146,23 @@ public void dispose(String topologyName, String oldestCheckpointPreserved,
     }
   }
 
+  private void cleanCheckpoints(String path, int remaining) throws StatefulStorageException {
+    if (FileUtils.isDirectoryExists(path) && FileUtils.hasChildren(path)) {
+      File[] children = new File(path).listFiles();
```
nwangtw: suggest to keep it consistent with dispose(), which uses:

String[] names = new File(topologyCheckpointRoot).list();

Sorting might be cheaper this way as well.

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
```diff
@@ -134,6 +146,23 @@ public void dispose(String topologyName, String oldestCheckpointPreserved,
...
+      Arrays.sort(children, Comparator.reverseOrder());
```
nwangtw: I would suggest to break the logic into two steps:
1. generate the toDelete list.
2. for each one in the list, deleteDir().

The benefit is that you can write a unit test for step 1 easily to make sure the order is correct.

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
```diff
@@ -134,6 +146,23 @@ public void dispose(String topologyName, String oldestCheckpointPreserved,
...
+
+      // only keep the latest N remaining files, delete others
+      for (int i = remaining; i < children.length; i++) {
```
nwangtw: if we do:
for (int i = 0; i < children.length - remaining; ++i) {
}

we dont need Comparator, and the sorting might be slightly faster because the list might be likely to be in the same order already.

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
nlu90: done

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
nlu90: done

kramasamy:
@nlu90 - Do you do background cleaning of snapshots or on demand basis? It might be easier to do a background cleaning.

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
nlu90: Providing a default value in code might cause confusion. I prefer to have this value comes from conf file. It's cleaner and easier to reason.

nlu90:
@kramasamy 
1. On demand basis cleaning is not working in localfsstorage because the checkpoints are stored locally and they can't receive the delete requests from tmaser by design.

2. What do you mean by "backgroud cleaning"? Spawn another thread and keeps monitoring # of checkpoints? This sounds too overkill.

nlu90:
@kramasamy This change only needed by local filesystem storage. For other storage systems, no change is needed and the original design works fine.

kramasamy:
Ok got it.

heron/statefulstorages/src/java/org/apache/heron/statefulstorage/localfs/LocalFileSystemStorage.java
nlu90: Added the default value

PR #2914, opened 2018-06-07
Add permissive licence to libunwind

Code0x58:
I tried running tests with Bazel 0.14.0 and the lack of licence was a complaint, so I added a permissive marker as the package is from [LLVM](https://en.wikipedia.org/wiki/University_of_Illinois/NCSA_Open_Source_License).


nwangtw:
@kramasamy

PR #2920, opened 2018-06-17
Correct case of acronyms

Code0x58:
[Do you like dags](https://www.youtube.com/watch?v=ZXILzUpVx7A)?

**Edit**:
I had a quick look around and noticed some inconsistencies with other acronyms:

- [x] `API server` ← `api server`/`apiserver`
- [x] `API` ← `api`
- [x] `HDFS` ← `hdfs`
- [x] `CPU` ← `cpu`/`Cpu`
- [x] `RAM` ← `ram`/`Ram`
- [x] `S3` ← `s3`
- [x] `JVM` ← `jvm`

If there aren't any objections I'll go on a cleaning spree and pile them up on this branch.

There are still things like `heron` but I am not feeling masochistic enough to start that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,6 @@ website/public/
 
 # integration_test
 results/
+
+# Vagrant
+vagrant/.vagrant/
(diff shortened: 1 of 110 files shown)


kramasamy:
Sure go for it!

Code0x58:
@kramasamy yep, I think it's good to go now

PR #2921, opened 2018-06-18
Add JDK/JAVA_HOME to bazel_configure.py

Code0x58:
I found that `jar` wasn't available when I did `bazel test heron/...`, so I thought it would be good to check for the presence of a JDK in configure.

I don't know what version requirements there are, so I didn't add a check for that, but the version can be found from `javac -version`.


kramasamy:
@Code0x58 - there is no need to run bazel configure anymore. It can be used to check whether all the requirements are present.

PR #2922, opened 2018-06-18
Fix integer types to avoid warnings

Code0x58:
The output was a bit noisy because of signed/unsigned comparisons so I cleaned them up.


kramasamy:
@Code0x58 - thanks for the cleanup.

PR #2924, opened 2018-06-19
Py3 heronpy

Code0x58:
I am thinking this may be all it takes to get Python3 topologies running, as it builds a universal package which can be included in a `.pex`, so I imagine it will just work if the workers have the same interpreter as the `.pex` is built for; I'll go on and manually test uploading a Python3 `.pex` some time in the next few days.

This includes the `six` library which could be used to replace the `try`/`except` imports.

- [x] make heronpy a universal package
- [x] test a python3 PEX locally
- [ ] include python3.7 in the executor
- [ ] test a python3 PEX on Kubernetes
- [ ] update documentation
  - python3 support
  - python3 PEX generation
  - note about PEX pinning to a specific (major, minor) version


scripts/packages/BUILD
```diff
@@ -474,9 +474,11 @@ genrule(
         'rm setup.py.template',
         'tree $$HERONPY_DIR',
         '/usr/bin/env python2.7 setup.py sdist',
-        '/usr/bin/env python2.7 setup.py bdist_wheel',
+        '/usr/bin/env python2.7 setup.py bdist_wheel --universal',
+        'cd -',
         'ls -l $$HERONPY_DIR/dist',
-        'cp $$HERONPY_DIR/dist/heronpy-*-py2-*.whl $$OUTPUT_DIR',
+        'cp $$HERONPY_DIR/dist/heronpy-*-py2.py3-*.whl $$OUTPUT_DIR',
+        'cp $$HERONPY_DIR/dist/heronpy-*-py2.py3-*.whl "$@"',
```
Code0x58: This wasn't being done although it is supposed to be a product of this rule so I added it.

kramasamy: @Code0x58 - wondering if you had a chance to test this with py3 topologies?

Code0x58: I haven't quite got to running any PEXs yet, but should have done so by the end of this weekend

Code0x58:
Once this does work locally, it'll need support on the schedulers, which I think is to update the executor image to include Python3. I will look into making `.pex` files pin to more generic interpreters (`python3` vs. `python3.6`), as otherwise that will be a pain causing you have to rebuild your PEXs every time the executor image's python 3 minor version increases.

scripts/packages/BUILD
Code0x58: The universal package works locally, so the next step is to make sure it can run on the executor image. I'll get stuck into changing the Dockerfiles to include python3.6 over the next few days. My thinking is that they should include the latest stable python, so PEXs only need to have python3 in their shebang and they'll just work.

_p.s. the WIP commit is ugly, I pushed it to switch machines for a few days_

kramasamy:
Sounds good @Code0x58

Code0x58:
As an update, I was waiting for changes around `./docker/` to happen before I go changing things in there for the executors. I'll probably use the first official Apache release as my trigger to get back on this.

kramasamy:
@Code0x58 - the docker changes are merged I think.

aaronstjohn:
Just getting into heron and am excited about the prospect of python 3 support. It looks like you've made some progress on this. Anything I can do to help push it forward?

nicknezis:
@Code0x58 How does this PR relate to #3431 ? Is this still needed or can it be closed?

Code0x58:
It looks like that PR and its parent exclusively do python 2 to 3 changes on scripts, then set the python interpreter in those scripts to python 3. I did a very quick rebase and look around, and I'm not sure that the goals of this PR have been met by give given PR (or previous merges) - but they probably help a great deal and may unblock work on this.

I think around the time I lost steam on this there was a fairly big issue where you could only use one version of python with Bazel, which may still be the case, but if python3 is being used in the workspace then this could probably continue. I probably won't have time to get back into this project for a while, but if python3 is confirmed as the one being in use in the workspace then that would probably coax me in sooner.

nicknezis:
Maybe I can continue the effort and leverage you for help as needed? I haven't really used Python before. Currently working through #3438. Once I complete that I can take a look at this. Have tests failing due to "not proper python names".

Code0x58:
That sounds great to me! I think as long as you can publish a heronpy that works with python3 and build a python3 pex which you can run in the executor docker image, that's job done (especially if python2 support is dropped).

As a heads up, you may find that the PEX uses an overly specific interpreter (e.g. `python3.8` rather than `python3`) which will cause issues if you have a version mismatch or don't have the specific version on the PATH.

Code0x58:
Closing as superseded by #3522

PR #2928, opened 2018-06-20
Clean up website gen + website python docs

Code0x58:
* don't use sudo during `make site` to install pip packages
 * update website/config.yaml to reflect latest release
   - included newer bazel version as it seems to work for me
 * fix python topologies docs
   - fixed imports
   - tried to make code style more consistent
   - changed PEXes to PEXs


kramasamy:
@Code0x58 - could you update the conflicts so that we can merge it?

Code0x58:
This is failing for the same reason as master now, but should be good to merge

Code0x58:
I pulled in #2937 so the tests pass, so this should probably be merged after it.

PR #2929, opened 2018-06-20
Make TopologyBuilder less permissive

Code0x58:
These jumped out at me as I was flicking through the code.

### DIFF
--- a/heronpy/api/topology.py
+++ b/heronpy/api/topology.py
@@ -336,6 +336,7 @@ class TopologyBuilder(object):
 
     self.topology_name = name
 
+    # TODO: make this a dict mapping from spec.name to spec
     self._specs = []
     self._topology_config = {}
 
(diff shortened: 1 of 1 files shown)


heronpy/api/topology.py
```diff
@@ -194,24 +194,25 @@ def get_heron_options_from_env():
...
 
-    ret = {}
-    heron_opt_list = heron_options_raw.replace("%%%%", " ").split(',')
-    for opt_raw in heron_opt_list:
-      opt = opt_raw.split("=")
-      if len(opt) == 2:
-        ret[opt[0]] = opt[1]
-    return ret
+    options = {}
+    for option_line in heron_options_raw.replace("%%%%", " ").split(','):
```
Code0x58: This feels a bit odd, I haven't seen the format elsewhere - is it used in Java? Maybe JSON would be better as it is just `os.environ["HERON_OPTIONS"] = json.dumps(config)` and `json.loads(os.envrion["HERON_OPTIONS"])` with `KeyError` and `ValueError` handling around the latter.

kramasamy: It might be worth trying and if it works - we can use json dumps and load.

Code0x58: I think I'd have to get stuck into at least all of the client libraries, heron-executor, and tests, which is a bit too much for me at the moment. I don't mind making an issue for it though.

nwangtw: it does look strange. From the code it seems this is used to support space in path. And the %%%% can be found in cpp, java and python with code search.

PR #2934, opened 2018-06-22
[aurora] add metadata support for aurora

huijunwu:
Motivation: In some scenarios, users would like to pass some metadata or data to the Aurora job. 

Approach: Users feed the metadata or data through `--config-property heron.scheduler.properties={aurora_config},{meta1},{meta2}...`. 'heron.aurora' picks up these config by `{{AURORA_METADATA_x}}`. All code changes are within Aurora scheduler scope

Example:
```
./bin/heron submit --verbose \
--config-property heron.scheduler.properties=heron2.aurora,huijunw,heron-examples,latest \
cluster/huijunwu/test ./examples/heron-api-examples.jar \
org.apache.heron.examples.api.WordCountTopology WordCountTopology
```


heron/schedulers/src/java/org/apache/heron/scheduler/aurora/AuroraField.java
```diff
@@ -35,5 +35,8 @@
   TIER,
   TOPOLOGY_ARGUMENTS,
   TOPOLOGY_NAME,
-  TOPOLOGY_PACKAGE_URI
+  TOPOLOGY_PACKAGE_URI,
+  AURORA_METADATA_1,
```
nlu90: 1. Why three such aurora configs, not less or more? 

2. Is it possible to assign them meaningful names instead of 1,2,3?

nwangtw: Agreed.

Also, is it possible to have a more general metadata argument like "SCHEDULE_META" so that this is for all schedulers? It is not ideal to have AURORA specific configs in heron/spi/common/Key.java to me.

Code0x58: _just speculating:_ the `SCHEDULER_PROPERTIES     ("heron.scheduler.properties", Type.PROPERTIES),` line in `Key.java` looks interesting:
 * does `Type.PROPERTIES` allow you do take a set/map of things instead of a string?
 * can `SCHEDULER_PROPERTIES` be used instead of adding these new options?

kramasamy: Let us use one parameter and comma separated. heron.scheduler.properties is a good one.

huijunwu:
put it on hold. lets find better way for aurora meta

heron/schedulers/src/java/org/apache/heron/scheduler/aurora/AuroraField.java
huijunwu: updated according to @kramasamy suggestion

huijunwu:
updated according to Karthik suggestion 'Let us use one parameter and comma separated. heron.scheduler.properties is a good one.'

heron/schedulers/src/java/org/apache/heron/scheduler/aurora/AuroraScheduler.java
```diff
@@ -279,6 +279,19 @@ public void removeContainers(Set<PackingPlan.ContainerPlan> containersToRemove)
     auroraProperties.put(AuroraField.CORE_PACKAGE_URI, heronCoreReleasePkgURI);
     auroraProperties.put(AuroraField.TOPOLOGY_PACKAGE_URI, topologyPkgURI);
 
+    if (config.containsKey(Key.SCHEDULER_PROPERTIES)) {
```
nwangtw: Still not idea to have AURORA_METADATA_1, AURORA_METADATA_2, and AURORA_METADATA_3.

I am thinking maybe in addition to auroraProperties, which is for standard properties with type Map[AuroraField, String], we can add another Map[String, String] extraProperties for these dynamic properties, so that we can avoid the hard coded names.

huijunwu: if i understand correctly, the new 'Map[AuroraField, String]' will be added in the method param list https://github.com/apache/incubator-heron/blob/0.17.8/heron/schedulers/src/java/com/twitter/heron/scheduler/aurora/AuroraController.java#L26. Thoughts to update the interface? @nlu90 @maosongfu

nwangtw: The new map would be the new 'Map[String, String]'. No need to add to param list I think since the values are generated from existing arguments, but I might be wrong.

nwangtw: nvm. i see it now.

huijunwu: after discussion, i am going to update the method to `boolean createJob(Map<AuroraField, String> auroraProperties, Map<String, String> extra);`

PR #2936, opened 2018-06-23
add integration topology test

yaoliclshlmch:
This test aims to test the launched topology structure correctness. The previous integration tests are focusing on the data correctness by fetching the real output data and comparing it with the expected output results, which is functional black box testing. For the topology structure test, white box testing should be used which means the test platform is aware of the internal topology structures. 

Topology structure test here includes the following aspects:
1. Topology graph
        a. components: spout type, bolt type and their parallelism;
        b. connectivity: data stream and connection correctness;
2. Instance states (only in stateful processing): loaded states in target rollback checkpoint.

For now, we focus on the implementation of topology component and connectivity, and the instance state test will be realized in the future.

Test cases:
1. Topology submission
3. Spouts/bolts parallelism scale up
4. Spouts/bolts parallelism scale down
5. Spouts/bolts parallelism scale up and down
2. Topology deactivation and activation
6. container failure and recovery

Running it locally:
scripts/run_integration_topology_test.sh


kramasamy:
@yaoliclshlmch - once scala PR gets merged, we should try to rebuild this.

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -262,6 +268,44 @@ void TController::HandleUpdateRuntimeConfigRequestDone(IncomingHTTPRequest* requ
...
+    http_server_->SendErrorReply(request, 400);
+    delete request;
+    return;
+  } else {
+    std::string pplanString;
+    tmaster_->getPhysicalPlan()->SerializeToString(&pplanString);
+    std::string pplanStringFixed = "";
+
+    // convert original pplan string to a new one which do not contains b'00000000
+    char const hex_chars[16] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B',
```
nwangtw: Please move this encoding code to its own function and add a unit test for it.

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -262,6 +268,44 @@ void TController::HandleUpdateRuntimeConfigRequestDone(IncomingHTTPRequest* requ
...
+  if (tmaster_->getPhysicalPlan() == NULL) {
+    http_server_->SendErrorReply(request, 400);
+    delete request;
+    return;
+  } else {
+    std::string pplanString;
+    tmaster_->getPhysicalPlan()->SerializeToString(&pplanString);
+    std::string pplanStringFixed = "";
+
+    // convert original pplan string to a new one which do not contains b'00000000
```
nwangtw: suggest changing the comment to:

SerializeToString() returns object in binary format which needs to be encoded.

integration_test/src/java/org/apache/heron/integration_topology_test/common/AbstractTestTopology.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  protected Options getArgOptions() {
+    Options options = new Options();
+
+    Option topologyNameOption = new Option("t", TOPOLOGY_OPTION, true, "topology name");
+    topologyNameOption.setRequired(true);
+    options.addOption(topologyNameOption);
+
+    // Option stateTokenOption = new Option("u", STATE_UPDATE_TOKEN, true,
```
nwangtw: to clean up?

scripts/run_integration_topology_test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# integration test binaries have to be specified as absolute path
+JAVA_INTEGRATION_TESTS_BIN="${PWD}/bazel-genfiles/integration_test/src/java/integration-topology-tests.jar"
+PYTHON_INTEGRATION_TESTS_BIN="${PWD}/bazel-bin/integration_test/src/python/integration_test/topology/heron_integ_topology.pex"
+SCALA_INTEGRATION_TESTS_BIN="${PWD}/bazel-genfiles/integration_test/src/scala/scala-integration-tests.jar"
+
+CORE_PKG="file://${PWD}/bazel-bin/scripts/packages/heron-core.tar.gz"
+
+set -e
+bazel build --config=darwin integration_test/src/...
```
nwangtw: replace darwin with `platform`

scripts/run_integration_topology_test.sh
```diff
@@ -0,0 +1,51 @@
...
+# building tar packages
+DIR=`dirname $0`
+source ${DIR}/detect_os_type.sh
+bazel run --config=`platform` -- scripts/packages:heron-install.sh --user
+bazel build --config=`platform` {heron/...,scripts/packages:tarpkgs,integration_test/src/...}
+
+# run the simple http server
```
nwangtw: are we going to need these lines?

Issue #2925, opened 2018-06-19
Flaky Java integration test


Code0x58:
This has failed a few times at least recently, and is why the current commit on master is red.

[/integration_test/src/java/org/apache/heron/integration_test/topology/non_grouping/NonGrouping.java](https://github.com/apache/incubator-heron/blob/0138b8d451c988f53e431f951f9a6fb9f2f4c116/integration_test/src/java/org/apache/heron/integration_test/topology/non_grouping/NonGrouping.java)
Example failure: [build](https://travis-ci.org/apache/incubator-heron/builds/393993192?utm_source=github_status&utm_medium=notification)
```
[2018-06-19 09:44:54 +0000] [INFO]: Fetching results for topology 20180619094300_IntegrationTest_NonGrouping_93bd575e-d7f0-4fb1-96d2-4b8bb7e4cd1f, retry count: 5
[2018-06-19 09:44:54 +0000] [INFO]: 200 GET /results/20180619094300_IntegrationTest_NonGrouping_93bd575e-d7f0-4fb1-96d2-4b8bb7e4cd1f (127.0.0.1) 0.99ms
[2018-06-19 09:44:54 +0000] [ERROR]: Actual result did not match expected result
[2018-06-19 09:44:54 +0000] [INFO]: Actual result ---------- 
['A', 'A', 'A', 'A', 'A', 'A', 'A', 'B', 'B', 'B', 'B', 'B', 'B', 'B']
[2018-06-19 09:44:54 +0000] [INFO]: Expected result ---------- 
['A', 'A', 'A', 'A', 'A', 'B', 'B', 'B', 'B', 'B']
[2018-06-19 09:44:54 +0000] [ERROR]: Checking result failed for 20180619094300_IntegrationTest_NonGrouping_93bd575e-d7f0-4fb1-96d2-4b8bb7e4cd1f topology :: Traceback (most recent call last):
  File "integration_test/src/python/test_runner/main.py", line 184, in run_test
    return results_checker.check_results()
  File "integration_test/src/python/test_runner/main.py", line 103, in check_results
    return self._compare(expected_results, actual_results)
  File "integration_test/src/python/test_runner/main.py", line 116, in _compare
    raise failure
TestFailure: ('Actual result did not match expected result', None)
```
[...]

PR #2937, opened 2018-06-24
Use latest scala rules to fix tests

Code0x58:
This is for https://github.com/bazelbuild/rules_scala/issues/532 which is breaking the builds at the moment.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -957,15 +957,17 @@ new_http_archive(
 )
 
 # scala integration
-rules_scala_version="5cdae2f034581a05e23c3473613b409de5978833" # update this as needed
+rules_scala_version="669ed8750b77fd99b758298f6467aaa5e6a9dabb" # update this as needed
 
 http_archive(
     name = "io_bazel_rules_scala",
-    url = "https://github.com/bazelbuild/rules_scala/archive/%s.zip"%rules_scala_version,
+    url = "https://github.com/bazelbuild/rules_scala/archive/%s.zip" % rules_scala_version,
     type = "zip",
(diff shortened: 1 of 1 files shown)


Code0x58:
It seems like the scala rules have changed a bit since the last pinned commit, maybe to include toolchains. Hopefully the SSL cert issue is fixed very soon so this doesn't have to be looked at again very soon, as I don't have time to get into this more.

erenavsarogullari:
Hi @Code0x58,

Firstly, thanks to having a look this.

It looks like _no matching toolchains found_ from latest travis log:
```
(18:43:50) ERROR: While resolving toolchains for target //heron/api/src/scala:api-scala: no matching toolchains found for types @io_bazel_rules_scala//scala:toolchain_type
(18:43:51) ERROR: Analysis of target '//heron/api/src/scala:api-scala' failed; build aborted: no matching toolchains found for types @io_bazel_rules_scala//scala:toolchain_type
```

From bazel doc, we can try setting default toolchain if no concern.
```
# WORKSPACE
# register default scala toolchain
load("@io_bazel_rules_scala//scala:toolchains.bzl", "scala_register_toolchains")
scala_register_toolchains()
```
Ref: https://github.com/bazelbuild/rules_scala/blob/master/README.md

kramasamy:
@Code0x58 @erenavsarogullari - is this good to go?

Code0x58:
Yep, I think this was merged as a part of #2928 though

Code0x58:
Yep, this is in. The build failure on master is down to #2925 now

PR #2938, opened 2018-06-26
add integration topology test

yaoliclshlmch:
This request is related to #2936, which was accidentally closed.


heron/common/src/cpp/basics/strutils.cpp
```diff
@@ -44,3 +44,38 @@ StrUtils::split(
...
+  }
+  return output;
+}
+
+std::vector<char> StrUtils::decode(const std::vector<char>& _input) {
+  std::vector<char> output;
+  int i = 0;
+  while (i < _input.size()) {
+    char chr_1 = _input[i];
+    if (chr_1 <= 57) {
```
nwangtw: Give 57, 48 and 55 const variables to be more readable.

heron/common/src/cpp/basics/strutils.cpp
```diff
@@ -44,3 +44,58 @@ StrUtils::split(
...
+}
+
+// Return decoded value in [0, 15]; return -1 if the character is illegal.
+int StrUtils::decodeHexChar(char c) {
+  const int char_zero = 48;
+  const int char_nine = 57;
+  const int char_A = 65;
+  const int char_F = 70;
+
+  if (c < char_zero) {
```
nlu90: could you simplify it to `if (0 to 9) ... else if (A to F) ... else return -1`?

scripts/run_integration_topology_test.sh
```diff
@@ -0,0 +1,51 @@
...
+set -e
+
+# building tar packages
+DIR=`dirname $0`
+source ${DIR}/detect_os_type.sh
+bazel run --config=`platform` -- scripts/packages:heron-install.sh --user
+bazel build --config=`platform` {heron/...,scripts/packages:tarpkgs,integration_test/src/...}
+
+# run the simple http server
+#${HTTP_SERVER} 8080 &
```
nlu90: any specific reason we keep the HTTP server, scala test,  python test commented out?

yaoliclshlmch: Http server will be used in instance state validation in stateful processing, and scala and python tests may be added in the future.

huijunwu:
ship after ci pass

PR #2941, opened 2018-06-27
add tmaster physical plan endpoint

yaoliclshlmch:


yaoliclshlmch:
Yes. It is weird this part is missing. Let me fix it.

heron/tmaster/src/cpp/manager/tcontroller.cpp
```diff
@@ -262,6 +269,38 @@ void TController::HandleUpdateRuntimeConfigRequestDone(IncomingHTTPRequest* requ
...
+
+    // SerializeToString() returns object in binary format which needs to be encoded
+    const unsigned char * encodeString = (unsigned char *)pplanString.c_str();
+    std::string pplanStringFixed = cereal::base64::encode(encodeString, pplanString.size());
+
+    const std::string message("Get current physical plan");
+    LOG(INFO) << message;
+    OutgoingHTTPResponse* response = new OutgoingHTTPResponse(request);
+    response->AddResponse(pplanStringFixed);
+    http_server_->SendReply(request, 200, response);
```
nlu90: In the `else {...}` case, `request` is not deleted. This will lead to memory leak issue.

yaoliclshlmch: Good point! Thanks! Just fixed it.

PR #2948, opened 2018-07-03
Universal heronpy

Code0x58:
This is the first part of #2924, it makes the heronpy library python3 compatable. I will up the compatible version to 3.7 in the other PR once I run it.

I am submitting this as a separate PR as it will probably be a couple of weeks before I get time to continue on the other PR, and there is some interest in using heronpy with version 3 at least locally.


kramasamy:
@Code0x58 - great. this is a needed.

scripts/packages/BUILD
```diff
@@ -474,9 +474,11 @@ genrule(
         'rm setup.py.template',
         'tree $$HERONPY_DIR',
         '/usr/bin/env python2.7 setup.py sdist',
-        '/usr/bin/env python2.7 setup.py bdist_wheel',
+        '/usr/bin/env python2.7 setup.py bdist_wheel --universal',
+        'cd -',
         'ls -l $$HERONPY_DIR/dist',
-        'cp $$HERONPY_DIR/dist/heronpy-*-py2-*.whl $$OUTPUT_DIR',
+        'cp $$HERONPY_DIR/dist/heronpy-*-py2.py3-*.whl $$OUTPUT_DIR',
+        'cp $$HERONPY_DIR/dist/heronpy-*-py2.py3-*.whl "$@"',
```
kramasamy: Why this extra copy?

Code0x58: This is because `"$@"` is the defined output of the rule, but it wasn't actually generated. I didn't change the existing one in case anything is depending/expecting on it

kramasamy: Ok. thanks.

Issue #2947, opened 2018-07-03
Stream manager not stopped gracefully


Glorfischi:
Note: I looked into the stream manager running on a local cluster, but as far as I can tell other parts of Heron and other kinds of clusters may also be affected.

When stopping a topology (`heron kill`) on a local cluster, the local scheduler sends a SIGTERM signal to all heron executors. The executor handles the signal and in turn sends a SIGTERM to all heron components running in its container. 

The stream manager however does not register a handler to handle this SIGTERM, which means it will be killed without calling its destructors or any cleanup code. This seems unintuitive and can lead to unexpected behavior . 

In my case, I tried to use the stream managers destructor to export additional latency metrics I collected and noticed that the destructor is never called. 

I suggest handling SIGTERM signals in the stream manager and exit gracefully. The eventlib used in the event loop can handle the signal and there is already a function in place that stops the event loop (`EventLoopImpl::loopExit()`), so the fix is quite simple and as far as I can tell should not have any side effects. 

@antiguru

kramasamy:
@Glorfischi - it will be great if you could send a quick PR.

antiguru:
See #2950 for the corresponding pull request.

PR #2950, opened 2018-07-04
Add signal handling to stream manager

Glorfischi:
This adds signal handling to the `EventLoopImpl` in order to add a SIGTERM handler to the stream manager. (See #2947)

I tried to handle signal events similarly to the other events.  This resulted in quite a lot of boilerplate code, I am not completely sure this is the right way to go.

### DIFF
--- a/heron/common/src/cpp/network/event_loop_impl.cpp
+++ b/heron/common/src/cpp/network/event_loop_impl.cpp
@@ -37,6 +37,12 @@ void EventLoopImpl::eventLoopImplWriteCallback(sp_int32 fd, sp_int16 event, void
   el->handleWriteCallback(fd, event);
 }
 
+// 'C' style callback for libevent on signal events
+void EventLoopImpl::eventLoopImplSignalCallback(sp_int32 sig, sp_int16 event, void* arg) {
+  auto* el = reinterpret_cast<EventLoopImpl*>(arg);
+  el->handleSignalCallback(sig, event);
+}
+
 // 'C' style callback for libevent on timer events
 void EventLoopImpl::eventLoopImplTimerCallback(sp_int32, sp_int16 event, void* arg) {
   // TODO(vikasr): this needs to change to VCallback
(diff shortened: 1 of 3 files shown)


heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -83,6 +89,39 @@ void EventLoopImpl::loop() {
...
+  // Now add it to the list of signals monitored by the mDispatcher
+  if (event_add(event->event(), NULL) < 0) {
+    delete event;
+    throw heron::error::Error_Exception(errno);
+  }
+  mSignalEvents[sig] = event;
+  return 0;
+}
+
+int EventLoopImpl::unRegisterSignal(int sig) {
```
nwangtw: where is this function hooked up?

heron/stmgr/src/cpp/server/stmgr-main.cpp
```diff
@@ -48,6 +48,12 @@ DEFINE_string(ckptmgr_id, "", "The id of the local ckptmgr");
 DEFINE_int32(ckptmgr_port, 0, "The port of the local ckptmgr");
 DEFINE_string(metricscachemgr_mode, "disabled", "MetricsCacheMgr mode, default `disabled`");
 
+EventLoopImpl ss;
+
+void sigtermHandler(int signum) {
+  ss.loopExit();
```
nwangtw: I am a bit concerned that if loopExit() is safe to call here.
- which thread it is running on and if there might be racing condition.
- if it is possible that loopExit() might get stuck.

heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -83,6 +89,39 @@ void EventLoopImpl::loop() {
...
+  if (mSignalEvents.find(sig) == mSignalEvents.end()) {
+    // This signal wasn't registed.  Hence we can't unregister it.
+    return -1;
+  }
+
+  // Delete the underlying event in libevent
+  if (event_del(mSignalEvents[sig]->event()) != 0) {
+    throw heron::error::Error_Exception(errno);
+  }
+  delete mSignalEvents[sig];
```
nwangtw: safer way is to use a temp var to make sure the object is erased from the map before deleting it.

heron/common/src/cpp/network/event_loop_impl.cpp
Glorfischi: This function is not actually used yet. I included it, because other events also have a unregister function and it might be useful sometime in the future.

I can also remove it.

heron/stmgr/src/cpp/server/stmgr-main.cpp
Glorfischi: I think the sigtermHandler should run the in the same thread as any other callback function and loopExit() directly calls event_base_loopbreak() which I think should be safe to call. [Here](http://www.wangafu.net/~nickm/libevent-book/Ref3_eventloop.html) in the libevent book they call event_base_loopbreak() in a very similar way.

But maybe someone with more experience with livevent could give some input on that.

heron/common/src/cpp/network/event_loop_impl.cpp
nwangtw: kk. It is ok to leave it there to me. Let's see if @srkukarni has any input.

kramasamy:
@nwangtw @Glorfischi - is this ready to be merged?

kramasamy:
@nwangtw - can you please check if this is ok to merge it? @Glorfischi - have you fixed all the feedback.

Glorfischi:
Sorry for the delay.

I have a fix ready for one of the feedback, but I'll wait for a response before I push it.  See the comment on the `unRegister*` functions. 

As for the `ss.loopExit()` call. I think this should be fine, but I hoped someone with more experience with libevent could give the ok.

heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -83,6 +89,40 @@ void EventLoopImpl::loop() {
...
+  if (mSignalEvents.find(sig) == mSignalEvents.end()) {
+    // This signal wasn't registed.  Hence we can't unregister it.
+    return -1;
+  }
+
+  // Delete the underlying event in libevent
+  if (event_del(mSignalEvents[sig]->event()) != 0) {
+    throw heron::error::Error_Exception(errno);
+  }
+  auto* event = mSignalEvents[sig];
```
kramasamy: do you need an auto* ? just auto event will be fine, isn't?

heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -136,8 +176,9 @@ int EventLoopImpl::unRegisterForRead(int fd) {
     // cout << "event_del failed for fd " << fd;
     throw heron::error::Error_Exception(errno);
   }
-  delete mReadEvents[fd];
+  auto* event = mReadEvents[fd];
```
kramasamy: same here and other places.

heron/common/src/cpp/network/event_loop_impl.cpp
Glorfischi: oh yes I think you're right. That * is useless.

PR #2951, opened 2018-07-04
Counting pending tuples

kalimfaria:
This commit allows us to count the all the tuples received in "HeronTupleSet2* _message" to maintain an accurate count of the number of tuples received by the stream manager per instance.


heron/proto/stmgr.proto
```diff
@@ -67,4 +67,6 @@ message TupleStreamMessage {
   required int32 task_id = 1;
   // serialized data
   required bytes set = 2;
+  // number of tuples in serialized data
+    required int32 num_tuples = 4;
```
huijunw: need @nwangtw @nlu90 comment on this
if i remember correctly, there was tuple count without this field. how did they count it? @nlu90

integration_test/README.md
```diff
@@ -29,8 +29,8 @@ the heron client:
 
 ```Bash
   bazel build --config=darwin integration_test/src/...
-  bazel run --config=darwin -- scripts/packages:heron-client-install.sh --user
-  bazel run --config=darwin -- scripts/packages:heron-tools-install.sh --user
+  bazel run --config=darwin -- scripts/packages:heron-install.sh --user
+  bazel run --config=darwin -- scripts/packages:heron-tests-tools-install.sh --user
```
huijunw: this change seems irrelevant to this pr?

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -413,15 +423,16 @@ void InstanceServer::DrainTupleStream(proto::stmgr::TupleStreamMessage* _message
...
 
-  if ( it != connection_buffer_length_metric_map_.end() )
-    connection_buffer_length_metric_map_
-      [instance_id]->scope("packets")->incr();
+    ConnectionBufferLengthMetricMap::const_iterator it =
+      connection_buffer_length_metric_map_.find(instance_id);
 
+    if ( it != connection_buffer_length_metric_map_.end() )
+      connection_buffer_length_metric_map_
+        [instance_id]->scope("packets")->incr_by(_message->data().tuples_size());
```
huijunw: do you use the same scope 'packets' but by two dimensions, one of num_tuples(), the other of tuples_size()? if i understand correctly, you are counting both count and size under one name?

kalimfaria: The tuples field `HeronTupleSet2` is actually a repeated field. So if I call the function `tuples_size()`, I am getting the size of the tuple array. On the other hand, the num_tuples field in `TupleStreamMessage` is just an integer that tells me how many tuples are in the serialized set field. So both functions just give me number of tuples and not size.

integration_test/README.md
kalimfaria: You're right, it's not relevant. I'll remove it.

heron/proto/stmgr.proto
nlu90: I think what you are referring is here: https://github.com/apache/incubator-heron/blob/571a90b7c8452f7922cdc47b91e52de66f5edd70/heron/stmgr/src/cpp/manager/instance-server.cpp#L446

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -74,7 +74,10 @@ const sp_string METRIC_TIME_SPENT_BACK_PRESSURE_AGGR = "__server/__time_spent_ba
 const sp_string METRIC_TIME_SPENT_BACK_PRESSURE_COMPID = "__time_spent_back_pressure_by_compid/";
 // Prefix for connection buffer's metrics
 const sp_string CONNECTION_BUFFER_BY_INSTANCEID = "__connection_buffer_by_instanceid/";
-// Prefix for connection buffer's length metrics
+// Prefix for connection buffer's length metrics. This is different
+// from METRIC_DATA_TUPLES_TO_INSTANCES as that counts
+// the tuples when they are sent to the instance -- this metric
+// will be used to count the tuples as they are received
```
nlu90: Why do we need such a separate metric to count tuples when they are received?  In effect, the sent tuples should be equal to received tuples. If the existing one can server the purpose, I suggest to use it directly instead of adding new metrics.

heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
```diff
@@ -141,6 +141,14 @@ bool StMgrClientMgr::SendTupleStreamMessage(sp_int32 _task_id, const sp_string&
   out = __global_protobuf_pool_acquire__(out);
   out->set_task_id(_task_id);
   out->set_src_task_id(_msg.src_task_id());
+  sp_int32 length = 0;
+  if (_msg.has_data()) {
```
nlu90: The `length` here represents both data tuple size and control tuple size. The mixing might cause confusions.

heron/proto/stmgr.proto
kalimfaria: It is not necessary that every `TupleStreamMessage` be converted into a `HeronTupleSet2` before it is added to the pending queue. In that case, the tuplestreammessage is processed separately (https://github.com/apache/incubator-heron/blob/571a90b7c8452f7922cdc47b91e52de66f5edd70/heron/stmgr/src/cpp/manager/instance-server.cpp#L402) and will not traverse through the code you have linked.

heron/stmgr/src/cpp/manager/instance-server.cpp
kalimfaria: Would the two not be different in case of backpressure?

nlu90: In the case of backpressure, only the spouts stop reading from the external data source(e.g, kafka stream). 

All the internal tuples are still sent and processed. 

Could you elaborate why they are different in such case?

heron/stmgr/src/cpp/manager/stmgr-clientmgr.cpp
kalimfaria: An alternate design would be to have separate optional length fields for control and data. I can do that or I can use the current field for data only.

heron/stmgr/src/cpp/manager/instance-server.cpp
kalimfaria: From what I understand, tuples can be dropped when they are being removed from the queue because of backpressure.
https://github.com/apache/incubator-heron/blob/571a90b7c8452f7922cdc47b91e52de66f5edd70/heron/stmgr/src/cpp/manager/instance-server.cpp#L432
In addition, I'd like to measure arrival rate of tuples at the stream manager and I think that because tuples are only drained when a certain number of tuples have arrived, that may not translate accurately?

kramasamy:
@nlu90 @kalimfaria - is this ready to be merged?

huijunwu:
ship after ci pass

PR #2953, opened 2018-07-05
Nwang/fix negative index caused by hash mod

nwangtw:
"hashCode() % taskIds.size()" can be negative which could cause "out of boundary" error. This has been handled in simulator/grouping/FieldsGrouping.java but a few others need the fix as well.


nwangtw:
Can you guys help to merge this PR? I dont have the permission yet. Thanks in advance.

PR #2956, opened 2018-07-10
add integration test for instance state in stateful processing

yaoliclshlmch:
This is a follow up of #2936  and #2938 .

Current white box topology structure integration test can check the DAG structure correctness. This PR aims to enable instance state check after checkpoint rollback in stateful processing. 

Test cases (following #2936 ):
7. Checkpoint rollback in stateful processing triggered by instance failure.

Running it locally:
scripts/run_integration_topology_test.sh


integration_test/src/java/org/apache/heron/integration_topology_test/common/AbstractTestTopology.java
```diff
@@ -82,12 +102,30 @@ public final void submit() throws AlreadyAliveException, InvalidTopologyExceptio
   }
 
   public final void submit(Config userConf) throws AlreadyAliveException, InvalidTopologyException {
-    TopologyBuilder builder = new TopologyBuilder();
-
-    Config conf = buildConfig(new BasicConfig());
-    if (userConf != null) {
-      conf.putAll(userConf);
+    if (this.httpServerResultsUrl == null) {
```
nlu90: this `if ... else ...` clause shares several common code, let's reuse the duplicate part.

integration_test/src/java/org/apache/heron/integration_topology_test/core/IStatefulBolt.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.apache.heron.api.bolt.BaseRichBolt;
+import org.apache.heron.api.bolt.OutputCollector;
+import org.apache.heron.api.state.State;
+import org.apache.heron.api.topology.IStatefulComponent;
+import org.apache.heron.api.topology.OutputFieldsDeclarer;
+import org.apache.heron.api.topology.TopologyContext;
+import org.apache.heron.api.tuple.Tuple;
+
+public class IStatefulBolt extends BaseRichBolt implements IStatefulComponent<String, Integer> {
```
nlu90: in heron's convention, the `I` prefix is used to indicate the type is an interface.

If it's not an interface, we can remove the `I` prefix.

integration_test/src/java/org/apache/heron/integration_topology_test/core/IStatefulSpout.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+
+import org.apache.heron.api.spout.BaseRichSpout;
+import org.apache.heron.api.spout.SpoutOutputCollector;
+import org.apache.heron.api.state.State;
+import org.apache.heron.api.topology.IStatefulComponent;
+import org.apache.heron.api.topology.OutputFieldsDeclarer;
+import org.apache.heron.api.topology.TopologyContext;
+
+public class IStatefulSpout extends BaseRichSpout implements IStatefulComponent<String, Integer> {
```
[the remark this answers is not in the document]
nlu90: In heron's convention, the `I` prefix means that it's an interface. 

If it's not, we can remove the `I` prefix.

integration_test/src/python/topology_test_runner/resources/test.json
```diff
@@ -4,49 +4,56 @@
   "role" : "dummy_role",
   "env" : "dummy_env",
   "cliConfigPath" : "$HOME/.heron/conf",
-  "processingType" : "non_stateful",
+  "processingType" : "stateful",
```
nlu90: With the changes to this file, the normal topologies aren't tested?

integration_test/src/java/org/apache/heron/integration_topology_test/core/IStatefulSpout.java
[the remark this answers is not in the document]
yaoliclshlmch: I prefix removed.

integration_test/src/python/topology_test_runner/resources/test.json
yaoliclshlmch: No. This will only enable instance state check in stateful topologies, but won't disable DAG structure check for both normal and stateful topologies.

yaoliclshlmch:
@huijunwu 
_"The IdentityBolt ABSpout IntegrationTopologyTestBolt IntegrationTopologyTestSpout
look similar to classes in inregation_test directory. is there a way to reuse code"_
It is a little bit difficult, because they inherit from different class/interface and have different processing logic.

integration_test/src/java/org/apache/heron/integration_topology_test/common/AbstractTestTopology.java
yaoliclshlmch: Thanks! Fixed.

maosongfu:
+1 to Neng.
Normally, in the description we will:
1. Current status/Problem statement (Why we need this pull request)
2. What this pull request does 
3. How will the pull request solve the problem (, and how the new status will look like)
4. Test coverage? etc

Thanks

PR #2960, opened 2018-07-12
joshfischer/update k8s docs

joshfischer1108:
@kramasamy @cckellogg  This pr corrects the Kubernetes proxy urls for version K8's v1.10.0.  It also adds a few explicit disclaimers stating the version of Kubernetes they were verified with.  Please review.


cckellogg:
should we update all the tools templates to the new proxy urls too?
https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/helm/templates/tools.yaml
https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/minikube/tools.yaml
https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/general/tools.yaml

joshfischer1108:
@cckellogg Yes we should and glad you mentioned them.  It totally slipped my mind.  Just pushed them up.

joshfischer1108:
Also found a file at `heron/deploy/kubernetes/helm/templates/NOTES.txt` that needed some changes.  Not sure if that's used to dump to the console or not, so I updated the values to be sure.  Also, integration tests have failed on me three times in a row.  Are there still some race conditions that could be causing this?

PR #2961, opened 2018-07-12
package c++ .h and .a files

kramasamy:
for c++ topology development


scripts/packages/BUILD
```diff
@@ -2,38 +2,6 @@ package(default_visibility = ["//visibility:public"])
 
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")
 
-load("//tools/rules:heron_client.bzl", "heron_client_bin_files")
```
Code0x58: it looks like these rules could be removed as well

PR #2962, opened 2018-07-12
[WIP] Enable instances and ckptmgr transferring state via local disk

nlu90:
During our test, we found that if a state is too large (>~64MB), the original checkpointing mechanism fails. The actual state will be packed in the protobuf message and transmitted through the network. Once it's large enough, the receiver will have problems parsing the IncomingPacket, which causes the connection to be closed. And it fails the whole checkpointing process.

This PR solves this problem by introducing the state spilling feature. With the feature turned on, the actual state will be spilled to local disk and only the state location information is transmitted through the network. This ensures the checkpointing works for even very large states.

Tested with internal topologies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,9 @@ env:
 
 before_install:
   # download and install bazel
-  - wget -q 'https://github.com/bazelbuild/bazel/releases/download/0.9.0/bazel-0.9.0-installer-linux-x86_64.sh'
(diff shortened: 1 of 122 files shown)


nlu90:
Closing this PR due to confusing commits and code not cleaned up. 
A new PR will be posted for this feature with better commit information and clean code.

PR #2966, opened 2018-07-16
Fix supervisor installation in Centos 7 Dockerfile

### LINKED ISSUE #2965: Dockerfile for Centos 7 tries to use apt-get
`docker/dist/Dockerfile.dist.centos7` includes the following line:

`RUN apt-get -y install supervisor; apt-get clean all`

which causes the Docker containers not to build.

thomas4g:
Fixes #2965

I tried the simpler `yum install supervisor` but the configured repos don't contain it. This is the simplest solution I found.


kramasamy:
@thomas4g - do we have to do for other docker containers as well?

thomas4g:
@kramasamy All the other containers are Debian based, so the `apt-get` install is fine and doesn't need to be changed.

kramasamy:
OK sounds good.

PR #2969, opened 2018-07-19
fixing download issue in Nomad scheduler

jerrypeng:
The heron downloaders API and behavior was updated causing the wrong config to be passed in the Nomad Scheduler


nlu90:
@huijunw Could you take a look?

PR #2976, opened 2018-08-09
Adding average Serialization/Deserialization time per tuple

kalimfaria:
Currently, the serialization metrics are reported as MultiCountMetric. This means that serialization times are reported as a sum of the serialization times for all tuples sent per unit time. It would be better if the metric was reported as average serialization time per tuple.


kalimfaria:
Instead of changing the old metric to a reduced type, now I've added an additional metric that measures average serialization/deserialization time per tuple.

PR #2978, opened 2018-08-10
first checkin for heron ml

skanjila:
Brought over all the ml code from samoa as is without building anything, this will serve as a starting point

### DIFF
--- a/heron/mlmgr/src/main/java/org/apache/heron/LocalStormDoTask.java
+++ b/heron/mlmgr/src/main/java/org/apache/heron/LocalStormDoTask.java
@@ -1,0 +1,83 @@
+package org.apache.samoa;
+
+/*
+ * #%L
+ * SAMOA
+ * %%
+ * Copyright (C) 2014 - 2015 Apache Software Foundation
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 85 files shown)


heron/mlmgr/src/main/java/org/apache/heron/LocalStormDoTask.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.samoa.topology.impl.StormSamoaUtils;
+import org.apache.samoa.topology.impl.StormTopology;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.commons.configuration.Configuration;
+
+import backtype.storm.Config;
```
nlu90: We should use `org.apache.heron.Config` instead of the `backtype.storm.Config`. And for any following files should be the same case.

heron/mlmgr/src/main/java/org/apache/heron/LocalStormDoTask.java
```diff
@@ -0,0 +1,82 @@
+package org.apache.heron;
```
nlu90: the package path should be `org.apache.heron.ml` if it's a general class needed by all machine learning runtime. If it's specific for samoa, then the package path should be `org.apache.heron.ml.samoa`. 

And this also applies to all the following files.

PR #2980, opened 2018-08-12
Extending update command to allow number of containers to be modified at runtime

kalimfaria:
Extended the update command to allow number of containers of a job to be updated at runtime. The only packing plan that supports this functionality currently is RoundRobinPackingPlan. Now, we can use a flag with the update command that specified the number of containers as follows:

/bin/heron update smf1/$USER/test jobName  --container-number 25  --verbose 

The previous semantics have been preserved, where either the runtime-config can be updated or either/both of component parallelism and number of containers can be updated at a time. 

The code has been tested locally and on Twitter's Aurora environment.


kramasamy:
@kalimfaria - how does this work with respect to non aurora environment?

huijunwu:
this pr does not touch schedulers. it works for both aurora and non-aurora

PR #2981, opened 2018-08-13
Metrics added to measure the rate at which tuples are added to the queue at spout gateway

kalimfaria:
This pull request includes metrics added that measure the rate at which tuples are added to the outgoing queues at spouts and the sizes of these queues. This allows us to measure whether the gateway thread pulls out tuples from the queue fast enough, thereby preventing the spout from becoming a bottleneck.


huijunwu:
could you pls copy pr description into comment in the code so that readers can easily understand why we have this metric

PR #2992, opened 2018-08-21
get back newParallelism ifelse branch

huijunwu:
In #2980, `updateTopologyHandler()`, if-else branch `newParallelism` was changed to cover both `newParallelism` and `newContainerNumber`, which failed several unit tests.
This PR gets `newParallelism` back, and puts `newContainerNumber` as the 3rd if-else branch.

### DIFF
--- a/heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
+++ b/heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
@@ -195,11 +195,11 @@ public class RuntimeManagerRunner {
       throws TopologyRuntimeManagementException, PackingException, UpdateDryRunResponse {
     assert !potentialStaleExecutionData;
     String newParallelism = updateConfig.getStringValue(RUNTIME_MANAGER_COMPONENT_PARALLELISM_KEY);
-    LOG.info("newParallelism " + newParallelism);
     String newContainerNumber = updateConfig.getStringValue(RUNTIME_MANAGER_CONTAINER_NUMBER_KEY);
-    LOG.info("newContainerNumber " + newContainerNumber);
     String userRuntimeConfig = updateConfig.getStringValue(RUNTIME_MANAGER_RUNTIME_CONFIG_KEY);
-    LOG.info("userRuntimeConfig " + userRuntimeConfig);
+    LOG.info("userRuntimeConfig " + userRuntimeConfig
+        + "; newParallelism " + newParallelism
+        + "; newContainerNumber " + newContainerNumber);
 
     // parallelism and runtime config can not be updated at the same time.
     if (((newParallelism != null && !newParallelism.isEmpty())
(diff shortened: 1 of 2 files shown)


heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -195,8 +195,11 @@ void updateTopologyHandler(String topologyName, Config updateConfig)
       throws TopologyRuntimeManagementException, PackingException, UpdateDryRunResponse {
     assert !potentialStaleExecutionData;
     String newParallelism = updateConfig.getStringValue(RUNTIME_MANAGER_COMPONENT_PARALLELISM_KEY);
+    LOG.info("newParallelism " + newParallelism);
```
nlu90: If these logs are for the testing purpose, remove them. 
If they will be kept, let's merge them into one line. `LOG.info(parallelise, container, runtimeConfig)`

heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -478,4 +551,16 @@ private static boolean parallelismChangeDetected(PackingPlans.PackingPlan curren
     }
     return false;
   }
+
+  private static boolean changeDetected(PackingPlans.PackingPlan currentProtoPlan,
```
nlu90: What's the difference between `changeDetected ` and `parallelismChangeDetected `?

heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -365,6 +405,39 @@ protected void cleanState(
     LOG.fine("Cleaned up topology state");
   }
 
+  @VisibleForTesting
+  PackingPlans.PackingPlan buildNewPackingPlan(PackingPlans.PackingPlan currentProtoPlan,
```
nlu90: Instead of replicating most of the code from the `buildNewPackingPlan(cp, cr, cn, t)`, can we reuse that method? Providing a special number for `cn` say -1, to indicate the old `repack` needs to be used.

heron/scheduler-core/src/java/org/apache/heron/scheduler/RuntimeManagerRunner.java
```diff
@@ -237,6 +237,46 @@ private int getCurrentContainerNumber(String topologyName) {
     return cPlan.getContainers().size();
   }
 
+  @VisibleForTesting
+  void updateTopologyComponentParallelism(String topologyName, String  newParallelism)
```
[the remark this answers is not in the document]
nlu90: +1 for Ning. We should reuse the `updatePackingPlan` instead of duplicating code. A special number of `newContainers` like -1 could be provided to indicate there's no container change.

PR #3004, opened 2018-08-23
Yaoli/fix stateful test

yaoliclshlmch:
Observed duplicated instance state info at http server during stateful integration test, which causes test failure. The issue is fixed in this PR.

### DIFF
--- a/integration_test/src/python/http_server/main.py
+++ b/integration_test/src/python/http_server/main.py
@@ -22,7 +22,6 @@ import tornado.ioloop
 import tornado.escape
 import tornado.web
 
-from sets import Set
 from heron.common.src.python.utils import log
 
 RESULTS_DIRECTORY = "results"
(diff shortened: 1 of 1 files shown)


nlu90:
One general question is why there'll be duplicated instance state info? Can we prevent the duplicated state being generated at the first place?

yaoliclshlmch:
Because checkpoint rollback is triggered by intended instance failure at a given component when the second checkpoint control tuple arrives. The failed instances will be restarted then and an additional checkpoint will be made afterwards which will result in unexpected instance state info.
What happens to the instances which can triggered checkpoint rollback:
1. instance started
2. successfully make checkpoint 1
3. throw exception to trigger instance failure when making checkpoint 2
4. restart this instance and rollback to checkpoint 1
5. successfully make checkpoint 3
6. throw exception to trigger instance failure when making checkpoint 4
......
the expected results are the instance states in checkpoint 1, so only those info at the http server is expected.
At the instance level, checkpoints will be made periodically and cannot be stopped or otherwise controlled.

PR #3009, opened 2018-08-28
fix version detection to avoid matching on single dot

julienledem:
For example: at end of "Anaconda Inc." versions (‘Python 3.6.4 :: Anaconda, Inc.’)

### DIFF
--- a/bazel_configure.py
+++ b/bazel_configure.py
@@ -149,7 +149,7 @@ def fail(message):
 
 # Assumes the version is at the end of the first line consisting of digits and dots
 def get_trailing_version(line):
-  version = re.search('([\d.]+)$', line)
+  version = re.search('([\d.]+\d)$', line)
   if version and '.' in version.group(0):
     return version.group(0)
 


bazel_configure.py
```diff
@@ -149,7 +149,7 @@ def fail(message):
 
 # Assumes the version is at the end of the first line consisting of digits and dots
 def get_trailing_version(line):
-  version = re.search('([\d.]+)$', line)
+  version = re.search('([\d.]+\d)$', line)
```
nwangtw: Hmm. It seems this new regex doesn't match "Python 3.6.4 :: Anaconda, Inc."?

nwangtw: or you just want to make sure the re doesn't match when the version string doesn't end with digits?

julienledem: ah right. maybe it doesn't fail when version is empty. This code needs to be hardened a little bit to verify what version matched is and whether it makes sense. Having a unit test with known version strings would help. Not sure why the version needs to be trailing.

julienledem: I instrumented the code and it seems to match:
first_line='Python 3.6.4 :: Anaconda, Inc.'
match('3.6.4','>=2.7.0')

nwangtw: Yeah. The version extraction is quite limited to specific format. Maybe we can update the code totally in future.

I tested with https://regex101.com/ and my local python and it doesn't match. I think "([\d.]+\d)$'" requires the last character to be digit in order to match.

>>> re.search('([\d.]+\d)$', "Python 3.6.4")
<_sre.SRE_Match object at 0x1047ca030>
>>> re.search('([\d.]+\d)$', "Python 3.6.4 :: Anaconda, Inc.")
>>>

joshfischer1108:
Thanks for the clean up @julienledem.  It's a good reminder that I don't go back through the old issues and clean up enough 😄

PR #3022, opened 2018-09-11
use RoundRobinPacking algorithm by default

nlu90:
As discussed in today's sync-up meeting, we decide to make the RoundRobinPacking algorithm as the default one.

This can avoid the submission failure problem when using the ResourceCompiliantRRPacking algorithm.

In long-term, there'll be fixes to the ResourceCompiliantRRPacking algorithm.


nlu90:
@nwangtw @jerrypeng

nwangtw:
sgtm. thanks

PR #3026, opened 2018-09-14
fix integration test failure with topo parallelism update

yaoliclshlmch:
Two integration test with topo parallelism update fails in travis ci because of unmatched expected and actual results.
During topology parallelism update, one instance (identified as container_id-task_id) may be one of the following case:
1. If it is not changed between the old and the new packing plans, it will not be killed, instead it will be deactivated and then activated which results in instance state before the update being reserved. 
2. If it is changed, it will be killed and start somewhere else, thus instance state before the update will be lost. 
The thing can happen to both spouts and bolts. But in the two tests, AT_LEAST_ONCE only requires tuples emitted at spouts to be a subset of tuples processed at bolts, thus clear spout state when update topology parallelism can guarantee test correctness no matter how packing plan changes.


yaoliclshlmch:
@nwangtw @nlu90 @huijunwu

PR #3027, opened 2018-09-18
Update python-doc-gen.sh

nwangtw:
- update mktemp command to be mac compatible
- install virtualenv automatically

### DIFF
--- a/website/scripts/python-doc-gen.sh
+++ b/website/scripts/python-doc-gen.sh
@@ -20,9 +20,10 @@ set -e
 HERONPY_VERSION=$1
 HERON_ROOT_DIR=$(git rev-parse --show-toplevel)
 INPUT=heronpy
-TMP_DIR=$(mktemp --directory)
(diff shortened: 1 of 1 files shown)


website/scripts/python-doc-gen.sh
```diff
@@ -20,9 +20,10 @@ set -e
 HERONPY_VERSION=$1
 HERON_ROOT_DIR=$(git rev-parse --show-toplevel)
 INPUT=heronpy
-TMP_DIR=$(mktemp --directory)
+TMP_DIR=$(mktemp -d)
 
-VENV="$(mktemp --directory)"
+VENV="$(mktemp -d)"
+pip install virtualenv
```
nlu90: Should we check if the `virtualenv` is already installed before using `pip install`?

nwangtw: I think pip doesn't reinstall package if it is already installed.

PR #3029, opened 2018-09-18
Nwang/add bolt based flatmap streamlet

nwangtw:
Streamlet flatmap function takes a function currently. Supporting Bolt in flatmap function makes it more flexible, and makes it easier to migration existing topology to streamlet API.

### DIFF
--- a/heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
+++ b/heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
@@ -21,7 +21,9 @@ package org.apache.heron.streamlet;
 
 import java.util.List;
 
+import org.apache.heron.api.bolt.IBasicBolt;
 import org.apache.heron.api.bolt.IRichBolt;
+import org.apache.heron.api.bolt.IWindowedBolt;
 import org.apache.heron.classification.InterfaceStability;
 
 /**
@@ -91,6 +93,18 @@ public interface Streamlet<R> {
    * @param bolt The rich bolt that is used to emit tuples from each input element
    */
   <T> Streamlet<T> applyBolt(IRichBolt bolt);
+
+  /**
+   * Return a new Streamlet by applying a bolt to each element of this Streamlet
+   * @param bolt The basic bolt that is used to emit tuples from each input element
+   */
+  <T> Streamlet<T> applyBolt(IBasicBolt bolt);
+
+  /**
+   * Return a new Streamlet by applying a bolt to each element of this Streamlet
+   * @param bolt The windowed bolt that is used to emit tuples from each input element
+   */
+  <T> Streamlet<T> applyBolt(IWindowedBolt bolt);
 
   /**
    * Return a new Streamlet by applying the filterFn on each element of this streamlet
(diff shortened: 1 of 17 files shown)


heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -85,6 +86,13 @@
   <T> Streamlet<T> flatMap(
       SerializableFunction<R, ? extends Iterable<? extends T>> flatMapFn);
 
+  /**
+   * Return a new Streamlet by applying a bolt to each element of this Streamlet and
+   * flattening the result
+   * @param flatMapBolt The rich bolt that is used to emit tuples from each input element
+   */
+  <T> Streamlet<T> flatMap(IRichBolt flatMapBolt);
```
nlu90: What is the use case for this new api?

This looks weird to me. When using the high-level DSL, users don't need to worry about bolt/spout.

@srkukarni Could you please take a look and provide some feedback?

jerrypeng: Ya I don't know if we should mix the two APIs. Kind of breaks the abstraction.

jerrypeng: Unless there is a strong user demand to help them transition from storm api -> streamlets

nwangtw: First of all, discussion is welcome in this PR (ok to move to mailing list as well). My thoughts are not firm yet but I do believe we need it.

Here are my thoughts:

Streamlet is not really the abstraction. My feeling is that Streamlet is good at the DAG layer but not flexible enough in the low level (operators). I would think it is like Scala vs Java(not the same, just some idea). Scala has the nice functional API but it is pretty useless in real life if procedural code is not allowed/supported.

Two reasons:
1. Migration is one major reason. There are quite some existing topologies written in low level API (for heron and storm). Streamlet is only friendly to new users, existing code such as KafkaSpout (it is spout, but same issue) in storm and some ML bolts has to be rewritten to take the readability/maintainability advantages.
2. Bolt/Spout are more flexible. They can do a lot more than a function provided by Streamlet API (initialization, config, checkpoint, etc). For examples, the stateful processing and component configs, they are not supported currently by Streamlet and if we add the features, it is likely user has to provide extra functions as parameters and the Streamlet API would became more and more complicated. Streamlet API will evolve but supporting Bolt/Spout could give us a lot room to design a clean API.

nwangtw: And with the new feature, it is still true that "When using the high-level DSL, users don't need to worry about bolt/spout.". The difference is that "they can if they need to".

srkukarni: Adding IRichBolt interface to the streamlet really mixes two things that stand for different things. Here is one potential compromise. Leave the streamlet interface the same, and maybe have something that inherits from streamlet but which has more interfaces like this? Would that leave pure streamlet pure while at the same time help the migration, etc?

nwangtw: Cool. Thanks for the idea! @srkukarni 

That sounds good to me. Let me see what I can do. I think the brother PR for Spout (https://github.com/apache/incubator-heron/pull/3032) is closer to the solution you suggested.

nwangtw:
Another pass of refactor is coming.

nwangtw:
Here is the design doc for CustomOperator (a more general feature of this PR):

PR #3032, opened 2018-09-20
Nwang/add spout based source streamlet

nwangtw:


heron/api/src/scala/org/apache/heron/streamlet/scala/Builder.scala
```diff
@@ -49,7 +50,14 @@ trait Builder {
...
+    * @return a Streamlet representation of the source object
     */
   def newSource[R](generator: Source[R]): Streamlet[R]
+
+  /**
+   * Creates a new Streamlet using the provided spout
+   * @param spout The spout that emits the tuples of the streamlet
+   * @return a Streamlet representation of the spout object
+   */
+  def newSource[R](spout: IRichSpout): Streamlet[R]
```
erenavsarogullari: @nwangtw Thanks for working on this PR.
**Integration Test** is also useful for end2end coverage. 

Currently, the following streamlet ITs exist: 
- **Python:** `word_count_streamlet` 
- **Scala:** `ScalaStreamletWithFilterAndTransform` and `ScalaStreamletWithMapAndFlatMapAndFilterAndClone`
- **Java Streamlet API** does not have any IT yet.

nwangtw: Agreed. And there could be some bugs in the API since so far it seems people are testing it but I haven't heard anyone using it in prod yet.

Let me prepare another PR to at least add the two Scala tests to java. https://github.com/apache/incubator-heron/issues/3101

And we definitely need better IT coverage for Streamlet API.

erenavsarogullari: @nwangtw Thanks for following this.
I developed Scala Integration Tests. If you want, i can also address `Java ITs` in the light of #3101. Also, please let me know to collaborate on other Streamlet Improvements ;)

heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
```diff
@@ -61,6 +62,14 @@ public BuilderImpl() {
     return retval;
   }
 
+  @Override
+  public <R> Streamlet<R> newSource(IRichSpout spout) {
+    StreamletImpl<R> retval = StreamletImpl.createSpoutStreamlet(spout);
+    retval.setNumPartitions(1);
```
erenavsarogullari: Number of partitions is already set at createSpoutStreamlet#SpoutStreamlet level so this looks redundant. 
Also, same for following source functions:
```
newSource(SerializableSupplier<R> supplier)
newSource(Source<R> generator)
```

heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
```diff
@@ -61,6 +62,14 @@ public BuilderImpl() {
     return retval;
   }
 
+  @Override
+  public <R> Streamlet<R> newSource(IRichSpout spout) {
```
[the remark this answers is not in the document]
nwangtw: good catch. will do.

heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
nwangtw: True.

erenavsarogullari:
Overall is ok for me. 
**IT** coverage will be useful so i will be addressing through #3106

PR #3034, opened 2018-09-21
Nwang/remove stateful from streamletoperator

nwangtw:
StreamletOperator is the base class of many operators (there are windowed operators which are different). Not all these operators support or need the stateful interface, such as FilterOperator. Supporting stateful interface or not should be decided in the implementation level instead of the base level.

Plus, StreamletWindowedOperator doesn't support stateful interface either.


nwangtw:
Thanks for pointing out. Cleaned up the "TODO"s.

The two classes has initState() function but didn't provide implementation for presave(), so I added TODO there. After checking the code, I don't see any code to be added for now.

PR #3037, opened 2018-09-28
Remove the heron-users google group from README

nlu90:
All discussion should happen in the Apache mailing list. This PR removes the deprecated heron-users google group.


nlu90:
@kramasamy Could you also delete the deprecated heron-users google group? This group has been receiving unrelated posts(e.g. furniture selling) for quite a long time and no one asks heron questions. Since we are preparing for the first Apache release, it's a good time to migrate all the discussion to the mailing list at this time.

PR #3040, opened 2018-10-02
Nwang/refactor grouping

nwangtw:
Refactor grouping logic from BoltDeclarer functions to classes so that it is easier for them to be reused by Streamlet API, from calling different function

boltDeclarer.XGrouping(parent, streamId, ...);

to:

StreamGrouping grouper = new X(...);
boltDeclarer.grouping(parent, streamId, grouper);

- No logic change.
- Old functions are still available.

### DIFF
--- a/heron/api/src/java/org/apache/heron/api/grouping/AllStreamGrouping.java
+++ b/heron/api/src/java/org/apache/heron/api/grouping/AllStreamGrouping.java
@@ -1,0 +1,40 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.heron.api.grouping;
+
+import org.apache.heron.api.generated.TopologyAPI;
+
+/**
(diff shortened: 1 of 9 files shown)


jerrypeng:
@nwangtw the change looks fine to me. However, I don't quite get the use case?  Why is change needed for the streamlets.  Can you elaborate some more? Thanks

nwangtw:
Thanks for reviewing!

Sure thing to explain a bit more:

Firstly, in order to support custom operator, we need to pass in the operator object as well as specifying the grouping strategy.

Currently, the implementation of Streamlet classes are like:

bldr.setBolt(getName(), new FlatMapOperator<R, T>(flatMapFn),
        getNumPartitions()).shuffleGrouping(parent.getName());


You can say the grouping is a function call with the parent name. It is ok for predefined operators because the grouping is likely to be decided, but this is tricky for custom operators (and there might be new operators that allow users to specify grouping strategy in future).

One way is to pass an enum, but this is not flexible.

jerrypeng:
@nwangtw thanks for the explanation. I not sure there will be such a use case.  The concept of of these types of grouping is really for the spout/bolt APIs.  The grouping for a DSL is really coupled with type of the operator.  At a DSL level operators like something "keyBy" is really the only operators that are needed.  I am also not aware of any other DSLs that allow users to have custom groupings.  I think that breaks the DSL abstraction.  

I also don't seem any harm with merging this PR.  We can still merge it and see in the future.

nwangtw:
Yeah. With the current standard operators, it is not necessary. However,

For the new CustomOperator I am working on, it is needed. Streamlet classes are responsible for creating the operator objects and hook up the grouping. However CustomOperatorStreamlet will be shared by all user defined CustomOperators.

Another case is that in your proposal to change the streamlet style to be like:
reduce(...)
  .windowed(...)
  .byKey()
   ... 
  .build();
then the byKey() implementation will likely to need the grouping strategies to have their own classes.

jerrypeng:
@nwangtw a KeyBy operator will always be a fieldgrouping

jerrypeng:
+1

nwangtw:
> @nwangtw a KeyBy operator will always be a fieldgrouping

Yeah. Very likely.

We dont have a grouping function yet currently. Custom grouping might be a separated function or uses keyBy() interface. We will see.

nlu90:
The refactoring looks fine to me. 

As Jerry mentioned, this might break the DSL abstraction. But we can merge it for now and see

PR #3041, opened 2018-10-05
Remove libunwind patches

nwangtw:
Try to avoid the patches.


third_party/libunwind/libunwind.BUILD
```diff
@@ -22,18 +22,12 @@ out_files = [
 ]
 
 exports_files([
```
nlu90: Since all the libunwind files are removed, do we still need this `exports_files` and the following `genrule` for `libunwind`?

nwangtw: genrule is still needed, but exports_files seems to be unnecessary now.

PR #3043, opened 2018-10-09
Refactor heron_executor to make it easier to customize

nwangtw:
Currently the main function and the function for generating java command are single functions. In case some users might need to customize it, it is hard to reuse code in the existing executor. After the refactor, users should be able to create their own executor by extending the HeronExecutor class and use it in the main function.

```
class UserHeronExecutor(oss_heron_executor.HeronExecutor):
  # pylint: disable=no-self-use
  def _get_jvm_instance_options(self, instance_id, component_name, remote_debugger_port):
    options = oss_heron_executor.HeronExecutor._get_jvm_instance_options(
        instance_id, component_name, remote_debugger_port)
    ....
    return options

def main():
  # Instantiate the executor
  parsed_args = oss_heron_executor.parse_args(sys.argv)
  executor = UserHeronExecutor(parsed_args, os.environ.copy())
  executor.initialize()

  # launch it
  oss_heron_executor.start(executor)

if __name__ == "__main__":
  main()
```

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -261,8 +261,7 @@ class HeronExecutor(object):
       parsed_args.jvm_remote_debugger_ports.split(",") \
         if parsed_args.jvm_remote_debugger_ports else None
 
-  def __init__(self, args, shell_env):
-    parsed_args = self.parse_args(args)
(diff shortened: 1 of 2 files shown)


nwangtw:
Only code move around. No logic change other than a few extra logs.

PR #3044, opened 2018-10-13
Catch exceptions in Streamlet operators

nwangtw:


jerrypeng:
@nwangtw was there an issue that was fixed by adding the try catches?

nwangtw:
No real known issue, but the current operators don't handle failures and I feel the try/catch could be necessary. Or it was by design that Streamlet operators couldn't fail?

nicknezis:
@huijunw Do you have any input on this merge request?

PR #3049, opened 2018-10-17
Nwang/add custom operator refactor

nwangtw:
This is the implementation of CustomOperator based on this design doc:

Todos in future PRs:
- add grouper
- add/refactor windowed custom operator
- scala support
- refactor/simplify existing operators (v2 design in doc)

Existing logic doesn't change.


heron/api/src/java/org/apache/heron/streamlet/impl/streamlets/CustomBasicStreamlet.java
```diff
@@ -0,0 +1,55 @@
...
+  private StreamletImpl<R> parent;
+  private IStreamletBasicOperator<R, T> op;
+
+  /**
+   * Create a custom streamlet from user defined CustomBasicOperator object.
+   * @param parent The parent(upstream) streamlet object
+   * @param op The user defined CustomeBasicOperator
+   */
+  public CustomBasicStreamlet(StreamletImpl<R> parent,
+                              IStreamletBasicOperator<R, T> op) {
```
nlu90: for custom operator, grouping strategy should also be provided

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -22,6 +22,8 @@
 import java.util.List;
 
 import org.apache.heron.classification.InterfaceStability;
+import org.apache.heron.streamlet.impl.operators.IStreamletBasicOperator;
```
nlu90: Given all the existing `StreamletOperator` extends `BaseRichBolt`, should we also enforce custom operator to be `BaseRichBolt`?

Thus we can reduce this redundant `IStreamletBasicOperator`.

heron/api/src/java/org/apache/heron/streamlet/impl/operators/IStreamletOperator.java
```diff
@@ -0,0 +1,30 @@
...
+ * under the License.
+ */
+
+
+package org.apache.heron.streamlet.impl.operators;
+
+import org.apache.heron.api.bolt.IRichBolt;
+
+/**
+ * The interface for custom operators: including new user defined operators as well as
```
nlu90: The comment here is inconsistent with the actual code.

This interface is actually for all the operators not only for the custom operators.

Need to update this comment or the code to keep them consistent.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -22,6 +22,8 @@
 import java.util.List;
 
 import org.apache.heron.classification.InterfaceStability;
```
nlu90: I think we may also need to add `CustomSource` to allow users to provide specific spout as the input?

heron/api/src/java/org/apache/heron/streamlet/impl/operators/CustomOperator.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.apache.heron.api.bolt.OutputCollector;
+import org.apache.heron.api.topology.TopologyContext;
+import org.apache.heron.api.tuple.Tuple;
+import org.apache.heron.api.tuple.Values;
+
+/**
+ * CustomOperator is the base class for all user defined operators.
+ * Usage:
+ * 1. Create user defined operator
```
nlu90: This part needs further discussion. IMO, users should only need to provide the Bolt to the streamlet API instead of having direct access to streamlet's operator implementation and being required to implement an additional `process` method.

Once the desired bolt is provided, streamlet should take care of the remaining stuff.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -22,6 +22,8 @@
...
+import org.apache.heron.streamlet.impl.operators.IStreamletBasicOperator;
+import org.apache.heron.streamlet.impl.operators.IStreamletOperator;
```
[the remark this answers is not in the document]
nwangtw: Make sense. Will do.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -201,6 +203,22 @@
   <T> Streamlet<T> transform(
       SerializableTransformer<R, ? extends T> serializableTransformer);
 
+  /**
+   * Returns a new Streamlet by applying the operator on each element of this streamlet.
+   * @param operator The operator to be performed
+   * @param <T> The return type of the transform
+   * @return Streamlet containing the output of the operation
+   */
+  <T> Streamlet<T> perform(IStreamletOperator<R, T> operator);
```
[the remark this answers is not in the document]
nwangtw: Yao and I thought about a few options and perform is the best we got.

apply() indeed is better, however it is a special function in Java/Scala so it cant be used. applyOperator() is valid, but maybe the "Operator" part sounds a bit redundant?

I am totally open for a better name. Let's see other thoughts.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
nwangtw: Yesh. I think we should enforce all StreamletOperator to be IRichBolt. CustomOperator will be the base class for all user defined operators (non-windowed onces) hence they will be IRichBolt.

IStreamletBasicOperator is not for CustomOperator. It is for reusing existing user Bolts (the ones extend from IBasicBolt instead of IRichBolt) only. That is why we have only IStreamletBasicOperator but no CustomBasicOperator.

heron/api/src/java/org/apache/heron/streamlet/impl/operators/IStreamletOperator.java
nwangtw: Good catch. Will do.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
nwangtw: Yeah. It is in another PR already. https://github.com/apache/incubator-heron/pull/3032 (may need refactor though)

heron/api/src/java/org/apache/heron/streamlet/impl/operators/CustomOperator.java
nwangtw: It is not an addition method users need to implement.

This class is for all "new" user defined streamlet operators and there is no Bolt concept in it.

Existing Bolts (or write operators in low level way) is supported but it another different use case.
```

class UserBoltOperator
    extends UserBolt
    implements ICustomOperator<InputType, OutputType> {
}

```
Then the bolt is converted to a streamlet operator and can be used in Streamlet API.
```
source
   ....
  .perform(new UserBoltOperator())
   ....

```

PR #3052, opened 2018-10-19
add back stmgr metrics

yaoliclshlmch:
The following Stmgr-Stmgr metrics are back in this PR:
__bytes_from_stmgrs
__bytes_to_stmgrs
__tuples_to_stmgrs
__tuples_from_stmgrs
__ack_tuples_to_stmgrs
__ack_tuples_from_stmgrs
__fail_tuples_to_stmgrs
__fail_tuples_from_stmgrs

Related to Issue #1908

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr-server.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr-server.cpp
@@ -30,37 +30,61 @@
 #include "network/network.h"
 #include "config/helper.h"
 #include "config/heron-internals-config-reader.h"
+#include "metrics/metrics.h"
 
 namespace heron {
 namespace stmgr {
 
 // Num data tuples received from other stream managers
-const sp_string METRIC_DATA_TUPLES_FROM_STMGRS = "__tuples_from_stmgrs";
+const sp_string METRIC_DATA_TUPLES_FROM_STMGRS = "__server/__tuples_from_stmgrs";
 // Num ack tuples received from other stream managers
-const sp_string METRIC_ACK_TUPLES_FROM_STMGRS = "__ack_tuples_from_stmgrs";
+const sp_string METRIC_ACK_TUPLES_FROM_STMGRS = "__server/__ack_tuples_from_stmgrs";
 // Num fail tuples received from other stream managers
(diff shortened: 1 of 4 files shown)


yaoliclshlmch:
@huijunw @nlu90 @nwangtw

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -30,37 +30,61 @@
...
 #include "config/helper.h"
 #include "config/heron-internals-config-reader.h"
+#include "metrics/metrics.h"
 
 namespace heron {
 namespace stmgr {
 
 // Num data tuples received from other stream managers
-const sp_string METRIC_DATA_TUPLES_FROM_STMGRS = "__tuples_from_stmgrs";
+const sp_string METRIC_DATA_TUPLES_FROM_STMGRS = "__server/__tuples_from_stmgrs";
```
nwangtw: maybe put "__server/" into a variable and use it in metrics:

const sp_string SERVER_SCOPE = "__server/";
....

during registration:
metrics_manager_client_->register_metric(
    SERVER_SCOPE | METRIC_DATA_TUPLES_FROM_STMGRS,
    tuples_from_stmgrs_metrics_);

Not a blocker though.

heron/stmgr/src/cpp/manager/stmgr-server.cpp
```diff
@@ -30,37 +30,61 @@
 #include "network/network.h"
 #include "config/helper.h"
 #include "config/heron-internals-config-reader.h"
+#include "metrics/metrics.h"
 
 namespace heron {
 namespace stmgr {
 
 // Num data tuples received from other stream managers
```
nwangtw: Please add a comment about why we have to register these metrics one by one instead of registering the scope.

nlu90:
In general, it looks good to me. One thing I suggest to try is verifying those metrics with an actual running topology in our environment.

yaoliclshlmch:
Yes, I tried the WordCount topology, and the "__bytes_*" and "__tuples_*" work well. There are no ack and fail in this topology, but the corresponding metrics should be fine.

PR #3094, opened 2018-10-27
Update pidhandler.py

windhamwong:
Fixing Bash syntax error in pid checking script.


nwangtw:
qq: on which OS you are seeing this error?

windhamwong:
Debian 9. The ` AWK` command should read `'{}'` as the output format.

nwangtw:
Ok. SGTM.

I am not a linux expert, maybe @nlu90 @huijunw can confirm and approve.

PR #3095, opened 2018-10-28
[Java Streamlet API] Extend validations

erenavsarogullari:
This PR aims to add more validations
- `BuilderImpl` level to fail fast with meaningful error message
- `Streamlet` `clone` function should accept as `numClones > 0`

### DIFF
--- a/heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
+++ b/heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
@@ -24,13 +24,12 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 
-import com.google.common.base.Preconditions;
-
 import org.apache.heron.api.topology.TopologyBuilder;
 import org.apache.heron.streamlet.Builder;
 import org.apache.heron.streamlet.SerializableSupplier;
 import org.apache.heron.streamlet.Source;
 import org.apache.heron.streamlet.Streamlet;
+import org.apache.heron.streamlet.impl.utils.StreamletUtils;
 
 /**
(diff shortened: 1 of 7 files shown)


heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
```diff
@@ -44,6 +46,7 @@ public BuilderImpl() {
 
   @Override
   public <R> Streamlet<R> newSource(SerializableSupplier<R> supplier) {
+    Preconditions.checkNotNull(supplier, "supplier must not be null.");
```
nwangtw: why use Preconditions? require() is not enough?

We are hoping to get rid of guava library since it is too big. :(

heron/api/src/java/org/apache/heron/streamlet/impl/StreamletImpl.java
```diff
@@ -216,7 +216,7 @@ protected StreamletImpl() {
 
   public void build(TopologyBuilder bldr, Set<String> stageNames) {
     if (built) {
-      throw new RuntimeException("Logic Error While building " + getName());
+      throw new RuntimeException("Logic Error While building stage: " + getName());
```
nwangtw: why "stage"?

heron/api/src/java/org/apache/heron/streamlet/impl/BuilderImpl.java
erenavsarogullari: Yes, `require` can be used by moving from `StreamletImpl` to `StreamletUtils` as the current patch

erenavsarogullari:
@nwangtw @huijunwu 
Thanks for reviews. It is ready to be merged.

PR #3099, opened 2018-11-01
Nwang/add streamlet operator interfaces with grouping v2

nwangtw:
Add stream grouping support in Steamlet API. Currently used by applyOperator() functions only.


heron/api/src/java/org/apache/heron/streamlet/StreamGrouper.java
```diff
@@ -0,0 +1,78 @@
...
+    append(Utils.DEFAULT_STREAM_ID, grouping);
+  }
+
+  /**
+   * Construct a Grouper object with specific grouping strategy and streamid
+   * @param streamId The stream id that the strategy to be applied to
+   * @param grouping The grouping strategy to be applied
+   */
+  public StreamGrouper(String streamId, StreamGrouping grouping) {
+    groupingMap = new HashMap<>();
```
erenavsarogullari: - `HashMap` accepts `null` K,V pairs so `ConcurrentHashMap` can be preferred to avoid `null` access (due to Public API)
- On the other hand, `streamId` also needs to be robust for blank values(`streamId` = " ")

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -209,6 +209,15 @@
...
   <T> Streamlet<T> applyOperator(IStreamletOperator<R, T> operator);
 
+  /**
+   * Returns a new Streamlet by applying the operator on each element of this streamlet.
+   * @param operator The operator to be applied
+   * @param grouper The grouper to be applied with the operator
+   * @param <T> The return type of the transform
+   * @return Streamlet containing the output of the operation
+   */
+  <T> Streamlet<T> applyOperator(IStreamletOperator<R, T> operator, StreamGrouper grouper);
```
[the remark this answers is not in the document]
nwangtw: That is a good question. Some others might need it as well.

As @jerrypeng said before, the grouping strategy for most (even though I am not sure about its most or some) operations are fixed, like .map doesn't need grouping in general, reduce needs a field grouping or custom grouping. Having a streamlet level setGrouper() could be convenient but it makes it less straightforward and cause confusion (which is why I am hesitate about the Builder API).

I am totally open to feedbacks and other options. Please feel free to discuss.

heron/api/src/java/org/apache/heron/streamlet/StreamGrouper.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.apache.heron.api.grouping.StreamGrouping;
+import org.apache.heron.api.utils.Utils;
+
+/**
+ * StreamGrouper is the class to config grouping strategy in Streamlet API. Note that
+ * One Operator could have more than one grouping strategies (when there are more than one
+ * incoming streams) although it has one in most cases.
+ */
+public class StreamGrouper {
```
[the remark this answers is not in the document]
nwangtw: Yeah. Document is important.

Planning to update/add a few new operators. Will update doc afterwards.

heron/api/src/java/org/apache/heron/streamlet/StreamGrouper.java
nwangtw: Good to know. Thx. Will update.

heron/api/src/java/org/apache/heron/streamlet/impl/StreamletImpl.java
```diff
@@ -522,9 +523,23 @@ public void toSink(Sink<R> sink) {
   public <T> Streamlet<T> applyOperator(IStreamletOperator<R, T> operator) {
```
nlu90: need some comments to specify when `NoneStreamGrouping()` is suitable for use

nwangtw: Done.

Issue #3104, opened 2018-11-08
NPE FullBoltMetrics.ackedTuple


thinker0:
- heron-0.18.7
```

[2018-11-09 00:13:28 +0900] [SEVERE] monitor: Exception propagated to the root monitor! 
java.lang.NullPointerException
	at com.twitter.heron.common.utils.metrics.FullBoltMetrics.ackedTuple(FullBoltMetrics.java:127)
	at com.twitter.heron.instance.bolt.BoltOutputCollectorImpl.admitAckTuple(BoltOutputCollectorImpl.java:159)
	at com.twitter.heron.instance.bolt.BoltOutputCollectorImpl.ack(BoltOutputCollectorImpl.java:91)
	at com.twitter.heron.api.bolt.OutputCollector.ack(OutputCollector.java:234)
	at org.apache.storm.task.OutputCollectorImpl.ack(OutputCollectorImpl.java:76)
```
```
[2018-11-09 00:13:06 +0900] [SEVERE] com.twitter.heron.instance.HeronInstance: Exception caught in thread: SlaveThread with id: 24 
java.lang.NullPointerException
	at com.twitter.heron.proto.system.HeronTuples$HeronControlTupleSet.isInitialized(HeronTuples.java:4158)
	at com.twitter.heron.proto.system.HeronTuples$HeronControlTupleSet$Builder.build(HeronTuples.java:4423)
	at com.twitter.heron.proto.system.HeronTuples$HeronTupleSet$Builder.setControl(HeronTuples.java:6144)
	at com.twitter.heron.instance.OutgoingTupleCollection.flushRemaining(OutgoingTupleCollection.java:193)
	at com.twitter.heron.instance.OutgoingTupleCollection.sendOutTuples(OutgoingTupleCollection.java:85)
	at com.twitter.heron.instance.AbstractOutputCollector.sendOutTuples(AbstractOutputCollector.java:102)
	at com.twitter.heron.instance.bolt.BoltInstance$1.run(BoltInstance.java:212)
	at com.twitter.heron.common.basics.WakeableLooper.executeTasksOnWakeup(WakeableLooper.java:186)
	at com.twitter.heron.common.basics.WakeableLooper.runOnce(WakeableLooper.java:105)
	at com.twitter.heron.common.basics.WakeableLooper.loop(WakeableLooper.java:95)
	at com.twitter.heron.instance.Slave.run(Slave.java:180)
[...]

PR #3107, opened 2018-11-09
Improve concurrency for needed parts.

thinker0:
Improve concurrency for needed parts.
- Change concurrentHashMap
- Improve thread safe OutputCollector
- Improve thread safe IOutputCollector

#3104

### DIFF
--- a/heron/api/src/java/org/apache/heron/api/metric/CountStatAndMetric.java
+++ b/heron/api/src/java/org/apache/heron/api/metric/CountStatAndMetric.java
@@ -19,9 +19,9 @@
 
 package org.apache.heron.api.metric;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.TimerTask;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
@@ -176,7 +176,7 @@ public class CountStatAndMetric implements IMetric<Long> {
   }
 
   synchronized Map<String, Long> getTimeCounts(long now) {
-    Map<String, Long> ret = new ConcurrentHashMap<>();
+    Map<String, Long> ret = new HashMap<>();
     long value = currentBucket.get();
     long timeSpent = now - bucketStart;
     ret.put("600", readApproximateTime(value, timeSpent, tmTime, tmBuckets, 600 * 1000));
--- a/heron/api/src/java/org/apache/heron/api/metric/CountStatAndMetric.java
+++ b/heron/api/src/java/org/apache/heron/api/metric/CountStatAndMetric.java
@@ -19,9 +19,9 @@
 
 package org.apache.heron.api.metric;
 
-import java.util.HashMap;
 import java.util.Map;
 import java.util.TimerTask;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
 
 /**
@@ -176,7 +176,7 @@ public class CountStatAndMetric implements IMetric<Long> {
   }
 
   synchronized Map<String, Long> getTimeCounts(long now) {
-    Map<String, Long> ret = new HashMap<>();
+    Map<String, Long> ret = new ConcurrentHashMap<>();
     long value = currentBucket.get();
     long timeSpent = now - bucketStart;
     ret.put("600", readApproximateTime(value, timeSpent, tmTime, tmBuckets, 600 * 1000));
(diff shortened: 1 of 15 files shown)


nwangtw:
That is the root cause of the exception? Nice find!

thinker0:
@nwangtw 

I created a Bolt with MoultiThread (finagle-http) async and ack all at once.

I think NPE is happening often.

heron-0.18.7 I can not create a maven package file, so I can not test it in my project.
Can you tell me how to package it so I can test it?

nwangtw:
I see. Make sense then when you use multi threading. In this case you might be better to synchronized {} to make sure ack() is not called many times at the same time. OutputCollector has a queue in it and I am not sure the queue is thread-safe. In Heron engine, the core assumes single thread environment.

Internally, we build library and use it directly, instead of the ones in Maven.

bazel build --compilation_mode=dbg --config=darwin scripts/packages:tarpkgs

heron/api/src/java/org/apache/heron/api/metric/MultiCountMetric.java
```diff
@@ -38,7 +38,7 @@ public CountMetric scope(String key) {
 
   @Override
   public Map<String, Long> getValueAndReset() {
-    Map<String, Long> ret = new HashMap<>();
+    Map<String, Long> ret = new ConcurrentHashMap<>();
```
nwangtw: ConcurrentHashMap might not be necessary in getValueAndReset(). This function is called every X seconds to sample the current metrics and export by Heron instance, so it is not called from multiple threads.

heron/api/src/java/org/apache/heron/api/metric/CountStatAndMetric.java
```diff
@@ -176,7 +176,7 @@ private synchronized void rotate(long value, long timeSpent,
   }
 
   synchronized Map<String, Long> getTimeCounts(long now) {
-    Map<String, Long> ret = new HashMap<>();
+    Map<String, Long> ret = new ConcurrentHashMap<>();
```
nwangtw: ConcurrentHashMap  is not needed here I think. This is a get function and it is already synchronized.

heron/api/src/java/org/apache/heron/api/metric/LatencyStatAndMetric.java
```diff
@@ -225,7 +225,7 @@ private synchronized void rotate(long lat, long count, long timeSpent, long targ
   }
 
   synchronized Map<String, Double> getTimeLatAvg(long now) {
-    Map<String, Double> ret = new HashMap<>();
+    Map<String, Double> ret = new ConcurrentHashMap<>();
```
nwangtw: ConcurrentHashMap  is not needed here I think. This is a get function and it is already synchronized.

storm-compatibility/src/java/org/apache/storm/metric/api/MultiCountMetric.java
```diff
@@ -36,7 +36,7 @@ public CountMetric scope(String key) {
   }
 
   public Object getValueAndReset() {
-    Map<String, Object> ret = new HashMap<>();
+    Map<String, Object> ret = new ConcurrentHashMap<>();
```
nwangtw: HashMap should be good here I think.

storm-compatibility/src/java/org/apache/storm/metric/api/MultiReducedMetric.java
```diff
@@ -39,7 +39,7 @@ public ReducedMetric scope(String key) {
   }
 
   public Object getValueAndReset() {
-    Map<String, Object> ret = new HashMap<>();
+    Map<String, Object> ret = new ConcurrentHashMap<>();
```
nwangtw: HashMap should be good here I think.

thinker0:
Please review the PR. @nwangtw

joshfischer1108:
The changes after Ning's comments look fine to me.  I'll leave it to @nwangtw to approve.

thinker0:
> I see. Make sense then when you use multi threading. In this case you might be better to synchronized {} to make sure ack() is not called many times at the same time. OutputCollector has a queue in it and I am not sure the queue is thread-safe. In Heron engine, the core assumes single thread environment.
> 
> Internally, we build library and use it directly, instead of the ones in Maven.
> 
> bazel build --compilation_mode=dbg --config=darwin scripts/packages:tarpkgs

Made the OutputCollector thread-safe.

@nwangtw,
```
synchronized
```
works fine without it.
And it works well in Production.

thinker0:
@joshfischer1108 PR Check ^^

Issue #3106, opened 2018-11-08
Add integration Tests for Java and Scala Streamlet APIs


erenavsarogullari:
Following transformations need IT coverage as well on both Java and Scala Streamlet APIs.
```
join
reduceByKeyAndWindow
union
applyOperator
```

PR #3109, opened 2018-11-13
Add withStream() in Streamlet to support stream selection

nwangtw:
Usage:

Assuming an operator creates two streams with ids "stream1" and "stream2" and it is managed by streamletWithTwoStreams (with type Streamlet<T>), it can be used like:

streamletWithTwoStreams
  .withStream("stream1")
  ......
  .log();

streamletWithTwoStreams
  .withStream("stream2")
  ......
  .log();


heron/api/tests/java/org/apache/heron/streamlet/impl/StreamletImplTest.java
```diff
@@ -81,6 +82,22 @@ public void testBasicParams() throws Exception {
     assertEquals(bStreamlet.getChildren().size(), 0);
   }
 
+  @Test
+  public void testWithStream() {
```
erenavsarogullari: Test coverage can be useful for the following cases:
- if `streamId` is set duplicate (`test_stream2 == test_stream3`)
- if `streamId` is set `Bolt` level again as different from `Spout`'s `streamId`

heron/api/src/java/org/apache/heron/streamlet/impl/streamlets/ConsumerStreamlet.java
```diff
@@ -34,10 +34,12 @@
  */
 public class ConsumerStreamlet<R> extends StreamletImpl<R> {
   private StreamletImpl<R> parent;
+  private String parentStream;
```
erenavsarogullari: may `parentStreamId` be more meaningful?

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -71,6 +71,22 @@
...
 
+  /**
+   * Set the id of the stream to be used by the children nodes.
+   * Usage (assuming source is a Streamlet object with two output streams: stream1 and stream2):
+   *   source.withStream("stream1").filter(...).log();
+   *   source.withStream("stream2").filter(...).log();
+   * @param streamId The specified stream id
+   * @return Returns back the Streamlet with changed stream id
+   */
+  Streamlet<R> withStream(String streamId);
```
erenavsarogullari: I think `streamId` offers to be distinguished streams, right? 
Also, wondering is there UI impact (if user would like to see `streamId` or not)?

nwangtw: Yeah. A component can generates multiple different streams (such as 3 integer streams: odd, even, all) and the next component needs to specify which one to accept data.

UI doesn't show this information currently and it could be a useful feature to add. And grouping strategy is another important information on the edges.

erenavsarogullari: Exactly. Vertices show Streamlet name and Edges can show both streamId and grouping strategy.

heron/api/src/java/org/apache/heron/streamlet/impl/StreamletImpl.java
```diff
@@ -205,11 +207,37 @@ public int getNumPartitions() {
...
+   * Set the id of the stream to be used by the children nodes.
+   * Usage (assuming source is a Streamlet object with two output streams: stream1 and stream2):
+   *   source.withStream("stream1").filter(...).log();
+   *   source.withStream("stream2").filter(...).log();
+   * @param streamId The specified stream id
+   * @return Returns back the Streamlet with changed stream id
+   */
+  @SuppressWarnings("HiddenField")
+  @Override
+  public Streamlet<R> withStream(String streamId) {
```
erenavsarogullari: Also wondering the case: if `Bolt` points invalid `streamId` (not matched with `parentStreamId`) so could **validation** be required? (to cover this case either **fail-fast** or **warning message** due to redundant Bolt)?

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
erenavsarogullari: As long-term, i think we can raise an UI issue to be addressed this (to be shown both **streamId** and **grouping strategy**).

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
```diff
@@ -71,6 +71,22 @@
...
+
+  /**
+   * Gets the stream id of this Streamlet.
+   * @return the stream id of this Streamlet
+   */
+  String getStreamId();
```
erenavsarogullari: Also IT coverage will be useful to verify end2end so will be addressing in the light of #3106

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
nwangtw: Yep.

heron/api/src/java/org/apache/heron/streamlet/impl/StreamletImpl.java
nwangtw: Agreed. Failfast is desirable for sure.

Validation is an option question for now. Currently all streamlet operators receive and emit tuples to default stream.

The stream ids are in run time information now. So we need to be careful about where the validation should be done and how to apply it.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
nwangtw: In fact, I am thinking of adding the split() operator now with this PR so that we can add a proper IT for it. I prefer to break PR down to single functionality but in this case it could be useful.

erenavsarogullari: Sure, IT coverage is too important due to aim end2end. I think it should be part of the code review checklist for incoming PRs ;) 

On the other hand, do you have template for `split` function signature?

nwangtw: Did a quick search, I think it is pretty close to the split operator in Flink (https://ci.apache.org/projects/flink/flink-docs-release-1.6/dev/stream/operators/).

In fact, select() seems to be a better name than "withStream". :D I will think about it.

Neng gives a feedback about the withStream() function as a new user and he made a good point that we need to handle both cases:
1.
  stream.map();
  stream.withStream("a").map();
  stream.map();   // this should be the default stream
  stream.withStream("a").map();

2.
  steam2 = stream.withString("a");
  stream2.map();   // this should be the new stream
  stream.map();   // this should be the default stream

And this is not supported with the current implementation since the this object is muted. I am going to see what is the best way and update this PR.

erenavsarogullari: Yep, i was having a look for Flink implementation as well. It supports child and parent Stream hierharchy to support `select` functions through `split` function.

I think more make sense `split` function naming convention and following can be as an alternative: 

```
Streamlet<Integer> numberStreamlets = ...
// user can apply his custom split logic
Map<String, Streamlet> splittedStreamlets = numberStreamlets.split(new SerializableSplitter())

splittedStreamlets.get("even")
splittedStreamlets.get("odd")
```

If we have proposal document by covering the all potential use-cases, that will be useful as well ;)

nwangtw: "Map<String, Streamlet<Integer>>"  won't really work because right now one Streamlet is mapped to one operator which is a Bolt component. In our case, there is only one component (hence one streamlet object) with multiple streams.

I had to add a StreamletShadow class to create companion objects of the main Streamlet object in my implementation.

nwangtw: In fact, still no....

The problem is the .setName and .setPartition functions, like:

numberStreamlets.split(splitter).setName("split").setPartition(10);

the split operator is one single bolt and all entries in Map<String, Streamlet> points to the same bolt. It might be ok to set name and partition like:
Map<String, Streamlet> splittedStreamlets  = numberStreamlets.split(splitter);

splittedStreamlets.get("even").setName("split")
splittedStreamlets.get("even").setPartition(partition)

but it looks really bad this way.

heron/api/src/java/org/apache/heron/streamlet/impl/StreamletImpl.java
nwangtw: Heron has a check already and it runs during submission time. Not as good as compile time but should be fast enough. :)

heron/api/src/java/org/apache/heron/streamlet/impl/operators/SplitOperator.java
```diff
@@ -0,0 +1,67 @@
...
+ * get the output stream id and emit to the specific streams.
+ * Note that one tuple can be emitted to multiple or zero streams.
+ */
+public class SplitOperator<R> extends StreamletOperator<R, R> {
+  private Map<String, SerializablePredicate<R>> splitFns;
+
+  public SplitOperator(Map<String, SerializablePredicate<R>> splitFns) {
+    this.splitFns = splitFns;
+    // Make sure stream ids are not empty
+    for (String stream: splitFns.keySet()) {
```
erenavsarogullari: This is already checked upper level

erenavsarogullari:
Thanks again @nwangtw for all your effort.
Overall looks ok for me and also, split function logic looks as common pattern with Flink.

nwangtw:
Resolved comments, except the UT for wrong stream id. And I am going to add a UT for the new StrealetShadow class.

nwangtw:
- Added stream id check in withStream() function
- Added unit test for calling withStream() with invalid (unavailable) stream id.
- Add unit test for StreamletShadow.

heron/api/src/java/org/apache/heron/streamlet/Streamlet.java
erenavsarogullari: I have just created for this: #3117

PR #3111, opened 2018-11-16
[Java Streamlet API] Extend Validations Part II

erenavsarogullari:
This PR aims to be more robust for `null` and `blank` arguments access by extending validations due to Public API


erenavsarogullari:
cc @nwangtw @nlu90

PR #3112, opened 2018-11-18
[Java Streamlet API] Support abstractions on Streamlet Operators

erenavsarogullari:
This PR aims to increase abstraction scope for Streamlet Operators

`prepare` function can be moved to `abstract StreamletOperator` and `abstract StreamletWindowOperator` classes so their sub-classes(a.k.a `Operators`) can extend it. This saves us from code duplication. Also, if an operator needs, it can still override for custom logic(e.g: `TransformOperator`).

### DIFF
--- a/heron/api/src/java/org/apache/heron/streamlet/impl/operators/TransformOperator.java
+++ b/heron/api/src/java/org/apache/heron/streamlet/impl/operators/TransformOperator.java
@@ -17,7 +17,6 @@
  * under the License.
  */
 
-
 package org.apache.heron.streamlet.impl.operators;
 
 import java.io.Serializable;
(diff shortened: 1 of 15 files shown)


heron/api/src/java/org/apache/heron/streamlet/impl/operators/TransformOperator.java
```diff
@@ -41,10 +40,10 @@
  */
 public class TransformOperator<R, T> extends StreamletOperator<R, T>
     implements IStatefulComponent<Serializable, Serializable> {
+
   private static final long serialVersionUID = 429297144878185182L;
   private SerializableTransformer<? super R, ? extends T> serializableTransformer;
 
-  private OutputCollector collector;
```
nwangtw: I am thinking in the prepare() function, instead of setting collector direction, it might be better to call super.prepare(...) instead. What do you think?

erenavsarogullari: @nwangtw yep, i agree and submitted new patch

PR #3116, opened 2018-11-21
Support utils.Time of kafka-client-spout

thinker0:
latest "storm-kafka-client" module must be supported for Storm Migration.

I think that you can do a lot of migration by supporting Spout which is the most used in Storm.


nwangtw:
Overall LGTM.

Will merge after fixing the ci errors.

thinker0:
@nwangtw fixed

PR #3122, opened 2018-12-08
Support bazel 0.20.0

joshfischer1108:
@kramasamy @nwangtw 

I expect this PR build to break.. I modified our bazel config to work with the newest release of Bazel.. I assuming that the changes I have made will not work with Bazel 0.14.1.  Just a FYI.

This PR allows Heron to be built with the latest version of Bazel (v0.20.0). There are a few file tree changes as well as global JDK arguments that are now passed as global options to each command to target JDK 8 now that Bazel version v0.20.0 defaults to using JDK 9.xxxx.  

The downside of this PR is that everyone who builds the system will have to update their workstation version of Bazel.  ~And CI will of course will have to have it's version  updated as well.~ <-- Took care of this

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -22,6 +22,7 @@ build --python2_path /usr/bin/python2.7
 build --spawn_strategy=standalone
 build --workspace_status_command scripts/release/status.sh
 
+
 # For centos
 # To use it: bazel build --config=centos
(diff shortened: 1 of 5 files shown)


joshfischer1108:
And of course the upside is that we can move away from Bazel version 0.14.1.

PR #3124, opened 2018-12-11
Nwang/fix streamwindowoperator generic types

nwangtw:
Currently the generic types are in different names and orders. For example:

GeneralReduceByKeyAndWindowOperator<K, V, VR>, where K is the key type, V is the incoming type, VR is the result type.

and
ReduceByKeyAndWindowOperator<K, V, R>, where K is the key type, V is the result type and R is the incoming type.

Also custom grouping classes are in the form of <ReturnType, ParameterType> which is the opposite of classes like SerializableFunction.

In this PR, the names and orders are updated to the regular names and order:
When there is only input and output, always use R as input type, T as output type and in <R, T> order.
When a key type is needed, always use R as input type, K as key type, T as output type and in <R, K, T> order.


xiaoyao1991:
Looks like the semantics for `T` and `R` go against the ones used in Java function types, say, [BiFunction](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html) where `T` is considered the input type and `R` as the result type.

nwangtw:
> Looks like the semantics for `T` and `R` go against the ones used in Java function types, say, [BiFunction](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html) where `T` is considered the input type and `R` as the result type.

Yeah. For some reason R is the default generic type in Streamlet API. Like https://github.com/apache/incubator-heron/blob/5ee2490c025c7684c097962cf6d25f54ead73cf4/heron/api/src/java/org/apache/heron/streamlet/impl/streamlets/SupplierStreamlet.java

I can update all of them if we decide to do so.

PR #3126, opened 2018-12-14
Add util function for registerTopologyTimerEvents in Config

nwangtw:
All other config functions have two versions: a static function and a member function, except registerTopologyTimerEvents(). I believe it was forgot and add it in this PR.


xiaoyao1991:
👍

PR #3127, opened 2018-12-18
Add CumulativeCountMetric and metrics to monitor instance task

nwangtw:
An internal user reported that in some edge case, spout instance could stop for a while. So far it is hard to tell the root cause but it seems like the spout task is not triggered. In this PR:
- added a new cumulative count metric.
- refactored SpoutMetrics and BoltMetrics.
- added metrics to monitor spout and bolt task runs and metric collector run.

Tested:
1. Build package and install cli.
2. Start an example topology: ~/.heron/bin/heron submit local   ~/.heron/examples/heron-streamlet-examples.jar   org.apache.heron.examples.streamlet.WindowedWordCountTopology   WindowedWordCountTopology
3. Open metrics log file and verify: vim ~/.herondata/topologies/local/nwang/WindowedWordCountTopology/metrics.json.metricsmgr-1.1
4. Stop topology:  ~/.heron/bin/heron kill local   WindowedWordCountTopology

### DIFF
--- a/heron/common/src/java/org/apache/heron/common/utils/metrics/FullSpoutMetrics.java
+++ b/heron/common/src/java/org/apache/heron/common/utils/metrics/FullSpoutMetrics.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.apache.heron.api.generated.TopologyAPI;
 import org.apache.heron.api.metric.CountMetric;
+import org.apache.heron.api.metric.CumulativeCountMetric;
 import org.apache.heron.api.metric.MeanReducer;
 import org.apache.heron.api.metric.MeanReducerState;
 import org.apache.heron.api.metric.MultiCountMetric;
@@ -41,9 +42,12 @@ import org.apache.heron.common.utils.topology.TopologyContextImpl;
  * 2. New them in the constructor
  * 3. Register them in registerMetrics(...) by using MetricsCollector's registerMetric(...)
  * 4. Expose methods which could be called externally to change the value of metrics
+ *
+ * This is a spout metrics object with more information and it is used in heron core. To change to a faster
+ * metrics object, go to SpoutInstance.java and replace "new FullSpoutMetrcs" to "new SpoutMetrics"
  */
 
-public class FullSpoutMetrics extends SpoutMetrics {
+public class FullSpoutMetrics implements ISpoutMetrics {
   private final MultiCountMetric ackCount;
   private final ReducedMetric<MeanReducerState, Number, Double> tupleSize;
   private final MultiReducedMetric<MeanReducerState, Number, Double> completeLatency;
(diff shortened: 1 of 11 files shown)


heron/common/src/java/org/apache/heron/common/utils/metrics/IBoltMetrics.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Bolt's metrics to be collect
+ * We need to:
+ * 1. Define the metrics to be collected
+ * 2. New them in the constructor
+ * 3. Register them in registerMetrics(...) by using MetricsCollector's registerMetric(...)
+ * 4. Expose methods which could be called externally to change the value of metrics
+ */
+
+public interface IBoltMetrics extends ComponentMetrics {
```
nlu90: Instead of introducing the `IBoltMetrics` and `ISpoutMetrics` interfaces, can we just simply add new methods into the existing `SpoutMetrics` and `BoltMetrics` class?

They are too specific and I don't see a special need to make them general interfaces.

heron/common/src/java/org/apache/heron/common/utils/metrics/FullSpoutMetrics.java
```diff
@@ -41,9 +42,12 @@
  * 2. New them in the constructor
  * 3. Register them in registerMetrics(...) by using MetricsCollector's registerMetric(...)
  * 4. Expose methods which could be called externally to change the value of metrics
+ *
+ * This is a spout metrics object with more information and it is used in heron core. To change to a faster
+ * metrics object, go to SpoutInstance.java and replace "new FullSpoutMetrcs" to "new SpoutMetrics"
  */
 
-public class FullSpoutMetrics extends SpoutMetrics {
+public class FullSpoutMetrics implements ISpoutMetrics {
```
nlu90: The original inheritance relationships is `FullSpoutMetrics extends SpoutMetrics implements ComponentMetrics`. But with the change here, the relationship chain is separate and now are `FullSpoutMetrics implements ISpoutMetrics` and `SpoutMetrics implements ISpoutMetrics`.

Many metrics originally available in `FullSpoutMetrics` which are inherited from `SpoutMetrics` are now missing.

heron/common/src/java/org/apache/heron/common/utils/metrics/IBoltMetrics.java
nwangtw: That is what I thought. :D 

But it turns out that FullSpoutMetrics is a re-implementation (focusing on accuracy and extra information) of SpoutMetrics and they are in parallel instead of parent-child.

heron/common/src/java/org/apache/heron/common/utils/metrics/FullSpoutMetrics.java
nwangtw: There is no inheritance between them.

PR #3129, opened 2018-12-19
ByteAmount String Format

xiaoyao1991:
Make ByteAmount string more accurate.


heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
```diff
@@ -23,12 +23,15 @@
  * Class that encapsulates number of bytes, with helpers to handle units properly.
  */
 public final class ByteAmount implements Comparable<ByteAmount> {
+  private static final long KB = 1024L;
   private static final long MB = 1024L * 1024;
```
nwangtw: "KB * 1024" might be better.

xiaoyao1991: 👍 addressed

PR #3132, opened 2018-12-21
Add sum/min/max reducers

nwangtw:
It is tricker to provide a sum() operation in Streamlet directly because the operation only works for specific types. In this PR, a few standard reducers are added for users to make it simple to implement sum.

Example:
instead of:
s.reduceByKey(x -> ..., x -> ..., (x, y) -> x + y);

user can do it like:
s.reduceByKey(x -> ..., x -> ..., StreamletReducers::sum);

### DIFF
--- a/examples/src/java/org/apache/heron/examples/streamlet/WindowedWordCountTopology.java
+++ b/examples/src/java/org/apache/heron/examples/streamlet/WindowedWordCountTopology.java
@@ -28,6 +28,7 @@ import org.apache.heron.examples.streamlet.utils.StreamletUtils;
 import org.apache.heron.streamlet.Builder;
 import org.apache.heron.streamlet.Config;
 import org.apache.heron.streamlet.Runner;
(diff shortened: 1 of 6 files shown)


xiaoyao1991:
Can we use `Math.max` and `Math.min` instead?

examples/src/java/org/apache/heron/examples/streamlet/WindowedWordCountTopology.java
```diff
@@ -71,7 +71,7 @@ public static void main(String[] args) throws Exception {
             // Value extractor (the value is always 1)
             word -> 1,
             WindowConfig.TumblingCountWindow(50),
-            (x, y) -> x + y
+            StreamletReducers::sum
```
xiaoyao1991: missing import

xiaoyao1991:
:+1:

xiaoyao1991:
:+1:

PR #3135, opened 2018-12-27
Refactor KVStreamlet to make it lighter and easier to convert from St…

nwangtw:
Originally KVStreamletImpl is a child class of StreamletImpl. More explicitly:
class KVStreamletImpl<K, V> extends StreamletImpl<KeyValue<K, V>>

The goal was to be able to add special functions in KVStreamlet.

A big blocker in this structure is that it is impossible to convert a StreamletImpl<KeyValue<K, V>> object (such as the MapStreamlet returned by map() operation) to KVStreamletImpl<K, V> although they contain the same data.

In this PR, KVStreamletImpl is replaced by KVStreamletShadow, which is a wrapper for StreamletImpl<KeyValue<K, V>> and converting StreamletImpl<KeyValue<K, V>> to KVStreamlet is as simple as "new KVStreamletShadow(stream)".

Also, after this change, StreamletImpl is the base class of all Streamlet classes again and the class structure is cleaner.

Note that StreamletImpl and KVStreamletImpl classes are wrappers already in Scala. So no change is needed.


xiaoyao1991:
👍

PR #3142, opened 2019-01-02
Validate resource constraint (RAM and CPU) in RoundRobinPacking

xiaoyao1991:
**This PR might break some mis-configured topologies in production**

This PR addresses some issues raised previously in the slack chat that the packing algorithm did not honor container-level and instance-level constraints on CPU resource. Additionally, we added resource constraint validation in RoundRobinPacking. Specifically:
- `topology.container.cpu` and `topology.container.ram` are honored and considered as the **hard cap** constraint for the corresponding resource in **one** container. 
- `topology.component.cpumap` and `topology.component.rammap` are honored and considered as the instance resource mapping in **one** container.
- The sum of instance resources in one container, plus padding, should not exceed the container-level constraint, otherwise the scheduling would fail.

### DIFF
--- a/examples/src/java/org/apache/heron/examples/api/ExampleResources.java
+++ b/examples/src/java/org/apache/heron/examples/api/ExampleResources.java
@@ -23,7 +23,9 @@ import org.apache.heron.common.basics.ByteAmount;
 
 public final class ExampleResources {
 
+  static final long DEFAULT_RAM_PADDING_PER_CONTAINER = 2 * 1024;
   static final long COMPONENT_RAM_MB = 1024;
+  static final double DEFAULT_CPU_PADDING_PER_CONTAINER = 1.0;
 
   static ByteAmount getComponentRam() {
     return ByteAmount.fromMegabytes(COMPONENT_RAM_MB);
@@ -35,12 +37,13 @@ public final class ExampleResources {
 
   static ByteAmount getContainerRam(int components, int containers) {
     final int componentsPerContainer = Math.max(components / containers, 1);
-    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * (componentsPerContainer + 2));
+    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * componentsPerContainer
+        + DEFAULT_RAM_PADDING_PER_CONTAINER);
   }
 
   static double getContainerCpu(int components, int containers) {
     final int componentsPerContainer = Math.max(components / containers, 1);
-    return componentsPerContainer + 1.0;
+    return componentsPerContainer + DEFAULT_CPU_PADDING_PER_CONTAINER;
   }
 
   private ExampleResources() {
--- a/examples/src/java/org/apache/heron/examples/api/ExampleResources.java
+++ b/examples/src/java/org/apache/heron/examples/api/ExampleResources.java
@@ -35,7 +35,12 @@ public final class ExampleResources {
 
   static ByteAmount getContainerRam(int components, int containers) {
     final int componentsPerContainer = Math.max(components / containers, 1);
-    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * componentsPerContainer);
+    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * (componentsPerContainer + 2));
+  }
+
+  static double getContainerCpu(int components, int containers) {
+    final int componentsPerContainer = Math.max(components / containers, 1);
+    return componentsPerContainer + 1.0;
   }
 
   private ExampleResources() {
--- a/heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
+++ b/heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
@@ -22,7 +22,7 @@ package org.apache.heron.common.basics;
 /**
  * Class that encapsulates number of bytes, with helpers to handle units properly.
  */
-public final class ByteAmount implements ResourceMeasure<ByteAmount> {
+public final class ByteAmount extends ResourceMeasure<ByteAmount, Long> {
   private static final long KB = 1024L;
   private static final long MB = KB * 1024;
   private static final long GB = MB * 1024;
@@ -34,10 +34,9 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
   private static final long MAX_KB = Math.round(Long.MAX_VALUE / KB);
 
   public static final ByteAmount ZERO = ByteAmount.fromBytes(0);
-  private final long bytes;
 
-  private ByteAmount(long bytes) {
-    this.bytes = bytes;
+  private ByteAmount(Long value) {
+    super(value);
   }
 
   /**
@@ -67,7 +66,8 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
 
   /**
    * Creates a ByteAmount value in gigabytes. If the gigabytes value
-   * is &gt;= Long.MAX_VALUE / 1024 / 1024 / 1024, the byte representation is capped at Long.MAX_VALUE.
+   * is &gt;= Long.MAX_VALUE / 1024 / 1024 / 1024,
+   * the byte representation is capped at Long.MAX_VALUE.
    *
    * @param gigabytes value in gigabytes to represent
    * @return a ByteAmount object repressing the number of GBs passed
@@ -85,7 +85,7 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return number of bytes
    */
   public long asBytes() {
-    return bytes;
+    return super.getValue();
   }
 
   /**
@@ -97,7 +97,7 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return returns the ByteValue in MBs or 0 if the value is &lt; (1024 * 1024) / 2
    */
   public long asMegabytes() {
-    return Math.round((double) bytes / MB);
+    return Math.round(value.doubleValue() / MB);
   }
 
   /**
@@ -109,7 +109,7 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return returns the ByteValue in KBs or 0 if the value is &lt; (1024) / 2
    */
   public long asKilobytes() {
-    return Math.round((double) bytes / KB);
+    return Math.round(value.doubleValue() / KB);
   }
 
   /**
@@ -121,15 +121,7 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return returns the ByteValue in GBs or 0 if the value is &lt; (1024 * 1024 * 1024) / 2
    */
   public long asGigabytes() {
-    return Math.round((double) bytes / GB);
-  }
-
-  /**
-   * Convenience methdod to determine if byte value is zero
-   * @return true if the byte value is 0
-   */
-  public boolean isZero() {
-    return ZERO.equals(this);
+    return Math.round(value.doubleValue() / GB);
   }
 
   /**
@@ -138,10 +130,11 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return a new ByteValue of this minus other ByteValue
    * @throws IllegalArgumentException if subtraction would overshoot Long.MIN_VALUE
    */
+  @Override
   public ByteAmount minus(ByteAmount other) {
-    checkArgument(Long.MIN_VALUE + other.asBytes() <= asBytes(), String.format(
-        "Subtracting %s from %s would overshoot Long.MIN_LONG", other, this));
-    return ByteAmount.fromBytes(asBytes() - other.asBytes());
+    checkArgument(Long.MIN_VALUE + other.value <= value,
+        String.format("Subtracting %s from %s would overshoot Long.MIN_LONG", other, this));
+    return ByteAmount.fromBytes(value - other.value);
   }
 
   /**
@@ -150,10 +143,11 @@ public final class ByteAmount implements ResourceMeasure<ByteAmount> {
    * @return a new ByteValue of this plus other ByteValue
    * @throws IllegalArgumentException if addition would exceed Long.MAX_VALUE
    */
+  @Override
   public ByteAmount plus(ByteAmount other) {
-    checkArgument(Long.MAX_VALUE - asBytes() >= other.asBytes(), String.format(
-        "Adding %s to %s would exceed Long.MAX_LONG", other, this));
-    return ByteAmount.fromBytes(asBytes() + other.asBytes());
+    checkArgument(Long.MAX_VALUE - value >= other.value,
+        String.format("Adding %s to %s would exceed Long.MAX_LONG", other, this));
+    return ByteAmount.fromBytes(value + other.value);
   }
 
   /**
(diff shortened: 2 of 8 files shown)


heron/common/src/java/org/apache/heron/common/basics/CPUShare.java
```diff
@@ -0,0 +1,125 @@
...
+  public int hashCode() {
+    return cpuShare.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other == null || getClass() != other.getClass()) {
```
nwangtw: "!(other instanceof CPUShare))" might be safer.

heron/common/src/java/org/apache/heron/common/basics/CPUShare.java
```diff
@@ -0,0 +1,125 @@
...
+ * under the License.
+ */
+
+package org.apache.heron.common.basics;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class CPUShare implements ResourceMeasure<CPUShare> {
+  private final Double cpuShare;
```
nwangtw: "share" or "value" might be better than "cpuShare" I feel. "cpuShare" feels more redundant. Not a hard request though.

heron/common/src/java/org/apache/heron/common/basics/CPUShare.java
```diff
@@ -0,0 +1,125 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.heron.common.basics;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class CPUShare implements ResourceMeasure<CPUShare> {
```
nwangtw: Can we move the implementation code into ResourceMeasure<T> and declare CPUShare as:

public final class CPUShare extends  ResourceMeasure<Double> {
}

And ByteAmount can be:
public final class ByteAmount extends  ResourceMeasure<Integer> {
}

The util function below might be a lot simpler this way.

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -133,16 +143,36 @@ private PackingPlan packInternal(int numContainer, Map<String, Integer> parallel
...
     ByteAmount containerDiskInBytes = getContainerDiskHint(roundRobinAllocation);
-    double containerCpu = getContainerCpuHint(roundRobinAllocation);
+    double containerCpuHint = getContainerCpuHint(roundRobinAllocation);
     ByteAmount containerRamHint = getContainerRamHint(roundRobinAllocation);
 
-    LOG.info(String.format("Pack internal: container CPU hint: %f, RAM hint: %s, disk hint: %s.",
-        containerCpu,
-        containerDiskInBytes.toString(),
-        containerRamHint.toString()));
+    LOG.info(String.format("Pack internal: container CPU hint: %.1f, RAM hint: %s, disk hint: %s.",
```
nwangtw: "%.3f" is more consistent to the toString function in CPUShare. Or maybe wrap it into CPUShare and call the toString() function like the containerRamHint var.

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -367,16 +410,16 @@ private ByteAmount getContainerRamHint(Map<Integer, List<InstanceId>> allocation
 
     return TopologyUtils.getConfigWithDefault(
         topologyConfig, org.apache.heron.api.Config.TOPOLOGY_CONTAINER_RAM_REQUESTED,
-        NOT_SPECIFIED_NUMBER_VALUE);
+        NOT_SPECIFIED_BYTE_AMOUNT);
   }
 
   /**
-   * Check whether the PackingPlan generated is valid
+   * Check whether the PackingPlan configured each instance to have at least minimum required RAM
```
nwangtw: There might be more checks in the function, like CPU and disk and maybe others. I feel validatePackingPlan() makes more sense here.

heron/common/src/java/org/apache/heron/common/basics/ResourceMeasure.java
```diff
@@ -0,0 +1,43 @@
...
+
+  T plus(T other);
+
+  T multiply(int factor);
+
+  T divide(int factor);
+
+  T increaseBy(int percentage);
+
+  boolean greaterThan(T other);
```
nlu90: The `Comparable` interface already provides the `compareTo` method. Do we really need these 4 comparison methods?

heron/common/src/java/org/apache/heron/common/basics/ResourceMeasure.java
```diff
@@ -0,0 +1,43 @@
...
+
+public interface ResourceMeasure<T> extends Comparable<T> {
+
+  boolean isZero();
+
+  T minus(T other);
+
+  T plus(T other);
+
+  T multiply(int factor);
```
nlu90: why the `factor` and `percentage` are `int`?

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -211,69 +248,75 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
...
-        if (!containerRamHint.equals(NOT_SPECIFIED_NUMBER_VALUE)) {
-          // remove RAM for heron internal process
-          ByteAmount remainingRam =
-              containerRamHint.minus(containerRamPadding).minus(usedRam);
+          if (remainingRes.lessOrEqual(zero)) {
+            throw new PackingException(String.format("Invalid packing plan generated. "
+                + "No enough %s to allocate for unspecified instances", resourceType));
+          }
 
           // Split remaining RAM evenly
```
nlu90: update comment

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -211,69 +248,75 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
...
-          individualInstanceRam = remainingRam.divide(instancesToAllocate);
+          individualInstanceRes = (T) remainingRes.divide(unspecifiedInstances.size());
         }
 
         // Put the results in instancesRam
```
nlu90: update comment

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -85,23 +87,31 @@
   private static final Logger LOG = Logger.getLogger(RoundRobinPacking.class.getName());
 
   // TODO(mfu): Read these values from Config
```
nlu90: update the TODO

heron/common/src/java/org/apache/heron/common/basics/CPUShare.java
nlu90: Do we really need to create a  separate `CPUShare` class to handle the CPU resource calculation? Is that possible to reuse what we already have in the ram resource calculation? Or will there be `RamShare` and `DiskShare` classes in following patches?

xiaoyao1991: It doesn't seem necessary to separate `RamShare` and `DiskShare` as RAM and disk are both measured in `ByteAmount`. CPU is measured in share of time used. 

Previously, we didn't take CPU resource constraints into consideration when composing packing plan. Now that we do -- namely we need to collect instance cpu resource mapping,  and the logic of that is very similar to that of instance RAM resource mapping (previously as `getInstancesRAMMapInContainer()`). But due to the difference in the type of measurement(`ByteAmount` vs `double`), we might end up with a lot of code duplicates. Hence, I abstract out `CPUShare` so that it looks very similar to what we have in `ByteAmount`, and thus we can have one `calculateInstancesResourceMapInContainer()` that works for both CPU and RAM (potentially disk if needed) instance resource mapping collection.

xiaoyao1991: 👍 Updated.

heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
```diff
@@ -221,42 +202,37 @@ public ByteAmount max(ByteAmount other) {
...
-    return Long.compare(asBytes(), other.asBytes());
-  }
-
   @Override
   public boolean equals(Object other) {
     if (this == other) {
       return true;
     }
-    if (other == null || getClass() != other.getClass()) {
+    if (!(other instanceof ByteAmount)) {
```
nwangtw: null check is still needed.

heron/common/src/java/org/apache/heron/common/basics/CPUShare.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    return retval;
+  }
+
+  @Override
+  public int hashCode() {
+    return value.hashCode();
+  }
+
+  @Override
```
nwangtw: can this function be implemented in base class?

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -288,7 +330,7 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
     }
 
     for (int i = 1; i <= numContainer; ++i) {
-      allocation.put(i, new ArrayList<InstanceId>());
+      allocation.put(i, new ArrayList<>());
```
nwangtw: this change seems to be unnecessary?

heron/packing/tests/java/org/apache/heron/packing/roundrobin/RoundRobinPackingTest.java
```diff
@@ -158,8 +177,12 @@ public void testContainerRequestedResources() throws Exception {
       Assert.assertEquals(1, resources.size());
       int instancesCount = containerPlan.getInstances().size();
       Assert.assertEquals(containerRam
-          .minus(RoundRobinPacking.DEFAULT_RAM_PADDING_PER_CONTAINER).divide(instancesCount),
+              .minus(RoundRobinPacking.DEFAULT_RAM_PADDING_PER_CONTAINER).divide(instancesCount),
```
nwangtw: this indent change is needed?

heron/packing/tests/java/org/apache/heron/packing/roundrobin/RoundRobinPackingTest.java
```diff
@@ -255,6 +282,41 @@ public void testCompleteRamMapRequested() throws Exception {
     Assert.assertEquals(totalInstances, packingPlanExplicitRamMap.getInstanceCount());
   }
 
+  /**
+   * Test the scenario CPU map config is completely set
+   */
+  @Test
+  public void testCompleteCpuMapRequested() throws Exception {
```
nwangtw: Can we add these cases:
- container cpu is more than needed by instances.
- container cpu is less than needed by instances.

also, other processes in container could need a fixed amount of CPU as well. So if container cpu == needed by instances, it might be better to return false.

heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
xiaoyao1991: Looks like `instanceof` automatically does null check, so it's redundant

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
xiaoyao1991: The IDE suggests the type declaration here is unnecessary. I figure just to clean it up

nwangtw: kk. sgtm.

heron/packing/tests/java/org/apache/heron/packing/roundrobin/RoundRobinPackingTest.java
xiaoyao1991: 👍 Added more tests to cover the mentioned scenarios

heron/common/src/java/org/apache/heron/common/basics/ByteAmount.java
nwangtw: The code below prints Yes.

     Thread x = null;
      if (x instanceof Thread) {
        System.out.println("Yes!");
      }

So when other is null, "if (!(other instanceof ByteAmount))" returns false, and .value will be called later.

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -133,16 +142,36 @@ private PackingPlan packInternal(int numContainer, Map<String, Integer> parallel
...
+        CPUShare.fromDouble(getContainerCpuHint(roundRobinAllocation)),
+        CPUShare.fromDouble(instanceCpuDefault),
+        CPUShare.fromDouble(containerCpuPadding),
+        CPUShare.fromDouble(0.0),
+        CPUShare.fromDouble(NOT_SPECIFIED_CPU_SHARE),
+        CPU);
 
     ByteAmount containerDiskInBytes = getContainerDiskHint(roundRobinAllocation);
-    double containerCpu = getContainerCpuHint(roundRobinAllocation);
+    double containerCpuHint = getContainerCpuHint(roundRobinAllocation);
```
nwangtw: containerCpuHint (configured container cpu, or max (padding + 1 per instance)) is not used. As the result, I think user specify container cpu is ignored. Maybe need a max(containerCpuHint, containerCpu) I feel?

xiaoyao1991: It is actually used down in `calculateInstancesResourceMapInContainer` as the `containerResHint`. It is used to validate the all instances + padding in `roundRobinAllocation` does not exceed the `containerResHint`. It's just not used anymore in `packingInternal`, similar to how we treated `containerRamHint` before the change

xiaoyao1991: hmm. I see. Then I think we don't even need to take the max out of 2. ContainerCpuHint is always gonna be greater no matter what.

nwangtw: But if container cpu is not configured, the getContainerCpuHint() would return padding + 1 * instance number, which is smaller than your new value, unless you update the getContainerCpuHint() function.

xiaoyao1991: Yeh you are right, I just figured that situation :) Updated

nwangtw: The original logic is messy and makes it hard to update. :(

examples/src/java/org/apache/heron/examples/api/ExampleResources.java
```diff
@@ -35,7 +35,12 @@ static ByteAmount getContainerDisk(int components, int containers) {
 
   static ByteAmount getContainerRam(int components, int containers) {
     final int componentsPerContainer = Math.max(components / containers, 1);
-    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * componentsPerContainer);
+    return ByteAmount.fromMegabytes(COMPONENT_RAM_MB * (componentsPerContainer + 2));
```
nlu90: give `2` a meaningful name instead of a magic number

PR #3143, opened 2019-01-03
Fix integration test script on mac

nwangtw:
It seems like Python on mac doesn't like '*' in command line. Change regex to avoid '*'.


xiaoyao1991:
👍

PR #3147, opened 2019-01-09
[UI] Move reset view button to left and change plan graph listener fr…

nwangtw:
…om click to dblclick

Motivation:
- "Reset View" button is hard to notice when it is aligned to the right side.
- click event is used to reset view but it is quite easy to mis-click, which is not intuitive. Also it is hard to select text on the page because the reset view is triggered on mouse up. Change the event to double click to avoid mis-click.

### DIFF
--- a/heron/tools/ui/resources/static/js/logical-plan.js
+++ b/heron/tools/ui/resources/static/js/logical-plan.js
@@ -393,6 +393,7 @@
           return d.color;
         })
         .on("click", planController.logicalComponentClicked)
+        .on("dblclick", planController.logicalComponentClicked)
         .on("mouseover", planController.logicalComponentHoverOver)
         .on("mouseout", planController.logicalComponentHoverOut);
 
@@ -402,6 +403,7 @@
         .attr("y", function (d) { return d.cy - d.r - 10; })
         .attr("class", "fade")
         .style("text-anchor", "middle")
+        .style("user-select", "all")
         .text(function (d) {
           if (d.isReal) {
             return d.name;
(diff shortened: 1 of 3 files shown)


heron/tools/ui/resources/static/js/logical-plan.js
```diff
@@ -402,6 +403,7 @@
         .attr("y", function (d) { return d.cy - d.r - 10; })
         .attr("class", "fade")
         .style("text-anchor", "middle")
+        .style("user-select", "all")
```
huijunw: what is this for?

heron/tools/ui/resources/static/js/plan-controller.js
```diff
@@ -184,8 +184,8 @@ function PlanController(baseUrl, cluster, environ, toponame, physicalPlan, logic
     highlightFocusedElements();
   };
 
-  // treat any click that bubbles up unhandled as intent to clear the focus
-  $(document).on('click', ".plans .graphics", function () {
+  // treat double click as intent to clear the focus
```
huijunw: besides comments, shall we add the instruction somewhere, so that users can easily find it and know how to operate the UI webpage?

heron/tools/ui/resources/static/js/logical-plan.js
nwangtw: select all text on click

heron/tools/ui/resources/static/js/plan-controller.js
nwangtw: We dont have an instruction page for heorn UI I think currently. Could be useful to have one though.

xiaoyao1991:
👍

PR #3150, opened 2019-01-10
Add a release-yaml-file argument to CLI

nwangtw:
In Heron UI, there is a "version" string for each topology. Currently the information is from a release.yaml file in CLI so it is just the CLI release information. The information is useful but users might want to have more in UI.

An extra "release-yaml-file" argument is added to CLI in this PR. User can generate a release info file and pass it to CLI and the version info will be displayed in UI.

~/.heron/bin/heron submit --release-yaml-file MY_OWN_RELEASE_INFO.yaml local   ~/.heron/examples/heron-streamlet-examples.jar   org.apache.heron.examples.streamlet.WindowedWordCountTopology   WindowedWordCountTopology

### DIFF
--- a/heron/tools/cli/src/python/args.py
+++ b/heron/tools/cli/src/python/args.py
@@ -165,6 +165,18 @@ def add_extra_launch_classpath(parser):
       help='Additional JVM class path for launching topology')
   return parser
 
+def add_release_yaml_file(parser):
+  '''
+  :param parser:
+  :return:
+  '''
+  parser.add_argument(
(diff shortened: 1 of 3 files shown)


huijunw:
to confirm, MY_OWN_RELEASE_INFO.yaml follows the same format as the CLI release.yaml?

nwangtw:
> to confirm, MY_OWN_RELEASE_INFO.yaml follows the same format as the CLI release.yaml?

Yes.

PR #3151, opened 2019-01-10
Update topology UI to make topology info more consistent across pages

nwangtw:
Currently the "all topologies" page and "topology" page are not consistent:
- it is "cluster" in one page and "dc" in the other
- there is no version info in the "topology" page
- "submitted by" vs "launched by"
- column order is different. I think "cluster" -> "role" -> "env" is the correct order.


xiaoyao1991:
👍

PR #3152, opened 2019-01-10
Fix RoundRobinPacking repack with no specified numContainers

xiaoyao1991:
In `RoundRobinPacking`, calling `repack` without providing a new `numContainer` should automatically determine the `numContainers` based on the updated parallelism while remaining the number of instances per container <= that of the old packing plan. The method is currently not following this logic.

### DIFF
--- a/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
@@ -390,10 +390,18 @@ public class RoundRobinPacking implements IPacking, IRepacking {
     }
   }
 
-  /*
-   * read the current packing plan with update parallelism to calculate a new packing plan
-   * the packing algorithm packInternal() is shared with pack()
+  /**
+   * Read the current packing plan with update parallelism to calculate a new packing plan.
+   * This method should determine a new number of containers based on the updated parallism
+   * while remaining the number of instances per container <= that of the old packing plan.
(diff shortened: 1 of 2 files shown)


heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -402,11 +410,10 @@ public PackingPlan repack(PackingPlan currentPackingPlan, Map<String, Integer> c
     int initialNumInstance = TopologyUtils.getTotalInstance(topology);
     double initialNumInstancePerContainer = (double) initialNumInstance / initialNumContainer;
 
-    Map<String, Integer> currentComponentParallelism = currentPackingPlan.getComponentCounts();
     Map<String, Integer> newComponentParallelism =
         getNewComponentParallelism(currentPackingPlan, componentChanges);
 
-    int newNumInstance = TopologyUtils.getTotalInstance(currentComponentParallelism);
+    int newNumInstance = TopologyUtils.getTotalInstance(newComponentParallelism);
```
nwangtw: this is a fix of some issue?

xiaoyao1991: Yes, this is exactly where the problem is. Previously, it is wrongfully using the old plan's component parallelism to calculate the new number of containers, but that's always going to return the same number of containers as the old plan...

nwangtw: KK.

PR #3153, opened 2019-01-11
Update metrics section in Heron UI

nwangtw:
When instance is selected, move instance name to title from bottom.


xiaoyao1991:
👍

PR #3155, opened 2019-01-13
Add tooltip for component connections

nwangtw:
Stream id and grouping strategy are included in the connection tooltip.


xiaoyao1991:
LGTM

PR #3157, opened 2019-01-14
Support instance healthcheck api

thinker0:
Support health check API of Mesos/Aurora/Marathon.

When I use Aurora, sometimes it is often the case that the Instance is not healthy.

* http://aurora.apache.org/documentation/latest/reference/configuration/#healthcheckconfig-objects
* https://mesos.readthedocs.io/en/latest/health-checks/
* https://mesosphere.github.io/marathon/docs/health-checks.html


nwangtw:
LGTM. @huijunw what do you think?

PR #3162, opened 2019-01-18
keep executor running when `heron update` adds containers

huijunwu:
The present `heron update` scale process:
1. deactivate topology
2. call `scheduler add containers`, which returns the new added container id
3. update the repacking plan (replace the assumed container id with actual container id)
4. send repacking plan to zk
5. the new container heron-executor gets the new repacking plan and update its heron-instances.

Observed race condition
In the log, found that the step 2 never returned, then the process timeout.
In the step 2, `scheduler add containers` tried to check the new container `wait-until running`. However, the new container heron-executor assertion failure and scheduler thought `--wait-until running` condition was not satisfied, and still in `waiting` status.
On the other hand, if the step 4 is not performed, the new container heron-executor cannot recover.
thus step 2 depends on step 4, while step 4 depends on step 2, which is race condition.

This PR let step 2 return by letting executor runs healthy


heron/executor/src/python/heron_executor.py
```diff
@@ -976,6 +979,14 @@ def get_commands_to_run(self):
     # During shutdown the watch might get triggered with the empty packing plan
     if len(self.packing_plan.container_plans) == 0:
       return {}
+    if self._get_instance_plans(self.packing_plan, self.shard) is None and self.shard != 0:
```
nwangtw: what is this for?

heron/executor/src/python/heron_executor.py
```diff
@@ -844,8 +844,11 @@ def _get_instance_plans(self, packing_plan, container_id):
       if container_plan.id == container_id:
         this_container_plan = container_plan
 
-    # make sure that our shard id is a valid one
-    assert this_container_plan is not None
+    # When the executor runs in newly added container by `heron update`,
+    # there is no plan for this container. In this situation,
+    # return None to bypass instance processes.
+    if this_container_plan is None:
+      return None
```
nwangtw: currently, executor uses the list to start instances, I think. If we skip here, where is this function called again to get the list and start the instances?

huijunwu: the new repacking plan triggers the executor to launch the actual instnaces

heron/executor/src/python/heron_executor.py
huijunwu: if there is no container meta for this shard, keep shell only for health-check

heron/executor/src/python/heron_executor.py
```diff
@@ -976,6 +979,14 @@ def get_commands_to_run(self):
     # During shutdown the watch might get triggered with the empty packing plan
     if len(self.packing_plan.container_plans) == 0:
       return {}
+    if self._get_instance_plans(self.packing_plan, self.shard) is None and self.shard != 0:
+      retval = {}
+      retval['heron-shell'] = [
+        '%s' % self.heron_shell_binary,
```
nlu90: why in this case a `heron-shell` command is returned?

huijunwu: health check in shell determines if aurora returns

heron/executor/src/python/heron_executor.py
nwangtw: kk. thx.

PR #3164, opened 2019-01-18
remove heron-core.tar.gz from docker image

yaoliclshlmch:
Remove /usr/local/heron/dist/heron-core.tar.gz from docker images to reduce image size.

@nwangtw @nlu90 @huijunwu ping for review.

### DIFF
--- a/docker/dist/Dockerfile.dist.ubuntu14.04
+++ b/docker/dist/Dockerfile.dist.ubuntu14.04
@@ -35,7 +35,8 @@ ADD artifacts /heron
 WORKDIR /heron
 
 # run heron installer
-RUN /heron/heron-install.sh
+RUN /heron/heron-install.sh \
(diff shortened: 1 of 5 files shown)


docker/dist/Dockerfile.dist.ubuntu14.04
```diff
@@ -35,7 +35,8 @@ ADD artifacts /heron
 WORKDIR /heron
 
 # run heron installer
-RUN /heron/heron-install.sh
+RUN /heron/heron-install.sh \
+    && rm -f /usr/local/heron/dist/heron-core.tar.gz
```
nwangtw: should "&& rm -f /heron/heron-install.sh \" be added as well?

yaoliclshlmch: It is in Line 60.

PR #3169, opened 2019-01-24
Add a spouts directory and README for spout implementations

nwangtw:
Heron community has been interested in maintaining spout implementations for users. There have been discussions in slack channel (#general) and dev@ mailing list.

There are multiple options and each has its own pros and cons. This PR is just a start of the effort to unblock contributors. The directory might be migrated to a different place after a final decision is made.

Please feel free to comment or join the discussion in dev@heron.apache.org.

### DIFF
--- a/contrib/bolts/README.md
+++ b/contrib/bolts/README.md
@@ -1,0 +1,36 @@
+
+This directory contains implementations of Heron bolts that can be used in user topologies.
+
+The bolt implementations are maintained by Heron community.
+
+## Bolt Development
+
+- https://apache.github.io/incubator-heron/docs/developers/python/bolts/
+- https://apache.github.io/incubator-heron/docs/developers/java/bolts/
+
+
+## Requirements
+
+### Directories and Files
+
+Bolt implementation files should be organized in this directory structure:
+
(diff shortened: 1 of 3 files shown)


huijunw:
there was a contrib dir, which included spouts.
https://github.com/apache/incubator-heron/tree/0.15.0/contrib
i think we may recover the contrib dir, and incubate codes such as spouts from there till they are migrated to other places. thoughts?

nwangtw:
> there was a contrib dir, which included spouts.
> https://github.com/apache/incubator-heron/tree/0.15.0/contrib
> i think we may recover the contrib dir, and incubate codes such as spouts from there till they are migrated to other places. thoughts?

Cool. Thanks!

nwangtw:
I am thinking maybe we should have both spouts and bolts. Both are for reusable components anyway. For example KafkaSpout reads from Kafka and KafkaBolt could be the writer.

simingweng:
@nwangtw definitely make sense to have both Spout and Bolt, Apache Storm has the same thing.

nwangtw:
@simingweng thx for the input!

PR #3172, opened 2019-01-25
Avoid squash docker image (an experimental feature)

nwangtw:


nwangtw:
image size is 400m bigger. :(  abort

PR #3174, opened 2019-01-25
Add build scripts for ci jobs

nwangtw:
Create scripts to be used by Heron release jenkins jobs.

### DIFF
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -1,0 +1,37 @@
+# Heron CI scripts
+This directory contains scripts used by CI jobs.
+
+## Build release packages
+
+Release packages include heron.tar.gz as well as installer script. Packages are platform dependent.
+
+Example:
+
+```bash
+set -e
+set -o pipefail
+
+# Install bazel (linux build) because CI hosts may not have it installed
+sh scripts/ci/setup_bazel.sh linux
+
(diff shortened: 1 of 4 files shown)


nwangtw:
Yeah. The travis folder has a few ci scripts. There differences are:
- the travis job try to build everything and then run integration tests. It is for continuous integration, not for builds. Also in the scripts there are some travis specific code.
- The scripts in this folder are general and can be easily used by different CI engines (like the description in README). It is for release only currently but will have some for testing.

We will need to migration the integration tests travis job to jenkins. I am thinking of creating a few new scripts for integration tests. Eventually the jenkins/travis folder (if there still is one) will be the config only, no scripts.

huijunw:
Agree the the ultimate goal is to unite CIs in one place. This may make the CI scripts more organized and easy to maintain. Besides scripts/travis/, scripts/applatix/ should be migrated to scripts/ci/ as well. For configs, scripts/ci/config is better than scripts/travis/.
To confirm, scripts/ci/ will handle all CIs including travis, jenkins, applatix; and no more scripts/travis/ or scripts/applatix/ ?

nwangtw:
"For configs, scripts/ci/config is better than scripts/travis/."

Agreed. 

Eventually yeah. jenkins and applatix will be using the same set of scripts with different configs.

PR #3175, opened 2019-01-26
Fix memory leak in stmgr when measuring tuple data size

nwangtw:
this pr fixed the `__global_protobuf_pool_acquire__` in that if/else branch


thinker0:
Nice +++++1

PR #3176, opened 2019-01-26
Another memory leak in stmgr-client

nwangtw:
Another lead in stmgr-client needs a fix.

This is related to the following PR:
https://github.com/apache/incubator-heron/pull/3175


huijunwu:
#3175

PR #3178, opened 2019-01-28
Soften padding constraint

xiaoyao1991:
To alleviate burdens on users to reconfigure topologies with the correct amount of resources, we will soften the padding constraint. i.e. we will only do hard validation to see if user configured container resource can accommodate all instances without padding. If there's no enough resource left for padding, we will only throw warning messages to alert users instead of failing the submission.

### DIFF
--- a/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
@@ -20,12 +20,14 @@
 package org.apache.heron.packing.roundrobin;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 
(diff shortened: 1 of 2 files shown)


heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -281,12 +294,24 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
...
-                + "Total instance %s in a container (%s) have exceeded "
+                + "Total instance %s (%s) in container#%d have exceeded "
                 + "the container-level constraint of %s.",
-            resourceType, usedRes.toString(), containerResHint));
+            resourceType, usedRes.toString(), containerId, containerResHint.toString()));
+      }
+
+      // Soft padding constraint validation: warn if padding amount cannot be accommodated
+      boolean paddingThrottling = false;
+      if (!containerResHint.equals(notSpecified)
```
nwangtw: this block can be refactored and combined with the "if (!containerResHint.equals(notSpecified) && usedRes.greaterThan(containerResHint))" above.

heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -347,6 +381,24 @@ private ByteAmount getContainerRamPadding(List<TopologyAPI.Config.KeyValue> topo
     return allocation;
   }
 
+
+  /**
+   * Sort the components in decreasing order based on their RAM requirements
+   *
+   * @return The sorted list of components and their RAM requirements
+   */
+  private ArrayList<RamRequirement> getSortedRAMInstances(Set<String> componentNames) {
```
nwangtw: - should be "components"
- maybe better to document why Ram only.

PR #3184, opened 2019-02-01
Consolidate packing code

xiaoyao1991:
Currently, different packing algorithms assume different things and follow different semantics for the user-set configs. Because of it, there exist some code duplicates and it's not friendly for new packing algorithms. This PR is an attempt to address this issue and consolidate the three existing packing algorithms. 

Major behavior changes are:
- `FirstFitDecreasingPacking` and `ResourceCompliantRoundRobinPacking` now take container-level resource constraints from `topology.container.cpu`, `topology.container.ram`, and `topology.container.disk` instead of from hints. 
- `FirstFitDecreasingPacking` and `ResourceCompliantRoundRobinPacking` now determine padding resource by taking `max(containerResource * padding%, paddingAmount)`

This PR also revamp tests and code reuse.

PR #3187, opened 2019-02-04
Consolidate Packing

xiaoyao1991:
Currently, different packing algorithms assume different things and follow different semantics for the user-set configs. Because of it, there exist some code duplicates and it's not friendly for new packing algorithms. This PR is an attempt to address this issue and consolidate the three existing packing algorithms. 

Major behavior changes are:
- `FirstFitDecreasingPacking` and `ResourceCompliantRoundRobinPacking` now take container-level resource constraints from `topology.container.cpu`, `topology.container.ram`, and `topology.container.disk` instead of from hints. 
- `FirstFitDecreasingPacking` and `ResourceCompliantRoundRobinPacking` now determine padding resource by taking `max(containerResource * padding%, paddingAmount)`
- More details can be found in `AbstractPacking` class

This PR also revamp tests and code reuse. Changed packing tests to all extend from `CommonPackingTest` and uses helper methods for each test scenarios. Assertion expectations are changed accordingly due to the behavior changes introduced by the new determination of padding resource amount.

PR #3188, opened 2019-02-04
Revamp Packing tests

xiaoyao1991:
Breaking down #3184 

This PR mainly revamped RoundRobinPackingTest so that it extends CommonPackingTest and uses assertion helpers just like how other packing tests are doing. 

This PR also re-placed some FFD and RCRR specific tests accordingly

### DIFF
--- a/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
+++ b/heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
@@ -46,6 +46,10 @@ import org.apache.heron.spi.packing.PackingException;
 import org.apache.heron.spi.packing.PackingPlan;
 import org.apache.heron.spi.packing.Resource;
 
+import static org.apache.heron.api.Config.TOPOLOGY_CONTAINER_CPU_REQUESTED;
+import static org.apache.heron.api.Config.TOPOLOGY_CONTAINER_DISK_REQUESTED;
+import static org.apache.heron.api.Config.TOPOLOGY_CONTAINER_RAM_REQUESTED;
+
 /**
  * Round-robin packing algorithm
  * <p>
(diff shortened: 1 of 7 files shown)


heron/packing/src/java/org/apache/heron/packing/roundrobin/RoundRobinPacking.java
```diff
@@ -463,11 +463,15 @@ private ByteAmount getContainerDiskHint(Map<Integer, List<InstanceId>> allocatio
    * @return Container RAM requirement
    */
   private ByteAmount getContainerRamHint(Map<Integer, List<InstanceId>> allocation) {
+    ByteAmount defaultContainerRam = instanceRamDefault
```
nwangtw: same as #3187, could it be helpful to refactor this logic into a function? since the logic is used in a few places.

heron/packing/tests/java/org/apache/heron/packing/AssertPacking.java
```diff
@@ -42,7 +42,7 @@
  */
 public final class AssertPacking {
 
-  private static final double DELTA = 0.1;
+  public static final double DELTA = 0.1;
```
nwangtw: same as #3187, is "public" necessary here?

heron/packing/tests/java/org/apache/heron/packing/AssertPacking.java
```diff
@@ -135,14 +156,10 @@ public static void assertContainers(Set<PackingPlan.ContainerPlan> containerPlan
...
-    int instancesFound = 0;
-    for (PackingPlan.ContainerPlan containerPlan : containerPlans) {
-      for (PackingPlan.InstancePlan instancePlan : containerPlan.getInstances()) {
-        if (instancePlan.getComponentName().equals(component)) {
-          instancesFound++;
-        }
-      }
-    }
+    int instancesFound = (int) containerPlans.stream()
+        .flatMap(containerPlan -> containerPlan.getInstances().stream())
```
nwangtw: the stream() is needed here?

xiaoyao1991: yeh, flatMap takes stream only as input

heron/packing/tests/java/org/apache/heron/packing/AssertPacking.java
xiaoyao1991: It's `public` in the test scope so it should be fine I think. I made it public because other test files involve comparing double too.

heron/packing/tests/java/org/apache/heron/packing/AssertPacking.java
nwangtw: I see. Thx.

PR #3189, opened 2019-02-05
Add metrics for total time spent on user logic

nwangtw:
executeTimeNs in FullBoltMetrics is quite useful for investigation. However it is not in BoltMetrics. It is added in this PR.

Also, there is no similar metrics in SpoutMetrics and FullSpoutMetrics. A new metric "__next-tuple-time-ns" is added as well.

### DIFF
--- a/heron/common/src/java/org/apache/heron/common/utils/metrics/BoltMetrics.java
+++ b/heron/common/src/java/org/apache/heron/common/utils/metrics/BoltMetrics.java
@@ -48,6 +48,7 @@ public class BoltMetrics implements IBoltMetrics {
   private final CountMetric failCount;
   private final CountMetric executeCount;
   private final ReducedMetric<MeanReducerState, Number, Double> executeLatency;
+  private final CountMetric executeTimeNs;
   private final CountMetric tupleAddedToQueue;
   // Time in nano-seconds spending in execute() at every interval
   private final CountMetric emitCount;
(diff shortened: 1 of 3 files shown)


heron/common/src/java/org/apache/heron/common/utils/metrics/BoltMetrics.java
```diff
@@ -127,6 +130,7 @@ public void failedTuple(String streamId, String sourceComponent, long latency) {
   public void executeTuple(String streamId, String sourceComponent, long latency) {
     executeCount.incr();
     executeLatency.update(latency);
+    executeTimeNs.incrBy(latency);
```
huijunw: is this equation correct ? executeLatency x executeCount = executeTimeNs

nwangtw: correct in most cases. One exception is that count.incr() and latency().update() are not atomic, so in theory they might be separated into two intervals. This

The more straightforward motivation is more like the FullBoltMetrics class has the metric but the others dont. This PR makes them more consistent.

joshfischer1108:
@nwangtw Is this PR still relevant?

nicknezis:
Can we merge this? Or should we close it?

nwangtw:
Not motivated enough for me to merge. I am going to drop it.

nicknezis:
I'd be willing to test it if you think it was ready to merge.

PR #3190, opened 2019-02-08
Change tmaster/stmgr uptime granularity to be 1 second

nwangtw:
Currently the uptime metric updates every 1 minute. This is not accurate enough in some case.

uptime metric granularity is changed to 1 second in tmaster and stmgr in this PR

### DIFF
--- a/heron/stmgr/src/cpp/manager/stmgr.cpp
+++ b/heron/stmgr/src/cpp/manager/stmgr.cpp
@@ -68,6 +68,7 @@ const sp_string METRIC_DROPPED_DURING_RESTORE = "__dropped_during_restore";
 const sp_string METRIC_TIME_SPENT_BACK_PRESSURE_INIT =
     "__server/__time_spent_back_pressure_initiated";
 const sp_int64 PROCESS_METRICS_FREQUENCY = 10_s;
+const sp_int64 UPTIME_METRIC_FREQUENCY = 1_s;
 const sp_int64 TMASTER_RETRY_FREQUENCY = 10_s;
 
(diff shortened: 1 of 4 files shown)


huijunw:
although stmgr/tmaster report uptime every 1 sec, the metricsmgr still report every 1 min? when a stmgr/tmaster is restarted, which still needs 1min for metricsmgr to flush metrics to expose this restarting event.

nwangtw:
> although stmgr/tmaster report uptime every 1 sec, the metricsmgr still report every 1 min? when a stmgr/tmaster is restarted, which still needs 1min for metricsmgr to flush metrics to expose this restarting event.

Yes. The only difference is that when the metric is reported, currently it could be 0, 60, 120, 180 or 0, 60, 60, 180 if the two process timer and reporting timer doesn't sync perfectly. After the PR, the value is more likely to be 0, 60, 118, 179, and so on

heron/stmgr/src/cpp/manager/stmgr.cpp
```diff
@@ -244,12 +250,13 @@ void StMgr::CheckTMasterLocation(EventLoop::Status) {
   }
 }
 
-void StMgr::UpdateProcessMetrics(EventLoop::Status) {
-  // Uptime
+void StMgr::UpdateUptimeMetric() {
```
huijunw: UpdateUptimeMetric() is called 60 times every min, UpdateProcessMetrics() is called every min.
is it overlap 1 in every 60 times?

nwangtw: yeah. that is right.

PR #3192, opened 2019-02-08
Make 0 stat visible in Heron UI

nwangtw:
Currently stat 0 looks the same as still loading. The only way to tell is mouse over the data bar. With this PR stat 0 is shown as a 1 pixel high bar so that it is a bit easier for users to tell the difference from UI.


xiaoyao1991:
👍

nwangtw:
Verified after the fix. Cant attach screenshot. :(

PR #3195, opened 2019-02-19
Refactor FirstFitDecreasingPacking to support CPU intensive topologies

xiaoyao1991:


heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
+/**
```
nwangtw: I am feeling that it is fine to keep the old name, since it is BinPacking anyway. Also if any user was using BinPacking before, renaming would require them to update their config which is not convenient. The new code works very similar to the old logic anyway.

Just my 2 cents.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+   * @return packing plan
+   */
+  @Override
+  public PackingPlan pack() {
+    PackingPlanBuilder ramFirstPlanBuilder = newPackingPlanBuilder(null);
+    PackingPlanBuilder cpuFirstPlanBuilder = newPackingPlanBuilder(null);
+
+    // Get the instances using FFD allocation
+    try {
+      numContainers = 0;
```
nwangtw: Some comments might be helpful here. I can imaging me trying to delete the second "numContainers = 0;" since it looks like a bug.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+      ramFirstPlanBuilder = get2DVectorBinAllocation(ramFirstPlanBuilder,
+          SortingStrategy.RAM_FIRST);
+      numContainers = 0;
```
nwangtw: can we avoid this numContainers?

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+
+      return pickBetter(ramFirstPlan, cpuFirstPlan);
+    } catch (ConstraintViolationException e) {
+      throw new PackingException("Could not allocate all instances to packing plan", e);
+    }
+  }
+
+  /**
+   * Pick the better plan by checking which plan dominates in terms of:
+   * overall resource utility, average container resource utility,
```
nwangtw: I think more detailed rules could be helpful here.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+   * container ram utility stdev, and container cpu utility stdev.
+   * @param plan1
+   * @param plan2
+   * @return the better plan
+   */
+  private PackingPlan pickBetter(PackingPlan plan1, PackingPlan plan2) {
```
nwangtw: I think we need a few unit tests for this function.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+    int plan1Points = 0;
+    int plan2Points = 0;
+
+    // Checking packing plan overall resource utility
+    PackingPlan.ResourceUtility plan1ResourceUtility = plan1.getPackingPlanResourceUtility();
```
nwangtw: Maybe a TODO, we can have an interface like PackingPlanEvaluationRule {} and refactor this code with a list of rules.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
nwangtw: Could we make this function a static function? I am feeling that we might move this function to a util or some other class in future.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
```diff
@@ -0,0 +1,379 @@
...
+
+    List<ResourceRequirement> resourceRequirements =
+        getSortedInstances(componentsToScaleDown.keySet(), sortingStrategy);
+
+    InstanceCountScorer instanceCountScorer = new InstanceCountScorer();
+    ContainerIdScorer containerIdScorer = new ContainerIdScorer(false);
+
+    for (ResourceRequirement resourceRequirement : resourceRequirements) {
+      String componentName = resourceRequirement.getComponentName();
+      int numInstancesToRemove = -componentsToScaleDown.get(componentName);
```
nwangtw: componentsToScaleDown keeps negative counts? It feels a bit intuitive this way to me.

heron/packing/src/java/org/apache/heron/packing/builder/ResourceRequirement.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Helper class that captures the RAM and CPU requirements of each component
+ */
+public class ResourceRequirement implements Comparable<ResourceRequirement> {
+
+  private String componentName;
+  private ByteAmount ramRequirement;
+  private double cpuRequirement;
+  private SortingStrategy sortingStrategy;
```
nwangtw: SortingStrategy is a bit strange to be here I feel. Does it make more sense to refactor it out?

heron/spi/src/java/org/apache/heron/spi/packing/PackingPlan.java
```diff
@@ -324,4 +390,39 @@ public String toString() {
       return str + "}";
     }
   }
+
+  public static class ResourceUtility implements Comparable<ResourceUtility> {
```
nwangtw: Suggest to move out to its own file.

Also, why disk is not included?

heron/spi/src/java/org/apache/heron/spi/packing/PackingPlan.java
```diff
@@ -324,4 +390,39 @@ public String toString() {
...
+    }
+
+    @Override
+    public int compareTo(ResourceUtility o) {
+      if (ramUtility > o.ramUtility && cpuUtility > o.cpuUtility) {
+        return 1;
+      } else if (ramUtility < o.ramUtility && cpuUtility < o.cpuUtility) {
+        return -1;
+      } else {
+        return 0;
```
nwangtw: hmm. So if ramUtil is greater and cpuUtil is less, the function would return 0? Seems strange. Maybe give ram and cpu different priorities. Like only compare cpu utils when the ram utils equal.

heron/spi/src/java/org/apache/heron/spi/packing/PackingPlan.java
nwangtw: And please add unit tests.

heron/packing/src/java/org/apache/heron/packing/binpacking/TwoDimensionVectorBinPacking.java
xiaoyao1991: That actually sounds perfect to me. I prefer not to have too many packing classes too.

heron/api/src/java/org/apache/heron/api/Config.java
```diff
@@ -215,6 +215,11 @@
    * spout0:12434,spout1:345353,bolt1:545356.
    */
   public static final String TOPOLOGY_COMPONENT_DISKMAP = "topology.component.diskmap";
+  /**
+   * Sorting strategy for FirstFitDecreasingPacking algorithm.
+   * RAM_FIRST (default), or CPU_FIRST
+   */
+  public static final String TOPOLOGY_FFD_SORTING_STRATEGY = "topology.ffd.sorting.strategy";
```
nwangtw: I am wondering if there is a better place for this config. @huijunw @nlu90 any thoughts?

Also if it will be kept here. I would suggest make the name and string like topology.packing.ffd....

heron/packing/src/java/org/apache/heron/packing/constraints/InstanceDensityConstraint.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class InstanceDensityConstraint implements PackingConstraint {
+  private final int maxNumInstancesPerContainer;
+
+  public InstanceDensityConstraint(int maxNumInstancesPerContainer) {
+    this.maxNumInstancesPerContainer = maxNumInstancesPerContainer;
+  }
+
+  @Override
+  public void test(Container container, PackingPlan.InstancePlan instancePlan)
```
nwangtw: maybe rename to "validate"?

xiaoyao1991: Sure. I named it `test` because I first wanted to use `Predicate` for this, but it's fine.

heron/packing/src/java/org/apache/heron/packing/builder/ResourceRequirement.java
xiaoyao1991: Refactored so that now `SortingStrategy` becomes a comparator itself.

PR #3196, opened 2019-02-22
Enable Valgrind Checks in Travis

xiaoyao1991:


scripts/travis/build.sh
```diff
@@ -108,6 +108,18 @@ python ${UTILS}/save-logs.py "heron_test_flaky.txt" bazel\
...
+# run valgrind tests on cpp tests
+# excluding tuple-cache_unittest because its results are flaky when running under valgrind
+T="heron valgrind test"
+start_timer "$T"
+python ${UTILS}/save-logs.py "heron_valgrind_test.txt" bazel\
+  --bazelrc=tools/travis/bazel.rc test\
+  --run_under=valgrind\
+  --test_summary=detailed --test_output=errors\
+  --config=$PLATFORM --jobs=0\
+  $(bazel query 'kind("cc_test rule", ...)' | grep -v //heron/stmgr/tests/cpp/util:tuple-cache_unittest)
```
nwangtw: tmaster is native code as well.

nwangtw:
Is there any output to share?

scripts/travis/build.sh
xiaoyao1991: `bazel query 'kind("cc_test rule", ...)'` basically returns all test targets that are cpp. tmaster tests are included as well

xiaoyao1991:
@nwangtw that test is excluded because it was flaky when I run it in my local environment. However, every test is flaky when running in travis. The previous two runs indicate failures at different places. Not very consistent.

PR #3197, opened 2019-02-23
Update build-docker.sh

windhamwong:
--squash should be removed as this parameter is for experimental setup only.

### DIFF
--- a/docker/scripts/build-docker.sh
+++ b/docker/scripts/build-docker.sh
@@ -61,7 +61,7 @@ run_build() {
   export HERON_VERSION
 
   echo "Building docker image with tag:$DOCKER_TAG"
-  docker build --squash -t "$DOCKER_TAG" -t "$DOCKER_LATEST_TAG" -f "$DOCKER_FILE" "$SCRATCH_DIR"
+  docker build -t "$DOCKER_TAG" -t "$DOCKER_LATEST_TAG" -f "$DOCKER_FILE" "$SCRATCH_DIR"
   # save the image as a tar file
   DOCKER_IMAGE_FILE="$OUTPUT_DIRECTORY/heron-docker-$HERON_VERSION-$TARGET_PLATFORM.tar"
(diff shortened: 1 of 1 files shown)


nwangtw:
I went back and forth about it too. :(  Without the flag the size is like doubled. @mjschmidt @joshfischer1108 how do you think about it?

joshfischer1108:
If it doesn't cause any harm I don't see anything wrong with it.. We are building with Bazel `0.14.x`.  You could consider that as experimental too.

joshfischer1108:
updated my comment for clarity.

nwangtw:
KK. Thanks.

In my test I remember they both work ok, except the size is ~600m vs 1.2G.

I am fine either way but we dont really use docker image inside Twitter and it is ok if you guys feel it ok. I will approve it later if no objection by the end of day.

joshfischer1108:
I would prefer to keep the `--squash` flag.  600m is better than 1.2G

nwangtw:
Another option is to make it a parameter for the script and the default is without that flag.

The only problem I met with that flag is that Apache jenkins hosts don't have the flag enabled. This could be a problem for users' ci environment as well.

nwangtw:
@windhamwong how do you think about making it a script argument?

mjschmidt:
Sorry just seeing this. Seems like you guys figured out a fine solution

mjschmidt:
I can live with it (:

mjschmidt:
I can live with either.

windhamwong:
Maybe we can add an option to the script that allows using this flag or not?

windhamwong:
Let me check again and make an update to this

windhamwong:
Updated the script with [-s|--squash] option.
Rephasing the default command help message

nwangtw:
Build failure is caused by zookeeper library getting removed. Dmitry is working on the fix.

PR #3201, opened 2019-02-28
Enable gperftool HEAPCHECK in integration test

xiaoyao1991:
Enable [gperftool HEAPCHECK](https://gperftools.github.io/gperftools/heap_checker.html) for CPP code in integration tests. Currently, it's simply printing heapcheck results in the heron-executor.log when the flag `ENABLE_HEAPCHECK` is set.


heron/common/src/cpp/network/event_loop_impl.cpp
```diff
@@ -76,7 +76,15 @@ EventLoopImpl::~EventLoopImpl() {
   event_base_free(mDispatcher);
 }
 
+static void handleTerm(evutil_socket_t _, sp_int16 what, void* ctx) {
```
nwangtw: Suggest to break this change into a different PR.

Also I am wondering if it is better to move the evsignal_new() to caller in case multiple event loop objects are created in an application.

heron/executor/src/python/heron_executor.py
```diff
@@ -881,10 +898,17 @@ def _run_process(self, name, cmd, env_to_exec=None):
...
-      proc.async_stream_process_stdout(process, stdout_log_fn(name))
+      if isinstance(cmd, dict) and cmd.has_key('env'):
+        Log.info("command %s env %s" % (' '.join(cmd['cmd']), ' '.join(cmd['env'])))
+        if env_to_exec is not None:
+          cmd['env'].update(env_to_exec)
+        process = subprocess.Popen(cmd['cmd'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+                                   env=cmd['env'], bufsize=1)
+        proc.async_stream_process_stdout(process, stdout_log_fn(name))
+      else:
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
```
nwangtw: so when cmd is a dict and it doesn't have 'env', this branch is called? Maybe it might be cleaner to construct a normalized_cmd object that contains 'cmd' and 'env' and then call popen to execute it?

PR #3204, opened 2019-03-01
align protopy to WORKSPACE

huijunwu:
https://github.com/apache/incubator-heron/blob/master/WORKSPACE#L835 shows 3.4.1, but proto py uses 3.4.0. 
update 3.4.0 to 3.4.1


nwangtw:
LGTM.

@kramasamy @srkukarni What do you think?

huijunwu:
looks like protobuf 3.4.1 is not available in pypi. close this pr

PR #3205, opened 2019-03-04
Update WORKSPACE info for zookeeper http archive.

jmark99:
Updated the new_http_archive entry for org_apache_zookeeper to zookeeper-3.4.13. Attempting to perform a clean installation of Heron failed due to the fact that the previous version (3.4.10) is no longer accessible at the url locations provided.

I updated to the current version and was able to successfully build and run heron.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -865,13 +865,13 @@ new_http_archive(
 new_http_archive(
     name = "org_apache_zookeeper",
     urls = [
-      "http://apache.cs.utah.edu/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz",
-      "http://apache.claz.org/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz",
-      "http://apache.mesi.com.ar/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz",
+      "http://apache.cs.utah.edu/zookeeper/zookeeper-3.4.13/zookeeper-3.4.13.tar.gz",
+      "http://apache.claz.org/zookeeper/zookeeper-3.4.13/zookeeper-3.4.13.tar.gz",
+      "http://apache.mesi.com.ar/zookeeper/zookeeper-3.4.13/zookeeper-3.4.13.tar.gz",
     ],
-    strip_prefix = "zookeeper-3.4.10",
+    strip_prefix = "zookeeper-3.4.13",
     build_file = "third_party/zookeeper/zookeeper.BUILD",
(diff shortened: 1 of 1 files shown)


nwangtw:
Thanks.

Another fix is to replace the link and keep the same version number. I am fine either way.

@kramasamy Any comment?

jmark99:
I was having difficulty finding the 3.4.10 version anywhere except the Apache archive site. The Apache site indicates that downloads should be made from a nearby mirror site rather than apache.org. Otherwise I would have taken that approach. If another reliable link can be found I can update the links rather than the version. But I did not find a reliable link.

nwangtw:
> I was having difficulty finding the 3.4.10 version anywhere except the Apache archive site. The Apache site indicates that downloads should be made from a nearby mirror site rather than apache.org. Otherwise I would have taken that approach. If another reliable link can be found I can update the links rather than the version. But I did not fiInd a reliable link.

I see. Thanks!

nlu90:
LGTM

dnrusakov:
@jmark99 @nwangtw When the new version released (for example, zookeeper-3.4.14), the current zookeeper-3.4.13 will go to archive (as it happened with zookeeper-3.4.10). The archive is not a part of any mirror, so we will have that problem again. Since they intentionally didn't add archive to any mirror, i assume it's safe to directly use the following link: http://archive.apache.org/dist/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz

nwangtw:
Thx.

If we change to zookeeper-3.4.13, when zk has a new release, it will be be removed from mirror sites and we will have to update again. I am curious what exactly the apache suggest means.

Mentors do you have any suggestion? @dave2wave @julienledem @ptgoetz 

FYI, Dmitry's PR for the issue is here: https://github.com/apache/incubator-heron/pull/3207

dave2wave:
I think it depends on whether or not you ask for a specific version or latest from Maven Central.

nwangtw:
Might be safer to stick with a specific version and manually upgrade (just my 2 cents). In this case, archive.apache.org is fine?

jmark99:
@nwangtw that sounds good? I’ll update the link to archive.apache.org and remove the current links. Would you prefer I revert back to 3.4.10 or stay with the most recent release?

jmark99:
@nwangtw I updated WORKSPACE to use archive.apache.org as the url for the zookeeper archive. I reverted back to the 3.4.10 version for now to keep the change to a minimum at this tmie. The version can be updated when desired without having to modify the url.

nwangtw:
Thanks!

Dave has some suggestions in the other similar PR. I am going to merge this one first to fix the build and see if we can improve it more.

PR #3208, opened 2019-03-07
Added logging an error status code when stream manager can not connect to tmaster

dnrusakov:
Added logging an error status code when stream manager can not connect to tmaster


xiaoyao1991:
👍

PR #3215, opened 2019-03-13
Support overriden configuration in ckpmanager

simingweng:
with this extra CLI argument added to `Checkpoint Manager`, it becomes possible to override `Checkpoint Manager` configuration via `-D` system properties that are specified to Heron API server process during deployment time.

A useful use case is to switch `heron.statefulstorage.classname` to a different stateful storage backend, such as `DlogStorage` when deployed on Kubernetes cluster, without having to build customized Heron docker image.


simingweng:
@nwangtw

PR #3216, opened 2019-03-15
Add an index column in all topologies page

nwangtw:
An index column could be helpful for users to count rows. Also updated width of release version since it could be slightly longer than other short columns.


xiaoyao1991:
👍

PR #3217, opened 2019-03-15
Add ack and fail to Streamlet API  (#2909)

### LINKED ISSUE #2909: Add ack and fail to Streamlet API
The default behavior for heron streamlet is not to ack/fail messages. Upon further inspection we realized this is not only the default, but the only option for Streamlet as it only uses emit, which will not ack and fail messages. This currently isn't configurable.

The code can be seen here: 
https://github.com/apache/incubator-heron/blob/6029c28943b7ad85be6fa175d20e0e908fb199b6/heron/api/src/java/org/apache/heron/streamlet/impl/sources/ComplexSource.java#L60

What steps would be required to add this functionality? This would aid greatly in us transitioning to Streamlet.

jmark99:
This PR updates Heron to allow for ack and fail behaviour with the Streamlet API.

Would request feedback, especially as to the caching of id's and tuples. 

These updates have been tested against a collection of streamlet examples. The repo located at https://github.com/jmark99/heron-examples contains a collection of many streamlet examples. The modifications have been run successfully against this collection. The README file for that repo describes the steps taken to exercise the changes.

This commit is intended to fix #2909

### DIFF
--- a/heron/api/src/java/BUILD
+++ b/heron/api/src/java/BUILD
@@ -1,5 +1,4 @@
 licenses(["notice"])
-
 package(default_visibility = ["//visibility:public"])
 
 load("//tools/rules:build_defs.bzl", "DOCLINT_HTML_AND_SYNTAX")
@@ -16,6 +15,7 @@ api_deps_files =  \
     heron_java_api_proto_files() + [
         ":classification",
         "//heron/common/src/java:basics-java",
+        "//third_party/java:guava"
     ]
 
 # Low Level Api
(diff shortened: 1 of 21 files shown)


nwangtw:
Is there a way to verify the behavior with unit tests?

jmark99:
I believe I will have to find a way to test via integration tests in order to verify programmatically. Even with integration tests I'm not sure how to test tuple failures. I'm open to ideas.

nwangtw:
Most operators don't have "failure" case (for example filter, flatmap). Only ack I remember. 

The only way to trigger fail() I can think of now is to write a simple bolt that has failure cases and then use the applyOperator() function.

I was thinking of catching exceptions as failures but it might have some unwanted side effects so it is not decided yet.

PR #3224, opened 2019-03-22
Make cppckecker independ of pcre

nwangtw:
Currently cppchecker relies on pcre which is not installed in some systems. Pcre (rules support) is not really need here so it is disabled here.


xiaoyao1991:
👍

PR #3226, opened 2019-03-25
Fix command line result when --dry-run is enabled

nwangtw:
Currently "heron submit .... --dry-run" returns 1 when the submission succeeds or fails.

It should return 0 when the submission succeeds.


xiaoyao1991:
👍

PR #3229, opened 2019-03-29
init instance state when necessary

nlu90:
init instance state only when the topology is stateful and the component is a stateful component.

If a statefulComponent instance is used in a non-stateful topology currently, the instance's local state map will be updated to `null` without this fix.


nlu90:
@nwangtw  just updated the description.

nwangtw:
KK. Thanks.

PR #3230, opened 2019-03-29
memory leak in tmaster fixed

dnrusakov:
The memory leak in TMaster was reported by one of the Twitter internal clients.
After running TMaster under Valgrind memory leaks detector, the leak was localized.
This PR aims to fix it. 

`GetStmgrsRegSummary()` now returns a smart pointer (`unique_ptr<>`). So, as soon as a caller no need the returned result anymore the memory is freed automatically.

Tested this PR locally under Valgrind and don't see the leak anymore.


thinker0:
+1

xiaoyao1991:
👍

PR #3244, opened 2019-04-25
Improve the error message (with numbers) for the 'More More container…

nwangtw:
…s allocated than instance' exception


xiaoyao1991:
👍

PR #3245, opened 2019-04-25
Fix build when the source code is not managed by git

nwangtw:
For this issue: https://github.com/apache/incubator-heron/issues/3243

If someone download source code tar and decompress it. The command below requires a HERON_BUILD_VERSION environment variable to build successfully. The code change adds a fallback to the current directory name when the variable is not set.


julienledem:
+1 looks good to me

nwangtw:
Thanks!

PR #3251, opened 2019-04-29
Check invalid number of container in round robin packing algorithm

nwangtw:
Currently if number of containers is set to an invalid value, we get the following exception which is not ideal.

```

[2019-04-29 15:17:05 -0700] [FINE] org.apache.heron.scheduler.SubmitterMain: Exception when submitting topology 
java.lang.NullPointerException
	at org.apache.heron.packing.roundrobin.RoundRobinPacking.getRoundRobinAllocation(RoundRobinPacking.java:385)
	at org.apache.heron.packing.roundrobin.RoundRobinPacking.packInternal(RoundRobinPacking.java:149)

```

After the change, we would expect "Invlaid number of container: %d" runtime exception.


xiaoyao1991:
👍

PR #3252, opened 2019-05-06
Joshfischer/asf site

joshfischer1108:
@nwangtw This is a first pass on a script to build and publish our static site.  I borrowed a lot of the steps from the incubator site build process here -> https://github.com/apache/incubator/blob/master/build_site.sh.  I've turned off the ci  job for now.  Will turn it on once we come to agreement on how the script should work.

Any feedback is welcome.

### DIFF
--- a/website/build_site.sh
+++ b/website/build_site.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+ROOT_DIR=`pwd`
+
+WORK_DIR=/website/public
+ME=`basename $0`
+
+rm -rf $WORK_DIR
+mkdir -p $WORK_DIR
+
+cd website
+
+make setup
+
+make site
+
+# push all of the results to asf-site branch
+git checkout asf-site
+git clean -f -d
+git pull origin asf-site
+
+#
+cd $ROOT_DIR
+
+rm -rf content
+mkdir content
+
+cp -a $WORK_DIR/* content
+cp -a .htaccess content
+git add content
+git commit -m "git-site-role commit from $ME"
+git push origin asf-site
--- a/.htaccess
+++ b/.htaccess
@@ -1,2 +1,1 @@
 Redirect	/index.html	https://apache.github.io/incubator-heron/
-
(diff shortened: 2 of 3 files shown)


joshfischer1108:
My thinking is that we will build off of master at some time period we decide on, similar to what the incubator site does.  We can then take the generated static files out of the `website/public` directory and move them to a folder called `content` in the root directory.  These changes can be committed to the `asf-site` branch.  We can then use apache infra to serve the files out of the `content` directory.  Thoughts?

nwangtw:
Currently the website is single version only. I feel eventually we will need to have web pages for each version. Therefore I am thinking that we can have a ci to make sure website is in good shape, and we should add it to our release process (generate website after a release vote is successful). Just my 2 cents.

website/build_site.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
nwangtw: License header.

website/build_site.sh
```diff
@@ -0,0 +1,32 @@
...
+rm -rf $WORK_DIR
+mkdir -p $WORK_DIR
+
+cd website
+
+make setup
+
+make site
+
+# push all of the results to asf-site branch
```
nwangtw: Maybe breaking the code below into a publish_website.sh could be more flexible.

website/build_site.sh
joshfischer1108: Ahh yes. Thanks

website/build_site.sh
joshfischer1108: I agree

joshfischer1108:
Is the release process manual right now?  Might be a good time to automate that while we are working on the website deployment.

nwangtw:
> Is the release process manual right now? Might be a good time to automate that while we are working on the website deployment.

There are jenkins jobs for building artifacts. Last three in this page: https://builds.apache.org/search/?q=heron&Jenkins-Crumb=edf72a5514450f709b23ec4c0f42f56a.

nwangtw:
> Is the release process manual right now? Might be a good time to automate that while we are working on the website deployment.

And the release process itself is manual currently after artifacts are built.

nwangtw:
LGTM~

PR #3254, opened 2019-05-13
Kafka spout bazel build

simingweng:
- convert the Kafka Spout Maven module to a Bazel target
- convert the sample topology Maven module to a Bazel target
- remove unnecessary Travis script step
- delete Maven POM files
- switch to `org.apache.heron` API


simingweng:
@nwangtw Please review. And, also, how do we produce the maven artifact out of this Bazel target, so that user will be able to use it as maven dependency?

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/test/java/BUILD
```diff
@@ -0,0 +1,44 @@
+heron_kafka_spouts_test_dep = [
+    "//contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java:heron-kafka-spout-java",
+    "@org_apache_kafka_kafka_clients//jar",
```
nwangtw: sort.

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/BUILD
```diff
@@ -0,0 +1,18 @@
+licenses(["notice"])
+
+package(default_visibility = ["//visibility:public"])
+
+kafka_spout_deps = [
+    "@org_apache_kafka_kafka_clients//jar",
```
nwangtw: sort

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/test/java/BUILD
simingweng: sort the order of the dependencies?

simingweng:
@nwangtw I've updated the order of the dependencies, please check whether it complies with the convention.

simingweng:
I've just pushed one more commit to add the unit tests to `scripts/travis/build.sh`.

simingweng:
There're a bunch of `checkstyle` violation errors reported from Travis. It will take some time for me to correct all of them.

By the way, I found two header txt files under `tools/..` directory, which is actually the one that `checkstyle` is using?

nwangtw:
> There're a bunch of `checkstyle` violation errors reported from Travis. It will take some time for me to correct all of them.
> 
> By the way, I found two header txt files under `tools/..` directory, which is actually the one that `checkstyle` is using?

Not sure, but from JavaCheckstyle.java, maybe both are used?

xiaoyao1991:
@simingweng mostly indentation errors. I guess that should be a quick fix by leveraging the IDE or editor to auto convert indentation to 2 spaces. There is a `https://github.com/apache/incubator-heron/blob/master/scripts/setup-intellij.sh` that setup the styling for IntelliJ for this repo. If you setup the project with that script, the style issues should mostly be avoided.

xiaoyao1991:
👍 LGTM when stylecheck is passed

simingweng:
@nwangtw @xiaoyao1991 `checkstyle` still gave me a lot of trouble, even I adjusted the `setup-intellj.sh` To include the `contrib` folder, and open the project in IntelliJ, but it seems the default formatting settings in IntelliJ does not align with Heron’s customized check style config.

So, what is the command I can use to locally build the `contrib/...` targets **with** `checkstyle` on? I tried running the same build command in Travis CI script, but it failed due to missing some static Protobuf lib on my Mac.

nwangtw:
> @nwangtw @xiaoyao1991 `checkstyle` still gave me a lot of trouble, even I adjusted the `setup-intellj.sh` To include the `contrib` folder, and open the project in IntelliJ, but it seems the default formatting settings in IntelliJ does not align with Heron’s customized check style config.
> 
> So, what is the command I can use to locally build the `contrib/...` targets **with** `checkstyle` on? I tried running the same build command in Travis CI script, but it failed due to missing some static Protobuf lib on my Mac.

Can you share the lint errors with us, maybe in slack channel?

For local style check, I believe "bazel build --compilation_mode=dbg --config=darwin --test_output=all heron/..." triggers style check. It is part of "darwin" config.

nwangtw:
> @nwangtw @xiaoyao1991 `checkstyle` still gave me a lot of trouble, even I adjusted the `setup-intellj.sh` To include the `contrib` folder, and open the project in IntelliJ, but it seems the default formatting settings in IntelliJ does not align with Heron’s customized check style config.
> 
> So, what is the command I can use to locally build the `contrib/...` targets **with** `checkstyle` on? I tried running the same build command in Travis CI script, but it failed due to missing some static Protobuf lib on my Mac.

Seems quite some indent error. I think the default indent is 4 spaces in Intellij. Heron requires 2-space indent.

PR #3256, opened 2019-05-15
Add kryo serializer in Heron API

nwangtw:
It was not supported in Heron API. As the result, users can't register
serializer like they did with Storm API. The issue is that Kryo is not
part of the Heron API and it is added in this PR.

Integration test is included and passed.

Default Serializer is still JavaSerializer which is different from Storm and Heron Streamlet. We can default to Kryo later after we make the decision.


heron/api/src/java/org/apache/heron/streamlet/impl/KryoSerializer.java
```diff
@@ -17,95 +17,12 @@
...
-  }
-
-  private class HashSetSerializer extends CollectionSerializer {
-    @Override
-    @SuppressWarnings("rawtypes") // extending Kryo class that uses raw types
-    public Collection create(Kryo k, Input input, Class<Collection> type) {
-      return new HashSet();
-    }
-  }
+public class KryoSerializer extends org.apache.heron.api.serializer.KryoSerializer {
```
nlu90: why not use the `KryoSerializer` from the `api.serializer` package directly?

nwangtw: This class can be used by Streamlet config currently. So I am leaving it available just in case someone is using it in their project.

Let me clarify it in comment.

nicknezis:
Should this pull request have had any website documentation updates? The website states that users would have to recompile Heron to add Kryo serializers. Does this change what the website states?

PR #3260, opened 2019-05-21
Heron API/Tracker update for storing Topologies in Array/List instead of Dict object

windhamwong:
Old version:
```
{
  (..),
  "result": 
  <cluster>: {
    <environment>: {
      <topology>: {
        (topology_details)
      }
    }
  }
}
```
New version:
```
{
  (..),
  "result": [{
    (topology_details)
  }]
}
```

The cluster, environment, topology variables are actually obtainable from the topology details. This PR is to store these data into array not in dict object so there will be less nested object inside and duplicated cluster/environment/topology can be also displayed on the UI. (Might be useful for multi-cluster with same configurations across all the clusters)

Still need some testing and recommendation. It works perfectly for me, but there are minor 500 errors (unrelated to this change) from the original source code might need to solve as well.

### DIFF
--- a/heron/tools/tracker/src/python/handlers/stateshandler.py
+++ b/heron/tools/tracker/src/python/handlers/stateshandler.py
@@ -65,7 +65,8 @@ class StatesHandler(BaseHandler):
 
     role = self.get_argument_role()
 
-    ret = {}
+    # ret = {}
+    ret = []
     topologies = self.tracker.topologies
     for topology in topologies:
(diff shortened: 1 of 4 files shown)


windhamwong:
`third_party/cppcheck/cppcheck.BUILD` looks like has been changed since `release/0.20.1-incubating-rc2`

windhamwong:
Major issue with topology details. will do another PR later

PR #3273, opened 2019-05-23
Make RoundRobin packing algo more forgivable when container resource …

nwangtw:
…is not set. Don't check if container resource is not set


xiaoyao1991:
👍

PR #3275, opened 2019-05-23
Bugfix kafka spout

simingweng:
a bunch of improvement and bug fixes


contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
```diff
@@ -330,11 +355,8 @@ private void manualCommit(TopicPartition topicPartition, NavigableMap<Long, Long
   private Iterable<ConsumerRecord<K, V>> poll() {
     ConsumerRecords<K, V> records = consumer.poll(Duration.ofMillis(200));
     if (!records.isEmpty()) {
-      //since the Kafka Consumer metrics are built gradually based on the partitions it consumes,
-      //we need to periodically check whether there's any new metrics to register after
-      //each polling.
       if (System.currentTimeMillis() - previousKafkaMetricsUpdatedTimestamp
-          > metricsIntervalInSecs) {
+          > metricsIntervalInSecs * 1000) {
```
nwangtw: why "* 1000"?

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
```diff
@@ -275,27 +266,58 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
             new Fields(consumerRecordTransformer.getFieldNames(s))));
   }
 
+  @Override
+  public Map<String, Object> getComponentConfiguration() {
+    return null;
+  }
+
+  private void initialize(Map<String, Object> conf) {
+    topologyReliabilityMode = Config.TopologyReliabilityMode.valueOf(
```
nwangtw: I am wondering if this valueOf(... toString()) is necessary?

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
```diff
@@ -275,27 +266,58 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
...
+    Map<String, List<Object>> tupleByStream = consumerRecordTransformer.transform(record);
+    //nothing worth emitting out of this record,
+    //so immediately acknowledge it if in ATLEAST_ONCE mode
+    if (tupleByStream.isEmpty() && topologyReliabilityMode
+        == Config.TopologyReliabilityMode.ATLEAST_ONCE) {
+      ack(new ConsumerRecordMessageId(new TopicPartition(record.topic(), record.partition()),
+          record.offset()));
+    }
+    tupleByStream.forEach((s, objects) -> {
+      if (topologyReliabilityMode != Config.TopologyReliabilityMode.ATLEAST_ONCE) {
```
nwangtw: I feel "switch(mode)" might be more readable.

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
```diff
@@ -275,27 +266,58 @@ public void declareOutputFields(OutputFieldsDeclarer declarer) {
...
-            //emit tuple with the message id
-            collector.emit(s, objects, consumerRecordMessageId);
-          }
-        });
+    Map<String, List<Object>> tupleByStream = consumerRecordTransformer.transform(record);
+    //nothing worth emitting out of this record,
+    //so immediately acknowledge it if in ATLEAST_ONCE mode
+    if (tupleByStream.isEmpty() && topologyReliabilityMode
+        == Config.TopologyReliabilityMode.ATLEAST_ONCE) {
+      ack(new ConsumerRecordMessageId(new TopicPartition(record.topic(), record.partition()),
```
nwangtw: better to ack after emit I think.

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
simingweng: because `metricsIntervalInSecs` has the unit of second, while the duration is in millisecond

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
simingweng: unfortunately, it's necessary because the value we put into the config is always a `String`, such as https://github.com/apache/incubator-heron/blob/9f05f47227aa87d0014621519e2d9c89f667a065/heron/api/src/java/org/apache/heron/api/Config.java#L547

And other places in the code base has to do the same thing, such as https://github.com/apache/incubator-heron/blob/9f05f47227aa87d0014621519e2d9c89f667a065/heron/instance/src/java/org/apache/heron/instance/AbstractOutputCollector.java#L71

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
simingweng: addressed

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
simingweng: This is an ack action when the Kafka consumer record gets translated into nothing, probably because it's not a record that user code is interested in. It is not meant to ack the record that results in tuple to be emitted.

But, I think it's better to add a `return` at the end of the if block, which makes it easier to read.

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
nwangtw: Ah. My bad. I dont know what I was thinking.

contrib/spouts/kafka/java/heron-kafka-spout-parent/heron-kafka-spout/src/main/java/org/apache/heron/spouts/kafka/KafkaSpout.java
nwangtw: You are right. I didn't notice the isEmpty() part. Sounds good then.

PR #3282, opened 2019-06-05
STREAMCOMP-2724: Stream Manager migrated to mainly use smart pointers instead of a manual memory management

dnrusakov:
This PR contains the following changes:

1. Stream Manager has been migrated to mainly use smart pointers instead of manual memory management.

2. The way how we use object pool has been changed. Previously we had several places where we released objects back to the pool, but those objects were not previously acquired from the pool. This might cause memory leaks. This PR reduces the overall amount of use cases where we use pool-based allocations. Actually, now there are just three most "high load" message types (`HeronTupleSet2`, `HeronTupleSet`, `TupleStreamMessage`) left which are managed by the memory pool. And for such use cases i verified that we release back only objects which has been previously acquired from the pool. All other message types are not on the "high load" path, so i switched them back to use system memory allocator and smart pointers.

Testing.
It successfully passed unit and integration tests.
Also tested both `StreamManager` and `TMaster` on my laptop by running a `WordCountTopology` under `valgrind` within 24 hours. I haven't found any memory leaks in the `valgrind` logs.

### DIFF
--- a/heron/common/src/cpp/metrics/tmaster-metrics.cpp
+++ b/heron/common/src/cpp/metrics/tmaster-metrics.cpp
@@ -35,7 +35,9 @@
 namespace heron {
 namespace common {
 
-TMasterMetrics::TMasterMetrics(const sp_string& sinks_filename, EventLoop* eventLoop) {
+using std::shared_ptr;
+
+TMasterMetrics::TMasterMetrics(const sp_string& sinks_filename, shared_ptr<EventLoop> eventLoop) {
   sinks_reader_ = new config::MetricsSinksReader(eventLoop, sinks_filename);
   std::list<std::pair<sp_string, sp_string> > metrics;
   sinks_reader_->GetTMasterMetrics(metrics);
--- a/heron/instance/src/cpp/boltimpl/bolt-instance.cpp
+++ b/heron/instance/src/cpp/boltimpl/bolt-instance.cpp
@@ -35,11 +35,11 @@
 namespace heron {
 namespace instance {
 
-BoltInstance::BoltInstance(EventLoop* eventLoop,
-                           std::shared_ptr<TaskContextImpl> taskContext,
-                           NotifyingCommunicator<google::protobuf::Message*>* dataToSlave,
-                           NotifyingCommunicator<google::protobuf::Message*>* dataFromSlave,
-                           void* dllHandle)
+BoltInstance::BoltInstance(std::shared_ptr<EventLoop> eventLoop,
+    std::shared_ptr<TaskContextImpl> taskContext,
+    NotifyingCommunicator<unique_ptr<google::protobuf::Message>>* dataToSlave,
+    NotifyingCommunicator<google::protobuf::Message*>* dataFromSlave,
+    void* dllHandle)
   : taskContext_(taskContext), dataToSlave_(dataToSlave),
     dataFromSlave_(dataFromSlave), eventLoop_(eventLoop), bolt_(NULL), active_(false),
     tickTimer_(-1) {
@@ -114,10 +114,11 @@ void BoltInstance::executeTuple(const proto::api::StreamId& stream,
   metrics_->executeTuple(stream.id(), stream.component_name(), endTime - startTime);
 }
 
-void BoltInstance::HandleGatewayTuples(proto::system::HeronTupleSet2* tupleSet) {
+void BoltInstance::HandleGatewayTuples(unique_ptr<proto::system::HeronTupleSet2> tupleSet) {
   if (tupleSet->has_control()) {
     LOG(FATAL) << "Bolt cannot get incoming control tuples from other components";
   }
+
   if (tupleSet->has_data()) {
     for (int i = 0; i < tupleSet->data().tuples_size(); ++i) {
       auto t = new proto::system::HeronDataTuple();
@@ -128,7 +129,7 @@ void BoltInstance::HandleGatewayTuples(proto::system::HeronTupleSet2* tupleSet) 
       executeTuple(tupleSet->data().stream(), tup);
     }
   }
-  delete tupleSet;
+
   if (dataFromSlave_->size() > maxWriteBufferSize_) {
     dataToSlave_->stopConsumption();
   }
--- a/heron/instance/src/cpp/gateway/stmgr-client.cpp
+++ b/heron/instance/src/cpp/gateway/stmgr-client.cpp
@@ -32,12 +32,13 @@
 namespace heron {
 namespace instance {
 
-StMgrClient::StMgrClient(EventLoop* eventLoop, const NetworkOptions& options,
-                         const std::string& topologyName, const std::string& topologyId,
-                         const proto::system::Instance& instanceProto,
-                         std::shared_ptr<GatewayMetrics> gatewayMetrics,
-                         std::function<void(proto::system::PhysicalPlan*)> pplanWatcher,
-                         std::function<void(proto::system::HeronTupleSet2*)> tupleWatcher)
+StMgrClient::StMgrClient(
+                   std::shared_ptr<EventLoop> eventLoop, const NetworkOptions& options,
+                   const std::string& topologyName, const std::string& topologyId,
+                   const proto::system::Instance& instanceProto,
+                   std::shared_ptr<GatewayMetrics> gatewayMetrics,
+                   std::function<void(std::unique_ptr<proto::system::PhysicalPlan>)> pplanWatcher,
+                   std::function<void(std::unique_ptr<proto::system::HeronTupleSet2>)> tupleWatcher)
     : Client(eventLoop, options),
       topologyName_(topologyName),
       topologyId_(topologyId),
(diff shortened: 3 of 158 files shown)


heron/common/src/cpp/metrics/tmaster-metrics.cpp
```diff
@@ -35,7 +35,9 @@
 namespace heron {
 namespace common {
 
-TMasterMetrics::TMasterMetrics(const sp_string& sinks_filename, EventLoop* eventLoop) {
+using std::shared_ptr;
```
nwangtw: the others use std::shared_ptr and here "using" is used. Better to be consistent.

heron/instance/src/cpp/boltimpl/bolt-instance.cpp
```diff
@@ -35,11 +35,11 @@
 namespace heron {
 namespace instance {
 
-BoltInstance::BoltInstance(EventLoop* eventLoop,
-                           std::shared_ptr<TaskContextImpl> taskContext,
-                           NotifyingCommunicator<google::protobuf::Message*>* dataToSlave,
-                           NotifyingCommunicator<google::protobuf::Message*>* dataFromSlave,
-                           void* dllHandle)
+BoltInstance::BoltInstance(std::shared_ptr<EventLoop> eventLoop,
+                         std::shared_ptr<TaskContextImpl> taskContext,
```
nwangtw: indent

heron/instance/src/cpp/gateway/stmgr-client.cpp
```diff
@@ -97,32 +98,35 @@ void StMgrClient::HandleClose(NetworkErrorCode code) {
   AddTimer([this]() { this->OnReconnectTimer(); }, reconnect_interval_ * 1000 * 1000);
 }
 
-void StMgrClient::HandleRegisterResponse(void*, proto::stmgr::RegisterInstanceResponse* response,
-                                         NetworkErrorCode status) {
+void StMgrClient::HandleRegisterResponse(
+                                        void*,
```
nwangtw: indent

heron/instance/src/cpp/slave/slave.cpp
```diff
@@ -79,19 +80,22 @@ void Slave::InternalStart() {
...
+    auto tupleSet = unique_ptr<proto::system::HeronTupleSet2>(
+            static_cast<proto::system::HeronTupleSet2*>(msg.release()));
+    HandleStMgrTuples(std::move(tupleSet));
   }
 }
 
-void Slave::HandleNewPhysicalPlan(proto::system::PhysicalPlan* pplan) {
-  taskContext_->newPhysicalPlan(pplan);
+void Slave::HandleNewPhysicalPlan(unique_ptr<proto::system::PhysicalPlan> pplan) {
+  std::shared_ptr<proto::system::PhysicalPlan> newPplan = std::move(pplan);
```
nwangtw: why convert unique to shared here? Or in another word, is unique + weak better here? But if the convention is to use shared ptr in member functions, then it is totally fine.

I am wondering which function was responsible for deleting this pointer?

heron/statemgrs/src/cpp/statemgr/heron-statemgr.cpp
```diff
@@ -33,7 +33,8 @@ namespace heron {
 namespace common {
 
 shared_ptr<HeronStateMgr> HeronStateMgr::MakeStateMgr(const std::string& _zk_hostport,
-                                           const std::string& _topleveldir, EventLoop* eventLoop,
+                                           const std::string& _topleveldir,
```
nwangtw: indent

heron/statemgrs/src/cpp/statemgr/heron-statemgr.h
```diff
@@ -64,7 +64,8 @@ class HeronStateMgr {
 
   // Factory method to create
   static shared_ptr<HeronStateMgr> MakeStateMgr(const std::string& _zk_hostport,
-                                             const std::string& _topleveldir, EventLoop* eventLoop,
+                                             const std::string& _topleveldir,
```
nwangtw: indent

heron/stmgr/src/cpp/manager/checkpoint-gateway.cpp
```diff
@@ -73,30 +72,34 @@ void CheckpointGateway::SendToInstance(sp_int32 _task_id,
...
   }
+
   sp_int32 task_id = _message->task_id();
   sp_uint64 size = _message->set().size();
-  CheckpointInfo* info = get_info(task_id);
-  _message = info->SendToInstance(_message, size);
-  if (!_message) {
+  CheckpointInfo& info = get_info(task_id);
+
+  proto::stmgr::TupleStreamMessage *raw_message = info.SendToInstance(_message.release(), size);
```
nwangtw: I am wondering about the difference of .release() and std::move()?

heron/stmgr/src/cpp/manager/checkpoint-gateway.h
```diff
@@ -85,7 +89,7 @@ class CheckpointGateway {
     std::deque<Tuple> ForceDrain();
     void Clear();
    private:
-    void add(Tuple _tuple, sp_uint64 _size);
+    void add(Tuple& _tuple, sp_uint64 _size);
     void add_front(Tuple _tuple, sp_uint64 _size);
```
nwangtw: maybe better to be "Tuple&" too here like the function above?

heron/stmgr/src/cpp/manager/instance-server.cpp
```diff
@@ -87,13 +87,14 @@ const sp_string CONNECTION_BUFFER_LENGTH_BY_INSTANCEID =
...
 
-InstanceServer::InstanceServer(EventLoop* eventLoop, const NetworkOptions& _options,
-                               const sp_string& _topology_name, const sp_string& _topology_id,
-                               const sp_string& _stmgr_id,
-                               const std::vector<sp_string>& _expected_instances, StMgr* _stmgr,
-                               heron::common::MetricsMgrSt* _metrics_manager_client,
-                               NeighbourCalculator* _neighbour_calculator,
-                               bool _droptuples_upon_backpressure)
+InstanceServer::InstanceServer(shared_ptr<EventLoop> eventLoop, const NetworkOptions& _options,
+                             const sp_string& _topology_name, const sp_string& _topology_id,
```
nwangtw: indent

heron/instance/src/cpp/gateway/stmgr-client.cpp
dnrusakov: Here and in other "indent" comments, if i align the indent, then the code in the line will hit the max line limit (100 chars) and, as a result, a compilation fails on a style-check step...

heron/instance/src/cpp/slave/slave.cpp
dnrusakov: In the line 98 we call ```TaskContextImpl::newPhysicalPlan(std::shared_ptr<proto::system::PhysicalPlan> pplan)``` which takes the shared pointer and then store it in the field of the class (`TaskContextImpl`).
Then `HandleNewPhysicalPlan` (after call of `TaskContextImpl::newPhysicalPlan(newPplan)`), the handler (`HandleNewPhysicalPlan`) proceeds to use `pplan` down the road, - that means we can not move `pplan` into `TaskContextImpl::newPhysicalPlan(...)` if it would be a `unique_ptr`.

> I am wondering which function was responsible for deleting this pointer?
Previously it was handled by destructor of `TaskContextImpl` class. Now `TaskContextImpl` stores `pplan` as a smart pointer and the memory will be freed automatically as soon as the instance of `TaskContextImpl` destructed.

heron/instance/src/cpp/gateway/stmgr-client.cpp
nwangtw: I see. In that case, it might be better to be 4-space indent:

void StMgrClient::HandleRegisterResponse(
    void*,
    proto::stmgr::RegisterInstanceResponse* response,
    NetworkErrorCode status) {
  // implementation
}

heron/instance/src/cpp/slave/slave.cpp
nwangtw: kk. Thanks.

heron/stmgr/src/cpp/manager/checkpoint-gateway.cpp
dnrusakov: We can not move here because `tuplestream_drainer_` and all the chain down the road expects just a raw pointer. That's why i converted `uniq_ptr` into a raw pointer using `release`. `release` basically tells that this concrete smart pointer is not responsible for managing this concrete raw pointer any more.

nwangtw:
@thinker0 Just FYI. This PR could potentially be helpful to your memory leak issue. Dmitry found some suspicious code. But we are not 100% sure since we haven't seen it here.

thinker0:
I'm testing it.

heron/instance/src/cpp/boltimpl/bolt-instance.cpp
dnrusakov: fixed

heron/instance/src/cpp/gateway/stmgr-client.cpp
dnrusakov: fixed

heron/statemgrs/src/cpp/statemgr/heron-statemgr.h
dnrusakov: fixed

heron/stmgr/src/cpp/manager/checkpoint-gateway.h
dnrusakov: fixed

heron/stmgr/src/cpp/manager/instance-server.cpp
dnrusakov: fixed

heron/statemgrs/src/cpp/statemgr/heron-statemgr.cpp
dnrusakov: fixed

heron/stmgr/src/cpp/manager/checkpoint-gateway.cpp
dnrusakov: discussed offline

PR #3283, opened 2019-06-06
Set to Kryo serializer when registerKryoSerialization is called

nwangtw:
In Storm, users just need to register class serializer. Kryo is the default serializer. However in Heron, the default is Java serializer which doesn't have the functionality. I think it is more consistent to set to KryoSerializer when the registration function is called.

Another option is to set Kryo serializer as the default one. The concern is that some users might be affected by the implicit change.

I am totally open to discussion.

### DIFF
--- a/heron/api/src/java/org/apache/heron/api/Config.java
+++ b/heron/api/src/java/org/apache/heron/api/Config.java
@@ -480,11 +480,19 @@ public class Config extends HashMap<String, Object> {
   }
 
   public static void registerKryoSerialization(Map<String, Object> conf, Class klass) {
+    // Set to Kryo Seralizer when this function is caled. In Storm API, Kryo seralizer is
+    // the default but it is not in Heron. This implicit set could make migration easier
+    // since it is more consistent.
+    setSerializationClassName("org.apache.heron.api.serializer.KryoSerializer");
     getRegisteredKryoSerializations(conf).add(klass.getName());
   }
 
   public static void registerKryoSerialization(
       Map<String, Object> conf, Class klass, Class<? extends Serializer> serializerClass) {
+    // Set to Kryo Seralizer when this function is caled. In Storm API, Kryo seralizer is
+    // the default but it is not in Heron. This implicit set could make migration easier
+    // since it is more consistent.
+    setSerializationClassName("org.apache.heron.api.serializer.KryoSerializer");
     Map<String, String> register = new HashMap<>();
     register.put(klass.getName(), serializerClass.getName());
     getRegisteredKryoSerializations(conf).add(register);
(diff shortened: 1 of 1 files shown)


heron/api/src/java/org/apache/heron/api/Config.java
```diff
@@ -480,11 +480,19 @@ private static List getRegisteredKryoSerializations(Map<String, Object> conf) {
...
+    // Set to Kryo Seralizer when this function is caled. In Storm API, Kryo seralizer is
+    // the default but it is not in Heron. This implicit set could make migration easier
+    // since it is more consistent.
+    setSerializationClassName("org.apache.heron.api.serializer.KryoSerializer");
     getRegisteredKryoSerializations(conf).add(klass.getName());
   }
 
   public static void registerKryoSerialization(
       Map<String, Object> conf, Class klass, Class<? extends Serializer> serializerClass) {
+    // Set to Kryo Seralizer when this function is caled. In Storm API, Kryo seralizer is
```
huijunwu: `called`

heron/api/src/java/org/apache/heron/api/Config.java
```diff
@@ -480,11 +480,19 @@ private static List getRegisteredKryoSerializations(Map<String, Object> conf) {
   }
 
   public static void registerKryoSerialization(Map<String, Object> conf, Class klass) {
+    // Set to Kryo Seralizer when this function is caled. In Storm API, Kryo seralizer is
```
huijunwu: `called`

huijunwu:
The Heron API is not bound to Storm API. The translation is done by Heron-Storm API. Ideally, the Storm compatibility is supposed to be done in Heron-Storm API.

heron/api/src/java/org/apache/heron/api/Config.java
```diff
@@ -480,11 +480,19 @@ private static List getRegisteredKryoSerializations(Map<String, Object> conf) {
...
+    // the default but it is not in Heron. This implicit set could make migration easier
+    // since it is more consistent.
+    setSerializationClassName("org.apache.heron.api.serializer.KryoSerializer");
```
nlu90: create a final field for this `"org.apache.heron.api.serializer.KryoSerializer"` string and reuse it everywhere

nwangtw:
> The Heron API is not bound to Storm API. The translation is done by Heron-Storm API. Ideally, the Storm compatibility is supposed to be done in Heron-Storm API.

Yeah. The translation works (although hacky). This is for native Heron API which works a little differently from the Heron-Storm API.

heron/api/src/java/org/apache/heron/api/Config.java
nwangtw: kk. done.

PR #3291, opened 2019-06-17
Docker build should not remove heron-core.tar.gz

windhamwong:
The file `/usr/local/heron/dist/heron-core.tar.gz` is used during `heron submit` command. This file should not be removed during docker build, otherwise `heron submit` will not succeed.


nwangtw:
KK. Make sense. It was removed to reduce image size, but "docker run" was used for testing instead of "heron submit". This change makes sense. We will find another way to reduce size later.

Thanks.

Issue #3287, opened 2019-06-12
Some components do not properly override config


rohanag12:
The Heron executor reads statemgr locations from the `statemgr.yaml` config file, but it does not read overrides from `override.yaml`. This causes it to connect to the wrong statemgr if a topology relies on the override config file.

The same applies to metrics manager. It picks out certain config from the override config, but ignores some important ones (like the list of metrics sinks to use).

Is this by design? If so, what is the reasoning behind it?
If not, can I submit a patch to make these components properly override config by reading from `override.yaml`?

nwangtw:
I think it is likely to be an in-completed feature and good to move forward.

@huijunw @nlu90 you can confirm?

huijunwu:
all components should work in the same way - read the default config yaml, then override the config with `override.yaml`

rohanag12:
Makes sense, thanks! I'll send out patches for some of the components I found that do not override the config.

nlu90:
Here is how the override config operation is done: [Config Override](https://github.com/apache/incubator-heron/blob/5ee2490c025c7684c097962cf6d25f54ead73cf4/heron/spi/src/java/org/apache/heron/spi/common/ConfigLoader.java#L61). By design, all the default configs will be overridden by the override config.

Could you be more specific which configs are not overridden correctly?

rohanag12:
@nlu90 The `loadConfig` function you linked looks ideal. However, it is not used in all Java code.

Take a look at https://github.com/apache/incubator-heron/compare/master...uhanaio:executor-overload-config for an example. In metrics manager, the list of metrics sinks to load is read separately, in the [MetricsSinksConfig](https://github.com/uhanaio/incubator-heron/blob/master/heron/metricsmgr/src/java/org/apache/heron/metricsmgr/MetricsSinksConfig.java) class.

Also part of that diff is a change to heron executor, where it does not override the statemgr config from `overrides.yaml`.

rohanag12:
Resolved by #3292 and #3293

PR #3292, opened 2019-06-17
Bug: Read override config file in heron executor

rohanag12:
Heron executor should read config from `override.yaml` when connecting to the statemgr. More details in #3287.

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -1109,10 +1109,15 @@ class HeronExecutor(object):
     Receive updates to the packing plan from the statemgrs and update processes as needed.
     """
     Log.info("Start state manager watches")
+
+    with open(self.override_config_file, 'r') as stream:
+      overrides = yaml.load(stream)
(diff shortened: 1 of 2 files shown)


rohanag12:
@nwangtw @nlu90

heron/executor/src/python/heron_executor.py
```diff
@@ -1109,10 +1109,15 @@ def start_state_manager_watches(self):
     Receive updates to the packing plan from the statemgrs and update processes as needed.
     """
     Log.info("Start state manager watches")
+
+    with open(self.override_config_file, 'r') as stream:
+      overrides = yaml.load(stream)
+    overrides["heron.statemgr.connection.string"] = self.state_manager_connection
```
xiaoyao1991: Can `yaml.load` fail in any case? If so, should we check null and initialize `overrides` if necessary?

rohanag12: You're right, if the file is empty, `yaml.load` returns `None`. Added a check, thanks!

xiaoyao1991:
👍

PR #3296, opened 2019-06-20
Feature/helm

the-scott-hand:
branched off my bug/helm branch as to not hopefully cause conflicts with the prometheus fix.

These are 3/4 features I'd like to add, mentioned in [this issue](https://github.com/apache/incubator-heron/issues/3288)

I did not add the feature for using bare metal with volumeClaimTemplates, for the time being I will just use minikube as my platform since I don't have full understanding of why those distinctions were made. 

Leaving this as WIP for now so I don't forget, just want to double check some stuff and ran out of time this morning.

Thanks guys!


the-scott-hand:
removed work in progress, seems to work for me.

nwangtw:
> removed work in progress, seems to work for me.

Cool. Thanks!

nwangtw:
Let me merge it~

PR #3299, opened 2019-06-24
Retry tunnel setup when it fails

nwangtw:
Tunnel setup could fail. Retry for a few times when there is an issue.

### DIFF
--- a/heron/statemgrs/src/java/org/apache/heron/statemgr/zookeeper/curator/CuratorStateManager.java
+++ b/heron/statemgrs/src/java/org/apache/heron/statemgr/zookeeper/curator/CuratorStateManager.java
@@ -63,7 +63,7 @@ import org.apache.zookeeper.Watcher;
 
 public class CuratorStateManager extends FileSystemStateManager {
   private static final Logger LOG = Logger.getLogger(CuratorStateManager.class.getName());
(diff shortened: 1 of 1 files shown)


heron/statemgrs/src/java/org/apache/heron/statemgr/zookeeper/curator/CuratorStateManager.java
```diff
@@ -83,17 +84,24 @@ public void initialize(Config newConfig) {
...
     if (tunnelConfig.isTunnelNeeded()) {
-      Pair<String, List<Process>> tunneledResults = setupZkTunnel(tunnelConfig);
+      for (int setupCount = 0;; ++setupCount) {
+        Pair<String, List<Process>> tunneledResults = setupZkTunnel(tunnelConfig);
+        String newConnectionString = tunneledResults.first;
+
+        // If tunnel can't be setup correctly. Retry or bail.
+        if (newConnectionString.isEmpty()) {
+          if (setupCount < TUNNEL_SETUP_RETRY) {
+            continue;  // Retry
```
nlu90: wait for some time before another retry will be more favorable.

nwangtw: make sense. thx.

PR #3301, opened 2019-06-25
Add license header to new website files, update license check script …

nwangtw:
The previous release has been removed from Apache Rat project and changed to 0.13.

- Updated the Apache Rat version
- Update scripts and doc to take a path to the Rat jar file instead of hard coded path
- Add license header to two new files to pass the license check.


xiaoyao1991:
👍

PR #3307, opened 2019-07-04
Add license header to md files

nwangtw:
license header was missing in Markdown files.

"sh ./scripts/release_check/license_check.sh ~/Downloads/apache-rat-0.12/apache-rat-0.12.jar" passed.

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -4,7 +4,6 @@
 # Data/doc files
 .*\.conf
 .*\.json
-.*\.md
 .*\.txt
 .*\.BUILD
 .*\.patch
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,3 +1,21 @@
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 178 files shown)


joshfischer1108:
@nwangtw / @xiaoyao1991 I think this PR broke  the new static site.  The website will serve from localhost but when I try to click on the `docs` link in the header the navigation does not work..  Is it possible these comments/headers broke the site?

joshfischer1108:
I've checked the Pulsar repo (they use the same web page framework).  They do not have license headers in their markdown files for their website.

nwangtw:
Header is a comment. Link shouldn’t be affected I think

发自我的 iPhone

joshfischer1108:
I agree the comment header shouldn't be a problem.. But the site doesn't work anymore.  The last commit I merged in worked.  Today the site does not work. I looked at the commit history, the only change was this PR.

joshfischer1108:
I think we need to back this commit out or just revert the changes in the `website2` directory. This broke the new site.  The Pulsar repo doesn't have apache headers in their site markdown files.=.  I'm guessing they experienced the same issue.  This folder should added the rat excludes file I believe.

PR #3309, opened 2019-07-06
Annotate 3rd party and UI libraries in rat excludes with licenses

nwangtw:
For issue https://github.com/apache/incubator-heron/issues/3305

license check passes:

sh ./scripts/release_check/license_check.sh ~/Downloads/apache-rat-0.12/apache-rat-0.12.jar

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -16,8 +16,9 @@ heron\.iml
 # Directories
 bazel-.*
 m4
-website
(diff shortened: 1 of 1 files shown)


nwangtw:
Thanks.

Yeah. Website and website2 are not part of the release since we will manage the publish separately. I will add another annotation for them.

PR #3310, opened 2019-07-08
Allow interpolate more spout config parameters in extra links

xiaoyao1991:


heron/tools/tracker/src/python/config.py
```diff
@@ -54,7 +57,7 @@ def load_configs(self):
 
     if SPOUT_EXTRA_LINKS_KEY in self.configs:
       for extra_link in self.configs[SPOUT_EXTRA_LINKS_KEY]:
-        self.spout_extra_links[extra_link[SPOUT_TYPE_KEY]] = [self.validate_extra_link(link) for link in extra_link[EXTRA_LINKS_KEY]]
+        self.spout_extra_links[extra_link[SPOUT_TYPE_KEY]] = extra_link[EXTRA_LINKS_KEY]
```
nwangtw: why not need to validate?

heron/tools/config/src/yaml/tracker/heron_tracker.yaml
```diff
@@ -74,12 +77,12 @@ statemgrs:
 # - spout.type: "kafka"
 #   extra.links:
 #   - name: "Viz"
-#     formatter: "http://127.0.0.1/kafka/viz/${ENVIRON}-${CLUSTER}-${TOPOLOGY}-${SPOUT_NAME}-${SPOUT_SOURCE}"
+#     formatter: "http://127.0.0.1/kafka/viz/${ENVIRON}-${CLUSTER}-${TOPOLOGY}-${spout.cluster}-${spout.source}"
 #   - name: "Alerts"
-#     formatter: "http://127.0.0.1/kafka/alerts/${ENVIRON}-${CLUSTER}-${TOPOLOGY}-${SPOUT_NAME}-${SPOUT_SOURCE}"
+#     formatter: "http://127.0.0.1/kafka/alerts/${ENVIRON}-${CLUSTER}-${TOPOLOGY}-${spout.cluster}-${spout.source}"
```
nwangtw: can we make the new vars more consistent?

heron/tools/tracker/src/python/config.py
xiaoyao1991: The way we were doing validation worked for extra links with only pre-defined params, but it's hard to pre-define all possible params beforehand. In addition, different spout types could have different sets of params.

nwangtw: kk. thx

Issue #3266, opened 2019-05-21
Java 9+ build support


rohanag12:
JDK 11 is the current LTS Java release. JDK 8 will no longer receive public security updates. Heron should be made compatible with Java 9+.

Currently state is that the Heron project does not compile with JDK 9 or higher. The first error that I see when trying to build with JDK 11 is this - 

```
bazel build heron/...
WARNING: Running Bazel server needs to be killed, because the startup options are different.
WARNING: Waiting for server process to terminate (waited 5 seconds, waiting at most 60)
Starting local Bazel server and connecting to it...
WARNING: /home/ubuntu/incubator-heron/heron/healthmgr/tests/java/BUILD:52:12: in srcs attribute of java_library rule //heron/healthmgr/tests/java:healthmgr-tests: please do not import '//heron/healthmgr/src/java:org/apache/heron/healthmgr/HealthManager.java' directly. You should either move the file to this package or depend on an appropriate rule there
INFO: Analyzed 609 targets (299 packages loaded, 4190 targets configured).
INFO: Found 609 targets...
INFO: Deleting stale sandbox base /home/ubuntu/.cache/bazel/_bazel_rohan/6ae13315cd4f4a72906d48bc12f08d52/sandbox
ERROR: /home/ubuntu/incubator-heron/heron/api/src/java/BUILD:22:1: Building heron/api/src/java/libapi-java-low-level.jar (130 source files) failed (Exit 1)
heron/api/src/java/org/apache/heron/api/Config.java:31: error: package javax.xml.bind does not exist
import javax.xml.bind.DatatypeConverter;
                     ^
heron/api/src/java/org/apache/heron/api/Config.java:698: error: cannot find symbol
    optsBase64 = DatatypeConverter.printBase64Binary(
                 ^
  symbol:   variable DatatypeConverter
  location: class Config
heron/api/src/java/org/apache/heron/api/Config.java:700: error: cannot find symbol
    componentBase64 = DatatypeConverter.printBase64Binary(
                      ^
  symbol:   variable DatatypeConverter
  location: class Config
INFO: Elapsed time: 78.103s, Critical Path: 53.08s
INFO: 66 processes: 40 local, 26 worker.
FAILED: Build did NOT complete successfully
```

Will try to fix this and build again, will keep this thread updated. If anyone has successfully tried building Heron with JDK 11, any pointers would be appreciated.

nwangtw:
I think the same error happens to Java11 as well.

[...]

PR #3311, opened 2019-07-09
Flip buffers using helper method

rohanag12:
Java 8 and 11 have a small incompatible change in the `java.nio.Buffer` API. This issue crops up if Heron topologies are compiled with JDK 11, but run with Java 8.

Until Java 8, calling the `flip()` method on any object of `Buffer` (or subclasses) returns a `Buffer`. For example, `ByteBuffer#flip()` returns a `Buffer` in Java 8.

In Java 11, all subclasses of `Buffer` return objects of the subclass instead of `Buffer`, i.e. `ByteBuffer#flip()` returns a `ByteBuffer` in Java 11.

To work around this incompatibility between Java versions, we can cast the buffers to `Buffer` before flipping, since we do not care about the return value anyways. This causes the resulting bytecode to be the same regardless of which Java version is used to compile it.

Progress towards #3266

### DIFF
--- a/heron/common/src/java/org/apache/heron/common/network/BufferHelper.java
+++ b/heron/common/src/java/org/apache/heron/common/network/BufferHelper.java
@@ -1,0 +1,45 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.heron.common.network;
+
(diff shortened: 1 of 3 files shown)


nwangtw:
/home/travis/.cache/bazel/_bazel_travis/be6dac4936703c7eedcb4f5cf38cdd65/execroot/org_apache_heron/heron/common/src/java/org/apache/heron/common/network/BufferHelper.java:28:1: Utility classes should not have a public or default constructor. [HideUtilityClassConstructor]

Need a private constructor for util classes.

heron/common/src/java/org/apache/heron/common/network/BufferHelper.java
```diff
@@ -0,0 +1,48 @@
...
+   * If this function is used, any object of {@link Buffer} (and subclasses) are first cast to
+   * {@link Buffer} objects, then flipped, thus avoiding the binary incompatibility.
+   *
+   * @param buffer The buffer to flip
+   */
+  static void flip(Buffer buffer) {
+    buffer.flip();
+  }
+
+  private BufferHelper() {
```
nwangtw: Suggest moving this function to the beginning. not a blocker.

rohanag12: Done, thanks!

rohanag12:
Build failed on an integration test this time.Pushed an empty commit to re-run it.

rohanag12:
@nwangtw As far as I can tell, the last few builds are failing due to the flaky integration tests. Is there anything else I can at the moment other than re-triggering the build?

nwangtw:
> @nwangtw As far as I can tell, the last few builds are failing due to the flaky integration tests. Is there anything else I can at the moment other than re-triggering the build?

You can ping committers to restart the build. :)

It is green now. Let me merge~ Thanks!

PR #3313, opened 2019-07-15
adding wait for zookeeper script

joshfischer1108:
@nwangtw @xiaoyao1991 @kramasamy This PR fixes the debian9 and ubuntu:14.04 images that are used in the helm chart..  It will allow for us to use newly created artifacts. Please review.

### DIFF
--- a/docker/dist/Dockerfile.dist.debian9
+++ b/docker/dist/Dockerfile.dist.debian9
@@ -82,4 +82,5 @@ RUN mkdir -p /opt/zookeeper/scripts
 ADD dist/scripts /opt/zookeeper/scripts
 RUN chmod +x /opt/zookeeper/scripts/generate-zookeeper-config.sh \
     && chmod +x /opt/zookeeper/scripts/zookeeper-ruok.sh \
-    && chmod +x /opt/zookeeper/scripts/start-zookeeper.sh
+    && chmod +x /opt/zookeeper/scripts/start-zookeeper.sh \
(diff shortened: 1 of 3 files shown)


joshfischer1108:
@jerrypeng @aahmed-se @kramasamy @cckellogg Could any of you review these changes please?

Thanks in advance.

joshfischer1108:
I imagine quorum would be what we wanted, I'm not sure to be honest.  The reason I added this script in is because the helm chart as it is in the repo today expects this script to be executed.  Upon startup of the heron tools the init containers are waiting for zookeeper.  See [here](https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/helm/templates/tools.yaml#L70).  I'm trying to pick up where the chart was left off at..  If you use the docker scripts in this [folder](https://github.com/apache/incubator-heron/tree/master/docker) to build a heron docker image with the base as debian9 the helm deployment will fail since this `wait-for-zookeeper.sh` script is not present in the container.  With this change I was able to get the helm chart to deploy successfully.  Not knowing the full story of the whats or why something was done, I wanted some review on this from someone(s) who may be more familiar with it.

joshfischer1108:
I agree that waiting for limited amount of time would probably be better, but I'm unsure of how we would handle that failed deployment. I'd like to merge this for now. Then find a better way to handle failures once we get more feedback from users.

nwangtw:
KK. SGTM. I am going to approve and see if other guys have any other concerns or not before merging to master.

PR #3316, opened 2019-07-19
When create a smart pointer we now allocate the most hot/frequent messages in the memory pool instead of the heap

dnrusakov:
When create a smart pointer we now allocate the most hot/frequent messages in the memory pool instead of the heap. It should fix the performance regression we have in the new Heron version.

### DIFF
--- a/heron/common/src/cpp/basics/basics.h
+++ b/heron/common/src/cpp/basics/basics.h
@@ -46,6 +46,9 @@
 #include "basics/sptest.h"
 #include "basics/mempool.h"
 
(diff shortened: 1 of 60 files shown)


dnrusakov:
@nwangtw This PR was sort of "work in progress" because we found it didn't fix the performance issue. I didn't know you were going to merge it immediately. Actually, it's my bad cause I didn't add "WIP" into the PR header... However, I don't think we have to revert it since it doesn't do anything bad. We may consider this PR like an independent optimization...

PR #3318, opened 2019-07-21
enabling prometheus sink for local use

joshfischer1108:
@nwangtw By default the prometheus metric sink is not enabled when running Heron locally.  To enable the sink locally users will have to recompile Heron.  This is not user friendly when trying to get familiar with the system. This PR will enable prometheus for local use in future releases.


joshfischer1108:
Closing so that I may do further testing before re-opening.

PR #3322, opened 2019-07-30
Optimize hot spots in Stream Manager

dnrusakov:
The performance degradation happened after we switched to use smart pointers in the Heron Stream Manager. I made some profiling to figure out why it happened. This is what I have found.

1. Logically, `unique_ptr`(which we mainly used) have no extra cost because it is just a RAII wrapper on top of the raw pointer. Every time we access it, `get()` method is called to reach the underlying raw pointer. However, the problem is **the compiler doesn't inline** that wrapper methods (like `get()`) by default and they become a real functions calls in the release build. A function call has some invocation cost and that has led us to the performance penalty. The solution is to force the compiler to inline small methods. To do it we have to build the release build with the "-O3" flag.

2. When I was profiling this code I also noticed that `std::unordered_map` in `tuple-cache` consumed a lot of CPU and it was actually number 1 in a slow methods list. I replaced it with `hopscotch_map` which is more efficient hash map implementations. After that `tuple-cache` disappeared from the hot spots list.


dnrusakov:
This is the difference in the CPU consumption when we were running the PR with `-O3`:

stmgr_cpu_utilization (1)

- Running with current master up until 14:20
- Running with --copt=-03 and hopscotch hashmap after 14:20

This chart and testing provided by guys from Heron community.
@Aditya and @rohanag12 thank you very much for your help!

PR #3324, opened 2019-08-03
Kafka bolt

simingweng:
implement a Kafka Bolt


contrib/bolts/kafka/src/java/org/apache/heron/bolts/kafka/KafkaBolt.java
```diff
@@ -0,0 +1,106 @@
...
+        producer.send(producerRecord).get();
+      } catch (InterruptedException e) {
+        LOG.error("interrupted while waiting for the record to be sent", e);
+        Thread.currentThread().interrupt();
+      } catch (ExecutionException e) {
+        LOG.error("error has occurred when sending record to Kafka", e);
+        throw new KafkaException(e);
+      }
+    } else {
+      producer.send(producerRecord, (recordMetadata, e) -> {
```
nwangtw: I am wondering if using "try {} catch {}" is more consistent.

contrib/bolts/kafka/test/java/org/apache/heron/bolts/kafka/KafkaBoltTest.java
```diff
@@ -0,0 +1,101 @@
...
+  @Test
+  public void cleanup() {
+    kafkaBolt.prepare(Collections.emptyMap(), null, outputCollector);
+    kafkaBolt.cleanup();
+    verify(producer).close();
+  }
+
+  @Test
+  public void executeATLEASTONCE() {
+    kafkaBolt.prepare(Collections.singletonMap(Config.TOPOLOGY_RELIABILITY_MODE, ATLEAST_ONCE),
```
nwangtw: suggest to test exactly once (exception) and at most once (no fail) as well.

contrib/bolts/kafka/src/java/org/apache/heron/bolts/kafka/KafkaBolt.java
simingweng: Could you please elaborate it a bit more? The idea here is that for `effective_once` mode, we need to do a blocking `send`, and throw some runtime exception (so I picked the KafkaException) if producer fails, so that the whole topology will be rolled back to the previous checkpoint; while in `atleast_once` mode, we can do an async `send`, then `ack` or `fail` the tuple so the spout will be signaled.

contrib/bolts/kafka/test/java/org/apache/heron/bolts/kafka/KafkaBoltTest.java
simingweng: I've added the extra 2 unit tests

nwangtw: thx~

PR #3330, opened 2019-08-22
[STREAMCOMP-2885] broadcast GCC completition

lpan:
## Overview
Details described in the [design doc](https://docs.google.com/document/d/1Shqp2z-uAhsOUN2erbc2d-72wdjA8anp/edit#heading=h.1egqt2p)

## Progress
- [x] TMaster broadcasts checkpoint saved to all Heron instances
- [x] add `I2PhaseCommitComponent` interface
- [x] Heron instance triggers `postSave` on receiving "checkpoint saved" message
- [x] Heron instance triggers `preRestore` on receiving "restore topology" message
- [x] block execute on `postSave` after `preSave` is invoked https://github.com/apache/incubator-heron/pull/3330/commits/e432c0e575acb0ef7d26d6af40f8f24473f21149
- [x] unit test for stateful bolt
  - [x] `preSave`
  - [x] `postSave`
  - [x] `preRestore`
  - [x] block execute on `postSave` for `I2PhaseCommitComponent` bolts
- [x] unit test for stateful spout
  - [x] `preSave`
  - [x] `postSave`
  - [x] `preRestore`
  - [x] block execute on `postSave` for `I2PhaseCommitComponent` spouts
- [x] address review comments
  - [x] fix the incorrectly formatted log statement @nlu90 
  - [x] change component name from `I2PhaseCommitComponent` to `ITwoPhaseStatefulComponent` @nwangtw 
  - [x] no need to check interface @xiaoyao1991 
  - [x] block readAnExecuteTuples in `spouts` as well @nlu90


nlu90:
Will review it once we have a TDD discussion meeting.

lpan:
Chatted with @nwangtw offline. We agreed that `I2PhaseCommitComponent` is a terrible name. Currently in the process of coming up with a better name 😛

nwangtw:
Also adding @xiaoyao1991

nwangtw:
Overall looks pretty good to me. 👍

heron/instance/src/java/org/apache/heron/instance/spout/SpoutInstance.java
```diff
@@ -296,9 +299,6 @@ public void run() {
           spoutMetrics.updateOutQueueFullCount();
         }
 
-        // Check if we have any message to process anyway
-        readTuplesAndExecute(streamInQueue);
-
```
lpan: ping @nwangtw

lpan: I made a small change to the message processing order for `SpoutInstance`.

Before, when the spout task is waken up by the event loop, it will attempt to emit tuples first. And then, it will process the control messages. Now it will process the control tuples first and then try to emit the tuples. Note that this change will only impact stateful topology (spout instances in non-stateful topology will not receive any control messages)

Implementing blocking emitting tuples until checkpoint is completed requires to process control messages first.

heron/instance/src/java/org/apache/heron/instance/IInstance.java
```diff
@@ -57,6 +57,21 @@
...
+   * @param checkpointId the ID of the checkpoint the instance will be restoring to
+   */
+  void preRestore(String checkpointId);
+
+  /**
+   * Inform the Instance that a particular checkpoint has become globally consistent
+   *
+   * @param checkpointId the ID of the checkpoint that became globally consistent
+   */
+  void onCheckpointSaved(String checkpointId);
```
nlu90: These two new interfaces are stateful topology specific. Instead of adding them here in the general `IInstance`, could we find a cleaner way to put them somewhere else?

heron/instance/src/java/org/apache/heron/instance/spout/SpoutInstance.java
nlu90: This change should be fine. 

But I think you will need to update the `realTuplesAndExecute` method in the SpoutInstance to actually implement the blocking mechanism, just like what you did for the `BoltInstance`. Currently, the blocking is implemented in the `isProduceTuple` method

heron/instance/src/java/org/apache/heron/network/StreamManagerClient.java
```diff
@@ -286,6 +289,17 @@ private void handleStartStatefulRequest(
     inControlQueue.offer(instanceControlMsg);
   }
 
+  private void handleCheckpointSaved(
+      CheckpointManager.StatefulConsistentCheckpointSaved message) {
+    LOG.info("Received a StatefulCheckpointSaved message:with checkpoint id: "
```
nlu90: remove the first `:`

heron/instance/src/java/org/apache/heron/instance/spout/SpoutInstance.java
lpan: `readTuplesAndExecute` in `SpoutInstance` only reads control tuples, and we don't need to block them. For spouts we only want to block `nextTuple` and for bolts we want to block `execute` (which takes place in the `readTuplesAndExecute` method).

heron/instance/src/java/org/apache/heron/instance/IInstance.java
lpan: Good point 👍 I will think about it

heron/instance/src/java/org/apache/heron/network/StreamManagerClient.java
lpan: > It seems the message loop is messing the later half part of sending ack from instance to tmaster to confirm the commit for a checkpoint id.

You mean "missing" right? We decided that we don't need to send `ack` back to TMaster. This is because since we will block executing tuples (or producing tuples for spouts) if the `postSave` callback on the task instance has not yet exited. As a result, the next checkpoint will never become globally consistent (since the blocked task will not have the chance to persist its state for the next checkpoint).

Therefore, all other tasks in the topology will be blocked on their `postSave` for the next checkpoint, which is the desired behaviour :)

heron/instance/src/java/org/apache/heron/instance/spout/SpoutInstance.java
nwangtw: The ordering change sounds ok to me.

One concern is the barrier though. I think each stream needs a separated blocker. I just want to confirm that this is included in the PR.

lpan: Chatted with Ning offline. He was worried about "barrier alignment" (eg. a task has multiple input streams). It is fine because "barrier alignment" is done at the stream manager 🙂

nwangtw: Yeah. Good point. Should be safe then. Thanks~

heron/instance/src/java/org/apache/heron/instance/IInstance.java
nwangtw: On the other side, init() and persistState() are in this interface, so I feel it might be ok.

heron/instance/src/java/org/apache/heron/instance/bolt/BoltInstance.java
```diff
@@ -264,14 +287,19 @@ public void run() {
     InstanceUtils.prepareTimerEvents(looper, helper);
   }
 
+  // do not process data tuple when it is a I2PhaseCommitComponent and waiting for postSave
+  private boolean isWaitingForCheckpointSaved() {
+    return bolt instanceof I2PhaseCommitComponent && waitingForCheckpointSaved;
```
xiaoyao1991: `instanceof` check does not seem necessary

xiaoyao1991:
looks good so far 👍

nwangtw:
How do you guys feel about merging this PR?

simingweng:
@nwangtw I just left a comment on the design doc, have we considered default methods instead of a whole new interface?

nwangtw:
> @nwangtw I just left a comment on the design doc, have we considered default methods instead of a whole new interface?

Thanks!

That is a good point. This implementation is more conventional and default methods might be cleaner.

lpan:
@nwangtw any updates?

nwangtw:
Let's merge and then iterate~

PR #3350, opened 2019-09-30
Clean up rate limit config and add an example topology

nwangtw:


heron/api/src/java/org/apache/heron/api/Config.java
```diff
@@ -1029,8 +1029,4 @@ public void setTopologyRemoteDebugging(boolean isOn) {
   public void setTopologyDropTuplesUponBackpressure(boolean dropTuples) {
     this.put(Config.TOPOLOGY_DROPTUPLES_UPON_BACKPRESSURE, String.valueOf(dropTuples));
   }
-
-  public void setTopologyComponentOutputBPS(long bps) {
-    this.put(Config.TOPOLOGY_COMPONENT_OUTPUT_BPS, String.valueOf(bps));
-  }
```
xiaoyao1991: Would the removal of these 2 methods break some existing topologies?

nwangtw: Shouldn't be. These two functions were added by mistake.

I double-checked the source code in topology-config-helper.cpp and it is supported to be used as component config only.

The concern is valid though. Do you feel it is safer to annotate with @Deprecated? Or is there a better annotation for it?

nwangtw:
Discussed offline. Keeping the functions with @Deprecated annotation to be safe.

xiaoyao1991:
👍

PR #3353, opened 2019-10-02
Make unknown arguments a warning instead of an error in heron executor

nwangtw:
This could give us more flexibility for debugging and customization.

Executor arguments don't change very often hence the risk is low. The warning output is in stderr and it should be easy to find.


xiaoyao1991:
👍

PR #3354, opened 2019-10-02
Separate gc log files for metrics manage and checkpoint manager by id

nwangtw:
Currently, if you are running a topology locally, all metrics managers write to the same GC log file which is not useful.

This PR separates the GC files for different metrics managers and checkpoint managers.


xiaoyao1991:
:

PR #3355, opened 2019-10-03
Add a cap to metrics communicator in TMasterSink and MetricsCacheSink

nwangtw:
Without the cap, when MetricsCacheSink is enabled but MetricsCache is not running, the communication will keep growing.

The same thing could happen with TMaster but it is less likely to have the problem.

### DIFF
--- a/heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
+++ b/heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
@@ -93,12 +93,20 @@ public class MetricsCacheClient extends HeronClient implements Runnable {
     Runnable task = new Runnable() {
       @Override
       public void run() {
-        while (!publishMetricsCommunicator.isEmpty()) {
-          TopologyMaster.PublishMetrics publishMetrics = publishMetricsCommunicator.poll();
+        TopologyMaster.PublishMetrics publishMetrics;
+        synchronized (publishMetricsCommunicator) {
+          publishMetrics = publishMetricsCommunicator.poll();
+        }
+        while (publishMetrics != null) {
           LOG.info(String.format("%d Metrics, %d Exceptions to send to MetricsCache",
               publishMetrics.getMetricsCount(), publishMetrics.getExceptionsCount()));
           LOG.fine("Publish Metrics sending to MetricsCache: " + publishMetrics.toString());
+
           sendMessage(publishMetrics);
+
+          synchronized (publishMetricsCommunicator) {
+            publishMetrics = publishMetricsCommunicator.poll();
+          }
         }
       }
     };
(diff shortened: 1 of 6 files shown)


heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
```diff
@@ -93,12 +93,20 @@ private void addMetricsCacheClientTasksOnWakeUp() {
     Runnable task = new Runnable() {
       @Override
       public void run() {
-        while (!publishMetricsCommunicator.isEmpty()) {
-          TopologyMaster.PublishMetrics publishMetrics = publishMetricsCommunicator.poll();
+        TopologyMaster.PublishMetrics publishMetrics;
+        synchronized (publishMetricsCommunicator) {
```
dnrusakov: I see that previously we didn't use `synchronized` during operations on `publishMetricsCommunicator`.
What have changed? Do we now work with `publishMetricsCommunicator` in different threads?

heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
```diff
@@ -93,12 +93,20 @@ private void addMetricsCacheClientTasksOnWakeUp() {
...
+          publishMetrics = publishMetricsCommunicator.poll();
+        }
```
xiaoyao1991: hmm, if we change the while loop condition, maybe we could reduce the code duplicate here?

heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
nwangtw: Yeah. Before there were only 1 producer (metrics manager) and 1 consumer (sink client) for this communicator from two different threads.

In the change, the producer can remove entries from head too so it is competing with the consumer thread. A simpler way is to avoid adding new entries into the queue, then we don't need to synchronize the producer and the consumer. However, I feel that the recent data could be more important than the old data.

heron/metricsmgr/src/java/org/apache/heron/metricsmgr/sink/metricscache/MetricsCacheClient.java
nwangtw: like while (synchronized (publishMetricsCommunicator) { publishMetrics = .... } ) {}?

As a non-java programmer, I don't know if it is ok. :D 

But this is a valid point. Let me see if I can improve it.

xiaoyao1991: 👍

xiaoyao1991:
👍

PR #3360, opened 2019-10-08
udpated yaml and updated tools

jemantheiy:
-updated the bookeeper and zookeeper configs to match a more production setting
-moved based url out of the tools.yaml and to the value.yaml to override.

### DIFF
--- a/deploy/kubernetes/helm/values.yaml.template
+++ b/deploy/kubernetes/helm/values.yaml.template
@@ -39,8 +39,9 @@ jobReplicas: 1
 
 # amount of memory to provide for API server
 apiServerMemory: 512M
-
-# Topologies uploader
+heron:
+  url:/api/v1/namespaces/{{ .Release.Namespace }}/services/{{ .Release.Name }}-ui:8889/proxy
+i Topologies uploader
 uploader:
   class: dlog # s3
   s3Bucket: ""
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/helm/values.yaml.template
```diff
@@ -53,18 +54,18 @@ uploader:
...
 bookieCpuMax: 250m
-bookieHeapMemory: 256M
-bookieDirectMemory: 128M
+bookieHeapMemory: 512M
+bookieDirectMemory: 256M
 bookieReadCacheSize: 32
 bookieWriteCacheSize: 32
-bookieJournalCapacity: 1G
-bookieStorageCapacity: 3G
+bookieJournalCapacity: 10G
```
nwangtw: 10G and 30G look quite big and could be a limitation for some people. Are they really necessary?

jemantheiy: we are finding using the shade plugin causes large jars and deploying a lot of jars the old configs were problematic.  This seems to be a happy spot.  I can reduce it.

deploy/kubernetes/helm/values.yaml.template
```diff
@@ -39,8 +39,9 @@ jobReplicas: 1
 
 # amount of memory to provide for API server
 apiServerMemory: 512M
-
-# Topologies uploader
+heron:
+  url:/api/v1/namespaces/{{ .Release.Namespace }}/services/{{ .Release.Name }}-ui:8889/proxy
+i Topologies uploader
```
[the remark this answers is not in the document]
jemantheiy: I moved the old proxy url out in the tools yaml and moved it to values.yaml so we can override it since our production env won't allow kubectl proxy

jemantheiy:
@nwangtw all changes made as requested

joshfischer1108:
LGTM

PR #3362, opened 2019-10-08
Add metrics for tracking bp caused by local instances and remote stmgrs

nwangtw:
These two metrics could be very helpful for locating bp issues.

Also, the existing bp by component id metrics are expensive to query and these two new ones are light weighted.


dnrusakov:
@nwangtw LGTM

PR #3365, opened 2019-10-16
publish site script

joshfischer1108:
@nwangtw This is the script that would need to be executed after the `make buildsite` command.  Please review.


joshfischer1108:
Seems the build keeps exceeding the allotted time limit.  I'll try and kick it off again.

`The job exceeded the maximum time limit for jobs, and has been terminated.`

PR #3366, opened 2019-10-18
Add a script to start docker container with development environment

nwangtw:
Mac dev environment could be tricky to set up since it requires Xcode's toolchain, and it is not easy to troubleshoot. This docker script can be useful for compiling source code without any setup other than installing Docker.

To use the script to create a docker image and run it:
$ sh docker/scripts/dev-env.sh

Write down the Docker container id (you can also find it with this command: "docker ps -a"). Next time you can run "docker container exec -it CONTAINER_ID bash" to start a new shell in the same container.

In the container:
$ ./bazel_config.py
$ bazel build --config=ubuntu heron/...

This is just the first version. There are a lot of room to improve. Any feedback is welcome~

### DIFF
--- a/docker/scripts/docker-dev-env.sh
+++ b/docker/scripts/docker-dev-env.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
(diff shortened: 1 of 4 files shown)


docker/scripts/docker-dev-env.sh
```diff
@@ -0,0 +1,75 @@
...
+#   cd heron
+#   bazel build --config=ubuntu heron/...
+
+set -o nounset
+set -o errexit
+
+# Default platform is ubuntu18.04. Other available platforms
+# include centos7
+TARGET_PLATFORM=${1:-"ubuntu18.04"}
+SCRATCH_DIR="docker"
```
joshfischer1108: Do you think we should make `SCRATCH_DIR` configurable by accepting an argument as the script is executed?

nwangtw: Make sense. Also "docker" is not ideal for the scratch dir. Updated and cleaned up in the new push.

nwangtw:
Still can't add @xiaoyao1991 so far. :(

PR #3375, opened 2019-10-22
Add instance resource usage in container

nwangtw:


heron/tools/ui/resources/static/js/physical-plan.js
```diff
@@ -63,16 +63,23 @@
...
       for (var i in containerPlan) {
         var packing = containerPlan[i];
         if (packing.id === container.index) {
           container.required_resources = packing.required_resources;
+          // Get instance resources
+          var instance_resources = {};
+          for (var j in packing.instances) {
+            var inst = packing.instances[j];
+            instance_resources[inst.component_name + '_' + inst.task_id] = inst.instance_resources;
+          }
```
xiaoyao1991: can probably be replaced with a `map`

heron/tools/ui/resources/static/js/physical-plan.js
```diff
@@ -86,6 +93,18 @@
...
+      if (byteSize > 1024 * 1024 * 1024 / 2) {
+        return (byteSize / (1024 * 1024 * 1024)).toFixed(2) + 'GB';
+      } else if (byteSize > 1024 * 1024 / 2) {
+        return (byteSize / (1024 * 1024)).toFixed(2) + 'MB';
+      } else if (byteSize > 1024 / 2) {
+        return (byteSize / 1024).toFixed(2) + 'KB';
+      } else {
+        return byteSize + 'B';
+      }
+    }
```
xiaoyao1991: There is a `pretty-bytes` npm package that does this.

nwangtw: Thanks.

nwangtw: hmm. maybe future TODO since I don't want to add new dependency now.

heron/tools/ui/resources/static/js/physical-plan.js
nwangtw: Maybe next time. :p 

Map function is not available for dict. For loop might be more straightforward here.

PR #3376, opened 2019-10-23
Fix license in health handler py

nwangtw:


joshfischer1108:
Thanks for fixing this one.  I wasn't sure what to do here.

PR #3377, opened 2019-10-24
Clean up license for jarjar help.txt

nwangtw:


LICENSE
```diff
@@ -248,10 +248,9 @@ See project link for details.
       -> scripts/packages/tests_template_bin.sh
       -> scripts/packages/bin_common.sh
       -> third_party/java/bazel/extra_actions_base.proto
-    Google
```
joshfischer1108: One question.  Any reason why you choose `jarjar` over `Google`?

nwangtw: This is my understanding only. The list here should have project names (jarjar) instead of organization or personal names (google).

The project source code is here: https://github.com/google/jarjar.

joshfischer1108: Ok. Makes sense to me.  Thanks!

PR #3378, opened 2019-10-25
Avoid duplicated metrics requests in Heron UI

nwangtw:
Same request might be sent three times which gives heron UI pressure and
makes UI slow.


nwangtw:
Suggest to enable "Hide whitespace changes" then review the PR.

Thanks.

PR #3381, opened 2019-10-26
Helm chart Bookkeeper fix

### LINKED ISSUE #3380: Bookkeeper not using hostname when run as Kubernetes Statefulset
The Kubernetes Statefulset of Bookkeeper needs to use hostname instead of IP. Because in Kubernetes, pod IPs can change, Bookkeeper needs to be configured to use the hostname. The Kubernetes yamls in the `general` folder have the needed config item, but the Helm chart does not.

nicknezis:
Fixes #3380


nwangtw:
@huijunw @nlu90 @xiaoyao1991 help to trigger a rebuild?

the-scott-hand:
This merge also requires the following change:
https://github.com/apache/incubator-heron/pull/3410

PR #3397, opened 2019-11-08
Re-organize UI elements in Heorn UI topology page

nwangtw:
In the topology page, create a navigator to show stats, counters and config.

The motivation is that counters have a lot of important information, but they are shown below the page and not very noticeable for users. They are also far from the plan diagrams so that it is hard to choose different component/instance to view the counters. Creating a "counters" tab make the counters easier to read.

Config was in a separated page which is not convenient. Now it is integrated in the same page under the "config" tab.

### DIFF
--- a/heron/tools/ui/resources/static/css/main.css
+++ b/heron/tools/ui/resources/static/css/main.css
@@ -2253,3 +2253,41 @@ nav {
 .navbar-custom .navbar-link:hover {
   color: #ffecb3;
 }
+
+/* Topology details */
+div#display-navigator {
+  border-style: solid;
+  border-width: 1px;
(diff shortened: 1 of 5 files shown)


nwangtw:
Screen Shot 2019-11-08 at 9 22 27 AM
Screen Shot 2019-11-08 at 9 27 33 AM
Screen Shot 2019-11-08 at 9 50 32 AM

nwangtw:
> keeping logical plan and packing plan visible is helpful for users to be aware of topology structure while they are navigating among tabs

You mean making the plan diagrams sticky? Yeah, I think that could be helpful. Let me think about it. Will be in a new PR. Thanks for the feedback.

PR #3401, opened 2019-11-14
Add parallelism calculator in Heron UI

nwangtw:
Simple parallelism calculator based on the current capacity utilization and execute count. Factors that could affect instance performance, such as GC and IO are ignored.


nwangtw:
Screen Shot 2019-11-14 at 10 53 23 AM

PR #3405, opened 2019-11-26
Update Kubernetes proxy URLs

ShawnZhong:
Fix https://github.com/apache/incubator-heron/issues/3404

Follow the Kubernetes proxy URLs documented [here](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#manually-constructing-apiserver-proxy-urls) (Note that `proxy` is placed after the service name instead of `/api/v1`):

> http://kubernetes_master_address/api/v1/namespaces/namespace_name/services/service_name[:port_name]/proxy

### DIFF
--- a/deploy/kubernetes/general/README.md
+++ b/deploy/kubernetes/general/README.md
@@ -75,7 +75,7 @@ $ kubectl proxy -p 8001
 2. Verify we can access the API server:
 
 ```shell
-$ curl http://localhost:8001/api/v1/proxy/namespaces/default/services/heron-apiserver:9000/api/v1/version
+$ curl http://localhost:8001/api/v1/namespaces/default/services/heron-apiserver:9000/proxy/api/v1/version
 {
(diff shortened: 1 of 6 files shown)


ShawnZhong:
Note that the helm configuration file at https://storage.googleapis.com/heron-charts needs to be updated as well. Otherwise, users following the [helm guide](https://apache.github.io/incubator-heron/docs/operators/deployment/schedulers/kubernetes-helm/) will not be able to get any stats from heron UI. 

image

I believe the file has already been updated in this repository (see below), but the file at https://storage.googleapis.com/heron-charts was not changed for a while (The last modified time is 2018-04-18). 

https://github.com/apache/incubator-heron/blob/d8db1c50b499c823e319fd16a4476a02f525a367/deploy/kubernetes/helm/templates/tools.yaml#L109-L115

joshfischer1108:
I'm going to let this sit for a day and see if anyone else has any feedback.

nwangtw:
I am not familiar with this part but the change looks reasonable. Thanks.

Please feel free to merge.

PR #3410, opened 2019-11-27
BUG: bookie statefulset helm template specifies incorrect service name

the-scott-hand:
The statefulset's service name:
https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/helm/templates/bookie.yaml#L74
Doesnt match the actual name of the service:
https://github.com/apache/incubator-heron/blob/master/deploy/kubernetes/helm/templates/bookie.yaml#L203
I believe bookie previously was using the podIPs to connect and the service wasnt actually getting used. Changing the useBookieHostNameAsID to true caused the helm chart to use the service for the first time to connect, which ultimately failed since it was named the wrong name.

### DIFF
--- a/deploy/kubernetes/helm/templates/bookie.yaml
+++ b/deploy/kubernetes/helm/templates/bookie.yaml
@@ -71,7 +71,7 @@ metadata:
     component: {{ .Release.Name }}-bookie
 spec:
 {{- if or (eq .Values.platform "gke") (eq .Values.platform "minikube") }}
-  serviceName: {{ .Release.Name }}-bookkeeper
+  serviceName: {{ .Release.Name }}-bookie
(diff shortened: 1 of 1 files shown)


the-scott-hand:
it might also be worth noting the yaml in general is doing this correctly, and also utilizing the following annotation:
https://github.com/apache/incubator-heron/blob/d8db1c50b499c823e319fd16a4476a02f525a367/deploy/kubernetes/general/bookkeeper.statefulset.yaml#L126-L127
which is the same annotation being utilized by pulsar's bookie:
https://github.com/apache/pulsar/blob/master/deployment/kubernetes/helm/pulsar/values.yaml#L97

the-scott-hand:
added an additional commit to add that annotation

PR #3411, opened 2019-11-30
Update zk version to 3.4.14 in dist and install zk in ubuntu 16/18 di…

nwangtw:
…st docker image

### DIFF
--- a/docker/base/Dockerfile.base.debian9
+++ b/docker/base/Dockerfile.base.debian9
@@ -27,9 +27,10 @@ RUN apt-get -y update && apt-get -y install \
     rm -rf /var/lib/apt/lists/*
 
 # install zookeeper
+ARG ZK_DIST=zookeeper-3.4.14
 ADD base/scripts /opt/heron-docker/scripts
 
-RUN sh /opt/heron-docker/scripts/install-zookeeper.sh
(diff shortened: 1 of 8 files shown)


nicknezis:
Looks good to me, but is there any value in keeping the Zookeeper version as a Docker build ARG? This change would lose that feature. Probably not a big deal.

nwangtw:
> Looks good to me, but is there any value in keeping the Zookeeper version as a Docker build ARG? This change would lose that feature. Probably not a big deal.

How it was used? It should be easy to add back.

joshfischer1108:
Overall this looks good to me.  Straight forward and easy to read.  I noticed several of the same changes in the different files for different distros which got me thinking.  Do we really need to support all of these different versions of ubuntu containers?  Or even further do we need to support  ubunutu, centos, and debian?  Wouldn't it be more efficient for us to have one "official" supported image?

nwangtw:
> Overall this looks good to me. Straight forward and easy to read. I noticed several of the same changes in the different files for different distros which got me thinking. Do we really need to support all of these different versions of ubuntu containers? Or even further do we need to support ubunutu, centos, and debian? Wouldn't it be more efficient for us to have one "official" supported image?

Agreed. This is a reasonable question to discuss.

PR #3424, opened 2019-12-19
Updates to make Zookeeper work

nicknezis:
Netcat was not included in the centos image.
Also the zookeeper scripts were not executable.


docker/dist/Dockerfile.dist.centos7
```diff
@@ -1,4 +1,3 @@
-#  Licensed to the Apache Software Foundation (ASF) under one
```
nwangtw: This line was deleted by mistake.

nicknezis: I fixed the error and updated my branch. Thank you!

nwangtw: > I fixed the error and updated my branch. Thank you!

thx~

PR #3426, opened 2019-12-28
Updated to support Kubernetes 1.16

nicknezis:
Updating to the latest Kubernetes Java client and API.


heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/AppsV1Controller.java
```diff
@@ -97,20 +95,13 @@ boolean submit(PackingPlan packingPlan) {
...
-      if (!response.isSuccessful()) {
-        LOG.log(Level.SEVERE, "Error creating topology message: " + response.message());
-        KubernetesUtils.logResponseBodyIfPresent(LOG, response);
-        // construct a message based on the k8s API server response
-        throw new TopologySubmissionException(
-            KubernetesUtils.errorMessageFromResponse(response));
-      }
-    } catch (IOException | ApiException e) {
+      final V1StatefulSet response =
+          client.createNamespacedStatefulSet(getNamespace(), statefulSet, null,
```
joshfischer1108: Should we check if the response is successful here?

nicknezis:
I just realized there is now a 7.0.0 release of the Kubernetes Java client. I'm upgrading to use that version. Will also update okhttp dependency from 2.7.5 to 3.4.14. This is the version referenced in the k8s client pom.xml.

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/AppsV1Controller.java
nicknezis: I believe the `ApiException` will have the status code because the `ApiClient` class checks the status code and throws the exception if there is an error. Is there any better logging you would suggest in the catch block when we deal with the `ApiException`?

Here is a link to the [`ApiClient.handleResponse()`](https://github.com/kubernetes-client/java/blob/d94b271a7dc9636d1b5d930b047b94273b58aa83/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java#L954) logic.

joshfischer1108: Nothing off the top of my head that would make it any better.  It looks like the client handles printing out enough information.  We can always come back and add more logging if we run into issues or someone requests it.

joshfischer1108:
@nicknezis I think you were doing some additional work on this PR.   Is this ready to review again or  should we wait?

nicknezis:
I think it's ready to review. I'm working on making the K8s client the internal cluster config, but that can be submitted as a separate pull request.

PR #3430, opened 2020-01-08
Helm update to v3

nicknezis:
Updated Helm to 3.0.2. Also updated the website documentation related to Helm.

### DIFF
--- a/website2/docs/schedulers-k8s-with-helm.md
+++ b/website2/docs/schedulers-k8s-with-helm.md
@@ -50,47 +50,43 @@ $ brew install kubernetes-helm
 You can install Helm on Linux using a simple installation script:
 
 ```bash
-$ curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get > install-helm.sh
-$ chmod 700 install-helm.sh
-$ ./install-helm.sh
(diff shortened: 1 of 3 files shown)


website2/docs/schedulers-k8s-with-helm.md
```diff
@@ -50,55 +50,52 @@ $ brew install kubernetes-helm
...
 ```bash
 $ kubectl config current-context
 ```
 
-If the context is correct, then you can get Helm running using just one command:
+Once you've installed the Helm client on your machine and gotten Helm pointing to your Kubernetes cluster, you need to make your client aware of the `heron-charts` Helm repository, which houses the chart for Heron:
 
 ```bash
-$ helm init
+$ helm repo add heron-charts https://storage.googleapis.com/heron-charts
```
joshfischer1108: I don't think this google bucket exists anymore.  I wonder if we should modify the instructions to build and deploy from the source.  Thoughts?

nicknezis: That's not a bad idea. I had to figure it out on my own and having the `bazel build` target to `generate-helm-charts` documented would be nice.

joshfischer1108: yeah I agree.   I'll wait to see if anyone else has any feedback.

PR #3412, opened 2019-12-04
make python code compatible with both python2 and python3

nlu90:
Given the deadline for python3 migration is drawing near, we need to move all python related code to python3 now.

This PR uses the 2to3 tools to update all the python files.

PR #3431, opened 2020-01-08
Upgrade to Python3 for the whole repo

nlu90:
Python2 has been sunsetting: https://pythonclock.org/

We also need to upgrade from python2  to python3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ addons:
       - g++-4.8
       - gcc-4.8-multilib
       - g++-4.8-multilib
-      - python2.7
+      - python3
       - wget
       - pkg-config
       - zip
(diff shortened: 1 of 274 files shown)


nlu90:
This is a follow-up PR for #3412

joshfischer1108:
I noticed travis only has the python package 2.7 listed.  Will this be a problem? https://github.com/apache/incubator-heron/blob/ab9c205a783d6e8b43144560080f2d75b9354145/.travis.yml#L22

nlu90:
> I noticed travis only has the python package 2.7 listed. Will this be a problem?
> 
> https://github.com/apache/incubator-heron/blob/ab9c205a783d6e8b43144560080f2d75b9354145/.travis.yml#L22

@joshfischer1108 Yeah, ideally we also need to update the travis config file. I will add a new commit and see how things work.

nlu90:
@joshfischer1108 With the Travis change, the ci job still passes

joshfischer1108:
@nlu90 That's fantastic.  Nice work.

nicknezis:
Should the `compile` and `dist` Dockerfiles also be updated to include the right Python packages?

Issue #3259, opened 2019-05-21
Update k8's Scheduler Code


joshfischer1108:
The scheduler code is out of date and needs to be updated to use the InCluster API in the Java Kubernetes client.  By updating the code we can remove the proxy  container that is being used today in the Helm chart.

#### This following class should be deprecated and use the AppsV1 APIs going forward.
https://github.com/apache/incubator-heron/blob/master/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/AppsV1beta1Controller.java#L52

#### the context class will have to change we no longer need a scheduler URI when running inside the cluster

https://github.com/apache/incubator-heron/blob/master/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java#L28

#### An example of getting the cluster config is below:
https://github.com/apache/pulsar/blob/master/pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/runtime/KubernetesRuntimeFactory.java#L254

nicknezis:
I'm working on this. Have changes needed for the 6.0.1 version of the Java API. Have not yet made changes to use the in cluster config, but I'll make that change also.

nicknezis:
#3426 and #3432 should complete this issue. Might still be worth updating the scheduler code to remove the scheduler URI context config item.

joshfischer1108:
@nicknezis 

> #3426 and #3432 should complete this issue. Might still be worth updating the scheduler code to remove the scheduler URI context config item.

@nicknezis Could you make an issue and describe what would need to be done to remove the scheduler URI context config item?

PR #3432, opened 2020-01-11
Updated to use default Kubernetes client discovery logic

nicknezis:
The defaultClient() call incorporates the following logic:
If `$KUBECONFIG` is defined, use that config file.
If `$HOME/.kube/config` can be found, use that.
If the in-cluster service account can be found, assume in cluster config.
Default to `localhost:8080` as a last resort.

With this code change, we'd no longer be using the scheduler config item for K8s scheduler URI. But the config item is used to create Job Links. So I couldn't get rid of that config item.

Any thoughts on how I could fully remove the scheduler URI config item?
Do you think the defaultClient() code is good enough? It seemed to work in my Minikube testing and I figured worst case that last option would point to kubectl-proxy  if we update it to use port 8080. I assume if someone needs to run the Heron API Server outside of the k8s cluster, then can configure with the `$KUBECONFIG` environment variable.

### DIFF
--- a/deploy/kubernetes/general/apiserver.yaml
+++ b/deploy/kubernetes/general/apiserver.yaml
@@ -27,7 +27,6 @@ metadata:
   namespace: default
 
 ---
-
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
@@ -44,7 +43,6 @@ subjects:
   namespace: default
 
 ---
-
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -97,13 +95,8 @@ spec:
               -D heron.uploader.dlog.topologies.namespace.uri=distributedlog://zookeeper:2181/heron
               -D heron.statefulstorage.classname=org.apache.heron.statefulstorage.dlog.DlogStorage
               -D heron.statefulstorage.dlog.namespace.uri=distributedlog://zookeeper:2181/heron
-        - name: kubectl-proxy
-          image: heron/kubectl:latest
-          command: ["sh", "-c"]
-          args:
-            - >
-              kubectl proxy -p 8001
----    
+
+---
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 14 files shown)


nicknezis:
Hopefully this addresses part of what was asked for in #3259

joshfischer1108:
I haven't looked at the code yet, but does this PR relate to the issue filed listed here? https://github.com/apache/incubator-heron/issues/3259

nicknezis:
Yes, I tested it in Minikube and was able to spin up a StatefulSet without the `kubectl-proxy`. I'll update this PR to remove that container from the Kubernetes deploy yamls and Helm chart. I wanted to also remove the Scheduler URI as was suggested in #3259, but the config item was used in the Job link. But the current implementation does not generate job specific job links.

joshfischer1108:
Ahh yeah I see you mentioned the issue right above where I mentioned the issue.  Consider that a palm to forehead moment 😄 .  I haven't tested the code, but It sounds like it addresses the issue after the proxy container is taken out.

joshfischer1108:
@cckellogg If you have some time could you review this?

joshfischer1108:
@nicknezis It looks like the build is failing because of checkstyle errors.

nicknezis:
I think the test failure might not be related to my changes. I am able to run the tests locally with success.

nwangtw:
> I think the test failure might not be related to my changes. I am able to run the tests locally with success.

//heron/schedulers/tests/java:KubernetesSchedulerTest                    FAILED in 1.2s
    FAILED  org.apache.heron.scheduler.kubernetes.KubernetesSchedulerTest.testGetJobLinks (0.0s)

Unit tests should be normally stable. Let me trigger another build and see.

nicknezis:
Looks like it failed again. I'll try to figure out why. Thanks!

nicknezis:
Is this PR ok to be accepted?

joshfischer1108:
@nicknezis LGTM.   I'll wait a bit to see if anyone else has feedback.

PR #3434, opened 2020-01-12
Fix command arguments.

thinker0:
Fix gc logging


nwangtw:
You need to rebase from master and also update the unit tests.

PR #3439, opened 2020-01-22
add downloader-core filegroup

nlu90:
Adding this new filegroup allows other people building downloaders jar without including unexpected dependencies, such as distributed-log etc.


integration_test/src/python/topology_test_runner/resources/test.json
```diff
@@ -47,13 +47,6 @@
       "restartArgs"  : "True",
       "expectedTopoResultRelativePath" : "fields_grouping/FieldsGroupingResults.json",
       "checkType" : "topology_structure"
-    },
```
nwangtw: This change might affect integration test?

nlu90: This integration test fails occasionally and cannot be reproduced locally. 
Temporary removing it will unblock other work.

PR #3441, opened 2020-01-29
Improve docker dev env script

nwangtw:
The existing script can be used to create a docker image/container of the heron dev environment. However, raw docker commands are needed to go back to the same container. A new script is added to make it easier to use the container dev environment, so that deverlopers can create a container with:

dev-env-create.sh NAME

and return to the container (or start a new shell) with:
dev-env-run.sh NAME

### DIFF
--- a/docker/scripts/dev-env-create.sh
+++ b/docker/scripts/dev-env-create.sh
@@ -16,39 +16,38 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# This is a script to start a docker container that has all
-# tools needed to compire Heron. Developer should be able to
-# compile Heron in the container without any setup works.
-# usage:
+# This is a script to create/start a docker container that has all
+# tools needed to build Heron. Developer should be able to
+# build Heron in the container without any other setup works.
+#
+# Usage:
 # To create a clean development environment with docker and run it,
 # execute the following scripts in the source directory of Heron:
-#   sh docker/scripts/dev-env.sh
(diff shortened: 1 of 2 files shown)


joshfischer1108:
It would be good to add this feature into the documentation somehow.  Otherwise new users will not know it exists.

nicknezis:
I really like this change. Thank you. I kept running `docker system prune` to clear out old containers.

I've been wondering if something like Vagrant would be a more full approach to do this. Either way this change makes the development cycle much more pleasant.

nwangtw:
> I really like this change. Thank you. I kept running `docker system prune` to clear out old containers.
> 
> I've been wondering if something like Vagrant would be a more full approach to do this. Either way this change makes the development cycle much more pleasant.

Thanks~ Vagrant is an interesting idea. It seems to be more powerful and can build a multiple nodes environment but it seems heavier than docker to me. Anyway, it could be useful to make development env setup easier.

nwangtw:
> It would be good to add this feature into the documentation somehow. Otherwise new users will not know it exists.

Yeah. I think now it is easy enough to use and will be included in the documentation.

joshfischer1108:
@nwangtw we can fix the CI build if we merge #3442 .  Then you can pull those changes in to this branch.

nwangtw:
> @nwangtw we can fix the CI build if we merge #3442 . Then you can pull those changes in to this branch.

cool.thx.

PR #3479, opened 2020-03-05
Python 3 upgrade effort

nicknezis:


Code0x58:
_Lead:_ it looks modules are mostly included in PEXs, but they are missing important `__init__.py` files so aren't being treated as modules.

Example:
```sh
# example generated PEX
pex=./bazel-bin/heron/tools/tracker/tests/python/topology_unittest_runner

for branch in master python-3-upgrade; do
    git checkout $branch
    bazel clean
    ./scripts/travis/ci.sh
    unzip -d /tmp/out "$pex"
    (cd /tmp/out && find *) > $branch.txt
    rm -fr /mt
done

diff -U0 master.txt python-3-upgrade
```

```diff
--- master.txt	2020-05-06 00:21:03.540305580 +0100
+++ branch.txt	2020-05-06 00:28:39.988906082 +0100
@@ -2 +1,0 @@
-heron/__init__.py
@@ -12 +10,0 @@
-heron/proto/__init__.py
@@ -31 +28,0 @@
-heron/tools/tracker/__init__.py
@@ -37 +33,0 @@
-heron/tools/tracker/tests/python/__init__.py
@@ -39 +34,0 @@
-heron/tools/tracker/tests/__init__.py
@@ -109 +103,0 @@
-heron/tools/tracker/src/python/__init__.py
@@ -120,2 +113,0 @@
-heron/tools/tracker/src/__init__.py
-heron/tools/__init__.py
@@ -123 +114,0 @@
-heron/tools/common/__init__.py
@@ -126 +116,0 @@
-heron/tools/common/src/python/__init__.py
@@ -131 +120,0 @@
-heron/tools/common/src/python/utils/__init__.py
@@ -146 +134,0 @@
-heron/tools/common/src/__init__.py
@@ -148 +135,0 @@
-heron/common/__init__.py
@@ -162 +148,0 @@
-heron/common/src/__init__.py
@@ -164 +149,0 @@
-heron/statemgrs/__init__.py
@@ -176 +160,0 @@
-heron/statemgrs/src/python/__init__.py
@@ -184 +167,0 @@
-heron/statemgrs/src/__init__.py
@@ -186 +168,0 @@
-heronpy/__init__.py
```

nicknezis:
@Code0x58 I believe it might be related to `Implicit Namespace Packages`: https://www.python.org/dev/peps/pep-0420/

Some more information with an example: https://stackoverflow.com/a/48804718

nicknezis:
Also, I found that the `bdist_wheel` command not working was fixed with something related to adding the `python3-wheel` package. Or `pip3 install wheel`. I can't remember exactly, but I think I went with the package approach and made progress.

Code0x58:
I think `bdist_wheel` not being found was due to a setup script of one of the pylint's transitive dependencies having a `setup.py` that has logic in it for building extensions, which broke up breaking encapsulation but went unnoticed as wheel was installed in the host python2 environment. On my branch I just added python3-wheel to the CI environment to skip over that issue.

Code0x58:
One "short term workaround"™ for getting the `pex_pytest` to work was to change it so the pex binaries+tests defaulted to be non-zip safe so they are extracted. There were still failures later, which I think is from pex_library consumption. I'll see if I can patch that too, then hopefully come up with a neater solution for all of the issues

nicknezis:
Closing this because the work has moved to #3522

PR #3483, opened 2020-03-10
updating docs for compiling

joshfischer1108:
This may be a bit strong of an opinion for the documentation.  But  Debian10 is the only current working docker solution we have.  I'd like to focus on it during incubation so we can start releasing convenience binaries with future releases.  Please review the wording and make sure it makes sense and that you agree with it.

cc @thinker0

### DIFF
--- a/website2/docs/compiling-docker.md
+++ b/website2/docs/compiling-docker.md
@@ -24,7 +24,7 @@ For developing Heron, you will need to compile it for the environment that you
 want to use it in. If you'd like to use Docker to create that build environment,
 Heron provides a convenient script to make that process easier.
 
-Currently, only Debian10 is actively being supported.  There is also limited support for Ubuntu 14.04, Ubuntu 18.04, Debian9, and CentOS 7. If you
+Currently Debian10 and Ubuntu 18.04 are actively being supported.  There is also limited support for Ubuntu 14.04, Debian9, and CentOS 7. If you
 need another platform there are instructions for adding new ones
 [below](#contributing-new-environments).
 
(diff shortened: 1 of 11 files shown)


nwangtw:
> This may be a bit strong of an opinion for the documentation. But Debian10 is the only current working docker solution we have. I'd like to focus on it during incubation so we can start releasing convenience binaries with future releases. Please review the wording and make sure it makes sense and that you agree with it.
> 
> cc @thinker0

I believe ubuntu work, but it is ok if decide that debian 10 is our main platform.

joshfischer1108:
@nwangtw I was under the impression ubuntu18.04 was not working.  I'm fine with either container.  I think we should pick one as the official released convenience binary.  It's too much to maintain multiple distros well.  I'll change the wording around before merging.

website2/docs/compiling-docker.md
```diff
@@ -130,8 +140,8 @@ After the commands, a new docker container is started with all the libraries and
 installed. The operation system is Ubuntu 18.04 by default. Now you can build Heron
```
nicknezis: Should this line mention Debian instead of Ubuntu?

joshfischer1108: Thanks for catching.  I reverted the script back ubuntu 18.04

joshfischer1108:
I've updated the documentation a bit.  Please give another review before we merge.

joshfischer1108:
If there are no concerns with the changes I'll merge the changes after 24 hours passes.

PR #3505, opened 2020-04-03
Change url to github.com of libunwind

thinker0:
Change url to github.com of libunwind

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -874,10 +874,10 @@ http_archive(
 
 http_archive(
     name = "org_nongnu_libunwind",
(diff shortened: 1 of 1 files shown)


nicknezis:
What was the reason for the update? New feature or just keeping things up to date? It seems like Travis CI is failing. I think this is not due to this change, but probably related to my update of Travis CI to Ubuntu 18 (from Ubuntu 14).

thinker0:
Ah, I don't think I need it. I thought the binary on github was the same thing. The content is different. I'll close.

nicknezis:
@thinker0 I did a local build and it seemed to work ok. Might be good to update it if you also think it works. I was merely asking out of curiosity.

PR #3518, opened 2020-04-23
Fix debian9 docker file and clean up centos8 and ubuntu18.04

nwangtw:
Finally found out the cause of the failure: pkg-config is missing.

After this PR I am going to test build again.

@thinker0


thinker0:
LGTM

PR #3520, opened 2020-04-30
Pin http_file and http_archive resources in WORKSPACE

Code0x58:
This was a tidy up after noticing warnings when running bazel


nwangtw:
Thanks~

PR #3521, opened 2020-04-30
Make Java optional in CLI and check JAVA_BIN

Code0x58:
This is so you only need the variables when they are used when submitting a topology, so you can still use other commands and check the `--help`.

It also uses `JAVA_BIN` before `JAVA_HOME`.

### DIFF
--- a/heron/tools/admin/src/python/main.py
+++ b/heron/tools/admin/src/python/main.py
@@ -121,8 +121,6 @@ def check_environment():
   Check whether the environment variables are set
   :return:
   '''
-  if not config.check_java_home_set():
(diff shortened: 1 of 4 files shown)


nwangtw:
lgtm. letting java experts to take a look.

joshfischer1108:
I'm no expert, but it looks good to me

Code0x58:
I made an amendment to use `Status.InvocationError` instead of `Status.HeronError`

joshfischer1108:
@Code0x58 It looks like there are some Python linting issues that are breaking the build.  Would mind fixing those before we merge?

PR #3525, opened 2020-05-17
Fix ubuntu16.04 docker file

nwangtw:


nwangtw:
build passed~

https://builds.apache.org/job/apache-heron-docker-image/44/

nicknezis:
Ahh yes. This is a good find. Much simpler solution than I originally thought would be needed. If I remember correctly, the patch that I apply to Zookeeper build files is so that it works with this. That allowed for Zookeeper source to build on old and newer OS. But this package was needed. I think I only added it in the newer OS flavors that I tested.

PR #3526, opened 2020-05-17
Fix ubunt14.04 docker file

nwangtw:


nwangtw:
Sounds good to me. We can move it out to a separated folder just for reference.

PR #3529, opened 2020-05-24
Build python package in virtualenv

Code0x58:


Code0x58:
I should mention I did this to reduce the requirements outside of the bazel environment, otherwise you need at least python-setuptools on the host

PR #3530, opened 2020-05-26
Update osx documentation (3453)

alpert:
As specified here: https://github.com/apache/incubator-heron/issues/3452#issuecomment-625435818, heron now requires bazel version 3.0.0


joshfischer1108:
@alpert Thank you for the contribution and helping keeping our docs up to date!

PR #3534, opened 2020-05-31
Fix helm chart

Code0x58:
the template string in the values.yaml wasn't rendered, so this patch renders the default value in the template

### DIFF
--- a/deploy/kubernetes/helm/templates/tools.yaml
+++ b/deploy/kubernetes/helm/templates/tools.yaml
@@ -17,6 +17,7 @@
 
 {{- $platform := .Values.platform -}}
 {{- $jobReplicas := .Values.jobReplicas -}}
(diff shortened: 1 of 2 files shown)


joshfischer1108:
This looks good to me.. I'll let it sit for a day to make sure no one else has any concerns.

Thank you @Code0x58

joshfischer1108:
@nicknezis could you review this please?

nicknezis:
Cool, I came up with a different solution using the `tpl` function as shown here. https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function

But I think I prefer this solution better. If it worked for your local testing, I'm good with this.

Code0x58:
:facepalm: I just noticed this should have contained `{{ .Values.heron.url | default $defaultUrl }}` otherwise it will disregard an explicit value

joshfischer1108:
> 🤦 I just noticed this should have contained `{{ .Values.heron.url | default $defaultUrl }}` otherwise it will disregard an explicit value

Not a big deal. Do you want to open another PR to address it?

PR #3536, opened 2020-06-05
fixing javadocs

joshfischer1108:
This PR fixes the Java doc build for the website.


examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java
```diff
@@ -1,131 +0,0 @@
...
- * Unless required by applicable law or agreed to in writing,
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
-
-
-package org.apache.heron.examples.streamlet;
```
joshfischer1108, line 21: This file caused a break in the build as the pulsar library wasn't included in the project anymore.  We can add it back at a later time when/if we add a packaged pulsar spout to the contrib code base

nicknezis: Want me to take a look at fixing this before the merge?

joshfischer1108: yeah that'd be great.  It's probably the better than just throwing it away.

joshfischer1108: Hey @nicknezis I'm still getting those Javadoc errors when building locally after your push

```
/Users/joshfischer/Source/apache/incubator-heron/examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java:34: error: package org.apache.pulsar.client.api does not exist
import org.apache.pulsar.client.api.Consumer;
                                   ^
/Users/joshfischer/Source/apache/incubator-heron/examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java:35: error: package org.apache.pulsar.client.api does not exist
import org.apache.pulsar.client.api.PulsarClient;
                                   ^
/Users/joshfischer/Source/apache/incubator-heron/examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java:36: error: package org.apache.pulsar.client.api does not exist
import org.apache.pulsar.client.api.PulsarClientException;
                                   ^
/Users/joshfischer/Source/apache/incubator-heron/examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java:52: error: cannot find symbol
    private PulsarClient client;
            ^
  symbol:   class PulsarClient
  location: class PulsarSource
/Users/joshfischer/Source/apache/incubator-heron/examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java:53: error: cannot find symbol
    private Consumer consumer;
            ^
  symbol:   class Consumer
  location: class PulsarSource
5 errors
```

joshfischer1108: To reproduce execute the below from the root directory of the project:
`cd website2/website && make javadocs`

maven_install.json
```diff
@@ -1,6 +1,6 @@
 {
     "dependency_tree": {
-        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": 607867898,
+        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": -512671146,
         "conflict_resolution": {},
```
joshfischer1108, line 4: A question outside the scope of this PR.  It looks like this file is auto generated from the maven rule.  Should this be checked into SCM?

nicknezis: I think so. So you can use the external_jvm_rules feature without this file, but anytime someone builds it will resolve the transitive dependencies and can lead to issues. Also performance and ability to do offline builds are impacted. Some good info on the feature here: https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader

One way to think of it is this. Before the WORKSPACE file had a longer list of Jar artifacts with SHA256, the tool now produces this file with the various files.

One key thing that I worry might trip someone up is that changes to the list of Maven dependencies in the WORKSPACE file aren't automatically honored. You have to first run `bazel run @unpinned_maven//:pin` to update the file. I put a comment in the WORKSPACE file as a reminder in case anyone tried to update versions of dependencies.

examples/src/java/org/apache/heron/examples/streamlet/SimplePulsarSourceTopology.java
nicknezis: Weird, I originally got the error, but then after the edit it went away. I do get a long list of warnings about invalid use of tags. I'll try a fresh `git clone` and see if it fails again.

nicknezis: Interesting. I was able to recreate the issue. Apparently at some point I had run `bazel build --config=darwin examples/src/java/...` and that made the problem go away. After running `bazel clean --expunge`, the issue came back. I'll keep looking into it.

nwangtw:
Interesting. Thanks for looking!

maven_install.json
joshfischer1108: Ok.  Thanks for the info.  I'll make sure to read up on this.

nwangtw:
Cool. Let me try to build again!

PR #3538, opened 2020-06-15
fixing website layout

joshfischer1108:
This PR fixes the site layout.  Credit for the fix goes to @windhamwong .


joshfischer1108:
Unfortunately.. This PR only fixes the issues when building locally. It seems some classes get overridden somehow and the error returns.  Right now the home page looks fine, but the docs page has the same issues.  *Do not merge this*

Issue #3542, opened 2020-06-18
Kubernetes additional work required for deploying topology


windhamwong:
Kubernetes version: 1.18.2+k3s1
I am trying out to deploy `heron-api-examples.jar` as suggested in the doc. After some troubleshooting (not just some but a lot) and found that the created pods (`acking-0`, `acking-1`, `acking-2`) require to connect together. The `acking-0` is responsible for launching servers for stream manager (6001/TCP), tmaster controller (6002/TCP) and tmaster state (6003/TCP). As k8s requires `Service` to expose ports to allow external or other containers to connect, we need additional code in Heron k8s source to generate these services as well. The following k8s yaml code has been tested and working for me.

I will keep on testing out as the topology is still not working as the demo expectation. I will come back to add more after.

```
apiVersion: v1
kind: Service
metadata:
  name: acking-0
spec:
  selector:
    statefulset.kubernetes.io/pod-name: acking-0
  ports:
    - name: stream-manager
      protocol: TCP
      port: 6001
      targetPort: 6001
    - name: tmaster-controller
      protocol: TCP
      port: 6002
      targetPort: 6002
    - name: tmaster-state
      protocol: TCP
      port: 6003
      targetPort: 6003
    - name: heron-shell
      protocol: TCP
      port: 6004
      targetPort: 6004
```

windhamwong:
Current situation:
Heron UI shows only the topology layout, but no details of config, count and instance log/info. Bolt seems to be working and showing logs of, 
```
[2020-06-18 17:27:28 +0000] [STDOUT] stdout: Bolt processed 64980000 tuples in 525055 ms
[2020-06-18 17:27:28 +0000] [STDOUT] stdout: Bolt processed 64990000 tuples in 525067 ms
```

I guess the Heron doc of k8s cluster setup, and the source for k8s require a lot of work to support the latest k8s version.

nicknezis:
Is this an issue with newer Kubernetes? How best can I reproduce the issue you are seeing. 

Kubernetes services are not needed for exposing ports, but are used for Service Discovery and for providing a stable IP as opposed to Pod IP which will change on pod restart. I believe the way the current Topologies are run as StatefulSets in Kubernetes, each pod has a specific name which can be referenced by the other pods in the StatefulSet. It might still be a good idea to create the Service, but I don't think this should be causing the Topology pods to fail to connect.

windhamwong:
[...]

PR #3550, opened 2020-06-29
Updates to Kubernetes scheduler to fix edge cases

nicknezis:
Fix for #3542 

Some notable changes:
1. Kubernetes scheduler now creates a Headless Service to pair with the StatefulSet.
2. Upgraded Kubernetes client from 7.0.0 to 8.0.0
3. Removed old compatibility code that deleted all pods. I don't see the code that used to launch individual pods, so I assume it is no longer needed. For now we only launch StatefulSets, although I do have thoughts on how to better manage the topology pods.
4. Heron Executor now checks for POD_NAME environment variable. If found, it assumed Kubernetes environment in which we use the fully qualified domain name which should be composed of pod name, headless service, namespace and cluster domain.


third_party/java/BUILD
```diff
@@ -227,10 +227,10 @@ java_library(
     name = "apache-pulsar-client",
     srcs = ["Empty.java"],
     exports = [
-        "@maven//:org_apache_pulsar_pulsar_client_shaded",
+        "@maven//:org_apache_pulsar_pulsar_client",
     ],
     deps = [
-        "@maven//:org_apache_pulsar_pulsar_client_shaded",
+        "@maven//:org_apache_pulsar_pulsar_client",
     ],
```
joshfischer1108, line 234: Why change the pulsar client in this PR?

nicknezis: We might want to put it back to the shaded jar. I think I made this change because of a compilation error. I just checked, and I think we changed the dependency in the WORKSPACE file. Here is the PR: https://github.com/apache/incubator-heron/pull/3536/files

nicknezis: Ok I put it back to the shaded version of the jar. Should look more like what it was before the PR we merged 23 days ago.

windhamwong:
Testing environment:
namespace: heron
No additional service added mannually.

Tested with the example topology `ackingTopology`. I am still seeing the `acking-1` complaining about:

metricsmgr-1.log.0
```
[2020-07-01 05:05:02 +0000] [INFO] org.apache.heron.common.network.HeronClient: Connecting to endpoint: acking-0:6001
[2020-07-01 05:05:02 +0000] [SEVERE] org.apache.heron.metricsmgr.sink.tmaster.TMasterSink: TMasterClient dies in thread: Thread[Thread-1117,5,main]
java.nio.channels.UnresolvedAddressException
	at java.base/sun.nio.ch.Net.checkAddress(Net.java:130)
	at java.base/sun.nio.ch.SocketChannelImpl.connect(SocketChannelImpl.java:675)
	at org.apache.heron.common.network.HeronClient.start(HeronClient.java:129)
	at org.apache.heron.metricsmgr.sink.tmaster.TMasterClient.run(TMasterClient.java:128)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

and

container_1_exclaim1_1.log.0
```
[INFO] org.apache.heron.network.StreamManagerClient: Stop writing due to not yet connected to Stream Manager.
```


It is still trying to resolve `acking-1` instead of `acking-1.acking.heron.svc.cluster.local` OR `acking1.acking`

nicknezis:
@windhamwong What environment are you running in? I'd like to recreate.

A few questions.
1. Do you see the Headless service created at submit time?
2. Does the `acking-0` Executor log not show the FQDN?
3. Does the `acking-1` Executor log also not show the FQDN?

nicknezis:
DrKNa in the Slack was able to test this and verified that it worked in his environment which previously did not work. But I'd like a review if the logic I added is the best option. Any suggestions before merging in?

PR #3552, opened 2020-06-30
Distribution Dockerfile missing ncat and other cleanup

### LINKED ISSUE #3548: Netcat missing from some of the distribution Dockerfiles
**Describe the bug**
Heron images don't work in Kubernetes because of missing `nc` command

**To Reproduce**

1. ./docker/scripts/build-artifacts.sh ubuntu20.04 0.20-20200625-ubuntu ~/workspace/heron-artifacts
2. ./docker/scripts/build-docker.sh ubuntu20.04 0.20-20200625-ubuntu ~/workspace/heron-artifacts --squash
3. Inspect resulting Docker image (with `dive` or some other method)

**Expected behavior**
Netcat package should be installed for Zookeeper shell script dependency

**Operating System**
 - OS: Ubuntu
 - Version: 20.04

nicknezis:
1. Fixes #3548 
2. Now using JRE instead of JDK for distribution images
3. Package install cleanup to reduce resulting docker image size. Especially when not squashed.
4. Removed `python-dev` packages
5. Removed `python3` packages

### DIFF
--- a/docker/dist/Dockerfile.dist.centos7
+++ b/docker/dist/Dockerfile.dist.centos7
@@ -17,19 +17,17 @@
 
 FROM centos:centos7
 
-RUN yum -y upgrade
-RUN yum -y install python; yum clean all
-RUN yum -y install unzip; yum clean all
-RUN yum -y install which; yum clean all
-RUN yum -y install curl; yum clean all
-RUN yum -y install nmap-ncat; yum clean all
-RUN yum -y install python python3-setuptools; yum clean all
-
-RUN yum install epel-release; yum clean all
-RUN yum update -y
-RUN yum install -y supervisor; yum clean all
-
-RUN yum -y install java-11-openjdk java-11-openjdk-devel; yum clean all
+RUN yum install epel-release \
+    && yum -y update \
+    && yum -y install \
+    curl \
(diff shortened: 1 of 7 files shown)


docker/dist/Dockerfile.dist.centos7
```diff
@@ -17,21 +17,17 @@
...
-RUN yum install -y supervisor; yum clean all
-
-RUN yum -y install java-11-openjdk java-11-openjdk-devel; yum clean all
-
-ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk
+RUN yum -y install epel-release \
+    && yum -y update \
+    && yum -y install \
+    curl \
+    java-11-openjdk-headless \
```
nwangtw, line 24: is there a jre version for centos7 as well?

nicknezis: Good question. I tried looking for one and couldn't find it. There is a `java-11-openjdk-devel` which I think is JDK, but when I install the package in the pull request, it still includes the `-devel` package. So I'm not sure.

nwangtw: KK. No worries. This is a huge improvement. Thanks~

Issue #3559, opened 2020-07-07
Use builtin python3 modules instead of externals


Code0x58:
These are leftovers from the python2 to python3 upgrade:
 * `mock` → `unittest.mock`
 * `unittest2` → `unittest`
 * `future`, `argparse`, `enum34`, `six` → :skull: 
 * `py` → ...

PR #3569, opened 2020-07-17
Strip old python requirements

Code0x58:
This is for #3559


scripts/resources/idea/libraries/hp_dep.xml
```diff
@@ -21,14 +21,12 @@ under the License.
 <component name="libraryTable">
```
Code0x58, line 21: I started making changes to this file, but realised that this is probably long out of date or dead, so avoided additional changes

PR #3576, opened 2020-07-19
Upgrade to Bazel 3.4.1

nicknezis:


WORKSPACE
```diff
@@ -363,25 +363,27 @@ http_archive(
 # end helm
 
 # for docker image building
-DOCKER_RULES_VERSION = "0.14.1"
```
nwangtw, line 366: could be cleaner to keep this variable? not a blocker.

PR #3577, opened 2020-07-20
try to fix setup-intellj.sh

huijunwu:
The GENERATED_PATHS includes only java_doc \*.java files (find bazel-bin/ -name "\*.java" ), which look the same with 4 artifact original \*.java files.
Not sure why we include them twice..  Bypass them..

### DIFF
--- a/scripts/get_all_heron_paths.sh
+++ b/scripts/get_all_heron_paths.sh
@@ -17,15 +17,12 @@
 
 set -eu
 
-function query() {
-  ./output/bazel query "$@"
-}
 
 set +e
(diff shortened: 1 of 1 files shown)


joshfischer1108:
I had a fresh build cache so I was able to rebuild and open intellij quickly. The change looks good to me.

To test I ran the following from the root folder of the project.

```
$ rm  heron.iml 
$ rm -rf .idea/
$ ./scripts/setup-intellij.sh 
```
Successful output is below:

Screen Shot 2020-07-19 at 9 31 22 PM

PR #3583, opened 2020-07-24
Upgrade virtualenv

Code0x58:
replace the 4 year old version with the latest version (released today)


nicknezis:
@Code0x58 Do you think we can close this if we might use the `venv` Pull Request?

PR #3584, opened 2020-07-25
Deep clean heron-admin - probably better to remove entirely

Code0x58:
I started cleaning up `heron-admin`'s code to make it a bit clearer, but along the way found that it was broken (at least because it tried to find nomad in the wrong place.

Given all it does is try to manage one of many schedulers, when you can already use the local "cluster" or deploy your own cluster (including minikube or kind which seem to have support within the community), I'm thinking it would be better to strip out the `heron-admin` binary all together to simplify the ecosystem and reduce debt/maintenance cost as well.

That said, the changes here are:
 * use [click](http://click.palletsprojects.com/) instead of argparse to make things more readable
   - parameters are explicitly passed instead of aways passing a mysterious `cl_args` dict everywhere
   - the interface is still the same as before, although docstrings may be different but also easier to change. I would be tempted to suggest reworking the interface to something more succinct if I didn't think it would be better to drop the whole thing
 * use `$VISUAL`, `$EDITOR`, or `vi` when editing the `inventory.yaml` in `heron-admin standalone set`
 * the code now at least passes a manual smoke test


Code0x58:
Closing as heron-admin removed in https://github.com/apache/incubator-heron/pull/3585

PR #3585, opened 2020-07-25
Purge heron-admin

Code0x58:
* remove heron-admin binary from release
 * remove nomad binary from release
 * remove "standalone" cluster config from release and documentation

I'm suggesting this as `heron-admin`:
 * lacks tests
 * is currently broken
 * has alternatives like self-managed minikube, kind, or heavier scheduler arrangement - simpler to give attention to other approaches instead
 * now a nice-to-have at best, but has cost/debt


nwangtw:
- remove heron-admin binary from release
I am all for it. Hasn't had a chance to use it in the past 3 years. @nlu90 @huijunw Is there any case heron-admin is needed? Even if there is a unique function in the tool, given how rare it is used, we may still want to consider maybe there is a better place for the unique function.

remove nomad binary from release
It could be useful for some people. I would suggest moving it into a new "extra" folder just in case some people use it.

remove "standalone" cluster config from release and documentation
Overall I think it is ok. I am wondering about the use cases. Does it mean running on a Baremetal machine?

huijunwu:
> * remove heron-admin binary from release
>   I am all for it. Hasn't had a chance to use it in the past 3 years. @nlu90 @huijunw Is there any case heron-admin is needed? Even if there is a unique function in the tool, given how rare it is used, we may still want to consider maybe there is a better place for the unique function.
> 
> remove nomad binary from release
> It could be useful for some people. I would suggest moving it into a new "extra" folder just in case some people use it.
> 
> remove "standalone" cluster config from release and documentation
> Overall I think it is ok. I am wondering about the use cases. Does it mean running on a Baremetal machine?

Nomad seems for running a set of machines ranther than single machine.
call @jerrypeng for comments

joshfischer1108:
I don't see a need for running Heron on Nomad.  I'll let this sit for a few more days, if no one objects we can merge it.

armadillojim:
@Code0x58 Thanks for the merge!  Say, I was working on another PR, and I noticed that some references to "standalone" haven't been removed from the docs.  The three files `schedulers-nomad.md`, `uploaders-http.md`, and `guides-troubeshooting-guide.md` in `website2/website/versioned_docs/version-0.20.0-incubating/` didn't get updated here in this PR.  Similarly, `website2/website/versioned_sidebars/version-0.20.0-incubating-sidebars.json` was missed.

Should I create a new PR for just these edits?  TIA.

Code0x58:
I think files in the `website2/website/versioned_*` directories shouldn't be modified in this case as they're for previous versions which legitimately contained the standalone scheduler (even if possibly broken).

Thanks for keeping an eye out @armadillojim!

PR #3580, opened 2020-07-23
Fix log-reader for Python3

thinker0:
Python3

@Code0x58 
```
[I 200722 23:06:34 web:1811] 200 GET /filedata/./log-files/container_36_system-log-to-es_36.log.0?offset=-1&length=-1 (10.25.178.125) 0.64ms
[E 200722 23:06:34 web:1407] Uncaught exception GET /filedata/./log-files/container_36_system-log-to-es_36.log.0?offset=2178106&length=50000 (10.25.178.125)
    HTTPServerRequest(protocol='http', host='LNMESOSS1513:31449', method='GET', uri='/filedata/./log-files/container_36_system-log-to-es_36.log.0?offset=2178106&length=50000', version='HTTP/1.1', remote_ip='10.25.178.125', headers={'Connection': 'close', 'Host': 'LNMESOSS1513:31449', 'Accept-Encoding': 'gzip'})
    Traceback (most recent call last):
      File "/var/lib/mesos/slaves/0114e468-0f07-485c-8673-14ff74e8b46c-S593/frameworks/c663397e-a472-43bd-92dd-d97027fcf6ce-0000/executors/thermos-www-release-heron-system-access-es-others-36-60361105-ae4e-4efc-84df-b70d3c717113/runs/f8aa692a-249d-40f8-a8db-e7e4a8cbdc8f/sandbox/.pex/installed_wheels/a92bf48f8e3ffa83220bafe58403d68e2d170148/tornado-4.0.2-cp36-cp36m-linux_x86_64.whl/tornado/web.py", line 1288, in _stack_context_handle_exception
        raise_exc_info((type, value, traceback))
      File "<string>", line 3, in raise_exc_info
      File "/var/lib/mesos/slaves/0114e468-0f07-485c-8673-14ff74e8b46c-S593/frameworks/c663397e-a472-43bd-92dd-d97027fcf6ce-0000/executors/thermos-www-release-heron-system-access-es-others-36-60361105-ae4e-4efc-84df-b70d3c717113/runs/f8aa692a-249d-40f8-a8db-e7e4a8cbdc8f/sandbox/.pex/installed_wheels/a92bf48f8e3ffa83220bafe58403d68e2d170148/tornado-4.0.2-cp36-cp36m-linux_x86_64.whl/tornado/web.py", line 1475, in wrapper
        result = method(self, *args, **kwargs)
[...]

PR #3586, opened 2020-07-26
fix shell download

huijunwu:
fix https://heronstreaming.slack.com/archives/C0VKQLKV0/p1595695491236400

The shell exception at str.decode https://github.com/apache/incubator-heron/blob/master/heron/shell/src/python/utils.py#L152


huijunwu:
close it,  duplicate #3580

PR #3587, opened 2020-07-26
Correct heron-executor CLI usage

Code0x58:
The usage was inaccurate/out of date for `heron-executor`, so this change leaves the parsing and documentation to [click](https://click.palletsprojects.com/) in a change that tried to be minimally invasive. The click options were made by automated substitution of the argparse arguments, then slightly organised (mostly alphabetical).


huijunwu:
instance may need similar refactoring https://github.com/apache/incubator-heron/blob/master/heron/instance/src/python/instance/st_heron_instance.py#L328

PR #3589, opened 2020-07-28
Reorder the init containers in the Helm chart

nicknezis:


nicknezis:
@joshfischer1108 The order that init containers run is sequentially. The check for Zookeeper to be up and healthy was happening after the Bookkeeper format call. Bookkeeper needs Zookeeper to be healthy and therefore startup issues would occur when installing using Helm (everything at once).

joshfischer1108:
> @joshfischer1108 The order that init containers run is sequentially. The check for Zookeeper to be up and healthy was happening after the Bookkeeper format call. Bookkeeper needs Zookeeper to be healthy and therefore startup issues would occur when installing using Helm (everything at once).

Gotcha.  Thanks for explaining.

PR #3594, opened 2020-07-29
Switch to pigz from gzip to enable multi-thread processing

windhamwong:
This helps to make the docker build faster with using multi-core/thread.
Ubuntu18 has pigz installed by default, but still need test on other distro.


windhamwong:
Messed up with the code, i will create another PR

PR #3595, opened 2020-07-29
Switch to pigz to support multi-core tar command

windhamwong:
pigz is a multi-core version of tar, which makes the docker build process much faster.

### DIFF
--- a/docker/scripts/build-artifacts.sh
+++ b/docker/scripts/build-artifacts.sh
@@ -95,7 +95,7 @@ trap cleanup EXIT
 
 generate_source() {
   echo "Generating source tarball"
-  tar --exclude-from=$DOCKER_DIR/.tarignore -C $PROJECT_DIR -czf $SRC_TAR .
+  tar --use-compress-program=pigz --exclude-from=$DOCKER_DIR/.tarignore -C $PROJECT_DIR -cf $SRC_TAR .
 }
 
 verify_source_exists() {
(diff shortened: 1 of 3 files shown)


nicknezis:
I'm not familiar with `pigz`. Is the resulting artifact equivalent to a gzip artifact? Does a user need `pigz` installed to `tar -xzf` the resulting files? Does that command even still work?

Code0x58:
A couple of tests to show equivalence for consumers:
With gzip:
```
$ cat txt 
oink
$ ls
txt
$ pigz txt
$ ls
txt.gz
$ gunzip txt.gz 
$ ls
txt
$ cat txt 
oink
```
With tar:
```
$ tar --remove-files --use-compress-program=pigz -cf out.tar.gz txt
$ ls
out.tar.gz
$ tar -zxf out.tar.gz
$ cat txt
oink
```

windhamwong:
https://zlib.net/pigz/
It is gzip format but with the parallel multi-core feature. The traditional gzip command uses only one core but modern CPU/vCPU has more than one.

nicknezis:
1. Have all containers been tested? Do some of them need `pigz` included in their Dockerfiles?
2. Should `bazel_configure.py` be updated to check for `pigz` if it is an external dependency?
3. Do any of the website pages need to be updated? For instance `brew install pigz` for someone compiling on their Mac.

windhamwong:
let me try out all docker builds tomorrow.

windhamwong:
Docker build testing result:
ubuntu18.04 - success
ubuntu20.04 - success
ubuntu16.04 - success
debian9 - fail - python3.6 not supported (unrelated to pigz)
debian10 - fail - protobuf issue (unrelated to pigz)
centos7 - success

PR #3596, opened 2020-07-30
Updating docker build for fixing python library click issue

windhamwong:
https://github.com/apache/incubator-heron/commit/39cf3a76b143b5c491de88cfe40cd83f29d2c0be#commitcomment-41026206

This PR adds 2 lines of ENV for applying the fix of click under Py36


nicknezis:
@huijunw @windhamwong  Is the underlying issue something that needs to be documented for those not using Docker images?

PR #3601, opened 2020-08-06
Dockerfiles Python 3 cleanup

nicknezis:


docker/compile/Dockerfile.ubuntu20.04
```diff
@@ -31,12 +31,13 @@ RUN apt-get update && apt-get -y install \
       libunwind8 \
       libcppunit-dev \
       patch \
+      python3 \
```
Code0x58, line 34: trivia: this is implicit from `python3-dev` [[dependencies](https://packages.ubuntu.com/focal/python3-dev)] and `python-is-python3`

docker/compile/Dockerfile.ubuntu20.04
```diff
@@ -31,12 +31,13 @@ RUN apt-get update && apt-get -y install \
...
       python3-dev \
-      python \
+      python3-pip \
```
Code0x58, line 36: is this needed for something? pip2/pip3 wasn't included before

docker/compile/Dockerfile.ubuntu20.04
nicknezis: Yeah I actually had this thought with python3-dev and python3, but had an issue when I tried it without. I was planning to revisit this though with some trial and error.

docker/compile/Dockerfile.ubuntu20.04
nicknezis: It might not be needed, but I ran into an issue when trying to `pip3` without it. This initial commit was just a placeholder really. I'm planning to go through each image type. Changes from `virtualenv` to `venv` will probably be a bigger change in another commit. Unless you want to just collaborate on that kind of change here. We can do something similar to your removal of `virtualenv` in #3583 .

docker/compile/Dockerfile.ubuntu20.04
```diff
@@ -31,12 +31,13 @@ RUN apt-get update && apt-get -y install \
...
+      python-is-python3 \
       pkg-config \
       wget \
       zip \
-      virtualenv \
```
nicknezis, line 39: @Code0x58 Do you think this package is needed? Seemed to work without it.

nwangtw:
Thanks. Python 3 is kinda annoying. :(

bazel_configure.py
```diff
@@ -413,7 +413,7 @@ def main():
   env_map['AUTOMAKE'] = discover_tool('automake', 'Automake', 'AUTOMAKE', '1.9.6')
   env_map['AUTOCONF'] = discover_tool('autoconf', 'Autoconf', 'AUTOCONF', '2.6.3')
   env_map['MAKE'] = discover_tool('make', 'Make', 'MAKE', '3.81')
-  env_map['PYTHON3'] = discover_tool('python3', 'Python3', 'PYTHON3', '3.4')
+  env_map['PYTHON3'] = discover_tool('python3', 'Python3', 'PYTHON3', '3.5')
```
Code0x58, line 416:
```suggestion
  env_map['PYTHON3'] = discover_tool('python3', 'Python3', 'PYTHON3', '3.6')
```

docker/compile/deprecated/Dockerfile.ubuntu14.04
```diff
@@ -33,7 +33,7 @@ RUN apt-get update && apt-get -y install \
       libssl-dev \
       git \
       libtool \
-      python3-dev \
```
Code0x58, line 36: python3.6 is the minimum supported version.

As the base OS is over a year past EOL, and there's no python3.6 packages, I think this should be removed, rather than leaving it in as moving to deprecated (unless there is a trivial python3.6 install - but I still don't think worth it)

docker/compile/deprecated/Dockerfile.ubuntu16.04
```diff
@@ -49,6 +49,8 @@ RUN apt-get update && apt-get -y install \
 
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
 
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
```
Code0x58, line 52: this image also has python3.5, so may as well delete this too. I think if anyone needed an old base OS image, they'd probably be maintaining their own Dockerfile for that or similar purposes anyway

docker/test/Dockerfile.ubuntu18.04
```diff
@@ -29,15 +29,17 @@ RUN apt-get update && apt-get -y install \
...
       wget \
       zip \
-      virtualenv \
       unzip \
       git \
       curl \
       tree \
       openjdk-11-jdk-headless
 
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
```
Code0x58, line 41: is linking `python` to `python3` needed - was it that transient dependency which needed it?

scripts/packages/BUILD
```diff
@@ -604,8 +604,9 @@ genrule(
         "export RELEASE_FILE_DIR=$$(pwd)",
         "export TMP_DIR=$$(mktemp -d -t heronpy.XXXXX)",
         "echo $$TMP_DIR",
-        "$(location @virtualenv//:virtualenv) --no-download --quiet --clear $$TMP_DIR/venv",
-        "PS1= source $$TMP_DIR/venv/bin/activate",
+        # Create the virtual environment
+        'python3 -m venv $$TMP_DIR/venv --clear',
```
Code0x58, line 608: :+1:

tools/rules/pex/BUILD
```diff
@@ -70,6 +69,7 @@ genrule(
     executable = True,
     message = "Bootstrapping pex",
     output_to_bindir = True,
-    tools = ["@virtualenv"],
+    # tools = ["@virtualenv"],
```
Code0x58, line 72: may as well delete these two lines now

bazel_configure.py
```diff
@@ -413,7 +413,7 @@ def main():
...
```
Code0x58, line 417: this should probably check that `python -m venv` returns non-zero (the version will be same as the python interpreter)

tools/rules/pex/BUILD
nicknezis: Thanks for catching this.

docker/test/Dockerfile.ubuntu18.04
nicknezis: I got this error:
```
ERROR: An error occurred during the fetch of repository 'pip_deps':
   pip_import failed:  (src/main/tools/process-wrapper-legacy.cc:75: "execvp(python, ...)": No such file or directory
)
ERROR: no such package '@pip_deps//': pip_import failed:  (src/main/tools/process-wrapper-legacy.cc:75: "execvp(python, ...)": No such file or directory
)
```

docker/compile/deprecated/Dockerfile.ubuntu16.04
nicknezis: I moved both 14.04 (Well past the End of Life) and 16.04 (about to reach End of Life) to the deprecated folder. I think they should be removed if they don't support Python 3.6. Both images won't build successfully.

bazel_configure.py
nicknezis: Should this be bumped up to 3.6?

docker/compile/deprecated/Dockerfile.ubuntu14.04
nicknezis: Good point. I've removed the image.

bazel_configure.py
nicknezis: Do you have a suggestion of how best to do this? Seems the `discover_tool` method wouldn't work for what you describe, right?

docker/test/Dockerfile.ubuntu18.04
Code0x58: That makes it sound like bazel (that `.cc` is a part of it) is trying to run `python` on the system, rather than `python3` for some reason, which would be a preexisting issue. Maybe it's because `test --host_force_python=PY3` isn't in `tools/bazel.rc`?

nicknezis: Just tested with the following, but it still fails:
```
test --genrule_strategy=standalone
test --host_force_python=PY3
test --ignore_unsupported_sandboxing
test --spawn_strategy=standalone
```

bazel_configure.py
nicknezis: I added the test function. Let me know if it looks good.

docker/test/Dockerfile.ubuntu18.04
nicknezis: Seems to be related to this line. Maybe we don't need it? Testing it now with it removed. https://github.com/apache/incubator-heron/blob/ee73293b2ce5f1c0aa0506a4398bb2efe696944f/WORKSPACE#L398

nicknezis: Looks like it's needed for some of the `tar` and `docker` rules.

PR #3605, opened 2020-08-13
adding bazel version check

joshfischer1108:
We often have issues with users trying to build Heron with an incompatible version of Bazel.  This file will hopefully help with instructing users how to get the correct version of Bazel to build Heron with.  For example: If you run Bazel in the root folder with a version other than 3.4.1 it will throw an error like below:

```
bazel --version
ERROR: The project you're trying to build requires Bazel 3.4.1 (specified in /Users/{user}/Source/apache/incubator-heron/.bazelversion), but it wasn't found in /Users/{user}/.bazel/bin.

Bazel binaries for all official releases can be downloaded from here:
  https://github.com/bazelbuild/bazel/releases

You can download the required version directly using this command:
  (cd "/Users/joshfischer/.bazel/bin" && curl -LO https://releases.bazel.build/3.4.1/release/bazel-3.4.1-darwin-x86_64 && chmod +x bazel-3.4.1-darwin-x86_64)
```


nicknezis:
Cool feature. I like the idea. My only fear is that it will be a hassle when someone uses a newer Bazel. For example, if it is auto updated using brew on OSX. Is it possible to provide a range (i.e. 3.4.x)?

Code0x58:
I think [bazelisk](https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run) is the preferred way of working with `.bazelversion` and [multiple versions](https://docs.bazel.build/versions/master/updating-bazel.html#managing-bazel-versions-with-bazelisk)

joshfischer1108:
The goal is to lock users into a specific version for this code.    Then using some like Bazelisk like Oliver mentioned people can adjust to their needed version.

PR #3618, opened 2020-09-14
Stop using "cluster admin" role and restrict to ns

the-scott-hand:
Heron doesnt need "cluster admin" level permissions, and based on helm chart set up also does not need to manage resources across kubernetes namespaces. Therefore we should be using a Role and RoleBinding combo to only allow heron apiserver to perform actions on statefulsets in its own namespace.

see https://github.com/apache/incubator-heron/issues/3615#issue-701399133

### DIFF
--- a/deploy/kubernetes/helm/templates/tools.yaml
+++ b/deploy/kubernetes/helm/templates/tools.yaml
@@ -217,20 +217,39 @@ metadata:
     k8s-app: {{ .Release.Name }}-apiserver
 
 ---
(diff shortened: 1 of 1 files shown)


nicknezis:
The Kubernetes scheduler also creates and deletes namespaced services using the `coreClient` as seen [here](https://github.com/apache/incubator-heron/blob/c6f634b75281df424bb402305ea3523db35a616d/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/V1Controller.java#L108).

nicknezis (approved):

PR #3619, opened 2020-09-17
Added ability to disable base_url in Helm chart

nicknezis:
This PR adds a new config item to support disabling the `base_url` parameter when installing Heron using the Helm chart.

### DIFF
--- a/deploy/kubernetes/helm/templates/tools.yaml
+++ b/deploy/kubernetes/helm/templates/tools.yaml
@@ -116,8 +116,8 @@ spec:
             - >-
               heron-ui
               --port=8889
-              {{- if eq .Values.heron.enable_url true }}
-              --base_url={{ .Values.heron.url | default $defaultUrl }}
+              {{- if not (kindIs "invalid" .Values.heron.url) }}
+              --base_url={{ eq .Values.heron.url "-" | ternary $defaultUrl .Values.heron.url }}
               {{- end }}
         - name: heron-apiserver
           image: {{ .Values.image }}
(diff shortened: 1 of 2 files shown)


nicknezis:
Closing this PR because the desired behavior can be attained by setting `.Values.heron.url` to `/`.

nicknezis:
I tested setting `.Values.heron.url` to `/` and things didn't work. I believe this PR is still needed to use Heron UI with k8s NodePorts.

Code0x58:
does setting `--base_url=` work as expected as I feel it should? If so, I'm thinking it would be cleaner to use something better than `default` when setting the url, and to instead use something like `value = config_value if config_value is not None else config_value` - that would avoid the new coupled variable currently introduced by the PR.

nicknezis:
@Code0x58 The issue I ran in when trying something similar to what you suggested, was that `""` and `null` both resolved to a type that was not a String. So my attempt to add a comparison in the template failed. I'm not sure how to check type and type convert in the Helm template. Any suggestions?

windhamwong (approved):
Looks good with the OR comparison to use the default value.

windhamwong:
https://github.com/Masterminds/sprig/issues/53#issuecomment-327904181
This could be an insight of checking null or "". 
Maybe checking if .Values.heron.url is empty or not? (eq nil?)

nicknezis:
I updated the logic to use more features I learned about [Sprig Functions](http://masterminds.github.io/sprig/).

This latest version will use the single variable. If set to `"-"` it will use the full k8s proxy URL. I've made this the default. If a user sets the `heron.url` to any string, it will set that as the value. This includes the empty string case `""`. The third option that I needed was the ability to disable the `base_url` parameter. This can now be achieved by setting `heron.url: null`.

Code0x58 (approved):

Issue #3640, opened 2020-12-01
Heron UI license issue


joshfischer1108:
An inconsequential file - ./heron/tools/ui/resources/templates/shell.snip.html has an unspecified license.  This needs to be fixed after the 0.23.3-incubating-rc8 vote process.

PR #3641, opened 2020-12-01
Add license to shell.snip.html

nicknezis:
Fixes: #3640

### DIFF
--- a/heron/tools/ui/resources/templates/application.html
+++ b/heron/tools/ui/resources/templates/application.html
@@ -1,4 +1,4 @@
-{% comment %}
+{#
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
 distributed with this work for additional information
@@ -15,7 +15,7 @@ software distributed under the License is distributed on an
 KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
-{% endcomment %}
+#}
 
 <!DOCTYPE html>
 <html lang="en" class="not-showing-stats no-colors">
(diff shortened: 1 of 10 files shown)


joshfischer1108 (approved):

heron/tools/ui/resources/templates/shell.snip.html
```diff
@@ -1,3 +1,22 @@
+<!--
```
Code0x58, line 1: This is used as a Jinja2 template, so it would be preferable to make it a [Jinja2 comment](https://jinja.palletsprojects.com/en/2.10.x/templates/#comments)

nicknezis: I assume this is true for all of the files in that folder?

Code0x58: Yep. Switching those would be nice too - the notice wasn't converted in the FastAPI+Jinja2 upgrade

joshfischer1108: Thanks for catching @Code0x58

heron/tools/ui/resources/templates/application.html
```diff
@@ -1,4 +1,4 @@
-<!--
+{% comment %}
```
Code0x58, line 1: isn't it [`{# … #}`](https://jinja.palletsprojects.com/en/2.10.x/templates/#comments)?

nicknezis: I found the syntax [here](https://blog.j95.io/3/). Apologies if it's the wrong format. I'll make the update.

Code0x58 (approved):

Issue #3638, opened 2020-11-29
fix macOS build way


sreev:
**Is your feature request related to a problem? Please describe.**
build macos artifacts in a consistent way. update/upgrade to a small dependency is breaking the build.

**Describe the solution you'd like**
detail out the dependencies and versions.
make optional as possible
create new flag/options to keep an existing feature.
for eg. make pex optional.

**Describe alternatives you've considered**
open.

**Additional context**
none.

sreev:
Build from source on macOS bigsur fails. python3.9.0_2 / pip3 20.2.4.

```
DEPRECATION: The -b/--build/--build-dir/--build-directory option is deprecated. pip 20.3 will remove support for this functionality. A possible replacement is use the TMPDIR/TEMP/TMP environment variable, possibly combined with --no-clean. You can find discussion regarding this at https://github.com/pypa/pip/issues/8333.
ERROR: Package 'pex' requires a different Python: 3.9.0 not in '>=2.7,<3.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*'
INFO: Elapsed time: 228.509s, Critical Path: 85.54s
INFO: 876 processes: 153 internal, 723 local.
FAILED: Build did NOT complete successfully

WARNING: incubator-heron/heron/healthmgr/tests/java/BUILD:52:13: in srcs attribute of java_library rule //heron/healthmgr/tests/java:healthmgr-tests: please do not import '//heron/healthmgr/src/java:org/apache/heron/healthmgr/HealthManager.java' directly. You should either move the file to this package or depend on an appropriate rule there. Since this rule was created by the macro 'java_library', the error might have been caused by the macro implementation

warning: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: archive library: bazel-out/darwin-fastbuild/bin/heron/proto/libproto-cxx.a the table of contents is empty (no object file members in the library define global symbols)

ERROR: incubator-heron/tools/rules/pex/BUILD:53:8: Bootstrapping pex //tools/rules/pex:pex_wrapper failed (Exit 1): bash failed: error executing command 
  (cd /private/var/tmp/_bazel_svaddi/5513d20b4906d6e334d9ceb49a2d11d7/execroot/org_apache_heron && \
  exec env - \
[...]

PR #3646, opened 2020-12-08
Upgrade WORKSPACE python packages for python3.9 support

Code0x58:
This upgrades a few python packages. One benefit is that infrastructure/tests work on python3.9 (due to upgrading pex to a version that says it can be installed in python <=3.9) so will fix the issue seen in #3638


nicknezis (approved):

nicknezis:
Looks good. What else is needed to remove from draft status?

DvalCaruso (approved):
Duplicate of #

joshfischer1108:
@Code0x58 What do we need to do to pick up work on this branch?  Anything I can help with?

Code0x58:
I think with these changes the python version in Travis blows up, or image (tested via `./scripts/release/build-images --all`) having versions that blow up during the build.

I can't recall what specifically required a change from 3.6 to something higher - probably just the newer libraries that work with 3.9.

One path forward is to upgrade all the python versions installed in the container images+CI (Travis + Ubuntu vagrant manifests) to some minimum version - you could use pyenv in the Vagrant machine to find what the lowest working version is. This would be a bit of a pain due to the multiple runtimes required to use Heron, so can't just use pre-existing base images (I think the JRE may be the first class citizen for that in the images). It's not a particularly appealing change, especially considering it may force python tennants to try upgrading their code+dependencies, but it's probably a reasonable thing to do without reconsidering the whole architecture or trying stick with Python 3.6-3.8 only.

nwangtw:
I am wondering which module uses more libraries? It might be ideal to
minimize the use of extra libraries considering the compatibility issues
with Python.

.travis.yml
```diff
@@ -29,7 +29,7 @@ addons:
       - libgoogle-perftools-dev
 
 env:
-  - BAZEL_VERSION=4.1.0 ENABLE_HEAPCHECK=1
+  - BAZEL_VERSION=4.1.0 ENABLE_HEAPCHECK=1 LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 LANGUAGE=en_US.UTF-8
```
surahman, line 32:
```suggestion
  - BAZEL_VERSION=4.1.0 ENABLE_HEAPCHECK=1
```

Should resolve error:

```bash
UnicodeEncodeError: 'ascii' codec can't encode characters in position 73-75: ordinal not in range(128)
```

surahman: Reverted. The build container is still running Python 3.6:

```bash
$ which python
/opt/pyenv/shims/python
The command "which python" exited with 0.
$ python -V
Python 2.7.17
The command "python -V" exited with 0.
$ which python3
/opt/pyenv/shims/python3
The command "which python3" exited with 0.
$ python3 -V
Python 3.6.9
The command "python3 -V" exited with 0.
```

surahman:
The build containers are still running older versions of Python:

```bash
$ which python
/opt/pyenv/shims/python
The command "which python" exited with 0.
$ python -V
Python 2.7.17
The command "python -V" exited with 0.
$ which python3
/opt/pyenv/shims/python3
The command "which python3" exited with 0.
$ python3 -V
Python 3.6.9
The command "python3 -V" exited with 0.
```
We will probably need to update the base Dockerfiles here: https://github.com/apache/incubator-heron/tree/master/docker/compile

https://github.com/apache/incubator-heron/blob/ee369ef7745df530c765e89d2e8a2eefc8d8b742/docker/compile/Dockerfile.ubuntu18.04#L23-L40

nicknezis:
#3642 Code was integrated into this PR.

Tracker Behavior changes:

- replace `/` → `/topologies` redirect with (OpenAPI on ReDoc) documentation
- deprecate `executiontime` and always return 0.0 for it in the response envelope - this could probably be put back in, but I don't think it's worth it
- inbound requests aren't terminated if they take over 120s. At the moment, every outbound request has a 5s limit (due to httpx default) so I wouldn't expect the timeout to be needed (even in metrics queries which may transform data). A replacement timeout could be added in if necessary.

nwangtw (approved):

nicknezis (approved):

PR #3650, opened 2020-12-13
Allow localhost graceful shutdown

thinker0:
#3581 Add gracefully shutdown

In Apache Aurora, it is necessary to allow a graceful shutdown on localhost.

### DIFF
--- a/heron/shell/src/python/handlers/killexecutorhandler.py
+++ b/heron/shell/src/python/handlers/killexecutorhandler.py
@@ -45,11 +45,7 @@ class KillExecutorHandler(tornado.web.RequestHandler):
       os.killpg(os.getppid(), signal.SIGTERM)
 
     def is_local():
-      remote_ip = self.request.remote_ip
-      if 'localhost' == remote_ip or '127.0.0.1' == remote_ip or '::1' == remote_ip:
-        return True
-      else:
-        return False
+      return self.request.remote_ip in ('localhost', '127.0.0.1', '::1')
 
     logger = logging.getLogger(__file__)
     logger.info("Received 'Killing process' request")
--- a/heron/shell/src/python/handlers/killexecutorhandler.py
+++ b/heron/shell/src/python/handlers/killexecutorhandler.py
@@ -44,13 +44,20 @@ class KillExecutorHandler(tornado.web.RequestHandler):
       logger.info("Killing parent executor")
       os.killpg(os.getppid(), signal.SIGTERM)
 
+    def is_local():
+      remote_ip = self.request.remote_ip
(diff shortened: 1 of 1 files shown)


heron/shell/src/python/handlers/killexecutorhandler.py
```diff
@@ -44,13 +44,20 @@ def kill_parent():
...
+        return False
+
     logger = logging.getLogger(__file__)
     logger.info("Received 'Killing process' request")
     data = dict(parse_qsl(self.request.body))
 
     # check shared secret
     sharedSecret = data.get('secret')
-    if sharedSecret != options.secret:
+    if not is_local() and sharedSecret != options.secret:
```
Code0x58, line 60: This change is really that the secret is disregarded if the last hop of the request comes from a local connection (which could happen if you use a reverse proxy like nginx).

If this change is to allow graceful shutown, does mean aurora always calls this endpoint without the secret and from a loopback address? So before this patch, aurora would call this endpoint and get a 403 and no action, instead of getting to send a SIGTERM to the executor process group? If so, I'm wondering if you could just make Aurora send the SIGTERM to avoid having to worry about secrets being ignored in certain cases.

thinker0: # Aurora proceeds in the following order. 
https://aurora-scheduler.readthedocs.io/en/latest/reference/task-lifecycle/
```
1. POST localhost/quitquitquit
sleep 10
2. POST localhost/aboutabout
sleep 10
3. SIGTERM pid
sleep 10
4. SIGKILL pid
```
admin port 31000 to 32000, Random port,

# current 403
1. Currently 403 and the next SIGTERM, SIGKILL proceeded.

thinker0:
Nice work.. ^^;
```
[I 201214 14:47:11 web:1811] 200 GET /health (127.0.0.1) 0.64ms
[I 201214 14:47:31 web:1811] 200 GET /health (127.0.0.1) 0.53ms
[I 201214 14:47:46 killexecutorhandler:51] Received 'Killing process' request
[I 201214 14:47:46 web:1811] 200 POST /quitquitquit (127.0.0.1) 1.08ms
[I 201214 14:47:46 killexecutorhandler:44] Killing parent executor
```

joshfischer1108:
@thinker0 or @Code0x58 Is this PR ok to be further reviewed and merged?

Code0x58 (approved):
It doesn't sound too bad to me that this disregards the secret if the last hop of the request appears to be from the local host, assuming this is deployed in containers which is the case in the supported schedulers).

At a glance I can't see the use of this secret being in the documentation, so I don't think there's anywhere to update, but it would be nice if the use of the secret was documented somewhere - even in the docstring of `KillExecutorHandler`.

Issue #3659, opened 2021-01-12
Add config toggle for turning off Verbose JVM GC output


nicknezis:
**Is your feature request related to a problem? Please describe.**
The current implementation will generate 5 x 100MB x # of components worth of log files with JVM GC debug output. This is in addition to the actual log files (5 x 100MB x # of components)

**Describe the solution you'd like**
I would like a simple toggle to enable/disable this GC debug output

**Describe alternatives you've considered**
We could also make the log configuration parameters configurable, but I would be happy enough with ability to toggle.

**Additional context**
The Heron executor also seems to have older JVM parameters (pre-JDK 9), but we are now on JDK 11, so I think we should also clean up the older code that is no longer needed.

nicknezis:
Resolved by #3663

PR #3663, opened 2021-01-12
Added config toggle for verbose GC logging

nicknezis:
I added a config item to toggle the verbose GC logging. Not sure if this is the best way to do it. There was another option to add it to the heron_internals.yml. In the end I thought this might be the better approach because it would allow for the analytic to set the value.

I also removed some of the pre-JDK9 config options because Heron is now on JDK 11.


heron/executor/src/python/heron_executor.py
```diff
@@ -342,6 +343,8 @@ def init_from_parsed_args(self, parsed_args):
     self.health_manager_mode = parsed_args.health_manager_mode
     self.health_manager_classpath = '%s:%s'\
         % (self.scheduler_classpath, parsed_args.health_manager_classpath)
+    self.enable_verbose_gc_log = parsed_args.enable_verbose_gc_log \
```
nwangtw, line 346: You need to use boolean instead of string.

The string 'false' is true in a if statement.

nicknezis: Thank you. Does that look better?

heron/executor/src/python/heron_executor.py
```diff
@@ -102,6 +102,7 @@
 @click.option("--topology-defn-file", required=True)
 @click.option("--topology-id", required=True)
 @click.option("--topology-name", required=True)
+@click.option("--enable-verbose-gc-log", required=False)
```
Code0x58, line 105: so you want to add a [boolean option to the click app](https://click.palletsprojects.com/en/7.x/options/#boolean-flags), which I think would be best as:
```suggestion
@click.option("--verbose-gc-log", is_flag=True)
```
Without the "enable" as that's implied, and using `is_flag` so that the parser stores a bool.

heron/executor/src/python/heron_executor.py
```diff
@@ -342,6 +343,8 @@ def init_from_parsed_args(self, parsed_args):
     self.health_manager_mode = parsed_args.health_manager_mode
     self.health_manager_classpath = '%s:%s'\
         % (self.scheduler_classpath, parsed_args.health_manager_classpath)
+    self.enable_verbose_gc_log = parsed_args.enable_verbose_gc_log \
+        if parsed_args.enable_verbose_gc_log else False
```
Code0x58, line 347: If you have click see this is a flag with the suggestion above, this becomes
```suggestion
    self.verbose_gc_log = parsed_args.verbose_gc_log
```

heron/executor/src/python/heron_executor.py
```diff
@@ -574,36 +577,17 @@ def _get_java_major_version(self):
...
-          '-XX:+UseGCLogFileRotation',
-          '-XX:NumberOfGCLogFiles=5',
-          '-XX:GCLogFileSize=100M',
-          '-XX:+PrintPromotionFailure',
-          '-XX:+PrintTenuringDistribution',
-          '-XX:+PrintHeapAtGC',
-          '-XX:+HeapDumpOnOutOfMemoryError',
-          '-XX:ParallelGCThreads=4',
-          '-Xloggc:' + self.log_dir + '/gc.' + gc_name + '.log']
+    gc_cmd += [
```
Code0x58, line 580:
```suggestion
    gc_cmd = [
```

Code0x58 (changes requested):

heron/executor/src/python/heron_executor.py
```diff
@@ -574,36 +577,17 @@ def _get_java_major_version(self):
...
+        '-XX:+UseG1GC',
+        '-XX:+ParallelRefProcEnabled',
+        '-XX:+UseStringDeduplication',
+        '-XX:MaxGCPauseMillis=100',
+        '-XX:InitiatingHeapOccupancyPercent=30',
+        '-XX:+HeapDumpOnOutOfMemoryError',
+        '-XX:ParallelGCThreads=4']
+    if self.enable_verbose_gc_log:
```
Code0x58, line 588: Following the suggested naming
```suggestion
    if self.verbose_gc_log:
```

nicknezis:
This fixes #3659

Code0x58 (commented):
At least the python side appears ok. I can't approve on the Java side without further reading, but I'd double check the naming and how it connects to the executor (particularly as this may be the first `is_flag` example so my earlier suggestion may case issues, and this may be the first Boolean option)

nicknezis:
I'm going to do some actual testing before merging, so hopefully will catch any name mismatches.

heron/executor/tests/python/heron_executor_unittest.py
```diff
@@ -157,19 +148,16 @@ def get_expected_instance_command(component_name, instance_id, container_id):
...
-           "-XX:+UseConcMarkSweepGC -XX:+CMSScavengeBeforeRemark -XX:TargetSurvivorRatio=90 " \
-           "-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps " \
-           "-XX:+PrintGCCause -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 " \
-           "-XX:GCLogFileSize=100M -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution " \
-           "-XX:+PrintHeapAtGC -XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
-           "-Xloggc:log-files/gc.%s.log " \
+           "-Djava.net.preferIPv4Stack=true " \
+           "-XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication " \
+           "-XX:MaxGCPauseMillis=100 -XX:InitiatingHeapOccupancyPercent=30 " \
+           "-XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
```
nwangtw, line 154: "Xloggc:log-files/gc.%s.log" and the `instance_name` are still necessary I feel since all the logs are in the same folder.

heron/executor/tests/python/heron_executor_unittest.py
```diff
@@ -105,50 +105,41 @@ def build_packing_plan(self, instance_distribution):
...
   def get_expected_healthmgr_command():
     return "heron_java_home/bin/java -Xmx1024M -XX:+PrintCommandLineFlags " \
-           "-Djava.net.preferIPv4Stack=true -verbosegc " \
-           "-XX:+UseConcMarkSweepGC -XX:+CMSScavengeBeforeRemark -XX:TargetSurvivorRatio=90 " \
-           "-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps " \
-           "-XX:+PrintGCCause -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 " \
-           "-XX:GCLogFileSize=100M -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution " \
-           "-XX:+PrintHeapAtGC -XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
-           "-Xloggc:log-files/gc.healthmgr.log " \
+           "-Djava.net.preferIPv4Stack=true " \
```
nwangtw, line 139: "-Xloggc:log-files" might be necessary.

heron/executor/tests/python/heron_executor_unittest.py
```diff
@@ -105,50 +105,41 @@ def build_packing_plan(self, instance_distribution):
...
-           "-XX:+UseConcMarkSweepGC -XX:+CMSScavengeBeforeRemark -XX:TargetSurvivorRatio=90 " \
-           "-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps " \
-           "-XX:+PrintGCCause -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 " \
-           "-XX:GCLogFileSize=100M -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution " \
-           "-XX:+PrintHeapAtGC -XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
-           "-Xloggc:log-files/gc.metricsmgr-%d.log " \
+           "-Djava.net.preferIPv4Stack=true " \
+           "-XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication " \
+           "-XX:MaxGCPauseMillis=100 -XX:InitiatingHeapOccupancyPercent=30 " \
+           "-XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
```
nwangtw, line 114: "-Xloggc:log-files" might be necessary.

heron/executor/tests/python/heron_executor_unittest.py
```diff
@@ -105,50 +105,41 @@ def build_packing_plan(self, instance_distribution):
...
-           "-XX:+UseConcMarkSweepGC -XX:+CMSScavengeBeforeRemark -XX:TargetSurvivorRatio=90 " \
-           "-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps " \
-           "-XX:+PrintGCCause -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=5 " \
-           "-XX:GCLogFileSize=100M -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution " \
-           "-XX:+PrintHeapAtGC -XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
-           "-Xloggc:log-files/gc.metricscache.log " \
+           "-Djava.net.preferIPv4Stack=true " \
+           "-XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication " \
+           "-XX:MaxGCPauseMillis=100 -XX:InitiatingHeapOccupancyPercent=30 " \
+           "-XX:+HeapDumpOnOutOfMemoryError -XX:ParallelGCThreads=4 " \
```
nwangtw, line 128: "-Xloggc:log-files" might be necessary.

nwangtw (approved):
Overall LGTM other than the question about "-Xloggc".

nicknezis:
@nwangtw Is it ok that I modified the Heron CLI to add the new argument? In the future I want to also add a flag to toggle the heap dump JVM feature. Would I also add that as a flag on the CLI? It's there a more generic way to provide overrides per analytic submission when submitting to a remote Heron server?

nwangtw:
> @nwangtw Is it ok that I modified the Heron CLI to add the new argument? In the future I want to also add a flag to toggle the heap dump JVM feature. Would I also add that as a flag on the CLI? It's there a more generic way to provide overrides per analytic submission when submitting to a remote Heron server?

Yeah. I think passing from CLI could be very useful. I remember we can pass extra topology configs from CLI now but I can't remember the details. @huijunw @xiaoyao1991 ?

joshfischer1108:
> > @nwangtw Is it ok that I modified the Heron CLI to add the new argument? In the future I want to also add a flag to toggle the heap dump JVM feature. Would I also add that as a flag on the CLI? It's there a more generic way to provide overrides per analytic submission when submitting to a remote Heron server?
> 
> Yeah. I think passing from CLI could be very useful. I remember we can pass extra topology configs from CLI now but I can't remember the details. @huijunw @xiaoyao1991 ?

I think it's here:

Something like `--config-property PROPERTY=VALUE` 

https://github.com/apache/incubator-heron/blob/0f0ba185430437b2439a40c2a0c442fde25dc5f8/heron/tools/cli/src/python/args.py#L120

heron/executor/tests/python/heron_executor_unittest.py
nicknezis: `-Xloggc` set of flags are all related to the GC logging. In my testing, I still saw JVM component log files in the `log-files` folder. This change should only remove the GC specific logging (unless someone chooses to enable at submit time).

I believe if the `verbose-gc` flag is not set, the changes I made to the code will not set those JVM flags. This unit test is testing with the defaults, so what you see here matches what was generated by the `heron_executor.py`.

heron/executor/tests/python/heron_executor_unittest.py
nicknezis: Same as comment above. I believe `-Xloggc:log-files` is a GC specific setting. This does not impact the default logging.

heron/executor/tests/python/heron_executor_unittest.py
nwangtw: Yeah. The flag is for gc logs only, but it separates different gc logs into different files. For example, the gc log for this process would be "log-files/gc.healthmgr.log" and each process has its own gc log file.

nicknezis: I think the `heron_executor` functionality that was setting that path and log file names is still there. It's just not there for the default submission. You have to use the flag to enable those parameters. 

These lines of code that you are commenting on are test results for the default case that does not have `verbose_gc` enabled. If there was a test for the `verbose_gc` use case, then we would have another version of this string that included the various GC log parameters. I'm not sure how much more work it will be to add another test case that captures that scenario, but I will look into making it.

nicknezis:
@Code0x58 Github is still showing that you have "1 change requested". I'm not sure if that is something I can resolve or if you need to re-review to clear that state.

nicknezis:
Also a general re-review is probably warranted. I was able to test running an analytic with and without the flag and it seemed to operate as I wanted.

Code0x58 (approved):

PR #3664, opened 2021-01-12
Improve the Kubernetes Resource handling

nicknezis:
The previous Heron Kubernetes Scheduler calculated a homogenous PackingPlan and used the resulting `ScheduledResource` which had the "max" cpu and memory. The `RequiredResource` was never used. This Pull Request updates the logic in the Heron Kubernetes Scheduler to use both sets of `Resource` objects. The `RequiredResource` will be used as the `V1ResourceRequirements` `Request` and the `ScheduledResource` will continue to be set as the `V1ResourceRequirements` `Limit`.


nwangtw (approved):
Container resources and max resources are a bit messy in Heron, especially in the schedulers.

nwangtw:
lgtm.

PR #3665, opened 2021-01-20
Removing defaulted HeapDumpOnOutOfMemory

nicknezis:
This merge request removes the flag from the Heron Executor code. In most of the java examples, we show how a topology can add the optional flag. When running java components in a Kubernetes container, the heap dump file isn't as useful.

Here is an [example](https://github.com/apache/incubator-heron/blob/1a0d198ea301fd6346e0b1b6b7600c834874d8a0/examples/src/java/org/apache/heron/examples/api/AckingTopology.java#L75) of how to enable the flag.


nwangtw (approved):

thinker0:
LGTM

PR #3673, opened 2021-02-03
remove explicit style ignore

joshfischer1108:
This pr fixes a broken style link in the static site.  

Before this change the drop down list items would not lay on top of each other.

Screen Shot 2021-02-02 at 8 26 28 PM

After this change the drop down will properly lay on top of each other.

Screen Shot 2021-02-02 at 8 26 00 PM


joshfischer1108:
Also the footer of the page was incorrectly laid out because of this. 

Before this PR..
Screen Shot 2021-02-02 at 8 33 16 PM


After this PR...
Screen Shot 2021-02-02 at 8 33 04 PM

nicknezis (approved):

PR #3678, opened 2021-02-24
Round Robin minimum RAM requirement descreasing to 16MB

windhamwong:
Decrease the minimum requirement of RAM of single instance to 16MB in Round Robin algorithm

The original minimum RAM setting is too high for micro spouts/bolts. This helps developers to split programs into much smaller than before.

16MB should fit most of the situation. Round Robin is about dividing the allocated RAM for the single container into required pieces. i.e. 800MB has 50 instances with 16MB RAM


nwangtw (approved):
LGTM

xiaoyao1991 (approved):

nicknezis (approved):

PR #3699, opened 2021-07-07
Added support for adding Kubernetes annotations to the topology pod and service

nicknezis:
Added support for passing in `--config-property heron.kubernetes.pod.annotation.nick=wasHere` which will result with an annotation of `nick: wasHere` on the Kubernetes pod. Also there is a matching `heron.kubernetes.service.annotation` config item. 

The Heron Kubernetes scheduler will support multiple of these config items being set.

### DIFF
--- a/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
+++ b/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
@@ -1,26 +1,29 @@
-/**
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing,
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
+/*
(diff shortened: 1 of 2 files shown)


Code0x58 (approved):

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
```diff
@@ -83,6 +88,11 @@
   public static final String HERON_KUBERNETES_CONTAINER_VOLUME_MOUNT_PATH =
       "heron.kubernetes.container.volumeMount.path";
 
+  public static final String HERON_KUBERNETES_POD_ANNOTATION =
+      "heron.kubernetes.pod.annotation.";
+  public static final String HERON_KUBERNETES_SERVICE_ANNOTATION =
+          "heron.kubernetes.service.annotation.";
```
nwangtw, line 94: nit: inconsistent indent

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
```diff
@@ -152,6 +162,36 @@ static String getContainerVolumeMountPath(Config config) {
     return config.getStringValue(HERON_KUBERNETES_CONTAINER_VOLUME_MOUNT_PATH);
   }
 
+  static Set<String> getConfigKeys(Config config, String key) {
```
nwangtw, line 165: `keyPrefix` might be more explicit than `key`

nwangtw (approved):

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
nicknezis: Fixed

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/KubernetesContext.java
nicknezis: I like the suggestion. I wasn't quite sure how to convey that it's a prefix.

PR #3704, opened 2021-07-30
Adding missing logic to Kubernetes Scheduler to properly set the Remote Debug ports

nicknezis:
The remote debug ports were being added to the Pod Spec, but the logic was never added to provide the port information to the Heron Executor which launches the Java processes.

Also there was an issue with the Kubernetes port name being larger than 15 characters. So I renamed it to fit the K8s spec.

### DIFF
--- a/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/V1Controller.java
+++ b/heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/V1Controller.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


thinker0 (approved):

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/V1Controller.java
```diff
@@ -297,15 +296,26 @@ void deleteStatefulSet() {
...
     final Map<ExecutorPort, String> ports =
         KubernetesConstants.EXECUTOR_PORTS.entrySet()
             .stream()
             .collect(Collectors.toMap(Map.Entry::getKey,
                 e -> e.getValue().toString()));
 
-    final Config configuration = getConfiguration();
-    final Config runtimeConfiguration = getRuntimeConfiguration();
+    if (TopologyUtils.getTopologyRemoteDebuggingEnabled(Runtime.topology(runtimeConfiguration))
+            && numOfInstances != 0) {
```
joshfischer1108, line 309: Do we want to get rid of this magic number here in the code? It's readable to me, but may not be to others.

joshfischer1108 (approved):

heron/schedulers/src/java/org/apache/heron/scheduler/kubernetes/V1Controller.java
nicknezis: I copied this code from another scheduler. If we change it, we should fix it everywhere. Any name suggestions? My gut feeling is that this one might be obvious, so ok to keep as is.

windhamwong (approved):

PR #3718, opened 2021-09-30
update heron-shell download handler logging

huijunwu:
update heron-shell download handler logging

### DIFF
--- a/heron/shell/src/python/handlers/downloadhandler.py
+++ b/heron/shell/src/python/handlers/downloadhandler.py
@@ -23,6 +23,7 @@
 import os
 import logging
 import tornado.web
+import anticrlf
 
 from heron.shell.src.python import utils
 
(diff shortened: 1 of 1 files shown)


joshfischer1108 (approved):

heron/shell/src/python/handlers/downloadhandler.py
```diff
@@ -34,7 +35,14 @@ class DownloadHandler(tornado.web.RequestHandler):
   def get(self, path):
     """ get method """
 
-    logging.debug("request to download: %s", path)
+    handler = logging.StreamHandler()
+    handler.setFormatter(anticrlf.LogFormatter('%(levelname)s:%(name)s:%(message)s'))
+    logger = logging.getLogger(__name__)
```
nwangtw, line 40: I am wondering if we should move this to a class variable so that no need to set the formatter every time. 

Maybe not a big deal fro downloader though since downloading doesn't happen very often.

PR #3721, opened 2021-10-25
Helm: Adding option to switch the apiserver service between NodePort and ClusterIP

windhamwong:
It gives an option for users to pick whether the apiserver service should be NodePort or ClusterIP.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -355,17 +355,17 @@ http_archive(
 http_archive(
     name = "helm_mac",
     build_file = "@//:third_party/helm/helm.BUILD",
-    sha256 = "05c7748da0ea8d5f85576491cd3c615f94063f20986fd82a0f5658ddc286cdb1",
+    sha256 = "3a9efe337c61a61b3e160da919ac7af8cded8945b75706e401f3655a89d53ef5",
     strip_prefix = "darwin-amd64",
-    urls = ["https://get.helm.sh/helm-v3.0.2-darwin-amd64.tar.gz"],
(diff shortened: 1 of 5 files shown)


nicknezis:
> > Is there a way to support asking for NodePort and letting k8s pick a random port, while also allowing for explicit port selection in the value.yml? Perhaps a property to enable/disable and another to set the value?
> 
> good idea, maybe
> 
> ```
> apiserver:
>   NodePort:
>     enabled: true
>     port: <leave empty for random>
> ```

Yeah I think that could work.

windhamwong:
@nicknezis have a look :D

nicknezis:
Going to do some testing and maybe add the ability to use a random port number. Also just noticed that this feature branch is not made off of the main branch. We should rebase so that it can merge without any of the ARM branch's dependencies.

nicknezis:
@windhamwong I made the branch off of `master` branch. 

I also realized that `helm` was stripping away the comments in the generated `values.yml` file. This issue was documented [here](https://github.com/helm/helm/issues/6951). It's fixed in Helm 3.1.0, so I updated the version of Helm that we are using.

I also did some minor refactoring of the values to start aggregating `apiserver`, `tracker` and `ui` sets of properties.

I also moved the `apiserver` Service up in `tools.yml` to group it with the other two `Service` definitions.

I also mirrored your logic to add `NodePort` support to the other two services.

I chose to keep the default behavior with k8s auto picking the NodePort.

nicknezis (approved):

windhamwong:
The pushed code looks alright to me.

Issue #3170, opened 2019-01-24
ClassNotFoundException: io.netty.channel.epoll.NativeStaticallyReferencedJniMethods


sautran:
I am using the kubernete deployment and go through the installation step.
After deployment, I am seeing the following error in api-server as well as the worker pod

Jan 24, 2019 6:27:39 PM org.apache.distributedlog.impl.BKNamespaceDriver getDefaultEventLoopGroup
WARNING: Could not use Netty Epoll event loop for bookie server:
java.lang.UnsatisfiedLinkError: failed to load the required native library
	at dlshade.io.netty.channel.epoll.Epoll.ensureAvailability(Epoll.java:78)
	at dlshade.io.netty.channel.epoll.EpollEventLoopGroup.<clinit>(EpollEventLoopGroup.java:38)
	at org.apache.distributedlog.impl.BKNamespaceDriver.getDefaultEventLoopGroup(BKNamespaceDriver.java:259)
	at org.apache.distributedlog.impl.BKNamespaceDriver.initializeBookKeeperClients(BKNamespaceDriver.java:270)
	at org.apache.distributedlog.impl.BKNamespaceDriver.initialize(BKNamespaceDriver.java:208)
	at org.apache.distributedlog.api.namespace.NamespaceBuilder.build(NamespaceBuilder.java:238)
	at com.twitter.heron.downloader.DLDownloader.download(DLDownloader.java:66)
	at com.twitter.heron.downloader.DownloadRunner.main(DownloadRunner.java:42)
Caused by: java.lang.NoClassDefFoundError: io/netty/channel/epoll/NativeStaticallyReferencedJniMethods
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1824)
	at java.lang.Runtime.load0(Runtime.java:809)
	at java.lang.System.load(System.java:1086)
	at dlshade.io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:36)
	at dlshade.io.netty.util.internal.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:278)
	at dlshade.io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:231)
	at dlshade.io.netty.channel.epoll.Native.loadNativeLibrary(Native.java:191)
[...]

PR #3729, opened 2021-11-11
Bump Netty to 4.1.70

nicknezis:
In production we see error messages about not finding the native epoll library. This may be fixed in newer versions of Netty as referenced here. (https://github.com/netty/netty/issues/10537)


nicknezis:
This may fix #3170

nwangtw (approved):

PR #3736, opened 2021-11-13
switch base image - centos

joshfischer1108:


joshfischer1108:
Base image repo can be found here: https://github.com/joshfischer1108/heron-centos7

joshfischer1108:
With this change we should be able to release centos7 containers next release.

docker/dist/Dockerfile.dist.centos7
```diff
@@ -16,22 +16,10 @@
 #  under the License.
 
 #syntax=docker/dockerfile:1.2
-FROM centos:centos7
+FROM joshfischer1108/heron-centos7
```
nwangtw, line 19: This is expected? Official image should be better I feel?

joshfischer1108: I'm up for any suggestions.  We can create a company or organization under Dockerhub, and store the base image there.  Or something else if you prefer.

joshfischer1108: We can create an org like pulsar did https://hub.docker.com/r/apachepulsar/pulsar, but at the end of the day it's still just a dockerhub account.

PR #3738, opened 2021-11-16
Protect file names in requests

nwangtw:


nicknezis:
Closing because we merged #3739

PR #3746, opened 2021-12-09
Fix dependency

thinker0:
Fix https://github.com/apache/incubator-heron/issues/3745

PR #3749, opened 2021-12-19
Updated to fix missing anticrlf dependency

nicknezis:
It looks like we added anticrlf to heron-shell for security reasons in this PR. https://github.com/apache/incubator-heron/pull/3718

But then removed it here 10 days ago: https://github.com/apache/incubator-heron/pull/3746

This PR fixes the issue by adding the `anticrlf` dependency.

### DIFF
--- a/heron/shell/src/python/BUILD
+++ b/heron/shell/src/python/BUILD
@@ -6,6 +6,7 @@ pex_library(
         ["**/*.py"],
     ),
     reqs = [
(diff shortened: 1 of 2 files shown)


windhamwong:
Looks good as it fixes the issue mentioned in #3746.

windhamwong (approved):

nwangtw (approved):

PR #3751, opened 2022-01-02
Python rules for Bazel

joshfischer1108:


joshfischer1108:
- [ ] Add native python rules to workspace
- [ ] Update docker rules because of a deprecated pip_import dependency 
- [ ] Remove pex rules from the project
- [ ] Add requirements.txt to support all of the python dependencies for the project.  This is only a first pass to loading python deps.  After the initial migration is done we can lazily load python deps for improved build times, if desired
- [ ] Remove tools/rules/newgenproto.bzl file as  it seems to be a duplicate.
- [ ] Swap out pex_library  with native py_libray rules 
- [ ] and pex_binary with native py_binary rules
- [ ] Investigate the use of starlark to replace python build scripts
- [ ] Investigate the use for py_proto_libaries  from google https://thethoughtfulkoala.com/posts/2020/05/08/py-protobuf-bazel.html

nicknezis:
If we are removing all use of pex, we will need to ensure we can still Launch HeronPy topologies. The Heron Executor code currently decides which binary type based on file extension. The logic in this file will need to be updated based on these changes. https://github.com/apache/incubator-heron/blob/fab089ce2c9cdbdd2d6a7ef7cda6cf22b201469c/heron/executor/src/python/heron_executor.py#L790

tools/python/requirements.txt
```diff
@@ -0,0 +1,25 @@
+click==7.1.2
+colorlog==2.6.1
+netifaces==0.10.6
+PyYAML==3.13
```
nicknezis, line 4:
```suggestion
PyYAML==5.4.1
```

surahman:
Draft removal of most references to `pex_binary` and `pex_library`.

nwangtw:
This would be really cool~

surahman:
There is no `zip_safe` and `resources` in `py_binary`. I have opted to disable `zip_safe` and placed the change in its own commit so it can be dropped if needed.

`resources` in [bazel_pex_rules](https://github.com/benley/bazel_rules_pex#pex_binary.data) appears to have switched to calling this `data`:

> `List of labels; Optional; Default is []`
>
> Files to include as resources in the final pex binary.
>
> Putting other rules here will cause the outputs of those rules to be embedded in this one. Files will be included as-is. Paths in the archive will be relative to the workspace root.

This aligns with [Bazel's](https://docs.bazel.build/versions/main/be/common-definitions.html#typical.data) definition of `data`:

> `List of labels; optional`
>
> Files needed by this rule at runtime. May list file or rule targets. Generally allows any target.

third_party/python/cpplint/BUILD
```diff
@@ -7,7 +7,8 @@ load("@heron_py_deps//:requirements.bzl", "requirement")
 
 py_binary(
     name = "cpplint",
-    entrypoint = "cpplint",
+    srcs = ["empty.py"],
```
joshfischer1108, line 10: @surahman / @nicknezis Wasn't exactly sure what to do here.  `py_binary` requires the `srcs` to have something in it.  So to get around this, I put an `empty.py` file in folder.

third_party/python/pylint/BUILD
```diff
@@ -7,6 +7,7 @@ load("@heron_py_deps//:requirements.bzl", "requirement")
 
 py_binary(
     name = "pylint",
+    srcs = ["empty.py"],
```
joshfischer1108, line 10: same here.

third_party/python/cpplint/BUILD
surahman: Thanks @joshfischer1108 I will look through the files and see where else I need to do this. I am still learning Bazel at this point so @nicknezis if you have the time please feel free to jump in anywhere at any time.

surahman:
I just reverted the commits on `pylint` and `cpplint`.

third_party/python/cpplint/BUILD
```diff
@@ -2,8 +2,14 @@ licenses(["notice"])
...
-pex_binary(
+load("@rules_python//python:defs.bzl", "py_binary")
+load("@heron_py_deps//:requirements.bzl", "requirement")
+
+py_binary(
     name = "cpplint",
-    entrypoint = "cpplint",
-    reqs = ["cpplint==1.3.0"],
+    srcs = ["empty.py"],
+    main = "cpplint",
```
surahman, line 11: Could this be causing an infinite recursion/expansion of the build rule?

joshfischer1108: I guess it could be possible. @nicknezis would have a better idea.   He's been our go to Bazel person,

surahman:
I think the best way to approach this enormous task is to completely remove a single PEX build rule at a time and make sure everything compiles and all tests pass. Start with the simpler build rules and move to the more complex ones. That way we can verify in a concrete fashion that the changes we are making are correct. It also means the work we are completing can be incrementally merged back into `master` so that there is progress.

Conceivably a Kanban board that tracks the rules to remove? We can clump some together for fewer PR's.

joshfischer1108:
I like the idea of removing a single build rule at a time.  After jumping into a lot this code I think we've learned a lot (a good time for growth 😄 ) Are you thinking we should abandon this branch and start fresh?

surahman:
> Are you thinking we should abandon this branch and start fresh?

We are thinking alike 😁

joshfischer1108:
Closing to account for smaller chunks of work.

PR #3753, opened 2022-01-03
Updated External JVM Rules plugin

nicknezis:
Updated `bazelbuild/rules_jvm_external` to version 4.2. Changelog can be found here. There were numerous enhancements and bug fixes.

One of the key new features is the addition of the `maven_install.fail_if_repin_required` flag which fails the build if the `maven_install` section was updated, but the pinned artifacts was not also updated.

This PR also replaced some outdated rules that were used in the original migration to the `rules_jvm_external` tool.


joshfischer1108:
Did you mean to delete the `maven_install.json`?  We don't have how this plugin works for Heron documented anywhere. Could you add something quick to the instructions on compiling Heron in the website so we don't lose track of this process/tool?

surahman (approved):
LGTM with Josh's comment above.

nicknezis:
> Did you mean to delete the `maven_install.json`? We don't have how this plugin works for Heron documented anywhere. Could you add something quick to the instructions on compiling Heron in the website so we don't lose track of this process/tool?

I didn't delete it. Only refreshed it with the newer version of the `rules_jvm_external` build tool. I think the logic for how dependencies are ordered changed, so that caused lots of changes in the json file. But that should be a one time change due to the newer build tool. 

General information about the pinning is here: https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader

With this updated version, if anyone ever updates the dependency list and doesn't repin, it will actually fail to compile and give the documentation in the console output for how to refresh.

joshfischer1108:
> > Did you mean to delete the `maven_install.json`? We don't have how this plugin works for Heron documented anywhere. Could you add something quick to the instructions on compiling Heron in the website so we don't lose track of this process/tool?
> 
> I didn't delete it. Only refreshed it with the newer version of the `rules_jvm_external` build tool. I think the logic for how dependencies are ordered changed, so that caused lots of changes in the json file. But that should be a one time change due to the newer build tool.
> 
> General information about the pinning is here: https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader
> 
> With this updated version, if anyone ever updates the dependency list and doesn't repin, it will actually fail to compile and give the documentation in the console output for how to refresh.

Sweet, sounds good.

joshfischer1108 (approved):

PR #3758, opened 2022-01-17
Helm chart cleanup

nicknezis:
1. Updated Helm to 3.7.2.
2. Added LICENSE, NOTICE and DISCLAIMER files to Helm chart
3. Updated the Helm chart.yaml `apiVersion` to `v2` (Helm 3.x)

Helm v2 is necessary to make the Helm chart discoverable in the apache.jfrog.org Helm repo.

I tried to find a way to instruct Helm to leave the commented LICENSE header in the Chart.yaml, but it seems the Chart.yaml is always processed to a final form with no comments.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -359,17 +359,17 @@ http_archive(
 http_archive(
     name = "helm_mac",
     build_file = "@//:third_party/helm/helm.BUILD",
-    sha256 = "3a9efe337c61a61b3e160da919ac7af8cded8945b75706e401f3655a89d53ef5",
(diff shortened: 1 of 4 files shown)


windhamwong (approved):

surahman:
_For posterity:_

> I tried to find a way to instruct Helm to leave the commented LICENSE header in the Chart.yaml, but it seems the Chart.yaml is always processed to a final form with no comments.

As per the conversation Nick initiated on dev@h.a.o on this issue, it is acceptable for a file that is generated or processed to not contain the ASF header. I also have been unable to find a way to work around the pre-processing performed server/client-side.

PR #3772, opened 2022-02-17
Add legacy CentOS7

thinker0:
As it is still used a lot in the legacy environment, a compiled CentOS 7 is required. CentOS 8 will be EOL...

https://www.centos.org/centos-linux-eol/

### DIFF
--- a/docker/compile/Dockerfile.centos7
+++ b/docker/compile/Dockerfile.centos7
@@ -19,6 +19,7 @@ FROM centos:centos7
 
 # This is passed to the heron build command via the --config flag
 ENV TARGET_PLATFORM linux
+ENV LC_ALL en_US.utf8
 
 RUN yum -y upgrade && yum -y install centos-release-scl-rh
 RUN yum -y install \
@@ -28,8 +29,8 @@ RUN yum -y install \
       cmake \
       openssl-devel \
       file \
-      gcc \
-      gcc-c++ \
+      devtoolset-8-gcc \
+      devtoolset-8-gcc-c++ \
       git \
       kernel-devel \
       libtool \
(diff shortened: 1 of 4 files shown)


nicknezis:
Yes Centos 8 update was a short term solution to the Python issues. We should switch to use CentOS Stream, which is not EOL. https://quay.io/repository/centos/centos?tab=tags

Other options are Rocky Linux and RHEL's UBI images. At work we are moving from Centos to RHEL and Rocky.

Do we really need to add the Centos 7 Docker image? At this point, no one should be using a Centos 7 Heron image when there are better alternatives.  Is the desire to add it just for compilation testing?

thinker0:
A Docker image is not required.

However, there are users who need legacy beta-metal environments, so the centos7 part is needed.

nicknezis:
Ahh ok. For testing purposes then. Thanks. I still wonder if it's our responsibility to maintain the older image. We don't build all images in any automated tests. The docker images and documentation were horrible out of date when we did our recent "fix the build" effort. I think this is why I'm questioning the need for it. But if you know of specific customers that need it, and having the Dockerfile helps, and we will maintain it... then I guess it's ok. We'll see how well we do.

We should also rename the `centos8` image to `rocky8`. Change the `FROM` line to `FROM rockylinux:8.5`. This will address the Centos 8 EOL concern. Then we'll have both Centos 7 and Rocky 8.

thinker0:
#3773

docker/compile/Dockerfile.centos7
```diff
@@ -0,0 +1,66 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+FROM centos:centos7
+
+# This is passed to the heron build command via the --config flag
+ENV TARGET_PLATFORM linux
+
```
nicknezis, line 22:
```suggestion

ENV LC_ALL en_US.utf8

```

nicknezis (changes requested):
We should add the Centos7 equivalent files in the `dist` and `test` folders.

thinker0:
```
% ./docker/scripts/test-unittest.sh centos 0.20.5
Build label: 4.2.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 2 18:15:58 2021 (1638468958)
Build timestamp: 1638468958
Build timestamp as int: 1638468958
Platform Linux
Using C compiler          : /opt/rh/devtoolset-8/root/usr/bin/gcc (8.3.1)
Using C++ compiler        : /opt/rh/devtoolset-8/root/usr/bin/g++ (8.3.1)
Using C preprocessor      : /opt/rh/devtoolset-8/root/usr/bin/cpp (8.3.1)
Using C++ preprocessor    : /opt/rh/devtoolset-8/root/usr/bin/cpp (8.3.1)
Using linker              : /usr/bin/ld.bfd
Using JDK                 : /usr/lib/jvm/java-11-openjdk
Using Automake            : /usr/bin/automake (1.13.4)
Using Autoconf            : /usr/bin/autoconf (2.69)
Using Make                : /usr/bin/make (3.82)
Using Python3             : /opt/rh/rh-python38/root/usr/bin/python3.8 (3.8.11)
Using Libtool             : /usr/bin/libtool (2.4.2)
Using archiver            : /usr/bin/ar
Using coverage tool       : /usr/bin/gcov
Using ant                 : /usr/bin/ant
Using dwp                 : /usr/bin/dwp
Using nm                  : /usr/bin/nm
Using objcopy             : /usr/bin/objcopy
Using objdump             : /usr/bin/objdump
Using strip               : /usr/bin/strip

INFO: Elapsed time: 1668.568s, Critical Path: 376.83s
INFO: 3608 processes: 1160 internal, 2448 local.
INFO: Build completed successfully, 3608 total actions
Test cases: finished with 1189 passing and 0 failing out of 1189 test cases

Executed 247 out of 247 tests: 247 tests pass.
INFO: Build completed successfully, 3608 total actions
Cleaning up scratch dir
```

nicknezis (approved):

PR #3773, opened 2022-02-17
Change centos to RockyLinux-8

thinker0:
https://www.centos.org/centos-linux-eol/


thinker0:
https://github.com/apache/incubator-heron/pull/3772#issuecomment-1042636469

nicknezis:
The filenames should be renamed to include `rocky8`. I can make this update and do some testing later today.

nicknezis:
The rebase/merge was a bit painful, so I had to do a force push. You might need to do a fresh checkout if you want to add more commits. Hopefully this captures the full combo of Centos 7 updates and Rocky 8 additions.

nicknezis (approved):

PR #3786, opened 2022-03-02
Fix for missing physical plan in UI

### LINKED ISSUE #3784: Heron UI fails to show the Physical Plan layout
**Describe the bug**
The Heron UI renders the logical plan, but the physical plan does not render.

### LINKED ISSUE #3785: Heron UI does not render timeline metrics
**Describe the bug**
Heron UI does not render the timeline metrics elements.

nicknezis:
Fixes #3784 and Fixes #3785


heron/tools/tracker/src/python/routers/metrics.py
```diff
@@ -119,7 +119,7 @@ async def get_metrics( # pylint: disable=too-many-arguments
   )
 
 
-@router.get("/metricstimeline", response_model=metricstimeline.MetricsTimeline)
+@router.get("/metrics/timeline", response_model=metricstimeline.MetricsTimeline)
```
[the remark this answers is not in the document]
surahman: There might also be cascading changes required elsewhere, for example in links within the UI.

nicknezis:
Nice catch. The tests seem to work for me, but there's definitely more to clean up as you mentioned. I guess I started making the code match this line: https://github.com/apache/incubator-heron/blob/4d1ab84eea99a374c6838cb9401d6a7fc908af24/heron/tools/ui/resources/static/js/stat-trendlines.js#L264 with the assumption we had broken something in the Tracker cleanup. But looking back at old versions of the code, this UI query may never have matched up with the tracker api. Very odd. I personally prefer the `/metrics/query` syntax, but might be easier to just roll back to `/metricsquery` style route. And instead I'll update the javascript to use the proper api call.

surahman:
> But looking back at old versions of the code, this UI query may never have matched up with the tracker api. Very odd.

My guess is technical debt.

> I personally prefer the `/metrics/query` syntax, but might be easier to just roll back to `/metricsquery` style route.

I agree. Good REST API design necessitates versioning of the API as well as nesting under path routes. All metrics queries should be nested under the `/metrics/` path.

I have never used Node.js and I am not sure if the URL needs to have the `?` embedded in it - I think the framework should deal with the query string delimiter itself. We typically do not have to do this in libraries for C++/Java/Golang.

https://github.com/apache/incubator-heron/blob/dd23309cba06853e495f378f5d2cf2c50706a41e/heron/tools/ui/resources/static/js/plan-stats.js#L378

Please ping when ready for review.

surahman (approved):

surahman (approved):
Great to pay down tech debt. I think there might be a non-functional Python docstring update required:

https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/query.py#L21

nicknezis:
I'm running into some other issues with parameter mismatch betwen `ui` and `tracker`. So I'm doing a bit of testing. One example is `topology` vs `topology_name`. There is an alias [here](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/metrics.py#L106), but it didn't seem to work. I was testing with a failed Tracker API call and got different response when using either `topology` or `topology_name`. Perhaps the `alias` isn't working? @Code0x58 might have some insight in helping answer that question.

Another issue I'm seeing is an exception on an API call because `role` isn't present. But it's listed as `Optional`. The parameter is treated differently between the two set of routes. [Example 1](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/container.py#L44) and [Example 2](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/topologies.py#L112).

surahman:
I have always had issues bringing up the UI so I can only look at the code.

> One example is `topology` vs `topology_name`. There is an alias [here](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/metrics.py#L106), but it didn't seem to work.

The naming convention for topologies has changed on K8s and I am not sure if this would affect the UI. `Labels` have the topology's name and the `StatefulSet` name has either `-manager` or `-executors` appended to it. I am not sure how the tracker manages the scraping and collecting of stats.

> Another issue I'm seeing is an exception on an API call because `role` isn't present. But it's listed as `Optional`. The parameter is treated differently between the two set of routes. [Example 1](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/container.py#L44) and [Example 2](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/topologies.py#L112).

According to the [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) docs, a value of `None` will be returned as required. How is a `role` value of `None` handled? Exception? What changed?

nicknezis:
So for testing `heron-tracker` and `heron-ui`, I realized doing the local install is much easier. You can try this:
1. `bazel run --config=darwin_nostyle -- scripts/packages:heron-install.sh --user `
2. Add `~/.heron/bin/` to your `$PATH`
3. Submit analytic: 
```
heron submit local ~/.heron/examples/heron-streamlet-examples.jar \  org.apache.heron.examples.streamlet.WindowedWordCountTopology \
WindowedWordCountTopology \
--deploy-deactivated
```
4. Run the tracker: `heron-tracker --type file --rootpath ~/.herondata/repository/state/local/`
5. Run the UI: `heron-ui`

Code0x58:
> I'm running into some other issues with parameter mismatch betwen `ui` and `tracker`. So I'm doing a bit of testing. One example is `topology` vs `topology_name`. There is an alias [here](https://github.com/apache/incubator-heron/blob/2baa27cdf68e531ec5e560bc61d3f26484d1e2ca/heron/tools/tracker/src/python/routers/metrics.py#L106), but it didn't seem to work. I was testing with a failed Tracker API call and got different response when using either `topology` or `topology_name`. Perhaps the `alias` isn't working? @Code0x58 might have some insight in helping answer that question.

It looks like alias may be a bit of a misnomer, it may be better to call it `source`, if this [reference](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#alias-parameters) covers everything

nicknezis:
I've been banging my head against this code. Things are definitely broken. I have more drastic changes, but I haven't pushed them yet because I'm still testing. The `exceptions` page is still broken. Somewhere we lost the combo of `component` and `instances` query parameters.

surahman:
I am time-constrained for the next few weeks and I still cannot bring up the UI. I typically find it faster to jump in with a debugger and find the issue causing the problem and address it there. Failing that it is time to find what caused the issue.

We need to isolate the commit in which this happened. I would start by checking out the last tagged release to see if the UI is working there. If it is, run a [`Git bisect`](https://git-scm.com/docs/git-bisect) between that commit and `master` branch's `HEAD` to find the problematic commit. You should have a better idea of what caused the issue. There were major changes to the backend of the UI in the PR that bumped Python to 3.8/3.9, so that is a potential candidate.

nicknezis:
I figured out one of my main blockers that I was banging my head against. The `Topology` object that I was referencing was a Protobuf object. Once I reference the proper `TopologyInfo` type, the code worked better. Finally making progress again. Will have more commits to this PR in the near future.

nicknezis:
I now have most of it working, except the `exceptions.json` call on `heron-ui`. I think I'm going to have to put some of the `ResponseEnvelope` stuff back. Running into an issue with the `ApiResponse` requiring a `dict`, but it's getting a `list` instead. Hopefully will have this all figured out soon.

nicknezis:
To give some context to one of the issues I ran into with this work. In the Heron Tracker, we are now using FastAPI. There is a limitation in the framework which does not allow for custom routers in a nested router. Some information can be found here: https://github.com/tiangolo/fastapi/issues/4110

So I think my options are as follows.
1. Remove the envelope which was part of each returned API call. Not sure if it's a great idea to have. We didn't really use the feature. We never set the timing info. We already have HTTP return codes for status.
2. Keep the envelope on API returns, but implement in each handler method. The custom API Router decoration doesn't seem to work.
3. I can technically hack it by making separate nested `App` instantiations, but this seems wrong.

In the PR, I've gone back and forth between option 1 and 2 as I worked through various issues. I'm currently leaning towards option 1. This is reflected in the PR as it currently stands. But I can be convinced that it is better to go back to the original API responses with the envelope on each Heron Tracker API result.

surahman:
> 1. Remove the envelope which was part of each returned API call. Not sure if it's a great idea to have. We didn't really use the feature. We never set the timing info. We already have HTTP return codes for status

If option 1 simplifies the code then I would agree with going down this route but my fear is that it might break something elsewhere. For option 2, how many/which handlers would we need to update to handle the encasing envelopes?

I am looking things over and I will try to assist as much as I can on this PR.

surahman:
I think we can safely remove the `ResponseEnvelope`. There may have been grander plans when it was for its use but they do not seem to have ever been implemented.

https://github.com/apache/incubator-heron/blob/8841d1c4d6b3156a40ffed856fb32c1ceb682855/heron/tools/tracker/src/python/utils.py#L56-L63

It is just a data struct with references to it restricted to:
https://github.com/apache/incubator-heron/blob/8841d1c4d6b3156a40ffed856fb32c1ceb682855/heron/tools/tracker/src/python/app.py
https://github.com/apache/incubator-heron/blob/8841d1c4d6b3156a40ffed856fb32c1ceb682855/heron/tools/tracker/src/python/utils.py
The values of importance are limited to the HTTP response codes, which are themselves restricted to `success` and `failure`. Nothing else in the data structure is ever used in a worthwhile way.

heron/tools/tracker/src/python/query_operators.py
```diff
@@ -476,7 +477,7 @@ async def execute(self, tracker, tmanager: TManagerLocation, start: int, end: in
       # Initialize with first metrics timeline and its instance
       met = Metrics(None, None, metric.instance, start, end, metric.timeline.copy())
       for timestamp in list(met.timeline.keys()):
-        v = self._f(met.timeline[timestamp], metrics2[""].timeline.get(timestamp))
+        v = self._f(met.timeline[timestamp], metrics2[""].timeline[timestamp])
```
surahman, line 480: Good change. It is better to use the `[]` instead of the `.get()` method throughout to access the dictionary elements for consistency.

heron/tools/tracker/src/python/routers/container.py
```diff
@@ -186,29 +189,30 @@ async def _get_exception_log_response(
...
+  for exception_log in exception_response.exceptions:
+    ret.append(ExceptionLog(
+      hostname = exception_log.hostname,
+      instance_id = exception_log.instance_id,
+      stacktrace = exception_log.stacktrace,
+      lasttime = exception_log.lasttime,
+      firsttime = exception_log.firsttime,
+      count = str(exception_log.count),
+      logging = exception_log.logging,
+    ))
```
surahman, line 214: Indentation and whitespace fix

```suggestion
    ret.append(ExceptionLog(
        hostname=exception_log.hostname,
        instance_id=exception_log.instance_id,
        stacktrace=exception_log.stacktrace,
        lasttime=exception_log.lasttime,
        firsttime=exception_log.firsttime,
        count=str(exception_log.count),
        logging=exception_log.logging,
    ))
```

heron/tools/tracker/src/python/app.py
```diff
@@ -80,35 +79,23 @@ async def shutdown_event():
...
-
 @app.exception_handler(Exception)
 async def handle_exception(_, exc: Exception):
-  payload = ResponseEnvelope[str](
-      result=None,
-      execution_time=0.0,
-      message=f"request failed: {exc}",
-      status=constants.RESPONSE_STATUS_FAILURE
-  )
+  message=f"request failed: {exc}"
```
surahman, line 84: Whitespace fix:

```suggestion
  message = f"request failed: {exc}"
```

heron/tools/tracker/src/python/tracker.py
```diff
@@ -89,7 +89,7 @@ def get_topology(
       cluster: str,
       role: Optional[str],
       environ: str,
-      topology_name: str,
+      topology_name: str,
```
surahman, line 92: Whitespace fix:

```suggestion
      topology_name: str,
```

heron/tools/tracker/src/python/routers/metrics.py
```diff
@@ -151,19 +152,19 @@ class TimelinePoint(BaseModel): # pylint: disable=too-few-public-methods
 
 class MetricsQueryResponse(BaseModel): # pylint: disable=too-few-public-methods
   """A metrics timeline over an interval."""
-  start_time: int = Field(..., alias="starttime")
-  end_time: int = Field(..., alias="endtime")
+  starttime: int = Field(..., alias="starttime")
+  endtime: int = Field(..., alias="endtime")
```
surahman, line 156: Would be best to switch these to `str` as well for consistency as you have switched their similar definitions from `int` to `str` elsewhere? Or are these used in some sort of numeric calculation somewhere?

heron/tools/tracker/src/python/routers/metrics.py
```diff
@@ -151,19 +152,19 @@ class TimelinePoint(BaseModel): # pylint: disable=too-few-public-methods
...
-@router.get("/metricsquery", response_model=MetricsQueryResponse)
+@router.get("/metrics/query", response_model=MetricsQueryResponse)
 async def get_metrics_query( # pylint: disable=too-many-arguments
     cluster: str,
-    role: Optional[str],
     environ: str,
     query: str,
+    role: Optional[str] = None,
     start_time: int = Query(..., alias="starttime"),
     end_time: int = Query(..., alias="endtime"),
```
surahman, line 169: If you switch over to `str` in the class definition these will need to be updated as well.

heron/tools/tracker/src/python/app.py
```diff
@@ -80,35 +79,23 @@ async def shutdown_event():
...
 async def clusters() -> List[str]:
-  return ResponseEnvelope[List[str]](
-      execution_time=0.0,
-      message="ok",
-      status="success",
-      result=[s.name for s in state.tracker.state_managers],
-  )
-
+  return (s.name for s in state.tracker.state_managers)
+
```
surahman, line 95: Whitespace fix:

```suggestion

```

surahman (commented):
There are a bunch of import and style check errors. I have corrected the style check errors inline but have not touched the following import-related errors:

```bash
heron/tools/tracker/src/python/metricstimeline.py:24:0: W0611: Unused Log imported from heron.common.src.python.utils.log (unused-import)
heron/tools/tracker/src/python/metricstimeline.py:25:0: W0611: Unused common_pb2 imported from heron.proto (unused-import)
heron/tools/tracker/src/python/routers/container.py:25:0: W0611: Unused constants imported from asyncio (unused-import)
heron/tools/tracker/src/python/routers/container.py:26:0: W0611: Unused CompletedProcess imported from subprocess (unused-import)
heron/tools/tracker/src/python/routers/container.py:28:0: W0611: Unused Log imported from heron.common.src.python.utils.log (unused-import)
heron/tools/tracker/src/python/routers/metrics.py:36:0: W0404: Reimport 'Query' (imported line 35) (reimported)
heron/tools/tracker/src/python/utils.py:31:0: W0611: Unused iscoroutinefunction imported from asyncio (unused-import)
heron/tools/tracker/src/python/utils.py:32:0: W0611: Unused wraps imported from functools (unused-import)
heron/tools/tracker/src/python/utils.py:37:0: W0611: Unused constants imported from heron.tools.tracker.src.python (unused-import)
heron/tools/tracker/src/python/utils.py:43:0: W0611: Unused APIRouter imported from fastapi (unused-import)
heron/tools/tracker/src/python/utils.py:44:0: W0611: Unused Field imported from pydantic (unused-import)
heron/tools/tracker/src/python/utils.py:45:0: W0611: Unused GenericModel imported from pydantic.generics (unused-import)
```

Let us get the build past these errors and see what else the test suite is able to highlight as problematic.

heron/tools/tracker/src/python/metricstimeline.py
```diff
@@ -73,21 +73,21 @@ async def get_metrics_timeline(
...
   # Parse the response from tmanager.
   response_data = tmanager_pb2.MetricResponse()
   response_data.ParseFromString(result.content)
 
-  if response_data.status.status == common_pb2.NOTOK:
-    if response_data.status.HasField("message"):
-      Log.warn("Received response from Tmanager: %s", response_data.status.message)
+  # if response_data.status.status == common_pb2.NOTOK:
+  #   if response_data.status.HasField("message"):
+  #     Log.warn("Received response from Tmanager: %s", response_data.status.message)
```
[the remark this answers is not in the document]
nicknezis: Committed these by accident. This is a portion of code I kept putting back each time I toggled through the approaches. I'll clean it up. Will leave the comment unresolved until I actually do it.

heron/tools/tracker/src/python/routers/metrics.py
nicknezis: I honestly don't remember the specifics, but I think the changes I made was due to an error I encountered. But it might have been self inflicted based on the various changes.  At the moment, I don't know the right answer, but I agree with your comment.

nicknezis:
Fixes #3801

heron/tools/ui/src/python/main.py
```diff
@@ -306,6 +309,34 @@ def file_download(
...
+        executiontime=time.time() - started,
+        result={},
+    )
+  else:
+    return ApiListEnvelope(
+        status="success",
+        message="",
+        executiontime=time.time() - started,
+        result=result,
+    )
```
surahman, line 339: The `if` clause here will always return `false` and the `else` is dead code.

heron/tools/ui/src/python/main.py
```diff
@@ -316,17 +347,25 @@ class ApiEnvelope(pydantic.BaseModel):
...
+        executiontime=time.time() - started,
+        result={},
+    )
+  else:
+    return ApiEnvelope(
+        status="success",
+        message="",
+        executiontime=time.time() - started,
+        result=result,
+    )
```
surahman, line 368: The `if` clause here will always return `false` and the `else` is dead code.

heron/tools/ui/src/python/main.py
```diff
@@ -614,9 +653,11 @@ def cli(
     host: str, port: int, base_url_option: str, tracker_url_option: str, verbose: bool
 ) -> None:
   """Start a web UI for heron which renders information from the tracker."""
-  global base_url, tracker_url
+  global base_url, tracker_url, Log
   base_url = base_url_option
-  log.configure(level=logging.DEBUG if verbose else logging.INFO)
+  log_level = logging.DEBUG if verbose else logging.INFO
+  log.configure(log_level)
+  Log = log.Log
```
surahman, line 660: I assume the purpose of this line was to update the globally defined `Log`? If this is the case, we need the `Log` variable to be defined as `global` as it is now.

heron/tools/tracker/src/python/main.py
```diff
@@ -125,6 +125,8 @@ def cli(
 
   log_level = logging.DEBUG if verbose else logging.INFO
   log.configure(log_level)
+  global Log
+  Log = log.Log
```
surahman, line 129: I assume the purpose of this line was to update the globally defined `Log`? If this is the case, we need the `Log` variable to be defined as `global` as it is now.

heron/tools/ui/src/python/main.py
nicknezis: I made my changes because the Debug logging didn't seem to work until I got a fresh copy of `log.Log` which is defined in `heron/common/src/python/utils/log.py`.  I can put these 3 lines back to the original line I replaced. Feel free to submit any edit suggestions. I've never really done any python coding, so not sure how best to handle the logging initialization.

heron/tools/ui/src/python/main.py
nicknezis: Should it be `if result is None:` or do you recommend I remove the `if` statement?

nicknezis:
I think I have most of the UI working again. But having issues with this menu on the Exception page. I'm not sure, but this might have been broken for a long time before. I'm inclined to wrap up this PR to get it into the `master` branch. We can fix the exceptions menu at a later point. Don't want to leave the main branch broken for too long.

image

heron/tools/ui/src/python/main.py
surahman: > I can put these 3 lines back to the original line I replaced. Feel free to submit any edit suggestions.

If you need a fresh copy of the logger here we will need to rename the variable or suppress the style check warnings. The issue that then crops up is that the `Log` variable is assigned in the method where it is declared, but it is never used. That is why I thought your intent is to update the globally defined `Log`.

heron/tools/ui/src/python/main.py
surahman: The call to `api_topology_list_json` makes a lambda call to `get_component_exceptions` here:
https://github.com/apache/incubator-heron/blob/f417981fa33c9a1d9990e41a241878898adcb928/heron/tools/ui/src/python/main.py#L446-L447

The `get_component_exceptions` can return a type of `Any`, and that includes `None`.
https://github.com/apache/incubator-heron/blob/af7053ec8162ec43ea7b58d4ddf746ff2b2ab43d/heron/tools/common/src/python/clients/tracker.py#L250-L266

If you would like to handle a `None` return type by returning a `failure` status then your suggestion is correct. I will push the change.

surahman:
> We can fix the exceptions menu at a later point. Don't want to leave the main branch broken for too long.

I agree, let us get this into a stable state and merge. We can create another set of issues for what we know is broken and outstanding.

heron/tools/tracker/src/python/query_operators.py
```diff
@@ -460,14 +460,15 @@ async def execute(self, tracker, tmanager: TManagerLocation, start: int, end: in
...
-          v = self._f(met.timeline[timestamp], metric.timeline.get(timestamp))
-          if v is None:
-            met.timeline.pop(timestamp, None)
-          else:
-            met.timeline[timestamp] = v
-        all_metrics.append(met)
+        if metrics:
+          met = Metrics(None, None, metric.instance, start, end, metrics[""].timeline.copy())
+          for timestamp in list(met.timeline.keys()):
+            v = self._f(met.timeline[timestamp], metric.timeline[timestamp])
```
surahman: The method `.get` in `dict` is safer than the brackets because it will return `None` if the key is not found. This is currently raising a `KeyError` in the test suite. Reverting.

heron/tools/tracker/src/python/query_operators.py
surahman: Reverting. Please see the comment above.

heron/tools/tracker/src/python/routers/container.py
```diff
@@ -143,8 +142,8 @@ class ExceptionLog(BaseModel):
   hostname: str
   instance_id: str
   stack_trace: str = Field(..., alias="stacktrace")
-  last_time: int = Field(..., alias="lasttime")
-  first_time: int = Field(..., alias="firsttime")
+  last_time: str = Field(..., alias="lasttime")
+  first_time: str = Field(..., alias="firsttime")
```
surahman, line 146: We need to be careful with this. One of the many oddities and quirks of Python is that assigning values of different types to a variable will convert the type of the variable.

```python
test: str = "test string"
print (type(test))

test = 6
print (type(test))

#####

number: int = -1
print(type(number))

test = "test string"
number = test
print(type(number))
```

heron/tools/tracker/tests/python/app_unittest.py
```diff
@@ -34,19 +34,31 @@ def test_clusters(client, tracker):
...
   response = client.get("/topologies", json={
       "cluster": [],
       "environ": [],
   })
-  assert response.json() == ok({})
+  assert response.json() == ok({'executiontime': 0.0,
+                                'message': 'ok',
+                                'result': {},
+                                'status': 'success',
+                                'tracker_version': '0.1.0'})
```
surahman, line 64: These tests should be updated to match what needs to be happening in the production code.

heron/tools/common/src/python/clients/tracker.py
```diff
@@ -122,27 +122,26 @@ def strip_whitespace(s):
...
+    Log.error("error from tracker: %s", response.status_code)
     return None
 
-  execution = data["executiontime"] * 1000
   duration = (end - start) * 1000
-  Log.debug(f"URL fetch took {execution:.2}ms server time for {url}")
   Log.debug(f"URL fetch took {duration:.2}ms round trip time for {url}")
 
-  return data["result"]
+  return data
```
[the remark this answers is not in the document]
surahman: I think `data` is a `dict` and what you are returning is the entire `JSON` object as a `dict` instead of the `result` entry. This is causing failures in the tests. Have you handled this in rotuines that depend on this return value? If so you need to update the test suite to reflect the expected results.

heron/tools/tracker/src/python/query_operators.py
```diff
@@ -150,11 +150,11 @@ async def execute(
       raise Exception(metrics["message"])
 
     # Put a blank timeline.
-    if not metrics.get("timeline"):
-      metrics["timeline"] = {
+    if not metrics.timeline:
+      metrics.timeline = {
```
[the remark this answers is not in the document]
nicknezis: What did you revert? The changes in the PR? Do we instead need to update the tests to match the changes we made?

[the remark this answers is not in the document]
surahman: The changes are in commit fc8efabef837d68a8be7175d2d326bc7a2ddc3cc.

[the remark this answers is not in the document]
nicknezis: Interesting. I'll have to compare the tests with the code. In some cases, `metrics` isn't a `dict`, but instead is a `MetricsTimeline` object which doesn't have a `get()` method.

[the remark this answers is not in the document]
surahman: It is the production code that is faulty and causing this error. The test merely calls it. If you choose to go back to your original code you will need to update this routine, and anything that depends on it, as well as the tests.

<details><summary>Build Log Failure</summary>

```bash
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <heron.tools.tracker.src.python.query_operators.Subtract object at 0x7fe237a53e20>
tracker = <Mock id='140609572909[840]()'>, tmanager = <Mock id='140609572912960'>
start = 100, end = 300
    async def execute(self, tracker, tmanager: TManagerLocation, start: int, end: int) -> Any:
      """
      Return _f applied over all values in [start, end] of the two operands. Scalars
      are expanded a timeseries with all points equal to the scalar.
    
      """
      metrics, metrics2 = await asyncio.gather(
          self._get_metrics(self.operand1, tracker, tmanager, start, end),
          self._get_metrics(self.operand2, tracker, tmanager, start, end),
      )
    
      # In case both are multivariate, only equal instances will get operated
      if self._is_multivariate(metrics) and self._is_multivariate(metrics2):
        all_metrics = []
        for key in metrics:
          if key not in metrics2:
            continue
          met = Metrics(None, None, key, start, end, {})
          for timestamp in list(metrics[key].timeline.keys()):
            value = self._f(
                metrics[key].timeline[timestamp],
                metrics2[key].timeline.get(timestamp),
            )
            if value is None:
              metrics[key].timeline.pop(timestamp, None)
            else:
              met.timeline[timestamp] = value
          all_metrics.append(met)
        return all_metrics
    
      # If first is univariate
      if not self._is_multivariate(metrics):
        all_metrics = []
        for key, metric in metrics2.items():
          # Initialize with first metrics timeline, but second metric's instance
          # because that is multivariate
          if metrics:
            met = Metrics(None, None, metric.instance, start, end, metrics[""].timeline.copy())
            for timestamp in list(met.timeline.keys()):
>             v = self._f(met.timeline[timestamp], metric.timeline[timestamp])
E             KeyError: 120
../../../../../../../../../../_tmp/2b5b8aef99713d16d1504f0eb43a6602/.pex/unzipped_pexes/9e6c937c2cca84dde46c456e33bc2dcb1a8c4dc0/heron/tools/tracker/src/python/query_operators.py:466: KeyError
- generated xml file: /home/travis/.cache/bazel/_bazel_travis/be6dac4936703c7eedcb4f5cf38cdd65/execroot/org_apache_heron/bazel-out/k8-fastbuild/testlogs/heron/tools/tracker/tests/python/query_operator_unittest/test.xml -
```

</details>

nicknezis:
Another realization I had thanks to some chatting with @surahman ... The Tracker client is not only used by the Heron UI. There seem to also be references in Heron Explorer (I'm not 100% sure what that code does). But we will need to ensure those API changes did not break those referencing components.

surahman:
This is a tricky set of issues to resolve and is compounded by a lack of access to design documents. There are a lot of moving parts and interdependencies.

nicknezis:
Verified that the changes didn't break Heron Explorer (I actually had never used it before, but it's pretty handy tool). The Tracker client code helped abstract away the changes.

nicknezis:
Updated `pylint` to resolve an issue with it failing on Python 3.9+. https://github.com/PyCQA/pylint/issues/3882

nicknezis:
I'm now updating to resolve the style fix suggestions.

surahman:
I will get started on the bottom end of the list of style errors to help out. Files I am looking at:

```bash
************* Module basics.base_instance
************* Module basics.bolt_instance
************* Module basics.spout_instance
************* Module instance
************* Module network.gateway_looper
************* Module network.heron_client
************* Module network.metricsmgr_client
************* Module network.protocol
************* Module network.socket_options
************* Module utils.metrics.metrics_helper
************* Module utils.metrics.py_metrics
************* Module utils.misc.pplan_helper
************* Module utils.misc.serializer_helper
************* Module utils.topology.topology_context_impl
```

heron/instance/src/python/utils/metrics/py_metrics.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, metrics_collector):
                     PY_GC_GENERATION_1_THRESHOLD: self.g1_threshold,
                     PY_GC_GENERATION_2_THRESHOLD: self.g2_threshold,
                     PY_GC_GENERATION_3_THRESHOLD: self.g3_threshold}
-    super(PyMetrics, self).__init__(self.metrics)
+    _ = super()
```
nicknezis, line 83: Should this have been `super().__init__(self.metrics)`? I don't know Python well enough, but that was the recommendation I saw online. I'm not familiar with the `_ = super()` syntax.

surahman: This syntax is analogous to what we have in Golang. I am essentially telling the Python3 interpreter and linter know "Yes, I am aware a value is returned from this function call. No, I do not need it, and yes I wish to ignore it."

Some linters will output a warning about this and it could potentially become an error down the road. Let us not generate/leave technical debt to be dealt with later.

surahman: I stand corrected, if calling `super().__init__` there is no return value. My IDE's linter was reporting warnings in other calls to `super()`.

surahman:
Resolving the merge conflict with changes from this branch.

surahman:
I think these style fixes will require editing most of the Python codebase. The errors Just keep propagating.

nicknezis:
> I think these style fixes will require editing most of the Python codebase. The errors Just keep propagating.

Yes. Lots of Python changes. But most of them seem to be related to the f-string syntax. Hopefully once we get through that, it won't be too bad. Just takes a lot of time to fix each one and repeat the process.

nicknezis:
@surahman Could you help me convert the deprecated `asyncore` to `asyncio`? I might just add the waiver for now, but we should update it because it will be removed in Python 3.12. What do you think? I don't know how easy it will be to switch over.

surahman:
> @surahman Could you help me convert the deprecated `asyncore` to `asyncio`?

I have not worked with either of those libraries but I can look into it when I have some time. It is best to address this before it becomes an issue.

surahman (commented):
Incredible work!

I have not had the time to deployment test this PR and it would be nice if someone else could assist with that. There are some old review comments that still need to be addressed a handful of new ones. Overall it is pretty much in a mergeable state.

surahman:
I just removed a reference to update the global `Log` that I added.

heron/tools/common/src/python/clients/tracker.py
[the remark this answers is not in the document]
nicknezis: Correct. The Tracker API no longer has the envelope. So now there is not an embedded `result` field. The decoded json is the result.

nicknezis:
I can't figure out why the timeline is not working. I've tracked my issue to this [line of code](https://github.com/apache/incubator-heron/blob/06e9f75988fd254053b640689879c3393703b0dd/heron/tools/common/src/python/clients/tracker.py#L575). Within the `compute_max` function, I can print the final `dict` that is returned. It has values. But when I print `data`, it only has an empty `dict`.

surahman:
I am not sure but I do not think there is a need to convert the zipped together values into a `list` and then a `dict`. But then I also do not believe this would affect the return value.

https://github.com/apache/incubator-heron/blob/06e9f75988fd254053b640689879c3393703b0dd/heron/tools/common/src/python/clients/tracker.py#L626

Maybe try this:

```python
return dict(zip(keys, values))
```

nicknezis:
> I am not sure but I do not think there is a need to convert the zipped together values into a `list` and then a `dict`. But then I also do not believe this would affect the return value.
> 
> https://github.com/apache/incubator-heron/blob/06e9f75988fd254053b640689879c3393703b0dd/heron/tools/common/src/python/clients/tracker.py#L626
> 
> Maybe try this:
> 
> ```python
> return dict(zip(keys, values))
> ```

I think `zip()` returns an `Iterator` and the `list()` call converts to an actual List. I was not familiar with the `zip` method, but it seems to be a [standard pattern](https://realpython.com/python-zip-function/#passing-n-arguments). When I print the value returned from that call I can see the value. It's only when that same result is returned from the `compute_max` that I get an empty `{}` in the `data` variable.

surahman:
> I think `zip()` returns an `Iterator` and the `list()` call converts to an actual List.

I believe `dict` might have a constructor to handle `zip` return values as seen [here](https://stackoverflow.com/questions/209840/how-do-i-convert-two-lists-into-a-dictionary).

Code0x58:
> I can't figure out why the timeline is not working. I've tracked my issue to this [line of code](https://github.com/apache/incubator-heron/blob/06e9f75988fd254053b640689879c3393703b0dd/heron/tools/common/src/python/clients/tracker.py#L575). Within the `compute_max` function, I can print the final `dict` that is returned. It has values. But when I print `data`, it only has an empty `dict`.

Did you do that print within the `len(filtered_ts) > 0 and len(filtered_ts[0]["timeline"]) > 0:` branch of `compute_max(...)`? I'm wondering if the `{}` you are seeing is from the fall-though branch.

Beware that `timelines` and `values` in `compute_max(...)` are generators, so will be exhausted if you do something with them before they are used in the `return dict(list(zip(keys, values)))`, so if you did `print(dict(list(zip(keys, values))))` it would consume `values` and lead to zip of the keys list and an empty generator which results in an empty zip (python 3.10 adds [`strict=True`](https://docs.python.org/3/library/functions.html#zip) to help detect these) and would so make the `return dict(...)` line return an empty dict.

_p.s._
> I am not sure but I do not think there is a need to convert the zipped together values into a `list` and then a `dict`. But then I also do not believe this would affect the return value.
> 
> https://github.com/apache/incubator-heron/blob/06e9f75988fd254053b640689879c3393703b0dd/heron/tools/common/src/python/clients/tracker.py#L626
> 
> Maybe try this:
> 
> ```python
> return dict(zip(keys, values))
> ```

both are equivalent and will be handled the same way due to python's duck typing, as both a generator (returned by zip, in type annotations it is compatible with [`typing.Iterator`](https://docs.python.org/3/library/typing.html#typing.Iterator) but is more strictly a [`typing.Generator`](https://docs.python.org/3/library/typing.html#typing.Generator)) and list will be treated as iterables. The use of `list(generator)` can be faster for some cases, but constructs an object which will be discarded so can be a waste of memory.

surahman:
Thank you for the detailed explanation!

nicknezis:
I think this PR is finally done. I'm now able to see the timeline in the UI and most of the functionality is working (as good as it probably was working in the past). There are some super deep menus I think I'd like to clean up in the future, but this is good to go. Thank you for the help @surahman @Code0x58 and @thinker0 !

thinker0:
+1

Test Success

thinker0:
#3814 typo

surahman (approved):
:star: :star: :star: :star: :star: :clap:

PR #3789, opened 2022-03-10
Remove fixed young generation heap space of instance

thinker0:
It seems that it is not suitable for JVM 11 or higher to statically allocate 50% heap.

Ref: 
```
For Java 11 and above
The Xmx value is 25% of the available memory with a maximum of 25 GB. 
However, where there is 2 GB or less of physical memory, the value set is 50% of available memory with 
a minimum value of 16 MB and a maximum value of 512 MB.
For Java 8
The Xmx value is half the available memory with a minimum of 16 MB and a maximum of 512 MB.
```

https://blog.openj9.org/2020/04/30/default-java-maximum-heap-size-is-changed-for-java-8/

### DIFF
--- a/heron/executor/src/python/heron_executor.py
+++ b/heron/executor/src/python/heron_executor.py
@@ -610,11 +610,20 @@ class HeronExecutor:
             java_version.startswith("1.6") or \
             java_version.startswith("1.5"):
       java_metasize_param = 'PermSize'
+    xmn_param = '-Xmn%dM' % xmn_size
+    if self._get_java_major_version() >= 11:
+        # For Java 11 and above.
+        # The Xmx value is 25% of the available memory with a maximum of 25 GB.
+        # However, where there is 2 GB or less of physical memory,
+        # the value set is 50% of available memory with a minimum value of 16 MB and a maximum value of 512 MB.
+        # For Java 8
+        # The Xmx value is half the available memory with a minimum of 16 MB and a maximum of 512 MB.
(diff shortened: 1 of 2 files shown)


nicknezis (approved):

heron/executor/src/python/heron_executor.py
```diff
@@ -610,11 +610,20 @@ def _get_jvm_instance_options(self, instance_id, component_name, remote_debugger
             java_version.startswith("1.6") or \
             java_version.startswith("1.5"):
       java_metasize_param = 'PermSize'
+    xmn_param = '-Xmn%dM' % xmn_size
+    if self._get_java_major_version() >= 11:
+        # For Java 11 and above.
+        # The Xmx value is 25% of the available memory with a maximum of 25 GB.
```
nicknezis, line 616: This information only applies to the Xmx default value. But it seems we are always setting Xmx on line 624. So perhaps this comment isn't needed? Or is it related to the Xmn change this PR introduces?

thinker0: Want to delete -Xmn.

If there is this setting, YoungHeap area is too limited, so if -Xmx4g, only 2g is used.

thinker0: Before deletion -Xmx=5.5g Used 2.25g
After deletion -Xmx=5.5g Used 4.9g

image

nicknezis (approved):

PR #3795, opened 2022-03-19
Remove unnecessary Bookkeeper format init container from Helm chart

### LINKED ISSUE #3771: Add a initialization toggle to the Helm chart
There is an issue with the way the Bookkeeper cluster is initialized. It is currently only initialized when running the `minikube` platform mode. This causes issues for any of the other platform modes. Also this causes an issue whenever the Init container runs again, because it wipes out any underlying data. This is an issue in Kubernetes clusters that may restart frequently. For example K3D and KinD clusters.

The proposed fix is to provide an `initialize` setting which will be defaulted to false, but can be explicitly set during installation of the Helm chart. This is how Pulsar Helm chart handles Bookkeeper initialization. Here is the Pulsar documentation that mentions the `initialize` setting: https://pulsar.apache.org/docs/en/helm-deploy/#deploy-pulsar-cluster-using-helm

nicknezis:
Fixes #3771

### DIFF
--- a/deploy/kubernetes/helm/Chart.yaml.template
+++ b/deploy/kubernetes/helm/Chart.yaml.template
@@ -18,7 +18,7 @@
 apiVersion: v2
 name: heron
 version: VERSION
-kubeVersion: >= 1.16
+kubeVersion: ">=1.16"
 appVersion: VERSION
 description: Heron is a fast distributed streaming engine for processing large data volumes with velocity
 type: application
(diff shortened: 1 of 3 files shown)


nicknezis:
Had to rebase to include the latest fix from master that resolved the Github unsecure API build fix.

joshfischer1108 (approved):

PR #3798, opened 2022-03-24
Replace Helm chart's generic Bookkeeper platform property with specific properties

### LINKED ISSUE #3770: Helm chart removal of platform setting
The design which uses the `.Values.platform` property throughout the template files. The Heron Helm chart toggles various settings based on this one property.

The proposed fix is to remove this `platform` setting. Instead we will expose a property for each underlying setting that is modified by this `platform` setting. We will enumerate each Platform value and provide an example values.yml file. An example can be found in the Pulsar helm chart repo: https://github.com/apache/pulsar-helm-chart/tree/master/examples

wromansky:
This pull request removes the antiquated use of `platform` property which assumed certain combinations of Bookkeeper settings based on early incarnations of Kubernetes. This PR will create unique configurable settings for each of the features. This will allow greater flexibility for Heron users choosing to create clusters in Kubernetes.

Fixes #3770

### DIFF
--- a/deploy/kubernetes/helm/templates/bookie.yaml
+++ b/deploy/kubernetes/helm/templates/bookie.yaml
@@ -45,7 +45,7 @@ data:
   BK_journalMaxBackups: "{{ .Values.bookieJournalMaxBackups }}"
   BK_journalMaxSizeMB: "{{ .Values.bookieJournalMaxSizeMB }}"
   BK_logSizeLimit: "{{ int64 .Values.bookieLogSizeLimit }}"
-  {{- if or (eq .Values.platform "gke") (eq .Values.platform "minikube") }}
+  {{- if .Values.useHostNameAsBookieID }}
   BK_useHostNameAsBookieID: "true"
   {{- end }}
   # TODO: Issue 458: https://github.com/apache/bookkeeper/issues/458
@@ -60,7 +60,7 @@ data:
 ## cannot be moved across different nodes.
 ## For this reason, we run BK as a daemon set, one for each node in the
 ## cluster, unless restricted by label selectors
-{{- if or (eq .Values.platform "gke") (eq .Values.platform "minikube") }}
+{{- if .Values.useStatefulSet }}
 apiVersion: apps/v1
 kind: StatefulSet
 {{- else }}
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/helm/values.yaml.template
```diff
@@ -33,6 +33,15 @@ image: apache/heron:VERSION
 # Heron image pull policy
 imagePullPolicy: IfNotPresent
 
+# Optional Variables
```
nicknezis, line 36: Let's move these under the `bookkeeper` section (@ line 109). Maybe insert at the end of that section (@ line 116). You will also need to update the property names in the yaml to reference `.Values.bookkeeper.*`

deploy/kubernetes/helm/templates/bookie.yaml
```diff
@@ -165,7 +165,7 @@ spec:
             path: /bookkeeper/data/ledgers
 {{- end }}
 
-{{- if or (eq .Values.platform "gke") (eq .Values.platform "minikube") }}
+{{- if .Values.formatDisks }}
```
nicknezis, line 168: Instead of `formatDisks`, let's use `.Values.bookkeeper.useVolumeClaimTemplate`
```suggestion
{{- if .Values.bookkeeper.useVolumeClaimTemplate }}
```

deploy/kubernetes/helm/templates/bookie.yaml
```diff
@@ -154,7 +154,7 @@ spec:
             - name: ledgers-disk
               mountPath: /bookkeeper/data/ledgers
 
-{{- if or (eq .Values.platform "aws") (eq .Values.platform "baremetal") }}
+{{- if .Values.mountLocalDisk }}
```
nicknezis, line 157: This name is ok, but to keep consistent with the other name change I suggested, let's make this one `.Values.bookkeeper.useHostPath`.
```suggestion
{{- if .Values.bookkeeper.useHostPath }}
```

nicknezis (changes requested):
Added various comments. Also you still need to remove the `platform` setting from the `values.yaml.template` file. Lines 22-28 can be removed.

nicknezis (approved):

PR #3799, opened 2022-03-25
Update build-docker.sh - Removing debug flag

windhamwong:
Removing the debug message on build-docker.sh

### DIFF
--- a/docker/scripts/build-docker.sh
+++ b/docker/scripts/build-docker.sh
@@ -58,7 +58,6 @@ run_build() {
       break
     fi
   done
-  echo $DOCKER_SQUASH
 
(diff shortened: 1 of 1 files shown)


docker/scripts/build-docker.sh
```diff
@@ -58,7 +58,6 @@ run_build() {
       break
     fi
   done
-  echo $DOCKER_SQUASH
```
joshfischer1108, line 61: What's the reason for removing this?

windhamwong: It looks like it is a debug flag here only, without meaning.
Most of the time (at least i never see other than this) is showing `False` and this is it.

joshfischer1108 (approved):

PR #3807, opened 2022-04-01
Add logback.xml

thinker0:
Add logback.xml


nicknezis:
The error message in TravisCI stated that there was no output within a specific time period. Restarting the build to see if it happens again.

joshfischer1108 (approved):

nicknezis:
TravisCI build passed. I also verified with a local k8s install. API Server much more improved from before!

PR #3819, opened 2022-04-18
Fix HealthManager

thinker0:
- Fix legacy of HealthManager
- Fix Sfl4jBridge (Unsupported)
- Add healthcheck API for heron-tracker, heron-ui
- Fix heron-tracker, heron-ui

### DIFF
--- a/heron/healthmgr/src/java/org/apache/heron/healthmgr/HealthManager.java
+++ b/heron/healthmgr/src/java/org/apache/heron/healthmgr/HealthManager.java
@@ -48,7 +48,6 @@ import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
-import org.apache.heron.api.utils.Slf4jUtils;
 import org.apache.heron.classification.InterfaceStability.Evolving;
 import org.apache.heron.classification.InterfaceStability.Unstable;
 import org.apache.heron.common.basics.SingletonRegistry;
@@ -153,7 +152,6 @@ public class HealthManager {
   }
 
   public static void main(String[] args) throws Exception {
-    Slf4jUtils.installSLF4JBridge();
     CommandLineParser parser = new DefaultParser();
     Options slaManagerCliOptions = constructCliOptions();
 
--- a/heron/tools/tracker/src/python/query_operators.py
+++ b/heron/tools/tracker/src/python/query_operators.py
@@ -20,17 +20,13 @@
 
 ''' query_operators.py '''
 import asyncio
-import logging
 import math
 
 from typing import Any, Dict, List, Optional, Union
 
-from heron.common.src.python.utils import log
 from heron.proto.tmanager_pb2 import TManagerLocation
 from heron.tools.tracker.src.python.metricstimeline import get_metrics_timeline
 
-Log = log.Log
-Log.setLevel(logging.DEBUG)
 
 #####################################################################
 # Data Structure for fetched Metrics
--- a/heron/tools/tracker/src/python/query_operators.py
+++ b/heron/tools/tracker/src/python/query_operators.py
@@ -20,13 +20,17 @@
 
 ''' query_operators.py '''
 import asyncio
+import logging
 import math
 
 from typing import Any, Dict, List, Optional, Union
 
+from heron.common.src.python.utils import log
 from heron.proto.tmanager_pb2 import TManagerLocation
 from heron.tools.tracker.src.python.metricstimeline import get_metrics_timeline
 
+Log = log.Log
+Log.setLevel(logging.DEBUG)
 
 #####################################################################
 # Data Structure for fetched Metrics
@@ -460,7 +464,7 @@ class _SimpleArithmaticOperator(Operator):
       for key, metric in metrics2.items():
         # Initialize with first metrics timeline, but second metric's instance
         # because that is multivariate
-        if metrics:
+        if "" in metrics:
           met = Metrics(None, None, metric.instance, start, end, metrics[""].timeline.copy())
           for timestamp in list(met.timeline.keys()):
             v = self._f(met.timeline[timestamp], metric.timeline.get(timestamp))
@@ -477,7 +481,10 @@ class _SimpleArithmaticOperator(Operator):
       # Initialize with first metrics timeline and its instance
       met = Metrics(None, None, metric.instance, start, end, metric.timeline.copy())
       for timestamp in list(met.timeline.keys()):
-        v = self._f(met.timeline[timestamp], metrics2[""].timeline.get(timestamp))
+        met2_value = None
+        if "" in metrics2:
+          met2_value = metrics2[""].timeline.get(timestamp)
+        v = self._f(met.timeline[timestamp], met2_value)
         if v is None:
           met.timeline.pop(timestamp, None)
         else:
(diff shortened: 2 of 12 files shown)


surahman (approved):

nicknezis:
What was the error encountered that required the change from `float` to `string` in the timeline object? I ask because there were various issues with the intermingling of the types. I was hoping we could keep it consistent, but recognize there might be use cases I didn't account for with the prior testing. So I'm interested to better understand what this PR fixes, and also does the timeline in the UI still work? If so, then I think I'm ok with this PR.

surahman:
> I ask because there were various issues with the intermingling of the types. I was hoping we could keep it consistent, but recognize there might be use cases I didn't account for with the prior testing.

I approved because Python typing will, rather annoyingly, allow you to achieve what follows. I agree with the need for consistency.

```python
float_Val: float = 9.19
print(float_Val)

str_Val: int = float_Val
print(str_Val, type(str_Val))

str_Val = 919
print(str_Val, type(str_Val))
``

**Comment by thinker0** [comment]
https://github.com/apache/incubator-heron/blob/700125f271b854a5ad2e3fe3204c1e2169ce069a/heron/tools/tracker/tests/python/query_operator_unittest.py#L48-L53

In the internal UnitTest, it seems to be a String.

**Comment by nicknezis** [comment]
Should the unit test be updated to be a float instead of string? I did not catch that.

When I started working on the last PR, it was originally an `int` which I think was incorrect. `str` might be valid option. I think I tried to make it work, but we had enough other issues in the PR. I trust your opinion on `str` vs `float`.

We should make sure we are setting the type consistently. Should this [line of code](https://github.com/apache/incubator-heron/blob/7db7c24733bd7e66ecfe704ea65f864d1fff4adc/heron/tools/tracker/src/python/query_operators.py#L46) also be updated to be a `str`?

**Comment by thinker0** [comment]
humm. Conflit is occurring in several places ㅠ.ㅠ

**Comment by thinker0** [comment]
When tested up to here, it worked without any problem in heron-ui. And Legacy HealthManger is also good, so I think it might be okay to change it up to this version.

**Comment by thinker0** [comment]
I fixed it on my CentOS 7.

**Comment by thinker0** [comment]
```
%  ./docker/scripts/test-unittest.sh centos7 0.20.5
================================================================================
INFO: Elapsed time: 1889.214s, Critical Path: 207.25s
INFO: 3826 processes: 1175 internal, 2651 local.
INFO: Build completed successfully, 3826 total actions
Test cases: finished with 1189 passing and 0 failing out of 1189 test cases

Executed 247 out of 247 tests: 247 tests pass.
INFO: Build completed successfully, 3826 total actions
Cleaning up scratch dir
```

**Comment by thinker0** [comment]
```
% ./docker/scripts/test-unittest.sh rocky8 0.20.5
================================================================================
INFO: Elapsed time: 1835.725s, Critical Path: 246.64s
INFO: 3826 processes: 1175 internal, 2651 local.
INFO: Build completed successfully, 3826 total actions
Test cases: finished with 1189 passing and 0 failing out of 1189 test cases

Executed 247 out of 247 tests: 247 tests pass.
INFO: Build completed successfully, 3826 total actions
Cleaning up scratch dir
```

**Comment by thinker0** [comment]
> Should the unit test be updated to be a float instead of string? I did not catch that.

Nothing


> We should make sure we are setting the type consistently. Should this [line of code](https://github.com/apache/incubator-heron/blob/7db7c24733bd7e66ecfe704ea65f864d1fff4adc/heron/tools/tracker/src/python/query_operators.py#L46) also be updated to be a `str`?

I don't think you need to change this part.
When 0.20.5 centos binary was applied to both tracker and ui in production, it worked in the UI regardless of this part.

**Comment by surahman** [comment]
There was a failing test for killing topologies but I have seen that fail intermittently before. I did restart the CI pipeline.

Is the downgrade of the `dev-toolset` related to an unsupported library causing issues?

I do not see any issues with using the `C++1y` flag instead of `C++14` because the other OSs have their version of GCC and G++ kept up to date. This means the flag will indicate compiling against the final `C++14` standard on those OSs. It will, however, compile using the draft standard on Centos7 (GCC 4.8.x) but given the build completes successfully I believe this should be okay.

I will request @nicknezis to have a look at the tracker change because he has spent a lot of time in it recently.

**Comment by thinker0** [comment]
Fixed some functions.
- When Topology is re-executed (updated), it is giving a message that some '/stats' data of Tmanager is missing. In this case, since some data was missing in the tracker, logic to check this part was added, and if there is no data in this part in the UI, a function to check the value was added to the part where the screen was drawn.  
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-479615841054520a32d26e4a6857c244f63a704b6355292f35d7a96969f3156dL463 
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-479615841054520a32d26e4a6857c244f63a704b6355292f35d7a96969f3156dL480 
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-84d3871b76c715140b59ab14bd4b10351c8bd997c34590cf63e0aef53a4e36bcR997

- Fixed because the current status could not be displayed in the UI
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-29e7366069fde10ea907fccddae5506e19a2d8dcaa5f515ce417fec4957e10ffR69

- - Added because there is no Instance Healthcheck API in Tracker and UI.
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-555d27350199002b67424764ea3ec5f2078e9d04c6d8067b1cb6c8a60b8fc70dR132
-- https://github.com/apache/incubator-heron/pull/3819/files#diff-761812c31ae5763145f0e4fc31e09a80d398d85a306498386d42132d6a3d40cbR617

**Comment by thinker0** [comment]
> There was a failing test for killing topologies but I have seen that fail intermittently before. I did restart the CI pipeline.
> 
> Is the downgrade of the `dev-toolset` related to an unsupported library causing issues?
> 
> I do not see any issues with using the `C++1y` flag instead of `C++14` because the other OSs have their version of GCC and G++ kept up to date. This means the flag will indicate compiling against the final `C++14` standard on those OSs. It will, however, compile using the draft standard on Centos7 (GCC 4.8.x) but given the build completes successfully I believe this should be okay.
> 
> I will request @nicknezis to have a look at the tracker change because he has spent a lot of time in it recently.

This was judged to be a cause other than the cause of gcc, so it was restored to its original state.

File: heron/healthmgr/src/java/org/apache/heron/healthmgr/HealthManager.java [hunk 1]
```diff
@@ -153,7 +152,6 @@ public HealthManager(Config config, AbstractModule baseModule) {
   }
 
   public static void main(String[] args) throws Exception {
-    Slf4jUtils.installSLF4JBridge();
```
nicknezis, line 156: This is no longer needed?

thinker0: Yes. In HealthManger , there was a problem and it was removed.

heron/tools/tracker/src/python/query_operators.py
```diff
@@ -20,13 +20,17 @@
...
+import logging
 import math
 
 from typing import Any, Dict, List, Optional, Union
 
+from heron.common.src.python.utils import log
 from heron.proto.tmanager_pb2 import TManagerLocation
 from heron.tools.tracker.src.python.metricstimeline import get_metrics_timeline
 
+Log = log.Log
```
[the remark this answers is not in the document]
thinker0: I removed it.

nicknezis (approved):

Issue #3820, opened 2022-04-23
Topology Zookeeper Update sync error


thinker0:
**Describe the bug**
The information seems to be broken in the process of updating the topology information received from Zookeeper.

Restarting HeronTracker works without any problems.
```
    raise KeyError(f"Topology not found for {cluster}, {role}, {environ}, {topology_name}, {len(topologies)}")
KeyError: 'Topology not found for lad, None, beta, heron-system-access-es-beta, 2'
```
```
[2022-04-23 16:37:17 +0900] [37m[DEBUG][0m: HTTP Request: POST http://shared-aurora-002-dev:31210/stats "HTTP/1.1 200 "[0m
[2022-04-23 16:37:17 +0900] [33m[WARNING][0m: Received response from Tmanager: Metrics not available for component `oa-message-parser`[0m
[2022-04-23 16:37:17 +0900] [33m[WARNING][0m: Received response from Tmanager: Metrics not available for component `oa-message-parser`[0m
[2022-04-23 16:37:17 +0900] [33m[WARNING][0m: Received response from Tmanager: Metrics not available for component `oa-message-parser`[0m
[2022-04-23 16:37:17 +0900] [33m[WARNING][0m: Received response from Tmanager: Metrics not available for component `oa-message-parser`[0m
[2022-04-23 16:37:17 +0900] [33m[WARNING][0m: Received response from Tmanager: Unknown component: __stmgr__[0m
```

**Screenshots**

**Operating System**
 - OS: centos7
 - Version: 12.3.1 (21E258)

**Additional context**
- heron-0.20.5 (master)
- Update heron topology 
- Tracker Error (watch Zookeeper)

PR #3830, opened 2022-05-12
Fix Update of topology

thinker0:
Fix Update of topology 
#3820 

In case of redeploy and update, the Topology Object information is conflicted in the Tracker, and the normal Topology update is not performed, and the Tracker/UI becomes abnormal.

When watching Zookeeper is used, Race Condition occurs at the same time because part [Link](https://github.com/apache/incubator-heron/pull/3830/files#diff-29e7366069fde10ea907fccddae5506e19a2d8dcaa5f515ce417fec4957e10ffR604-R631)/[Call](https://github.com/apache/incubator-heron/blob/d1cc8dd69223fcce41fccdbc2472864e8c1a5222/heron/tools/tracker/src/python/topology.py#L629-L647) is executed at the same time.
There are cases where Topology is registered as a duplicate. A lock was used to remove the part where the race condition occurred. [Link](https://github.com/apache/incubator-heron/pull/3830/files#diff-08d60d80f8e06cd2706c6b1ac3e791e2a57e67178f420d89d69780969bc8eda9R131-R146)

### DIFF
--- a/heron/tools/tracker/src/python/tracker.py
+++ b/heron/tools/tracker/src/python/tracker.py
@@ -66,8 +66,7 @@ class Tracker:
     def on_topologies_watch(state_manager: StateManager, topologies: List[str]) -> None:
       """watch topologies"""
       topologies = set(topologies)
-      Log.info(f"State watch triggered for topologies of {state_manager.name}.")
-      Log.debug(f"Received topologies: {state_manager.name}, {topologies}")
+      Log.debug("Received topologies: %s, %s", state_manager.name, topologies)
       cached_names = {t.name for t in self.get_stmgr_topologies(state_manager.name)}
       Log.debug(f"Existing topologies: {state_manager.name}, {cached_names}")
       for name in cached_names:


heron/tools/tracker/src/python/metricstimeline.py
```diff
@@ -84,7 +84,7 @@ async def get_metrics_timeline(
 
   # Form and send the http request.
   url = f"http://{tmanager.host}:{tmanager.stats_port}/stats"
-  Log.debug(f"Making HTTP call to fetch metrics: {url}")
+  # Log.debug(f"Making HTTP call to fetch metrics: {url}")
```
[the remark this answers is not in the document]
thinker0: There are too many logs during DEBUG, so it would be better to delete them.

I guess I'll have to do something else later to reduce the DEBUG log.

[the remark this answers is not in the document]
thinker0: Reduce debug log https://github.com/apache/incubator-heron/pull/3836

nicknezis (approved):

thinker0:
Review @windhamwong @Code0x58

windhamwong:
The code looks good to me, but just wonder why the issue happens before this PR. I believe the lock works here well, but finding the root cause might be the key to fix the fundamental issue

thinker0:
In case of redeploy, the Topology Object information is conflicted in the tracker, and the normal Topology update is not performed, and the Tracker/UI becomes abnormal.

heron/tools/tracker/src/python/topology.py
```diff
@@ -589,30 +601,31 @@ def _update(
       scheduler_location=...,
     ) -> None:
     """Atomically update this instance to avoid inconsistent reads/writes from other threads."""
```
Code0x58, line 603: If there was one topology per thread, then I don't think the lock would have an impact, but that's a big _if_ given I can't recall the threading structure. If that was the case, you'd probably want the lock for a consistent read, e.g. while doing a comparison/`__eq__`.

heron/tools/tracker/src/python/tracker.py
```diff
@@ -64,17 +66,18 @@ def sync_topologies(self) -> None:
...
-      for name in cached_names - topologies:
-        Log.info("Removing topology: %s in rootpath: %s",
-                 name, state_manager.rootpath)
-        self.remove_topology(name, state_manager.name)
-
-      for name in topologies - cached_names:
-        self.add_new_topology(state_manager, name)
+      Log.info(f"State watch triggered for topologies of {state_manager.name}.")
+      Log.debug(f"Received topologies: {state_manager.name}, {topologies}")
+      cached_names = [t.name for t in self.get_stmgr_topologies(state_manager.name)]
```
Code0x58, line 71: was there a reason to convert this from a set to a list? the code below still ends up doing what are essentially set operations, just implicitly and slower

heron/tools/tracker/src/python/tracker.py
```diff
@@ -64,17 +66,18 @@ def sync_topologies(self) -> None:
...
-      Log.debug("Existing topologies: %s", cached_names)
-      for name in cached_names - topologies:
-        Log.info("Removing topology: %s in rootpath: %s",
-                 name, state_manager.rootpath)
-        self.remove_topology(name, state_manager.name)
-
-      for name in topologies - cached_names:
-        self.add_new_topology(state_manager, name)
+      Log.info(f"State watch triggered for topologies of {state_manager.name}.")
+      Log.debug(f"Received topologies: {state_manager.name}, {topologies}")
```
Code0x58, line 70: it's generally recommended to leave the logging in the form `(format_str, *args)`, as it avoids unnecessary stringification and interpolation (i.e. when the message level is lower than the configured logging level), so would be good to change back to that format

heron/tools/tracker/src/python/tracker.py
```diff
@@ -100,11 +103,15 @@ def get_topology(
                   and t.cluster == cluster
                   and (not role or t.execution_state.role == role)
                   and t.environ == environ]
-    if len(topologies) != 1:
+    if len(topologies) == 0:
```
Code0x58, line 106: as this is a behaviour change, is it known to rectify anything? now it warns if there is more than one topology, but not I wonder if that is a but. I don't recall if this would cause issues (without reading around a lot), so not sure it should be included unless it is a known improvement

heron/tools/tracker/src/python/tracker.py
```diff
@@ -114,37 +121,37 @@ def get_stmgr_topologies(self, name: str) -> List[Any]:
...
-    # avoid races due to concurrent execution
-    self.topologies.append(topology)
-
-    # Set watches on the pplan, execution_state, tmanager and scheduler_location.
-    state_manager.get_pplan(topology_name, topology.set_physical_plan)
-    state_manager.get_packing_plan(topology_name, topology.set_packing_plan)
-    state_manager.get_execution_state(topology_name, topology.set_execution_state)
-    state_manager.get_tmanager(topology_name, topology.set_tmanager)
-    state_manager.get_scheduler_location(topology_name, topology.set_scheduler_location)
+    with self.lock:
```
Code0x58, line 130: I guess `state_manager.name` can't change and `self.config` can't change

heron/tools/tracker/src/python/tracker.py
```diff
@@ -114,37 +121,37 @@ def get_stmgr_topologies(self, name: str) -> List[Any]:
...
+      if topology not in self.topologies:
+        Log.info(f"Adding new topology: {topology_name}, state_manager: {state_manager.name}")
+        self.topologies.append(topology)
+
+      # Set watches on the pplan, execution_state, tmanager and scheduler_location.
+      state_manager.get_pplan(topology_name, topology.set_physical_plan)
+      state_manager.get_packing_plan(topology_name, topology.set_packing_plan)
+      state_manager.get_execution_state(topology_name, topology.set_execution_state)
+      state_manager.get_tmanager(topology_name, topology.set_tmanager)
+      state_manager.get_scheduler_location(topology_name, topology.set_scheduler_location)
```
Code0x58, line 140: I suppose here you might want a lock on the `state_manager` for a consistent read, unless there's lock or immutability somewhere else to keep things consistent

heron/tools/tracker/src/python/topology.py
```diff
@@ -241,8 +244,16 @@ def __init__(self, name: str, state_manager_name: str, tracker_config: Config) -
...
+    # it is repopulated every time one of the pb2 properties is updated
     self.info: Optional[TopologyInfo] = None
+    self.lock = threading.RLock()
+
+  def __eq__(self, o):
+    return isinstance(o, Topology) \
+           and o.name == self.name \
+           and o.state_manager_name == self.state_manager_name \
+           and o.cluster == self.cluster \
+           and o.environ == self.environ
```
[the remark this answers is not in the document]
thinker0:
```
if name not in topologies:
```

Since @dataclass compares all values and judges that only the exact same Object is the same, we want some values to be the same.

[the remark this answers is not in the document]
thinker0: If this part is missing, topologies are added as duplicates.

heron/tools/tracker/src/python/tracker.py
thinker0: Updated

heron/tools/tracker/src/python/tracker.py
thinker0: Updated

heron/tools/tracker/src/python/tracker.py
thinker0: Updated

heron/tools/tracker/src/python/tracker.py
thinker0: ???

heron/tools/tracker/src/python/tracker.py
thinker0: okay. I'll erase it.

heron/tools/tracker/src/python/topology.py
thinker0: ???

thinker0: If the code below is,

When watching Zookeeper is used, Race Condition occurs at the same time because part [Link](https://github.com/apache/incubator-heron/pull/3830/files#diff-29e7366069fde10ea907fccddae5506e19a2d8dcaa5f515ce417fec4957e10ffR604-R631) is executed at the same time.

thinker0:
```
heron update lad/www/beta heron-system-access-prometheus --component-parallelism parser:1
heron update lad/www/beta heron-system-access-prometheus --component-parallelism parser:2
```
Lock is required because the above ```_update``` is executed at the same time when instance update occurs.

heron/tools/tracker/src/python/tracker.py
thinker0:
```
bazel test --config=stylecheck heron/tools/tracker/...
```
```
C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
```
Reverted

Code0x58: is that the case for things like `Log.debug("Existing topologies: %s", cached_names")`? That is different from `"Existing topologies: %s" % cached_names` which is where I'd expect to see what linter warning

nicknezis: Correct. I don't think this was a style check issue. When I did the conversion to f-string during the Python upgrade, I left some of these Log statements in the (format_str, &args) format because it is more efficient. Doing the f-string first and then passing into a log statement that might not render is wasted effort. The Log statement will only do the string interpolation if the log statement needs to be rendered. Because of this, I felt it was better to keep it in the original form.

heron/tools/tracker/src/python/tracker.py
```diff
@@ -64,17 +66,18 @@ def sync_topologies(self) -> None:
     def on_topologies_watch(state_manager: StateManager, topologies: List[str]) -> None:
       """watch topologies"""
       topologies = set(topologies)
-      Log.info("State watch triggered for topologies.")
-      Log.debug("Topologies: %s", topologies)
+      Log.info(f"State watch triggered for topologies of {state_manager.name}.")
+      Log.debug(f"Received topologies: {state_manager.name}, {topologies}")
```
nicknezis, line 70:
```suggestion
      Log.debug("Received topologies: %s, %s", state_manager.name, topologies)
```

nicknezis (approved):

PR #3838, opened 2022-05-30
Removed random long in filename which caused leaking in upload storage

### LINKED ISSUE #3769: Uploader implementations may leak storage when resubmitting the same analytic multiple times
**Describe the bug**
The Uploader logic in Heron adds a Random Long to the filename that causes memory leak in Bookkeeper.
The Uploader filename generation logic is [here](https://github.com/apache/incubator-heron/blob/abb2767e3df3ca6eba009f46efe1f1e83695617a/heron/spi/src/java/org/apache/heron/spi/utils/UploaderUtils.java#L62).

And the Dlog config that disables time and size based retention is [here](https://github.com/apache/incubator-heron/blob/abb2767e3df3ca6eba009f46efe1f1e83695617a/heron/uploaders/src/java/org/apache/heron/uploader/dlog/DLUploader.java#L122-L128).

Eventually the Bookkeeper log runs out of space and no more topologies can be submitted.

The proposed fix is to remove the Random.long that is being added to the filename.

nicknezis:
Fixes #3769 

If you exec into the Bookkeeper pod, you can run the following command in `/opt/bookkeeper/bin`
`./dlog tool watch -u distributedlog://heron-zookeeper:2181/heron`

When I upload the same package, I can see it deleting the old one, and then uploading the new one
```
Old streams :
acking-root-tag-0.tar.gz
New streams :

Old streams :
New streams :
acking-root-tag-0.tar.gz
```

Without this fix, this [bit of code](https://github.com/apache/incubator-heron/blob/f3762f65cd101459d32a431977e980b69155f4ad/heron/uploaders/src/java/org/apache/heron/uploader/dlog/DLUploader.java#L181) is never executed. The package name never matches, so it never deletes the old package.

### DIFF
--- a/deploy/kubernetes/helm/templates/bookie.yaml
+++ b/deploy/kubernetes/helm/templates/bookie.yaml
@@ -43,9 +43,6 @@ data:
   BK_indexDirectories: "/bookkeeper/data/ledgers" 
   BK_zkServers: {{ .Release.Name }}-zookeeper:{{ .Values.zookeeper.clientPort }}
   BK_autoRecoveryDaemonEnabled: "true"
-  BK_journalMaxBackups: "{{ .Values.bookkeeper.journal.maxBackups }}"
-  BK_journalMaxSizeMB: "{{ .Values.bookkeeper.journal.maxSizeMB }}"
-  BK_logSizeLimit: "{{ int64 .Values.bookkeeper.logSizeLimit }}"
   {{- if .Values.bookkeeper.useHostNameAsBookieID }}
   BK_useHostNameAsBookieID: "true"
   {{- end }}
--- a/deploy/kubernetes/helm/values.yaml.template
+++ b/deploy/kubernetes/helm/values.yaml.template
@@ -89,8 +89,6 @@ bookkeeper:
   useStatefulSet: true
   affinityPods: false
   useVolumeClaimTemplate: true
-
-  logSizeLimit: 10000000
   useHostNameAsBookieID: true
   tolerateUnreadyEndpoints: true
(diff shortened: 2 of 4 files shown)


nicknezis:
Also reverting [this change](https://github.com/apache/incubator-heron/pull/3628) which was added 2 years ago as a bandaid to remedy #3769 .

windhamwong:
The commit looks good. Just wonder what was the original reason of using a random name in bookkeeper upload?

nicknezis:
> The commit looks good. Just wonder what was the original reason of using a random name in bookkeeper upload?

It looks like a bit of code added early in this [PR](https://github.com/apache/incubator-heron/pull/601). Someone suggested they use a timestamp instead of random to see which package was newest. But it doesn't seem to have much thought put behind the original Random Long. I'll have to look further back in the history to determine when it was added.

nwangtw:
Based on the unit test: `testGenerateFilename`, it seems like the file name is expected to be unique.

nicknezis:
> Based on the unit test: `testGenerateFilename`, it seems like the file name is expected to be unique.

Interesting, the test still passes. I'll have to dig into it further. But looking in each Uploader, they seem to have logic for uploading a package with the same name. But this logic would never be needed if the intention was to always generate a unique filename. I really think this is a relic from the old Twitter code base. It doesn't seem to provide any value.

nwangtw:
Yeah. It sounds a little strange. Maybe the logic is assuming manual cleanup or some kind of TTL?

nwangtw (approved):

PR #3843, opened 2022-06-29
Remove stray space characters between \ and \n

jmtd:
The example bash shell commands in this document are wrapped onto multiple lines with
backslashes for legibility. However, there are space characters inserted after the backslash
escapes, breaking the commands: the spaces are being escaped, not the newlines. An
attempt to copy and paste the commands directly from the documentation will fail.

Remove the spaces to fix the commands.

### DIFF
--- a/website2/docs/getting-started-docker.md
+++ b/website2/docs/getting-started-docker.md
@@ -30,16 +30,16 @@ The official Apache Heron Docker image is located at the link below
 In one terminal execute to start Heron in a container
 
 ```bash
-$ docker run -it  --rm \ 
-   -p 8889:8889 \ 
-   -p 8888:8888 \ 
-   --name local-heron \ 
+$ docker run -it  --rm \
+   -p 8889:8889 \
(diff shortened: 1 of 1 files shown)


jmtd:
I haven't yet figured out if/how the documentation at https://hub.docker.com/r/apache/heron is populated, but it suffers the same bug.

joshfischer1108:
Good catch and thank you for the fix.  I can work to get the docker stuff updated today.

joshfischer1108:
@jmtd , please check to verify the issue has been fixed. https://hub.docker.com/repository/docker/apache/heron

jmtd:
@joshfischer1108 thank you, I can confirm this issue is fixed on the docker hub too

(a separate issue is the example fails - but I'll report that separately once I've played around with it a bit more)

Thanks!

joshfischer1108:
Ha! Open source work never ends 😄 .  Thank you for looking into it.

PR #3868, opened 2023-01-08
Updated to fix broken file download

### LINKED ISSUE #3867: File download link does not work in the Heron UI
**Describe the bug**
When attempting to download log files from the UI, the resulting file has an "Internal Server Error" message instead of the file content. The logs are viewable in the UI, but the actual download is broken. This is true for both the download link on the log view page, and also the download links from the file browser.

**Expected behavior**
File should download.

nicknezis:
Fixes #3867


Code0x58 (approved):

Code0x58:
LGTM, just need to appease the linter which is why CI failed: heron/tools/tracker/src/python/routers/container.py:84:0: C0303: Trailing whitespace (trailing-whitespace)

